PR #16, opened 2017-03-21
Cocoapods support

HazAT:
~~This is currently on hold because there are a few open issues with react-native itself.
https://github.com/facebook/react-native/pull/12089#issuecomment-276654535
Waiting for 0.43 release.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 *Requirements:*
 
-* `react-native >= 0.41`
+* `react-native >= 0.38`
 * sentry-cli > 0.26 (`brew install getsentry/tools/sentry-cli`)
 
 With this SDK, Sentry is now able to provide mixed stacktraces. This means that if a JavaScript call causes a crash in native code, you will see the last call from JavaScript before the crash. This also means that with the new SDK, native crashes are properly handled on iOS.
--- a/SentryReactNative.podspec
+++ b/SentryReactNative.podspec
@@ -1,0 +1,26 @@
+require 'json'
+
+Pod::Spec.new do |s|
+  # NPM package specification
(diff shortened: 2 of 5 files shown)


oblador:
Hi, I believe your issue is unrelated. It works with React Native 0.42 if you also add `s.dependency 'React'`.

HazAT:
I am pretty sure you will get an ancient version of react if you do this ...
http://cocoapods.org/pods/react
It's marked as deprecated ... the new way of installing react as a pod is ...
`npm install react-native`

And then in your `Podfile`
```
pod 'React', :path => './node_modules/react-native', :subspecs => [
      'Core',
      'RCTImage',
      'RCTNetwork',
      'RCTText',
      'RCTWebSocket',
      # Add any other subspecs you want to use in your project
    ]
```

oblador:
Hi, yeah, you have to do that in your main Podfile then this podspec would use that one. If you haven't defined that one then yes it'd download version 0.11 or similar from the cocoapods repo.

oblador:
I guess there's no way around using CocoaPods for RN if you want to use it for other dependencies that depend on RN. And if you aren't using the local node-modules version then it won't work anyway.

oblador:
The PR you're referencing seems to be released in 0.42 already btw: https://github.com/facebook/react-native/commit/2981560e400c0e4982e992c582d08fef4edb1cde

HazAT:
I am getting this error ...
screen shot 2017-03-21 at 17 46 38
Haven't found a fix for it yet ... Changing the Header Search Paths doesn't do anything.

oblador:
Adding `s.dependency 'React'` fixes that error for me :-)

HazAT:
@oblador Thx that actually works 💯 
Will finish the PR tomorrow ...

PR #27, opened 2017-03-30
Remove superfluous calls to RCTConvert

eliperkins:
These calls to RCTConvert are inferred based on the types passed into
the macro, according to the RN docs.

http://facebook.github.io/react-native/releases/0.42/docs/native-modules-ios.html#argument-types


eliperkins:
Ah, just noticed these were added in a036f682edc60a5c4e427c9edaea9a0d7473c9f2. Is there a reason these are necessary?

HazAT:
Thanks for pointing that out 👍

PR #40, opened 2017-04-25
Add out of the box support for Expo

brentvatne:
This resolves various issues that I discussed with Daniel and Armin on Slack. Changes to `lib/Sentry.js` depend on new configuration options which I have PR'd to raven-js.

Wait until https://github.com/getsentry/raven-js/pull/938 lands to merge.

### DIFF
--- a/lib/Sentry.js
+++ b/lib/Sentry.js
@@ -1,7 +1,7 @@
 import {
     NativeModules
 } from 'react-native';
-import Raven from 'raven-js';;
+import Raven from 'raven-js';
 
 const {
     RNSentry
--- a/lib/Sentry.js
+++ b/lib/Sentry.js
@@ -96,9 +96,6 @@ export class Sentry {
 
     static setDataCallback(callback) {
       Sentry._ravenClient.setDataCallback(callback);
-      if (Sentry.isNativeClientAvailable()) {
-          console.warn('setDataCallback not supported on native client');
-      }
     }
 
     static setUserContext(user) {
--- a/lib/Sentry.js
+++ b/lib/Sentry.js
@@ -94,6 +94,13 @@ export class Sentry {
         if (Sentry.isNativeClientAvailable()) Sentry._nativeClient.nativeCrash();
     }
 
+    static setDataCallback(callback) {
+      Sentry._ravenClient.setDataCallback(callback);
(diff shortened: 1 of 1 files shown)


HazAT:
Hey, just discussed this with @mitsuhiko what we gonna do here.
We don't like the idea that we have a specific cloudfront url in the sentry codebase.
Mainly because it means changing the url results in updating the sentry client.
What we propose instead is that we expose the `setDataCallback` function which expo then can use to overwrite whatever is needed.
Does this make sense @brentvatne ?

Will add a separate PR for this if this is fine.

brentvatne:
> What we propose instead is that we expose the setDataCallback function which expo then can use to overwrite whatever is needed.
> Does this make sense @brentvatne ?

Yeah this makes sense -- the plan was to only have this temporarily and then make native changes in Expo to be able to remove the URL but callbacks work too. It's worth noting that we will need to have two callbacks: one to process the error before it gets passed into TraceKit, and another to be called in place of normalizeUrl if provided.

brentvatne:
@HazAT - force pushed over my previous commits! I ended up just exposing setDataCallback and disabling the timer and eventTarget instrumentation (you will want this for react-native in general too).

lib/Sentry.js
```diff
@@ -90,6 +94,13 @@ export class Sentry {
         if (Sentry.isNativeClientAvailable()) Sentry._nativeClient.nativeCrash();
     }
 
+    static setDataCallback(callback) {
+      Sentry._ravenClient.setDataCallback(callback);
+      if (Sentry.isNativeClientAvailable()) {
```
HazAT: I would remove this `if` completely, we don't need this on the native client since the event gets mutated before its sent to the native code.

HazAT:
Also about the the TraceKit issue ... we should fix this directly in tracekit in our `raven-js` repo and not expose a second callback.
https://github.com/getsentry/raven-js/blob/master/vendor/TraceKit/tracekit.js

brentvatne:
@HazAT - I ended up not adding a second callback, I just exposed `setDataCallback`

lib/Sentry.js
brentvatne: removed

PR #56, opened 2017-05-15
Fix missing keys in Android event context

Amwam:
I was getting errors running on Android. The keys for 'user' and 'tags' weren't defined (at least at startup time), so I've wrapped their access in a check.

I wasn't sure if they should be defined at this point in time, I'd be happy to change this PR, if you could point me in the right direction

### DIFF
--- a/android/src/main/java/io/sentry/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/RNSentryModule.java
@@ -132,14 +132,15 @@ public class RNSentryModule extends ReactContextBaseJavaModule {
                     .withLogger(event.getString("logger"))
                     .withLevel(eventLevel(event.getString("level")));
 
-            eventBuilder.withSentryInterface(
-                    new UserInterface(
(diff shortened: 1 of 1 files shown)


android/src/main/java/io/sentry/RNSentryModule.java
```diff
@@ -132,14 +132,15 @@ public void captureEvent(ReadableMap event) {
...
+                UserInterface userInterface = new UserInterface(
+                        event.getMap("user").getString("userID"),
+                        event.getMap("user").getString("username"),
+                        null,
+                        event.getMap("user").getString("email")
+                );
+                eventBuilder.withSentryInterface(userInterface);
+            }
 
             helper.helpBuildingEvent(eventBuilder);
```
HazAT: Can you also add a check for extra?

Amwam: @HazAT Done and rebased. Thanks!

HazAT:
Thank you 👍

PR #72, opened 2017-05-29
Initial support for codepush

mitsuhiko:
This requires support in sentry-swift that does not exist yet.


HazAT:
I think we can close this in favor of the latest changes in https://github.com/getsentry/react-native-sentry/pull/69

HazAT:
This works now like described in the docs
screen shot 2017-05-30 at 16 51 54

Issue #55, opened 2017-05-13
allow calls to sentry without install


alxyuu:
The plain javascript client allowed sentry calls without installing. this was extremely helpful when separating production from development environments, where reporting doesn't make sense. Is it possible to have this same behavior on the react-native client?

PR #85, opened 2017-06-05
#55 Allow calls to be made to Sentry when not installed.

mitch104:
Resolves issue #55


HazAT:
Hey, thx for the PR.
Somehow is this PR very "cluttery", please clean it up first I am not sure if I can merge it that way.

Also few notes, we don't need the fix for sentry-cli in `package.json` anymore.

And instead of doing this in every function ...
```
if (Sentry._ravenClient) {
  Sentry._ravenClient.setDataCallback(callback);
} else {
  if (Sentry.options.logLevel >= 2) {
    console.log('react-native-sentry (setDataCallback):', callback);
  }
}
```
refactor to this:
```
Sentry._log('react-native-sentry (setDataCallback):', callback);
if (Sentry._ravenClient) Sentry._ravenClient.setDataCallback(callback);
```

It's ok to log everytime, since this only should log if `logLevel` is `debug`.
Also if we ever want to change anything how we log, we can do it in one place.

mitch104:
Superseded by https://github.com/getsentry/react-native-sentry/pull/122

PR #115, opened 2017-06-21
Streamline exception handling

### LINKED ISSUE #110: Exception doesn't get sent to sentry while in dev mode?
I'm forcing my app to crash and also doing 
```
Sentry.captureException(new Error('Oops!'), {
  logger: 'my.module'
});
```
but I don't see the exception getting logged in Sentry dashboard.

I see the error and stacktrace in the `raven-js` module: 

https://github.com/getsentry/raven-js/blob/master/src/raven.js#L1498

but when going deeper to the Objective-C native module. I see the event here:

https://github.com/getsentry/react-native-sentry/blob/master/ios/RNSentry.m#L343

But the `RCTExceptionsManagerDelegate` delegate never gets fired:

https://github.com/getsentry/react-native-sentry/blob/master/ios/RNSentry.m#L397-L415

On the iOS simulator, i see the react native red box error dialog.

Is it not sending to sentry because i'm in development mode?

`captureMessage` works perfectly.

HazAT:
This streamlines the exception handling and always uses raven-js to capture the exception, we just use the native client as a transport layer.

- Fix #110

- This is required https://github.com/getsentry/sentry-java/pull/428


vafada:
@HazAT 

If you just use native code as a transport layer then what's the advantage of using native code vs using raven-js to send message/data to sentry server?

also does this change capture native errors/stacktrace (not JS errors, Java or Objective-C error) and report to sentry server?

HazAT:
Hey, we recommend using the native code as a transport layer because you get a lot more useful information out of your crashes. Mostly information about the device.
I've updated the readme and added more screenshots why it's better :)

PR #122, opened 2017-06-27
[Revamped] Allow calls to be made to Sentry when not installed

### LINKED ISSUE #55: allow calls to sentry without install

benhowes:
changed lib/Sentry.js to not require the raven client to be installed following advice from #85. Closes #55.

Previous PR got in a bit of a pickle, so I've opened this one with the relevant commit cherry picked out.

Thanks

### DIFF
--- a/lib/Sentry.js
+++ b/lib/Sentry.js
@@ -93,7 +93,7 @@ export class Sentry {
     return Sentry._nativeClient !== undefined;
   }
 
-  static _log(..args) {
+  static _log(...args) {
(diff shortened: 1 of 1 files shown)


HazAT:
🎉  Thx @benhowes @mitch104

benhowes:
Tested on a local iOS app, without installing raven

PR #142, opened 2017-07-07
Bumping sentry-cocoa version to 3.1.2

VojtechBartos:
Currently cocoapods install max version 3.0.12 of `sentry-cocoa`, but in latest version of `react-native-sentry` you are already using latest interface. Especially `shouldSendEvent` callback of `SentryClient`, which has been introduced in version 3.1.0. So we are not able to build it ;) See screenshot below

screen shot 2017-07-07 at 11 48 36


HazAT:
Oh, thx :)

PR #146, opened 2017-07-10
Do not search recursively in glob for ios.

mitch104:
This issue has been brought up before, there is an ignore list for the glob `['node_modules/**', 'ios/Pods/**', '**/Pods/**']` however I have a git submodule within my project hence this does not get ignored. So changed it such that the glob only searches 1 directory down for the `*.xcodeproj/project.pbxproj` file.


benhowes:
This does not address the same potential issue in android code bases, but we weren't sure how to tackle that one @HazAT

Issue #163, opened 2017-07-22
Crash on Android


ovr:
Hey!

```
java.lang.NullPointerException: null
    at io.sentry.RNSentryModule.captureBreadcrumb(RNSentryModule.java:172)
    at java.lang.reflect.Method.invokeNative(Method.java)
    at java.lang.reflect.Method.invoke(Method.java:515)
    at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:363)
    at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:166)
    at com.facebook.react.bridge.queue.NativeRunnable.run(NativeRunnable.java)
    at android.os.Handler.handleCallback(Handler.java:733)
    at android.os.Handler.dispatchMessage(Handler.java:95)
    at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:31)
    at android.os.Looper.loop(Looper.java:146)
    at com.facebook.react.bridge.queue.MessageQueueThreadImpl$3.run(MessageQueueThreadImpl.java:194)
    at java.lang.Thread.run(Thread.java:841)
```

Thank

dioxide:
same problem!

ovr:
refs https://github.com/ovr/ghubber/issues/40

HazAT:
This should be fixed with `0.14.9`

PR #164, opened 2017-07-23
captureBreadcrumb(), data: Skip toString for null, refs #163

ovr:


ovr:
refs https://github.com/ovr/ghubber/issues/40

ovr:
Was null for status_code, will be safety to add `null` check, cc @HazAT

HazAT:
Thx 👍

PR #172, opened 2017-08-02
Fixed Compilation Error in React Native 0.47.0

deathemperor:
createJSModules was removed in 0.47.0 release here https://github.com/facebook/react-native/releases/tag/v0.47.0
RN code change here https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8

I only remove @Override keyword, not sure if keeping the function would make it compatible with older versions of RN. Please check.


HazAT:
Thx 👍

PR #214, opened 2017-08-30
Improve performance on stacktrace merging

### LINKED ISSUE #158: Sentry is delaying react-navigation transition
I was investigating https://github.com/react-community/react-navigation/issues/608 and it looks like Sentry takes very long to parse an error during a transition. On a physical iPhone 6, it was delaying a transition by over 3 seconds. I found this by profiling with Instruments on macOS. Here's a screenshot of the stacktrace, where you can see that Sentry calls are blocking the transition for a long time. Removing Sentry from the app fixes the problem. I tried upgrading to the latest version, problem persists. Let me know if I can help with further details.

bildschirmfoto 2017-07-19 um 16 50 39

HazAT:
Fixes #158 

For react-native > 0.45 they no longer provide the `moduleID` <-> `moduleName` mapping we use to not trace every stacktrace that crosses the native bridge.
This result is a massive performance impact when for example an animation happens and react-native sends an update over the native bridge for every frame (almost 60 updates / sec) which is way to much when we always create a `new Error().stack`.

This PR adds a workaround; We are able to fetch the `moduleName` on the native side, I additionally send the `moduleID` over the bridge and build an index on native with `moduleID` <-> `moduleName`.
After an update to this, I propagate the changes to javascript (EventEmitter) and build our `_ignoredModules` in javascript which we then use to decide what we send over the native bridge. 
The result of this is that we build the ignore index on the fly which improves performance greatly.

### DIFF
--- a/ios/RNSentry.m
+++ b/ios/RNSentry.m
@@ -14,7 +14,7 @@ NSString *const RNSentrySdkName = @"sentry-react-native";
 
 @interface RNSentry()
 
-@property (nonatomic, strong) NSDictionary *lastReceivedException;
+@property (nonatomic, strong) NSMutableDictionary *moduleMapping;
 
 @end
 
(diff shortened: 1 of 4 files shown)


akshay2604:
Will this PR also solve the react-navigation issue on android?

HazAT:
@akshay2604 Whats the issue on android?

akshay2604:
Well navigation transitions in android are very slow using react-navigation.

PR #223, opened 2017-09-06
feat(typescript): Add typescript definitions

### LINKED ISSUE #194: TypeScript definitions
Do you plan to add TypeScript definitions?

HazAT:
Fixes #194

### DIFF
--- a/lib/Sentry.d.ts
+++ b/lib/Sentry.d.ts
@@ -7,90 +7,90 @@
 type SentryBreadcrumbType = "navigation" | "http";
 
 interface SentryBreadcrumb {
-    message?: string;
-    category?: string;
-    level?: SentrySeverity;
-    data?: object;
-    type?: SentryBreadcrumbType;
+  message?: string;
+  category?: string;
+  level?: SentrySeverity;
+  data?: object;
+  type?: SentryBreadcrumbType;
 }
 
 export enum SentrySeverity {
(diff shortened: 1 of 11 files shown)


lib/Sentry.d.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    static captureMessage(message: string, options?: object): void;
+
+    static captureException(ex: Error, options?: object): void;
+
+    static captureBreadcrumb(breadcrumb: SentryBreadcrumb): void;
+
+    static clearContext(): void;
+
+    static context(func: Function, ...args: any[]): void;
```
jan-auer: This and some other functions actually call through to `raven-js` and return its return value (which happens to be `void` in these cases). Once [typescript#6606](https://github.com/Microsoft/TypeScript/issues/6606) is closed, we could import raven-js' typescript definition and use the `typeof` operator to infer the return type

typescript/Sentry-tests.ts
```diff
@@ -0,0 +1,11 @@
+// import Sentry = require('./Sentry'); // or import * as Sentry from '..'
+import {Sentry, SentryLog, SentrySeverity} from '../lib/Sentry';
```
jan-auer: Prefer:
```ts
import Sentry, { SentryLog, SentrySeverity } from '../lib/Sentry';
```

PR #266, opened 2017-10-13
upd build tools

kesha-antonov:


HazAT:
Hey, thanks for submitting this PR, we've changed this already back and forth a few times 😅 
I think Facebook still recommends using version `23.0.1` see:
https://facebook.github.io/react-native/docs/android-building-from-source.html
So before they don't update, we really can't since people who create new projects get `23.0.1` by default.

HazAT:
So thx for contributing but I will close this for now.
Whenever facebook decides to bump their build tools we will be on board.

PR #327, opened 2018-01-12
fix: Clean up sentry requested sourcemap

dr-js:
Sentry requested sourcemap (like for release bundle) should be auto deleted after upload is complete.

For our usage, the resulting `release.apk` is ~1MB smaller, except with an extra line in `\assets\index.android.bundle`:

    //# sourceMappingURL=index.android.bundle.map


HazAT:
Thank you for this awesome fix/feature :)

JetCyC:
Why not to delete the sourcemaps upon successful upload by default in ios?
I think it's a basic feature.

PR #334, opened 2018-01-21
Don't crash on _log when Sentry is not configured.

confiks:
When Sentry is not configured and you for example call `Sentry.captureMessage`, `_log` tries to pull `logLevel` out of `undefined`, as `Sentry.options` is only set when calling `Sentry.config` with a valid DSN.


HazAT:
Thx 👍

PR #335, opened 2018-01-23
Add proguard rule

kazy1991:
Fixes a case when Android build if the RN project enables proguard.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -4,6 +4,10 @@ apply plugin: 'com.android.library'
 android {
     compileSdkVersion 23
     buildToolsVersion '23.0.1'
+    
+    defaultConfig {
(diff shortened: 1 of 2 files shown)


HazAT:
@kazy1991 Thank you for your PR.
Wouldn't this enable proguard by default then?

kazy1991:
@HazAT Thank you replying. XD

> Wouldn't this enable proguard by default then?

This changes wouldn't  enable proguard by default.  But if someone want to build project that depends `react-native-sentry` with proguard,  proguard rules that this PR added are loaded automatically.

HazAT:
@kazy1991 Thx for the clarification 👍 
(I am not the Android pro 😅)

PR #340, opened 2018-01-27
Update sentry-cli path on manual-setup.rst

lucascarvalho:
I'm currently using `react-native-sentry` version `0.33.0`.
I don't know if we should be more explicit about the versions here.


HazAT:
Thx

PR #355, opened 2018-02-15
Fixing uncatchable, crashing, exception on Android

cworsley4:
NOTE: This does not seem to occur on iOS. We believe that it fails silently which is also an error. We have not done extensive testing there.

---

We found an issue on Android (8.1 was used) where if we used the public DSN URL we would get an exception citing that we were missing the Sentry secret. In react native, this gave us a red box with the error but it crashed our app. We tried to wrap in a try/catch but no luck. 

The error was happening asynchronously [in the Sentry Java SDK](https://github.com/getsentry/sentry-java/blob/97fc170d0eec16906e56262cfedd075f0153fa72/sentry/src/main/java/io/sentry/dsn/Dsn.java#L204) and not being broadcasted over the [RN bridge like you would expect](https://facebook.github.io/react-native/docs/native-modules-android.html#callbacks) (note this line `errorCallback.invoke(e.getMessage());`, where `errorCallback` is of type Callback from the RN Callback class), causing the error to get lost and our app unable to recover.

To fix this we wanted to `await` on the asynchronous operation. So we created a new method [`async customInit()`](https://github.com/getsentry/react-native-sentry/compare/master...scootnetworks:master?expand=1#diff-57506f97aab5514d97ada99662ff14ebR59) that returns a Promise that we await on. 

In that Promise, we resolve/reject based on if the [error](https://github.com/getsentry/react-native-sentry/compare/master...scootnetworks:master?expand=1#diff-2c23c9bdba652ccf3c4f364415d1c35cR104)/[success](https://github.com/getsentry/react-native-sentry/compare/master...scootnetworks:master?expand=1#diff-2c23c9bdba652ccf3c4f364415d1c35cR154) callback is called.

`await`ing on the `install()` method allows us to wrap that call in a try/catch in case something does throw either Native-ly or in JavaScript.

This was a very concerning bug to us, that our Error Reporting tool would throw an _**uncatchable**_ error on initialization. 

If your SDK is throwing exceptions upon startup, how will it behave now that this PR fixes the uncaught exception? If we catch this error and then another part of our code tries to capture an exception with sentry what happens? Have you thought about that?

If we had properly set the secret we never would have caught this! And in the future, if we improperly updated the secret or fat fingered the DSN URL then we would have users with essentially bricked apps that would have lost as customers.

While we have forked the repo we would hope that this would be either merged or otherwise patched in the official react-native-sentry repository.

### DIFF
--- a/android/src/main/java/io/sentry/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/RNSentryModule.java
@@ -91,7 +91,7 @@ public class RNSentryModule extends ReactContextBaseJavaModule {
     }
 
     @ReactMethod
-    public void startWithDsnString(String dsnString, final ReadableMap options, Callback errorCallback, Callback successCallback) {
+    public void startWithDsnString(String dsnString, final ReadableMap options, Callback successCallback, Callback errorCallback) {
         if (sentryClient != null) {
             logger.info(String.format("Already started, use existing client '%s'", dsnString));
             return;
--- a/android/src/main/java/io/sentry/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/RNSentryModule.java
@@ -7,6 +7,7 @@ import android.support.annotation.NonNull;
 
 import com.facebook.react.ReactApplication;
 import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.Callback;
 import com.facebook.react.bridge.Promise;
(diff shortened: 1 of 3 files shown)


HazAT:
Hey @cworsley4
Thanks for this PR, this makes absolute sense.
I will merge this to master and create a new PR for iOS, before that I cannot release it.

PR #358, opened 2018-02-16
fix: Bugfix async start / Clone Object before send over bridge

HazAT:
This makes the start call async more info here:
https://github.com/getsentry/react-native-sentry/pull/355
I've changed it so it works for iOS and Android.

Fixes https://github.com/getsentry/react-native-sentry/issues/347

### DIFF
--- a/lib/NativeClient.js
+++ b/lib/NativeClient.js
@@ -75,24 +75,28 @@ export class NativeClient {
     });
   }
 
+  _cloneObject(obj) {
+    return JSON.parse(JSON.stringify(obj));
+  }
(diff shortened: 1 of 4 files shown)


lib/NativeClient.js
```diff
@@ -77,36 +71,40 @@ export class NativeClient {
           }
         });
       }
-      RNSentry.setLogLevel(options.logLevel);
+      RNSentry.setLogLevel(this.options.logLevel);
```
cworsley4: Nice! We missed this.

PR #360, opened 2018-02-19
Make sentry.gradle include all ABIs version codes as dist

angly-cat:
Hi!

I've encountered an issue that I haven't find a way how to resolve.

My Android app (and any other React Native app that uses default settings) have builds for 2 ABIs: `"x86"` and `"armeabi-v7a"`. App versions with different ABIs should have different version codes as per Google Play Store requirements.

The issue is that on each build `react-native-sentry` took only one version code on each build, which makes source maps work only for a version of the app with "lucky" ABI (I believe it's the last of alphabetically sorted list) which version code was used as `--dist` argument on source maps upload. 

But for source maps to work on all ABIs, all of their version codes should be accounted.

There's a code in `sentry.gradle`:
```
versionCodes.each { versionCode ->
    args.add("--dist");
    args.add(versionCode);
}
```

However, to date `versionCodes` isn't an array of any sort but a single value.

I wonder why haven't anyone run into this issue before.

This PR should fix this issue. It may break flavor support, I'm not sure, but with your help I think we could fix my issue in the best way possible.

### DIFF
--- a/sentry.gradle
+++ b/sentry.gradle
@@ -11,7 +11,12 @@ gradle.projectsEvaluated {
         def releaseName = "${variant.getApplicationId()}-${variant.getVersionName()}";
         variant.outputs.each { output ->
             def versionCode = output.getVersionCode();
-            releases[variant.getName()] = [output.getName(), releaseName, versionCode];
+            def variantName = variant.getName();
+            def outputName = output.getName();
+            if (releases[variantName] == null) {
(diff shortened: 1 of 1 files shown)


HazAT:
Hey, thx for the PR and sorry for the late reply.
I merged this and wait for the tests in master to run through (there are no actual tests running from outside contributions :/)
👍

PR #362, opened 2018-02-21
RN 54rc3 fix

cworsley4:


cworsley4:
Whoops, wrong repo. But we are seeing an issue where in RN rc3 where `stacktrace.getArray('frames')` needs to be cast to a ReadableNativeArray to avoid a compiler error. Sorry for posting on the wrong repo.

PR #373, opened 2018-03-09
Fixed null object error

Vaintti:
Fixed null object error from calling currentVariants.size() when currentVariants is null by moving null check before size call


HazAT:
thx

PR #410, opened 2018-05-16
fix: Fixed sentry.gradle to do the case insensitive search of build...

Wolfenberg:
…variant

I've encountered a problem in my Android project that sentry didn't upload source maps. It turned out that it was comparing the key of release with current release by usual equality, while my current release was like myRelease and in releases array it was MyRelease. So by ignoring case that shouldn't happen anymore.


HazAT:
🏅Thx 🏅

PR #414, opened 2018-05-22
change codepush command

chenzeyu:
In react-native, codepush is been exported as `codePush` instead of `CodePush`.

The release command has also been changed to be part of appcenter command.


HazAT:
Thanks! We will add full support for (CodePush) -> AppCenter in the coming week.

PR #427, opened 2018-06-19
feat: Remove react application parameter

HazAT:


Gregoirevda:
Was this causing a memory leak?

PR #466, opened 2018-08-17
Fix codepush docs

karltaylor:
Updated the appcenter CLI examples so they are up to date with the current commands.

```bash
$ appcenter codepush release-react YourApp --output-dir ./build
```

exports the bundle files to `./build/codePush/<files>` so this is now reflected in 

```
$ sentry-cli react-native appcenter YourApp ios ./build/codePush
```

(related: https://github.com/getsentry/sentry-cli/issues/233)


HazAT:
Thanks, docs moved and this will be merged with this PR
https://github.com/getsentry/sentry-docs/pull/293

PR #537, opened 2019-01-31
ci: Up build tools

HazAT:


Titozzz:
Hey, any chance you can release after this PR?

Thank you so much for the work ⚡️

HazAT:
Yeah, I just wanted to make the build green but it works locally.
Will do a release now, I think the build will work after the release because the way our tests are set up.

Titozzz:
So nice, thank you 😍

PR #557, opened 2019-03-14
Export types

romain-faust:
Exports all the types from the TypeScript declaration file.
It is sometimes useful to be able to type variables.

_Example :_
```ts
const options: SentryOptions = { ... }
```


HazAT:
I am closing this, please consider using the new `@sentry/react-native` SDK which is unified and written in Typescript.

PR #569, opened 2019-04-02
Fix compile warnings in RNSentry module

sryze:
This patch fixes a few warnings in RNSentry.m (uknonwn selector and type conversion warnings).

### DIFF
--- a/ios/RNSentry.m
+++ b/ios/RNSentry.m
@@ -5,6 +5,16 @@
 #else
 #import "RCTConvert.h"
 #endif
+#if __has_include(<React/RCTBridgeMethod.h>)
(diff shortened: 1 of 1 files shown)


ios/RNSentry.m
```diff
@@ -5,6 +5,16 @@
 #else
 #import "RCTConvert.h"
 #endif
+#if __has_include(<React/RCTBridgeMethod.h>)
```
sryze: I'm not sure if it's better to import private React headers or use NSSelectorFromString()

HazAT:
I am closing this, we will update this in a different way.

PR #572, opened 2019-04-18
Fix ‘cannot read property “internal” of undefined’ bug

tronin:
Hi guys, we use your service in every app and now we have a problem with ```setRelease``` method.

```Sentry.setRelease(`<project_name>@${DeviceInfo.getVersion()}.${DeviceInfo.getBuildNumber()}`);```

If you will try to set release version you catch the error

```ExceptionsManager.js:82 Unhandled JS Exception: Cannot read property 'internal' of undefined```

I fixed that in current PR.


HazAT:
Thanks for the effort of pulling up a PR!
Please upgrade to the new SDK `@sentry/react-native` the problem should be gone there.

PR #573, opened 2019-04-18
Improve TS typings

HenriBeck:
This PR fixes a couple of things:
- Add `types` field to package.json
- Properly export the default export from the main module
- Strictness of some TS types

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,2 +1,5 @@
 'use strict';
 export * from './lib/Sentry';
+import Sentry from './lib/Sentry';
+
+export default Sentry;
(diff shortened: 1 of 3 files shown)


jessestuart:
> Properly export the default export from the main module

THANK YOU this has always bothered me — and seems to be a common source of user error when using this package ([you don't have to look far](https://github.com/expo/sentry-expo/blob/master/index.d.ts) to find erroneous references to `import Sentry from 'react-native-sentry'`).

Just for reference, from the [MDN docs](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#Description):
> The following syntax does not export a default export from the imported module:
> ```es6
> export * from …;
> ```

PR #591, opened 2019-05-27
feat: Unified API + Transport + iOS

HazAT:


src/ios/RNSentry.m
```diff
@@ -1,299 +0,0 @@
...
-}
-
-RCT_EXPORT_METHOD(setLogLevel:(int)level)
-{
-    [SentryClient setLogLevel:[SentryJavaScriptBridgeHelper sentryLogLevelFromJavaScriptLevel:level]];
-}
-
-RCT_EXPORT_METHOD(setTags:(NSDictionary *_Nonnull)tags)
-{
-    SentryClient.sharedClient.tags = [SentryJavaScriptBridgeHelper sanitizeDictionary:tags];
```
jpshelley: Sorry I realize this is months old now @HazAT but it appears this was removed in the new API and therefor impacts react-native ability to report context on native crashes, see #694 - I can push up a PR if this was accidental.

However, a bit confused on how that would work in this new API, seeing as `setTags` is just exported directly from `@sentry/core` - is it possibly to create some sort of proxy so native context is added as well?

HazAT: We will fix this, instead of setting all properties individually we will just persist the Scope.
Keep an eye on the issue.

jpshelley: Thanks @HazAT

Issue #487, opened 2018-10-02
NoSuchKeyException: stacktrace getNullableValue


sibelius:
OS:
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

_Platform:_
- [ ]  iOS
- [x]  Android

Output of `node -v && npm -v && npm ls --prod --depth=0`

```
v10.7.0
6.4.1
├── @expo/react-native-action-sheet@1.1.2
├── @ptomasroos/react-native-multi-slider@0.0.14
├── date-fns@1.29.0
├── emoji-mart-native@0.3.0-beta
├── fbemitter@2.1.1
├── formik@1.3.0
├── hoist-non-react-statics@1.2.0
├── idx@2.4.0
├── intl@1.2.5
├── path-match@1.2.4
├── path-to-regexp@2.4.0
├── prop-types@15.6.2
├── query-string@4.3.4
├── react@16.5.0
├── UNMET PEER DEPENDENCY react-native@0.57.1
├── react-native-action-button@2.8.5
├── react-native-button@2.3.0
├── react-native-camera@1.0.2
├── react-native-camera-roll-picker@1.2.3
├── react-native-checkbox-field@2.0.2
├── react-native-code-push@5.4.2
├── react-native-collapsible@0.13.0
├── react-native-communications@2.2.1
├── react-native-device-info@0.17.4
├── react-native-document-picker@2.1.0
├── react-native-emoji@1.3.1
├── react-native-fabric@0.5.2
├── react-native-firebase@5.0.0
├── react-native-gifted-chat@0.4.3
├── react-native-globalize@1.2.4
├── react-native-i18n@2.0.15
├── react-native-icon-badge@1.1.3
├── react-native-image-crop@1.0.2 invalid
├── react-native-image-resizer@1.0.0
├── react-native-keyboard-aware-scroll-view@0.5.0
├── react-native-keyboard-spacer@0.4.1
├── react-native-linear-gradient@2.4.2
├── react-native-localization@0.1.32
├── react-native-looped-carousel@0.1.13
├── react-native-mime-types@2.2.1
├── react-native-modal-datetime-picker@6.0.0
├── react-native-picker-select@4.4.0
├── react-native-root-toast@3.0.1
├── react-native-sentry@0.39.0
├── react-native-simple-radio-button@2.7.3
├── react-native-slider@0.11.0
├── react-native-splash-screen@3.1.1
├── react-native-svg@6.5.2
[...]

PR #594, opened 2019-06-03
Checking if stacktrace exists before accessing it

MLDimitry:
Ran into this react native error even when not using Sentry. Sentry will crash (and crash your RN app with it) if stacktrace doesn't exist. This adds a check for stacktrace just like the checks being performed in the immediate vicinity of this code.

Related to issue #487


tonyo:
Thanks for the fix 👍

PR #607, opened 2019-06-18
Add `androidx` support.

jgreen01su:
- Replace `android.support` with `androidx` in import.

### DIFF
--- a/android/src/main/java/io/sentry/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/RNSentryModule.java
@@ -3,7 +3,7 @@ package io.sentry;
 import android.content.Context;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
-import android.support.annotation.NonNull;
+import androidx.annotation.NonNull;
 
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.Promise;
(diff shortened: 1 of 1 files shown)


maieonbrix:
Hello :) 

Maybe we can simply remove the annotation ? Like what you did here : https://github.com/itinance/react-native-fs/pull/691

Just to ensure the two-ways compatibility.

jgreen01su:
Thanks @maieonbrix and @se-bastiaan ! I removed the annotation and androidx import.

HazAT:
Thanks for this. We just released beta version of the new version of our SDK which has the fix already in there.
See: https://github.com/getsentry/sentry-react-native/issues/612

If really necessary I can make a patch release.

se-bastiaan:
I think that completely depends on the planning for the v1 of the SDK. I would say that this, as a breaking change, is the perfect moment to only support react-native v0.60 and higher as you already mentioned in #612. However, if v1 takes longer than expected it would be nice to have a patch release. For now, no need imo. We'll have to wait for other packages and their updates to be able to update to RN 0.60 anyways.

muhammadsr:
Can we merge this PR?

HazAT:
I will close this PR since it's outdaten, please migrate / use our new SDK which has support for AndroidX
https://github.com/getsentry/sentry-react-native/issues/612

PR #640, opened 2019-08-22
Remove old rnpm key from package.json

rstims:
Not sure if this was being left in the package file for a reason but here's a PR to take it out. Thanks!


HazAT:
I afraid we can't merge this since we also want to support react native version < 0.60 (that do not support auto linking)

rstims:
@HazAT no worries, I figured there may have been a reason.  Thanks!

PR #688, opened 2019-10-09
fix(breadcrumbs): Fix missing `type`, miscast `status_code` entries in Android breadcrumbs

lobsterkatie:
This PR fixes two small bugs in our handling of Android breadcrumbs.

1) Currently, Android breadcrumbs miss copying over the `type` attribute of breadcrumbs, leading all breadcrumbs to be of type `"default"`. Among other things, this affects their display in the UI:

Without `type`:

image

With `type`:

image

2) Somehow or other, the response status code for XHR requests is getting cast as a double (you can see it in the first screenshot above), which is confusing people. This strips the decimals off so that the status again appears like `200` rather than `200.0`.

### DIFF
--- a/android/src/main/java/io/sentry/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/RNSentryModule.java
@@ -185,6 +185,10 @@ public class RNSentryModule extends ReactContextBaseJavaModule {
                 if (breadcrumb.hasKey("category")) {
                     breadcrumbBuilder.setCategory(breadcrumb.getString("category"));
                 }
+                
+                if (breadcrumb.hasKey("type")) {
+                    breadcrumbBuilder.setType(breadcrumb.getString("type"));
+                }
 
                 try {
                     if (breadcrumb.hasKey("data") && breadcrumb.getMap("data") != null) {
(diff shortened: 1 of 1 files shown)


android/src/main/java/io/sentry/RNSentryModule.java
```diff
@@ -185,13 +185,27 @@ public void sendEvent(ReadableMap event, Promise promise) {
...
                         }
+                        
+                        // in case a `status_code` entry got accidentally stringified as a float
+                        if (newData.containsKey("status_code")) {
+                              String value = newData.get("status_code");
+                              newData.put(
+                                  "status_code",
+                                  value.endsWith(".0") ? value.replace(".0", "") : value
+                              );
+                        }
```
lobsterkatie: I don't know if this is the most elegant way to do this. Open to other suggestions.

bruno-garcia: This works. But ideally we'd just make sure Java doesn't get a stringyfied float in the first place and avoid this altogether.

lobsterkatie: 100% agreed. This is a total band-aid. If someone who is more familiar with the way the JS talks to the Java wants to dig into that and make suggestions or fix the problem at its root, I will not be offended in the least. This just seemed like the most straightforward path to at least _some_ solution. ¯\_(ツ)_/¯

bruno-garcia: Makes sense. I'd say happy to merge. Unless @HazAT thinks otherwise

HazAT:
Thank you!!

agungjk-okadoc:
still occur on 1.0.8
"error: incompatible types: String cannot be converted to Type"

@HazAT

lobsterkatie:
You are correct, @agungjk-okadoc, sorry about that! Fixing now...

lobsterkatie:
@agungjk-okadoc , the fix has been released in `1.0.9`.

PR #693, opened 2019-10-11
fix(types): Fix type mismatch when copying breadcrumbs

lobsterkatie:
In Java breadcrumbs,  [the `type` field](https://github.com/getsentry/sentry-java/blob/4dbeeb8dce08d1a70591a20612e6bafea8a1b415/sentry/src/main/java/io/sentry/event/Breadcrumb.java#L18) is [an enum](https://github.com/getsentry/sentry-java/blob/4dbeeb8dce08d1a70591a20612e6bafea8a1b415/sentry/src/main/java/io/sentry/event/Breadcrumb.java#L87), not [a string](https://github.com/getsentry/sentry-react-native/blob/d2cf9ceaaf99fd99b7737cf07170603c8a7b0ea0/android/src/main/java/io/sentry/RNSentryModule.java#L190). This fixes the type mis-match.


bruno-garcia:
Would be nice to add some tests eh? But since we're moving to the new Android SDK in a month I'm not sure it's worth the effort to setup the infra

lobsterkatie:
> Would be nice to add some tests eh?
@bruno-garcia Ideally yes. Let's coordinate offline on that, if not for this PR then for future ones.

PR #741, opened 2020-01-09
feat: Add local source context via devserver

HazAT:


src/js/integrations/debugsymbolicator.ts
```diff
@@ -121,25 +118,47 @@ export class DebugSymbolicator implements Integration {
...
+            !frame.file.includes("node_modules") &&
+            !frame.file.includes("native code");
+
+          const newFrame: StackFrame = {
+            colno: frame.column,
+            filename: frame.file,
+            function: frame.methodName,
+            in_app: inApp,
+            lineno: inApp ? frame.lineNumber : undefined, // :HACK
+            platform: inApp ? "javascript" : "node" // :HACK
```
lobsterkatie: Given that we're now filtering out both node modules _and_ native code, would it be better to make the default `undefined`? Or to test for `frame.file.includes("native code")`?

src/js/integrations/debugsymbolicator.ts
```diff
@@ -158,4 +177,34 @@ export class DebugSymbolicator implements Integration {
...
+        method: "GET"
+      }
+    );
+
+    const content = await response.text();
+    const lines = content.split("\n");
+
+    const line = Math.max(frame.lineno! - 1, 0);
+    frame.pre_context = lines.slice(line - 6, line);
+    frame.post_context = lines.slice(line, line + 5);
```
lobsterkatie: Maybe I'm reading this wrong, but is there a reason you're grabbing six lines of pre-context but only five lines of post-context? Also, shouldn't the post-context start on the line _after_ the line in question?

HazAT:
I've refactored the code in our code packages since we had this functionality already.
I've also added tests:
https://github.com/getsentry/sentry-javascript/pull/2376

src/js/integrations/debugsymbolicator.ts
HazAT: https://github.com/getsentry/sentry-react-native/pull/741#issuecomment-573024666

lobsterkatie: 👍🏻

PR #773, opened 2020-02-19
fix: Envelope header bytes length

bruno-garcia:
@HazAT it was supposed to be a draft PR but.. This is an attempt (untested) to fix: https://github.com/getsentry/sentry-android/issues/273

The size issue exists because currently, we use the string length which is the number of characters. The envelope spec expects the number of bytes (which in UTF-8 vary depending on the character).

Ideally, we'd implement this in Java (we need it there anyway given the new hybrid spec) and RN will just call in with the event.

Fixes: https://github.com/getsentry/sentry-react-native/issues/775

### DIFF
--- a/android/src/main/java/io/sentry/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/RNSentryModule.java
@@ -147,7 +147,7 @@ public class RNSentryModule extends ReactContextBaseJavaModule {
     }
 
     @ReactMethod
-    public void getStringBytesLength(String str) {
+    public int getStringBytesLength(String str) {
         return str.getBytes("UTF-8").length;
     }
 
--- a/android/src/main/java/io/sentry/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/RNSentryModule.java
@@ -146,6 +146,11 @@ public class RNSentryModule extends ReactContextBaseJavaModule {
         promise.resolve(true);
     }
 
+    @ReactMethod
+    public void getStringBytesLength(String str) {
+        return str.getBytes("UTF-8").length;
+    }
+
     private static PackageInfo getPackageInfo(Context ctx) {
         try {
             return ctx.getPackageManager().getPackageInfo(ctx.getPackageName(), 0);
--- a/android/src/main/java/io/sentry/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/RNSentryModule.java
@@ -15,6 +15,7 @@ import com.facebook.react.module.annotations.ReactModule;
 
 import java.io.File;
 import java.io.FileOutputStream;
+import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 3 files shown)


android/src/main/java/io/sentry/RNSentryModule.java
```diff
@@ -146,6 +146,11 @@ public void captureEnvelope(String envelope, Promise promise) {
         promise.resolve(true);
     }
 
+    @ReactMethod
+    public void getStringBytesLength(String str) {
```
marandaneto: I'd assume that you need a `Promise promise` here

android/src/main/java/io/sentry/RNSentryModule.java
```diff
@@ -146,6 +146,11 @@ public void captureEnvelope(String envelope, Promise promise) {
...
+        return string.getBytes(str)
```
marandaneto: I think it would be something like `str.getBytes("UTF-8").length`

bruno-garcia: string, `string` is also C# btw :)

bruno-garcia: And the build passed lol

android/src/main/java/io/sentry/RNSentryModule.java
bruno-garcia: Didn't add since there's nothing async here, but no idea if it's needed.
there's `setLogLevel` which doesn't take a Promise. Tomorrow when Vienna is awake we can discuss this change properly. Wanted to make a pointer to where the problem is and suggest some fix

marandaneto: yep, got it, but if you don't have a `return` type on the method declaration, it won't work, got your point, will discuss with @HazAT today, thanks for checking it out

android/src/main/java/io/sentry/RNSentryModule.java
marandaneto: lol, true

android/src/main/java/io/sentry/RNSentryModule.java
bruno-garcia: yeah next time I'll pay attention what I'm doing :)

bruno-garcia:
@marandaneto @HazAT regardless how we're fixing this (this PR or other) we need to address this since it's broken for RN users right now.
Can we please discuss this here or on a meeting?
Also I'd appreciate some walk through because I can't even get this to build locally. `npm run build` does `tsc`, Makefile is for publishing, etc

HazAT:
I fixed the code and bumped the SDK version. I will do a release.

HazAT:
Was able to repro it, fixed will do a release.
Thank you all.

PR #754, opened 2020-02-04
fix: Pass the correct sourcemap to sentry-cli

Swatinem:
The way the gradle hook works is very fragile, and it extracted the wrong sourcemap reference
when working with a multi-stage build process (as used for hermes).

Issue #772, opened 2020-02-19
Android sourcemapOutput detection broken?


muxe:
OS:
- [ ]  Windows
- [X]  MacOS
- [ ]  Linux

Platform:
- [ ]  iOS
- [X]  Android

SDK:
- [X]  `@sentry/react-native`
- [ ]  `react-native-sentry`

SDK version: 1.3.0
`react-native` version: 0.59.10

Init Code:

```
Sentry.init({
  dsn: 'https://...@sentry.io/...'
});
```

---
I have following issue:

My bundle is not producing any sourcemaps and then the upload of the sourcemap fails.
I think there is an issue with forcing sourcemap output:
My default bundle command does NOT have a --sourcemap-output in.
It starts here https://github.com/getsentry/sentry-react-native/blob/master/sentry.gradle#L45
Then here https://github.com/getsentry/sentry-react-native/blob/master/sentry.gradle#L212 it tries to get the sourcemapOutput path.
So these lines wont find a sourcemapOutput https://github.com/getsentry/sentry-react-native/blob/master/sentry.gradle#L180-L186 BUT then here 
https://github.com/getsentry/sentry-react-native/blob/master/sentry.gradle#L195-L198 some weird sourcemapOutput is returned.
So the script thinks there was a --sourcemap-output defined and does not force the arg, hence no sourcemap will be generated but it still expects a sourcemap to be there.

Or am I missing something?

**Steps to reproduce:**
- use rn 0.59.10 and sentry 1.3.0

**Actual result:**
- no sourcemap generated

**Expected result:**
- a sourcemap generated

muxe:
https://github.com/getsentry/sentry-react-native/pull/754#issuecomment-588246525

xwartz:
cc @HazAT

HazAT:
Sorry about that, we'll take a look.

marandaneto:
@muxe hey, I've raised a PR #777 could you just copy/paste that part of the code and try this out? thanks :)

muxe:
Thanks. I'll give it a go, but not before monday though.

I am not quite sure, but I think my --bundle-output is set but not the --sourcemap-output.
This should then still return a sourcemapOutput !== null from extractBundleTaskArguments which again leads to no args added here https://github.com/getsentry/sentry-react-native/blob/fix/sourcemaps_folder/sentry.gradle#L220-L224

But as I said this only by looking at code, I would need to run this to see how it behaves.

[...]

PR #777, opened 2020-02-21
fix source maps folder detection

marandaneto:
Fix #754 and #772

### DIFF
--- a/sentry.gradle
+++ b/sentry.gradle
@@ -205,10 +205,10 @@ static extractBundleTaskArguments(cmdArgs, Project project) {
         // react-native < 0.60.1
         def pattern = Pattern.compile("(/|\\\\)intermediates\\1sourcemaps\\1react\\1")
         Matcher matcher = pattern.matcher(sourcemapOutput)
-        // so if its intermediates/sourcemaps/react then it should be generated/sourcemaps/react
+        // if its intermediates/sourcemaps/react then it should be generated/sourcemaps/react
         if (matcher.find()) {
             project.logger.info("sourcemapOutput has the wrong path, let's fix it.")
-            // replacing from bundleOut which is more reliable
+            // replacing from bundleOutput which is more reliable
             sourcemapOutput = bundleOutput.replaceAll("(/|\\\\)generated\\1assets\\1react\\1", "\$1generated\$1sourcemaps\$1react\$1") + ".map"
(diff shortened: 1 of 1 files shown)


sentry.gradle
```diff
@@ -192,9 +192,15 @@ static extractBundleTaskArguments(cmdArgs){
     // In this function here, we only grep through the first `metro` step, which only generates an intermediate sourcemap,
     // which is wrong. We need the final one. Luckily, we can just generate the path from the `bundleOutput`, since
     // the paths seem to be well defined.
-    if (bundleOutput != null) {
+    // if sourcemapOutput is not null, it means react-native passed --sourcemap-output and we are good to go
```
Swatinem: hm, doesn’t it *always* pass that flag?

At least in the version that I developed against, there were two cases:

1. without hermes:
   * metro -> output.map
2. with hermes:
   * metro -> intermediate1
   * hermes -> intermediate2
   * merging the two -> output.map

We always want to get the `output.map`. How? It seems super fragile to grep and manipulate paths of the `metro` command line. Is there no other way to get that output map?

marandaneto: no, it's not always, that's why we check its nullability.
I could not find any other way than `--sourcemap-output` and if it's not there, it's a guessing game, so I check dirs existence.

Swatinem: But in its current state, your patch is incorrect, because in case 2 above, it does pass an explicit sourcemap, which in that case would be `intermediate1`, which is the wrong one.
I would recommend to try different alternatives, check that the file exists, and maybe just fall back to the one you detected via the args.

marandaneto: which version of RN are you testing with? I have a fresh RN setup and Sentry and never got `intermediate1` coming from `--sourcemap-output`

Swatinem: https://github.com/facebook/react-native/blob/25ea35632e18aa41bc753effaf3ad36c1fc84897/react.gradle#L158
I was looking at the `react-native` gradle file all the time. I don’t claim to completely understand it though :-D

marandaneto: @Swatinem yep, me too, even the history of it, but again, which RN version were you working on?

muxe:
I checked locally and can confirm that the current state fixes my issue #772

marandaneto:
@muxe thanks for testing it :)

amogh-jrules:
Have these changes been included in the latest release?

PR #800, opened 2020-03-17
Only if lower than 0.60

wouterds:
On the website it's correctly documented though: https://docs.sentry.io/platforms/react-native/#linking


wouterds:
Nvm 🙈

PR #837, opened 2020-05-08
Sample project

bruno-garcia:
I followed [the ideas on this post](https://medium.com/@charpeni/setting-up-an-example-app-for-your-react-native-library-d940c5cf31e4) on how to add a sample project in the library while being able to change the lib and see the result in the app.

It worked on Android but I'm having a hard time with xcode.
Opening this draft in case it's obvious to anyone.

TODO:
 - [x] Fix `xcworkspace` to build `sample` as well as `@sentry/react-native` so that changes to the native code shows up on the sample app.
 - [ ] Change UI to have some sample buttons


sample/.buckconfig
```diff
@@ -0,0 +1,6 @@
+
+[android]
```
marandaneto, line 2: if we use `Gradle` for Android, we don't need `Buck`

sample/android/app/_BUCK
```diff
@@ -0,0 +1,55 @@
+# To learn about Buck see [Docs](https://buckbuild.com/).
```
marandaneto, line 1: same as https://github.com/getsentry/sentry-react-native/pull/837/files#r422104024

sample/.buckconfig
bruno-garcia: This came up with the react-native init project

sample/android/app/build.gradle
```diff
@@ -0,0 +1,227 @@
...
+
+    debugImplementation("com.facebook.flipper:flipper:${FLIPPER_VERSION}") {
+      exclude group:'com.facebook.fbjni'
+    }
+
+    debugImplementation("com.facebook.flipper:flipper-network-plugin:${FLIPPER_VERSION}") {
+        exclude group:'com.facebook.flipper'
+    }
+
+    debugImplementation("com.facebook.flipper:flipper-fresco-plugin:${FLIPPER_VERSION}") {
```
marandaneto, line 207: guess this won't be any useful, fresco is an image loading lib., we dont need to debug this

sample/android/app/_BUCK
bruno-garcia: Could be an improvement but I'd like to land the working sample in first.
Android actually works already

sample/android/app/build.gradle
bruno-garcia: I believe this is used to hot reload sstuff, it seems this flipper is all over. I'm not sure yet, it's just the template sample.

sample/.buckconfig
marandaneto: yes I know, just afraid to make the sample heavier and even harder to set up

sample/android/app/build.gradle
marandaneto: flipper it is, but not all the modules of flipper, eg the `flipper-network-plugin` makes sense, but not a library loader debugger

sample/.buckconfig
bruno-garcia: I'm afraid of spending more hours "refactoring" this. I'd like to get this working first. We can remove all unecessary code after. I'd focus on adding cocoa 5.0 before that.

sample/android/app/build.gradle
marandaneto: yes yes, but we could clean it up, that's what I mean, happy to merge as it is

bruno-garcia: Thanks

HazAT:
Thank you @bruno-garcia

PR #848, opened 2020-05-12
sentry-cocoa 5.0

bruno-garcia:
To note:

 * Even though not public API in react-native, it was public on the bridge but now removed:
`crashedLastLaunch`

* Now passing the whole dictionary of options which might not be wise. Needs a review. Should we whitelist what can go through? Or perhaps blacklist what we know should only stay at the JS layer?
* Dropping `enableNativeCrashHandling` which means we're always bringing in native crashes. Can we get feedback on this? Is it a real use case where users need only JS crashes?

TODO:
- [x] Pending adding `deviceContexts` to get context added on JS-only errors.
- [ ] `setLogLevel` should be removed. Other SDKs have `DiagnosticLevel` to configure the level when `Debug=true`.
- [ ] Implement `captureEnvelope` to replace `captureEvent`
- [ ] Review files to ignore e.g: `sample/android/.project`, `sample/android/.settings/`

### DIFF
--- a/ios/RNSentry.m
+++ b/ios/RNSentry.m
@@ -38,12 +38,6 @@ RCT_EXPORT_METHOD(startWithDsnString:(NSString * _Nonnull)dsnString
 {
     NSError *error = nil;
     
-    SentryOptions *sentryOptions = [[SentryOptions alloc] initWithDict:options didFailWithError:&error];
-    if (error) {
-        reject(@"SentryReactNative", error.localizedDescription, error);
-        return;
-    }
-    
     SentryBeforeSendEventCallback beforeSend = ^SentryEvent*(SentryEvent *event) {
         // We don't want to send an event after startup that came from a Unhandled JS Exception of react native
         // Because we sent it already before the app crashed.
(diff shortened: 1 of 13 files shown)


bruno-garcia:
Context data is available
image

Even though it's a working solution, we should discuss and go ahead with a new approach for cocoa based on envelopes.

ios/RNSentry.m
```diff
@@ -31,56 +31,70 @@ + (BOOL)requiresMainQueueSetup {
...
+    [SentrySDK configureScope:^(SentryScope * _Nonnull scope) {
+        NSDictionary<NSString *, id> *serializedScope = [scope serialize];
+        // Scope serializes as 'context' instead of 'contexts' as it does for the event.
+        NSDictionary<NSString *, id> *contexts = [serializedScope valueForKey:@"context"];
+#if DEBUG
+        NSData *data = [NSJSONSerialization dataWithJSONObject:contexts options:0 error:nil];
+        NSString *debugContext = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
+        NSLog(@"Contexts: %@", debugContext);
+#endif
+        resolve(contexts);
```
HazAT, line 80: It's not stupid if it works.
And tbh we probably need to do a few of those "hacks" in order to make it work as needed so I am OK

PR #870, opened 2020-05-22
pass enableAutoSessionTracking to android

bruno-garcia:
propagate `enableAutoSessionTracking` from JS to the android sdk


jenshandersson:
@bruno-garcia this works fine but the types are missing in ReactNativeOptions interface https://github.com/getsentry/sentry-react-native/blob/2f1bef799edf57cedde6593645841e2057510925/src/js/backend.ts#L33. Thanks 🙌

Issue #694, opened 2019-10-11
Context values not getting attached to native crashes


lobsterkatie:
OS:
- [ ]  Windows
- [X ]  MacOS
- [ ]  Linux

_Platform:_
- [X ]  iOS
- [ ]  Android

SDK:
- [X ]  `@sentry/react-native` 1.0.6
- [ ]  `react-native-sentry`

`react-native` version: 0.61.2

**Init Code:**

```
Sentry.init({
  dsn: 'https://xxxxxx@sentry.io/xxxxx',
  environment: 'prod',
  attachStacktrace: true,
  debug: true,
});
Sentry.setRelease('friday');

Sentry.configureScope(scope => {
  scope.setTags({dogs: 'yes', cats: 'maybe'});
  scope.setExtras({things: 'stuff'});
  scope.setUser({id: 26, email: 'maisey@dogsrule.com'});
});
```

---
**I have following issue:**

None of the values set on the scope come through on native crashes.

**Steps to reproduce:**
Add context to scope and trigger a native crash.

**Actual result:**
None of the scope data is displayed:

![image](https://user-images.githubusercontent.com/14812505/66687125-88431e80-ec36-11e9-831b-c0795affee81.png)

[no image of extra data to show here because it doesn't come through]

**Expected result:**
Scope data is added to event, as it is for JS events:

![image](https://user-images.githubusercontent.com/14812505/66687192-b9235380-ec36-11e9-9366-438a996dcc71.png)

![image](https://user-images.githubusercontent.com/14812505/66687806-dc4f0280-ec38-11e9-822a-a60a324e1e1b.png)

**EDIT**: Per @HazAT, this applies to breadcrumbs as well.

HazAT:
To add to this, `scope.setContext` or `Sentry.setContext` also currently doesn't work.

anguyen1817:
I'm experiencing the same issue.

halilb:
This is also a huge problem for us. We can't track any tag data on native crashes.

The data coming from `beforeSend` method is also not sent along with the native crash log.

mattijsf:
Same thing for us. We set the user id on the javascript side like this:

```javascript
Sentry.setUser({ id: AppSettings.deviceLogin })
```

But there is a difference between native errors / js errors:
`Sentry.nativeCrash()` = Missing expected user id 👎 
`Sentry.captureException(new Error("JS Error"))` = User id is present and correct 👍 

[...]

PR #902, opened 2020-06-03
feat: Sync RN scope to native

jennmueng:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

Fixes https://github.com/getsentry/sentry-react-native/issues/694

## :scroll: Description

Implemented a new class `ReactNativeScope extends Scope` that calls native scope setters equivalent methods on every call to `setUser`, `addBreadcrumb`, `setExtra`, `setExtras`, `setTag`, `setTags`, `setContext`, `clearBreadcrumbs`.

Syncing the React Native scope to native events would allow users to actually group them together and see all the breadcrumbs from React Native that led to that event in native.

Created extensive scope assignment tests in the sample app.

## :pencil: Checklist

- [x] All tests passing


CHANGELOG.md
```diff
@@ -370,7 +371,7 @@ To activate it set:
 
 ```js
 Sentry.config("___DSN___", {
-  deactivateStacktraceMerging: false,
```
bruno-garcia, line 373: Please try to avoid these unrelated changes to the PR. This way the reviewers can focus on what really matters here.
Feel free to open a new PR for any styling or refactor changes.

android/src/main/java/io/sentry/RNSentryModule.java
```diff
@@ -108,6 +116,27 @@ public void startWithOptions(final ReadableMap rnOptions, Promise promise) {
                 } catch (Exception e) {
                     // We do nothing
                 }
+
+                // Add on the correct event.origin tag.
+                // it needs to be here so we can determine where it originated from.
+                SdkVersion sdkVersion = event.getSdk();
```
bruno-garcia, line 122: We could merge this on its own PR already, separately from the scope sync.

android/src/main/java/io/sentry/RNSentryModule.java
```diff
@@ -190,4 +219,114 @@ private Level logLevel(int level) {
                 return Level.OFF;
         }
     }
+
+    @ReactMethod
+    public void setUser(final ReadableMap user) {
+        if (user == null) {
+            Sentry.setUser(null);
```
bruno-garcia, line 226: Can we use `configureScope` instead? Ideally we'd drop these top level methods eventually.

```suggestion
            Sentry.configureScope(s -> { s.setUser(null) });
```

Same for other `Sentry.setUser` calls below

android/src/main/java/io/sentry/RNSentryModule.java
```diff
@@ -190,4 +219,114 @@ private Level logLevel(int level) {
...
+        } else {
+            User userInstance = new User();
+
+            // only support these defined keys for now.
+            if (user.hasKey("email")) {
+                userInstance.setEmail(user.getString("email"));
+            } else {
+                userInstance.setEmail(null);
```
bruno-garcia, line 234: why do we explicitly set `null` here when it's a newly created instance?
Same for the following lines

android/src/main/java/io/sentry/RNSentryModule.java
```diff
@@ -190,4 +219,114 @@ private Level logLevel(int level) {
...
     }
+
+    @ReactMethod
+    public void setUser(final ReadableMap user) {
+        if (user == null) {
+            Sentry.setUser(null);
+        } else {
+            User userInstance = new User();
+
+            // only support these defined keys for now.
```
bruno-garcia, line 230: We'd need to have this clear in our documentation.

This could be a source of support requests where user use our public documented API such as setting `user.data` and not getting it through while the other fields go through correctly

android/src/main/java/io/sentry/RNSentryModule.java
```diff
@@ -190,4 +219,114 @@ private Level logLevel(int level) {
...
+                case "error":
+                    breadcrumbInstance.setLevel(SentryLevel.ERROR);
+                    break;
+                default:
+                    breadcrumbInstance.setLevel(SentryLevel.ERROR);
+                    break;
+            }
+        }
+
+        if (breadcrumb.hasKey("data")) {
```
bruno-garcia, line 298: Can we do this for `user` too?

android/src/main/java/io/sentry/RNSentryModule.java
```diff
@@ -190,4 +219,114 @@ private Level logLevel(int level) {
...
+            ReadableMap data = breadcrumb.getMap("data");
+            ReadableMapKeySetIterator it = data.keySetIterator();
+            while (it.hasNextKey()) {
+                String key = it.nextKey();
+                String value = data.getString(key);
```
bruno-garcia, line 303: We can't assume this is `String`. We need to test the type somehow or try/catch this straight to string and log out the errors.
Users can add complex objects as values or pretty much anything. We could extend this to test for `Serializable` or other types in the future but for now just please gracefully handle the fact it might not be `String`

src/js/sdk.ts
```diff
@@ -63,7 +67,7 @@ export function init(
                 : `${appPrefix}/${frame.filename}`;
           }
           return frame;
-        },
```
bruno-garcia, line 66: Is this a `.editorconfig` or else that automatically strips these when the file is saved? We'd ideally share the linter/formatter rules etc so these things don't change on each PR.

android/src/main/java/io/sentry/RNSentryModule.java
jennmueng: Ah, in the javascript SDK I was looking at user at we allow any key, 
```
export interface User {
  [key: string]: any;
  id?: string;
  ip_address?: string;
  email?: string;
  username?: string;
}
```
So I wasn't sure at first how to pass this to native, but I just looked at `User` in Android and it looks like there is a `setOthers` method, I can take every other key that isn't in the defined 4 keys and pass it to others then.

android/src/main/java/io/sentry/RNSentryModule.java
jennmueng: All first-level values in data that aren't strings are actually passed through `JSON.stringify` in the `wrapper.ts` so they would be strings before they get here. I can do the same for all custom values passed to user then.

```js
const stringifiedData: { [key: string]: any } = {};
    if (typeof breadcrumb.data !== "undefined") {
      Object.keys(breadcrumb.data).forEach((dataKey) => {
        if (typeof breadcrumb.data !== "undefined") {
          const value = breadcrumb.data[dataKey];
          stringifiedData[dataKey] =
            typeof value === "string" ? value : JSON.stringify(value);
        }
      });
    }

    // tslint:disable-next-line: no-unsafe-any
    return RNSentry.addBreadcrumb({
      ...breadcrumb,
      data: stringifiedData
    });
```

src/js/sdk.ts
jennmueng: I have prettier on my VSCode that formats on save, I actually made a config that made it leave trailing commas alone but I think I accidentally removed it on a recent rebase I did, my bad!

android/src/main/java/io/sentry/RNSentryModule.java
jennmueng: 🙃

jennmueng: Thank you for telling me that completely went over my head that it would be null by default.

android/src/main/java/io/sentry/RNSentryModule.java
jennmueng: I think this is merged, just didn't do a rebase on the latest master yet.

jennmueng: Sorry about all these issues I'm not used to making PRs

jennmueng: Thanks for all the comments; I will get everything resolved tomorrow!

ios/RNSentry.m
```diff
@@ -132,6 +132,103 @@ + (BOOL)requiresMainQueueSetup {
...
+        [scope setTagValue:value forKey:key];
+    }];
+}
+
+// RCT_EXPORT_METHOD(setFingerprint:(NSArray<NSString *> *)fingerprint)
+// {
+//     [SentrySDK configureScope:^(SentryScope * _Nonnull scope) {
+//         [scope setFingerprint:fingerprint];
+//     }];
+// }
```
HazAT, line 230:
```suggestion
```

src/js/scope.ts
```diff
@@ -0,0 +1,124 @@
...
+import { timestampWithMs } from "@sentry/utils";
+
+import { NATIVE } from "./wrapper";
+
+export class ReactNativeScope extends Scope {
+  /**
+   * @inheritDoc
+   */
+  public setUser(user: User | null): this {
+    this._user = user || {};
```
HazAT, line 12: can't you just call `super.setUser` always? same for other functions

src/js/scope.ts
```diff
@@ -0,0 +1,77 @@
...
+  public setUser(user: User | null): this {
+    NATIVE.setUser(user);
+    return super.setUser(user);
+  }
+
+  /**
+   * @inheritDoc
+   */
+  public setTag(key: string, value: string): this {
+    // Stringify the value as user could pass other types such as a number which would crash native.
```
HazAT, line 19:
```suggestion
```
I think this can go. Moved to native wrapper

src/js/scope.ts
```diff
@@ -0,0 +1,77 @@
...
+    return super.setTag(key, value);
+  }
+
+  /**
+   * @inheritDoc
+   */
+  public setTags(tags: { [key: string]: string }): this {
+    // As native only has setTag, we just loop through each tag key.
+    Object.keys(tags).forEach((key) => {
+      // Stringify the value as user could pass other types such as a number which would crash native.
```
HazAT, line 30:
```suggestion
```

src/js/wrapper.ts
```diff
@@ -1,4 +1,4 @@
-import { Event, Response } from "@sentry/types";
+import { Event, Response, User, Breadcrumb } from "@sentry/types";
```
HazAT, line 1: Linter should complain that this needs to be alphabetically 🤔

cruzach:
I think this may have introduced some breaking changes- should the [`enableNative`](https://github.com/getsentry/sentry-react-native/blob/7fb3d4c06fca5fa42200c4a73da47613cff00479/src/js/backend.ts#L21) setting be checked before trying to call into these native methods?

jennmueng:
@cruzach Yes it does look like not checking for `enableNative` is a breaking bug. Will see if I can get something out ASAP.

Issue #911, opened 2020-06-04
Unhandled JS Errors are not registered in iOS


gramotei:
**Platform:**
- [x]  iOS
- [ ]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 1.4.2

**`react-native` version:** 0.62.1

**Are you using Expo?**
- [ ]  Yes
- [x]  No

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)
- [ ]  on-premise

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  dsn:
    'https://14....@o74....ingest.sentry.io/5.....',
  debug: true,
});
```

---
**I have following issue:**

Can't get Sentry to report unhandled JS crashes

**Steps to reproduce:**
- Install & init Sentry
- make React Native build and put to TestFlight (Archive/Release)
- throw Error in on of Screens

**Actual result:**

After getting to screen with Error the app crashes immediately without report to be sent

**Expected result:**

App is sending the report before termination

dehlen:
Also using 1.4.2 and am seeing this behavior. Downgrade to 1.4.1 did not resolve the issue, however 1.3.7 did work for me as expected. Any ideas for a quick fix? We would love to upgrade to the latest version as soon as possible. If you need any further information let me know :)

jennmueng:
Hmm, that's very interesting. I'm not sure why `1.4.x` wouldn't work for you while `1.3.7` does. When you were upgrading, did you ensure the installation is all correct (`pod install` and everything, along with maybe trying a full `node_modules` delete and reinstall?).

Does it work as intended when `Sentry.captureException()` or `Sentry.captureMessage()` is called directly?

dehlen:
Pretty sure installation is all fine. I am testing with our CI build which always is a clean build. Also it seems to work just fine for the Android app. I'll check whether it works correctly when calling the two methods from above correctly.

dehlen:
Can confirm that calling Sentry.captureException() does indeed report the correct error event and that I can see the stack trace.

gramotei:
> Hmm, that's very interesting. I'm not sure why `1.4.x` wouldn't work for you while `1.3.7` does. When you were upgrading, did you ensure the installation is all correct (`pod install` and everything, along with maybe trying a full `node_modules` delete and reinstall?).
> 
[...]

PR #917, opened 2020-06-05
fix: We only store the event in release mode

HazAT:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

We only want to store "fatal" events in Release mode to make sure we use fewer resources. We can't rely on always sending it. That means after a restart we will pick up the event and send it.

Potential fix for https://github.com/getsentry/sentry-react-native/issues/911 in combination with https://github.com/getsentry/sentry-cocoa/pull/571
## :pencil: Checklist

- [x] All tests passing

### DIFF
--- a/ios/RNSentry.m
+++ b/ios/RNSentry.m
@@ -119,17 +119,27 @@ RCT_EXPORT_METHOD(sendEvent:(NSDictionary * _Nonnull)event
                   resolve:(RCTPromiseResolveBlock)resolve
                   rejecter:(RCTPromiseRejectBlock)reject)
 {
-    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0ul), ^{
-        if ([NSJSONSerialization isValidJSONObject:event]) {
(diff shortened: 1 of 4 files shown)


ios/RNSentry.m
```diff
@@ -119,17 +119,27 @@ + (BOOL)requiresMainQueueSetup {
...
+                                                           options:0
+                                                             error:nil];
+        
+        SentryEvent *sentryEvent = [[SentryEvent alloc] initWithJSON:jsonData];
+#if DEBUG
+        [SentrySDK captureEvent:sentryEvent];
+#else
+        // We check this against the dictionary
+        if ([event[@"level"] isEqualToString:@"fatal"]) {
+            // In release for fatal crashes the only thing we do is store the event to send it after restart
```
bruno-garcia, line 133:
```suggestion
            // captureEvent queues the event for submission, so storing to disk happens asynchronously
            // We need to make sure the event is written to disk before resolving the promise.
            // This could be replaced by SentrySDK.flush() when available.
```

PR #919, opened 2020-06-05
fix: Fix Native Wrapper not checking enableNative setting

jennmueng:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Fix the native wrapper not checking for `enableNative` option in the init options. Also does not display the warning to use `enableNative: false` if it is already set to false.

Addresses https://github.com/expo/sentry-expo/issues/112#issuecomment-639845260
Also #920

Ran on both iOS and Android simulators that no warnings/errors appear when setting scope, or any of the functions in the sample app with either `enableNative:false` and `enableNative:true`. All events show up on sentry.io other than the native crash which is disabled.

## :pencil: Checklist

- [x] All tests passing

### DIFF
--- a/src/js/wrapper.ts
+++ b/src/js/wrapper.ts
@@ -1,5 +1,5 @@
 import { Breadcrumb, Event, Response, User } from "@sentry/types";
-import { SentryError } from "@sentry/utils";
+import { logger, SentryError } from "@sentry/utils";
 import { NativeModules, Platform } from "react-native";
 
 import { ReactNativeOptions } from "./backend";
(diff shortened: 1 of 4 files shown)


jennmueng:
Fixed the wrapper tests that were failing the CI build and also added some tested for disabled native.

src/js/wrapper.ts
```diff
@@ -133,26 +149,28 @@ export const NATIVE = {
    * @param value string
    */
   setUser(user: User | null): void {
-    if (!this.isNativeClientAvailable()) {
-      throw this._NativeClientError;
-    }
+    if (this.enableNative) {
```
HazAT, line 152: Instead of wrapping everything in an `if` and therefore intend all the code below
you could do
```
if (!this.enableNative) {
  return; // or return Promise.reject(new SentryError("..."));
}
```

jennmueng:
Also moved the native disabling functionality to `startWithOptions` as it should be handled there instead of with an if statement in `backend.ts` as before. We also call `logger.warn` if native is disabled for debug purposes.

PR #945, opened 2020-06-30
test: Add basic end-to-end tests workflow

jennmueng:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Adds end to end tests that run on every PR and every push to master. Currently it only tests that the events sent from `captureMessage` and `captureException` are present on the Sentry API. The unhandled exception test is currently left out for now as on iOS the app crashes instantly with no time for the event id to be fetched by the tests, will look for a solution for this as soon as possible.

We should look to adding more tests to test the properties of the events later on. 

There seems to be a [weird issue on production](https://sentry.slack.com/archives/G011TNTMZ99/p1597311491001100) where events sent from this workflow take 10-20 minutes to process, even though running the exact same tests or sending events from the sample app locally take no time. Currently the tests seem to take around 20-40 minutes with caching enabled, we might need to switch to using relay instead of production if we want faster tests.

## :pencil: Checklist

- [x] All tests passing


sample/__tests__/initiale2e.test.js
```diff
@@ -0,0 +1,144 @@
...
+describe('End to end tests for common events', () => {
+  test('captureMessage', async () => {
+    expect(await driver.hasElementByAccessibilityId('captureMessage')).toBe(
+      true,
+    );
+
+    const element = await driver.elementByAccessibilityId('captureMessage');
+    await element.click();
+
+    await driver.sleep(500);
```
jennmueng, line 87: These sleeps are here to wait for the event id to show on the ui

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,108 @@
...
+      - name: Start Appium Server
+        run: yarn run appium --log-timestamp --log-no-colors --allow-insecure chromedriver_autodownload > appium.log &
+        working-directory: ./sample
+      - name: Run Android Emulator
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: 29
+          script: cd sample && react-native run-android --variant=release && curl --output /dev/null --silent --head --fail http://127.0.0.1:4723/wd/hub/sessions && yarn test
+      - name: Upload Appium logs
+        if: ${{ always() }}
```
bruno-garcia, line 49: Would be good to have a comment to clarify why this condition exists

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,108 @@
...
+            sample/ios/Pods
+            ~/Library/Developer/Xcode/DerivedData
+          key: ${{ runner.os }}-${{ hashFiles('sample/ios/Podfile.lock') }}
+      - name: Install Dependencies
+        if: steps.deps-cache.outputs['cache-hit'] != 'true'
+        run: yarn install
+      - name: Install Sample Dependencies
+        if: steps.deps-cache.outputs['cache-hit'] != 'true'
+        run: yarn install
+        working-directory: ./sample
```
bruno-garcia, line 83: At first glance looks like `yarn install` twice on the same folder (this step and above). Maybe flip these?

```suggestion
        working-directory: ./sample
        run: yarn install
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,108 @@
...
+      - name: Build and run iOS emulator
+        run: react-native run-ios --configuration Release
+        working-directory: ./sample
+      - name: Start Appium Server
+        run: yarn run appium --log-timestamp --log-no-colors --allow-insecure chromedriver_autodownload > appium.log &
+        working-directory: ./sample
+      - name: Poll until Appium Server starts
+        run: |
+          until $(curl --output /dev/null --silent --head --fail http://127.0.0.1:4723/wd/hub/sessions); do
+            sleep 5
```
bruno-garcia, line 98: Android doesn't need pooling?
I'm afraid if it buggies up, we'll end up here until GH Actions kills our test.
What about we try up add a counter for each interation here and bail if the server doesn't come up in X seconds?
Maybe worth adding a script to a folder in the repo and executing it?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,108 @@
...
+          done
+      - name: Run Tests
+        run: yarn test
+        working-directory: ./sample
+      - name: Upload Appium logs
+        if: ${{ always() }}
```
bruno-garcia, line 104: Not sure what this does, maybe we get around without `if` altogether here?

sample/test/e2e.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  //   const eventIdElement = await driver.elementByAccessibilityId('eventId');
+  //   const eventId = await eventIdElement.text();
+
+  //   await driver.sleep(10000);
+
+  //   const sentryEvent = await fetchEvent(eventId);
+
+  //   expect(sentryEvent.eventID).toMatch(eventId);
+  // });
```
bruno-garcia, line 111: Deadcode spotted:

```suggestion
```

sample/utils/fetchEvent.ts
```diff
@@ -0,0 +1,56 @@
...
+        if (retries < 20) {
+          setTimeout(() => {
+            retries++;
+            console.log(`Retrying api request. Retry number: ${retries}`);
+            resolve(
+              fetch(request)
+                .then((res) => res.json())
+                .then(retryer),
+            );
+          }, 30000);
```
bruno-garcia, line 39: Maybe lift this value to a const somewhere above? It'll be easier to spot/tune if we need.
Would be nice to give it a name too like `waitTimeoutMs` or something.

.github/workflows/e2e.yml
jennmueng: Usually it takes around ~5 seconds to start, because on Android the appium server is started before the build so it's assumed to have started by the time the build is done, while on iOS it's started right before the test. I think I'll move it to start before the build on iOS as well for consistency

.github/workflows/e2e.yml
jennmueng: Got it, it's there so it uploads the logs regardless of whether the previous steps failed or not. Without it, it wouldn't upload the logs if anything failed.

.github/workflows/e2e.yml
jennmueng: - Explained in another reply

jennmueng:
Thanks for the review @bruno-garcia 🎉 committed everything in.

PR #991, opened 2020-07-23
fix: Attach sdk field to envelope header instead of sdk_info

jennmueng:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Fixes incorrect `sdk_info` spec and bump android to `2.2.2` so it actually is used.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/wrapper.ts
+++ b/src/js/wrapper.ts
@@ -25,7 +25,7 @@ export const NATIVE = {
     if (NATIVE.platform === "android") {
       const header = JSON.stringify({
         event_id: event.event_id,
-        sdk_info: event.sdk,
+        sdk: event.sdk,
       });
 
       (event as any).message = {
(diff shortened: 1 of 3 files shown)


src/js/wrapper.ts
```diff
@@ -25,7 +25,7 @@ export const NATIVE = {
     if (NATIVE.platform === "android") {
       const header = JSON.stringify({
         event_id: event.event_id,
-        sdk_info: event.sdk,
```
marandaneto, line 28: H: more like a question, what's the data structure of `event.sdk`? were you using the right object but the wrong key only? we attached the `sdk_info` previously. which is a different data structure of `sdk`

jennmueng: Yep, I just attached it to the wrong key but used the correct sdk object as `event.sdk` from the JS sdk matches this spec: https://develop.sentry.dev/sdk/event-payloads/sdk/

PR #993, opened 2020-07-24
fix: don't initiate DeviceContext if enableNative is false

cruzach:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
`DeviceContext` calls into native, which can result in extraneous error logs if the user has already stated they don't want to enable the native SDK. 
I also silenced the console warning if `enableNativeNagger` is false

This is an issue for our integration of `sentry-expo` https://github.com/expo/sentry-expo/issues/120 

Ran `yarn test` and tested in an expo project with sentry, no more bad logs! 🎉 

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Unreleased
 
 - fix: Use `LogBox` instead of `YellowBox` if possible. #989
+- fix: Don't add `DeviceContext` default integration if `enableNative` is set to `false`. #993
+- fix: Don't log "Native Sentry SDK is disabled" if `enableNativeNagger` is set to `false`. #993
 
 ## 1.6.3
 
--- a/src/js/sdk.ts
+++ b/src/js/sdk.ts
@@ -65,20 +65,15 @@ export function init(
             }
           }
           return frame;
-        }
-      }),
-      new DeviceContext()
+        },
+      })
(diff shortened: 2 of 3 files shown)


luizjr:
Congratulations on the correction.
I would like to know when it will be available in the repository:
https://github.com/expo/sentry-expo

lukebars:
Looks like you have to specify `enableNative: true`, else it will be false now?

cruzach:
@lukebars I don't think so. It should default to `true`

lukebars:
@cruzach hmm, I've tried to send example crash to Sentry without specifying `enableNative: true` and DeviceContext was completely empty...

cruzach:
Oh you're right, I assigned default options **after** that deviceContext check 🤦 

In sdk.ts: 

```
if (options.enableNative) {
      options.defaultIntegrations.push(new DeviceContext());
}
options = assignDefaultOptions(options);
```

should be 

```
options = assignDefaultOptions(options);
if (options.enableNative) {
      options.defaultIntegrations.push(new DeviceContext());
}
```

bruno-garcia:
Looks like we have no tests for that eh :/

PR #1007, opened 2020-07-29
fix: assign default options before enableNative check

cruzach:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
the changes i made in https://github.com/getsentry/sentry-react-native/pull/993 result in `deviceContext` not initializing by default, since the default `enableNative:true` isn't set until right after

https://github.com/getsentry/sentry-react-native/pull/993#issuecomment-665649047

existing tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/sdk.ts
+++ b/src/js/sdk.ts
@@ -71,10 +71,10 @@ export function init(
         },
       })
     );
+    options = assignDefaultOptions(options);
     if (options.enableNative) {
       options.defaultIntegrations.push(new DeviceContext());
     }
-    options = assignDefaultOptions(options);
   }
 
   // tslint:enable: strict-comparisons
--- a/src/js/sdk.ts
+++ b/src/js/sdk.ts
@@ -38,6 +38,10 @@ export function init(
 ): void {
   const reactNativeHub = new Hub(undefined, new ReactNativeScope());
   makeMain(reactNativeHub);
+  options = {
+    ...DEFAULT_OPTIONS,
+    ...options
+  }
 
   // tslint:disable: strict-comparisons
   if (options.defaultIntegrations === undefined) {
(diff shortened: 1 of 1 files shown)


bruno-garcia:
Note sure how CI breaks:
https://travis-ci.com/github/getsentry/sentry-react-native/jobs/366223716#L274

This change is unrelated

jennmueng:
> Note sure how CI breaks:
> https://travis-ci.com/github/getsentry/sentry-react-native/jobs/366223716#L274
> 
> This change is unrelated

This actually breaks because typescript can't be sure that `defaultIntegrations` isn't reassigned by the function call.

src/js/sdk.ts
```diff
@@ -71,10 +71,10 @@ export function init(
         },
       })
     );
+    options = assignDefaultOptions(options);
```
jennmueng, line 74:
```suggestion
    options = {
        ...DEFAULT_OPTIONS,
        ...options
    }
```
Instead of using a function call, I would do this and define the `DEFAULT_OPTIONS` as a constant object.

jennmueng: This file needs a massive refactor/clean now that I look at it, but I'll take care it later.

cruzach: good idea 👍 
addressed in the [most recent commit](https://github.com/getsentry/sentry-react-native/pull/1007/commits/d6e81a5607fdf29f66921d5ec69a4ee4a3f58fc0)

jennmueng: Oh the build is failing still, this would be fixed if you move the default options setting code before the if statement that sets the default integrations above it. Otherwise everything else looks good.

cruzach: gotchya, fixed that & the build is passing now

PR #1009, opened 2020-07-31
fix: Use the release and dist set in init options over native release

jennmueng:

- [x] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

The `release` and `dist` set in the init options are now used over the native build's release. However we keep `setRelease` and `setDist` as legacy methods to support users who currently use it along with users who use `codePush.getMetadata` to set their release later in the session rather than at the start. We will migrate the documentation to prefer setting the release and dist in init options over using `setRelease` and `setDist` and list the downside of using them.

Using `setRelease` and `setDist` does not set the release on the native Android/Cocoa SDKs, causing release health and sessions to be attributed to the build's release. This is not ideal for apps that use custom release versions such as apps with CodePushed bundles. 

Wrote `release.test.ts`

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Will be followed up with a docs pr

### DIFF
--- a/src/js/sdk.ts
+++ b/src/js/sdk.ts
@@ -100,14 +100,16 @@ export function init(
 }
 
 /**
- * Sets the release on the event.
+ * Legacy. Sets the release on the event.
+ * NOTE: Does not set the release on sessions.
  */
 export function setRelease(release: string): void {
   setExtra("__sentry_release", release);
 }
 
 /**
- * Sets the dist on the event.
+ * Legacy. Sets the dist on the event.
+ * NOTE: Does not set the dist on sessions.
  */
 export function setDist(dist: string): void {
   setExtra("__sentry_dist", dist);
--- a/src/js/sdk.ts
+++ b/src/js/sdk.ts
@@ -100,16 +100,18 @@ export function init(
 }
 
 /**
- * Legacy. Sets the release on the event.
+ * Deprecated. Sets the release on the event.
  * NOTE: Does not set the release on sessions.
+ * @deprecated
  */
 export function setRelease(release: string): void {
   setExtra("__sentry_release", release);
 }
 
 /**
- * Legacy. Sets the dist on the event.
+ * Deprecated. Sets the dist on the event.
  * NOTE: Does not set the dist on sessions.
+ * @deprecated
  */
 export function setDist(dist: string): void {
   setExtra("__sentry_dist", dist);
--- a/src/js/integrations/release.ts
+++ b/src/js/integrations/release.ts
@@ -34,12 +34,22 @@ export class Release implements Integration {
         // Something went wrong, we just continue
       }
 
-      // If __sentry_release or __sentry_dist it should be stronger because the user set it
-      if (event.extra && event.extra.__sentry_release) {
+      const options = getCurrentHub().getClient()?.getOptions();
+
+      /*
+        __sentry_release and __sentry_dist is set by the user with setRelease and setDist. If this is used then this is the strongest.
+        Otherwise we check for the release and dist in the options passed on init, as this is stronger than the release/dist from the native build.
+      */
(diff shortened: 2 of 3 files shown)


src/js/sdk.ts
```diff
@@ -100,14 +100,16 @@ export function init(
 }
 
 /**
- * Sets the release on the event.
+ * Legacy. Sets the release on the event.
```
marandaneto, line 103: L: is adding a doc comment the only way of making it legacy/deprecated? in Java, we have the `@Deprecated` annotation.

jennmueng: There is the `@deprecated` tag, but I couldn't find anything about a `@legacy` tag

marandaneto: mm but what's the meaning of legacy here? do we intend to remove it in the future in favor of options.setRelease or what? because if legacy means, we are keeping it for compatibility but there's another way to do it, I see that being `@deprecated` then

jennmueng: Hmm I agree then, I'll change it to `@deprecated`!

cruzach:
Quick question about this- I'd like to be able un-set the `dist` in the init options. I'm hitting some weird behavior, here are the steps:

- build the native app, Sentry auto-sets the release to the iOS build number & the android version code and uploads sourcemaps, errors in sentry dashboard find sourcemaps correctly 👍 
- send out an over the air update (similar to codepush), **without** setting a distribution for the release
  - errors in sentry dashboard for **android only** cannot find sourcemaps (ios works fine)
  - I guess anroid falls back to the previously set value? not sure

I think this is occurring because on android, the new release (after the OTA update) do not have a `dist` set, so we see "sourcemaps could not be found". I tried passing an empty string to `dist` in the init options, but then i get this error:

Screen Shot 2020-08-04 at 9 44 14 AM

jennmueng:
@cruzach Hmm I see, if you do not have a dist set it would fall back to the value set in the native build from Android/iOS. I will add in functionality to be able to un-set the `dist` by passing `null`.

jennmueng:
@cruzach Sorry for the late reply, we had our Hackweek at Sentry last week.

Ok so for source maps to be parsed both the `dist` and the `release` needs to exactly match on the event and the source map. So if the event does not have a `dist` set then the source map must also not have a `dist` set for them to correctly match.

But, on the topic of un-setting a `dist`, I've discussed this internally with others at Sentry and we should not allow an empty `dist` or `release` to be set as this would cause Release Health to be reported incorrectly. For OTA/codepush updates I suggest the method of setting both the `release` and `dist` together with each update and uploading a source map for each OTA update with the matching values, although this method could have the side effect of native errors not having the correct dsym due to the mismatch. Another solution (although extremely tedious) would be to have a source map for each `release` and `dist` combination

jennmueng:
@danmaas That method works as well but this means that if you set the release and dist in `Sentry.init` it would have to wait for the promise to resolve and any errors that happens before that will not be caught by Sentry. If you use `Sentry.setRelease` and `Sentry.setDist` those methods are legacy and will not be sent down to the native Android/iOS SDKs and Release Health will not be correctly reported.


A good implementation without these side effects is what I use for my own app of just setting the `release` and `dist` to some value such as the version from `package.json`. This way you don't need to wait for a promise and the version is packaged with the js bundle.

```js
import { version } from "package.json"

Sentry.init({
    ...
    release: version,
    dist: version
});
```
or you can store it in something like `app.json`
```js
import { release, dist } from "app.json"

Sentry.init({
    ...
    release,
    dist
});
```

cruzach:
@jennmueng thank you so much! using the `version` form package.json is exactly what I decided to go with, really appreciate your help here 😄  Hope the hackweek went well!

PR #1019, opened 2020-08-10
meta: Move from Travis CI to Github Actions

jennmueng:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


.github/workflows/test.yml
```diff
@@ -0,0 +1,102 @@
+name: Build & Test
+
+on:
+  push:
+    branches: [master]
+  pull_request:
```
marandaneto, line 6:
```suggestion
  pull_request:
    branches:
      - '*'
```

its just more "explicit"

.github/workflows/test.yml
```diff
@@ -0,0 +1,102 @@
...
+
+jobs:
+  job_test:
+    name: Test
+    runs-on: ubuntu-latest
```
marandaneto, line 11: does it compile/test for iOS even being ubuntu? doesn't need it to be macos?

.github/workflows/test.yml
```diff
@@ -0,0 +1,102 @@
+name: Build & Test
```
marandaneto, line 1: name is `Build & Test` but file is called `test.yml`, I'd rename the file though

.github/workflows/test.yml
marandaneto: ps: this only makes sense if you want to run those jobs in all the PRs, otherwise I'd remove the `pull_request` altogether

.github/workflows/test.yml
jennmueng: Ah so the Bundle step only builds the JS code bundle and not building the Java/Obj-C code. I was considering adding that as a test to this too but I think it would be redundant with the e2e tests already testing that.

jennmueng:
Thanks for the review and suggestions @marandaneto!

Issue #1055, opened 2020-09-06
Breadcrumbs for a native crash have log statements logged as error


koenpunt:
**OS:**
- [ ]  Windows
- [x]  macOS
- [ ]  Linux

**Platform:**
- [x]  iOS
- [ ]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 5.2.2 (RNSentry 1.7.2)

**`react-native` version:** 0.62.2

**Are you using Expo?**
- [ ]  Yes
- [x]  No
---
**I have following issue:**

Breadcrumbs for a native crash have log statements logged as error

![image](https://user-images.githubusercontent.com/351038/92328136-ef43a000-f05e-11ea-81a6-6b11639b9682.png)

I've tracked this down to [`RNSentry.addBreadcrumb`](https://github.com/getsentry/sentry-react-native/blob/2e957df8d6fa105f3f61ef805a2b83095c2ad02f/ios/RNSentry.m#L172) being called with a `level` of `"log"`, but that's not a case handled there, so it defaults to `kSentryLevelError`.

I've searched in this repo for the string `"log"`, but wasn't able to find anything, so it think it originates from one of the Sentry JS modules, that automatically forwards `console.log` invocations.

PR #1063, opened 2020-09-09
fix: Don't send over Log and Critical levels over native bridge

jennmueng:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Adds a `_processLevel` method to the native wrapper to convert `Severity.Log` and `Severity.Critical` levels to `Severity.Debug` and `Severity.Fatal` respectively. This method is called in `sendEvent` and `addBreadcrumb` in the native wrapper.

This is to handle this issue for now until we remove these levels from the javascript SDK. This also means that we can remove `Log` and `Critical` from android, @marandaneto.

https://github.com/getsentry/sentry-react-native/issues/1055

Added tests for `_processLevel`

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/wrapper.ts
+++ b/src/js/wrapper.ts
@@ -1,5 +1,5 @@
 /* eslint-disable max-lines */
-import { Breadcrumb, Event, Response, User } from "@sentry/types";
+import { Breadcrumb, Event, Response, Severity, User } from "@sentry/types";
 import { logger, SentryError } from "@sentry/utils";
 import { NativeModules, Platform } from "react-native";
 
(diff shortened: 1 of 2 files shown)


bruno-garcia (approved):
Nice! Thanks

src/js/wrapper.ts
```diff
@@ -315,6 +322,22 @@ export const NATIVE = {
...
+   * @param level
+   * @returns More widely supported Severity level strings
+   */
+  _processLevel(level: Severity): Severity {
+    if (level === Severity.Critical) {
+      return Severity.Fatal;
+    }
+    if (level === Severity.Log) {
+      return Severity.Debug;
+    }
```
marandaneto, line 336:
```suggestion
    if (level === Severity.Log) {
      return Severity.Debug;
    }
    if (level === Severity.Critical) {
      return Severity.Fatal;
    }
```

marandaneto: I believe its more common to have the `log` instead of `critical`, so we spare a comparison in most of the cases

marandaneto (approved):
thanks for doing this.
left a comment, but other than that LGTM

PR #1068, opened 2020-09-11
feat: Support RN macOS

bruno-garcia:


jennmueng:
@bruno-garcia We should check what is an appropriate version to use as the minimum requirement

bruno-garcia:
@jennmueng I defined the min version the Sentry SDK supports.
Possibly RN will require higher but not sure we should be pinning to that on our podspec. I.e: They could change things in the future

bruno-garcia:
@jennmueng @HazAT goodto merge? Please do if so

jennmueng (approved):

PR #1075, opened 2020-09-14
test: new org dsn

bruno-garcia:
DSN for new test org for SDK dev.

Sorry the auto format, disabled 2 VS Code integrations still don't know which one is doing that on JS files

### DIFF
--- a/sample/App.js
+++ b/sample/App.js
@@ -26,13 +26,13 @@ import {
   ReloadInstructions,
 } from 'react-native/Libraries/NewAppScreen';
 
-import {Provider} from 'react-redux';
+import { Provider } from 'react-redux';
 import * as Sentry from '@sentry/react-native';
 
-import {version as packageVersion} from './package.json';
+import { version as packageVersion } from './package.json';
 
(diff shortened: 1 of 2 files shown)


jennmueng (changes requested):
Should also update to the org and project slugs here https://github.com/getsentry/sentry-react-native/blob/5ce5e2eb8086ace05b9d0b2174c48c1dba36955a/sample/utils/fetchEvent.ts#L6

bruno-garcia:
Right, then also the auth key in the github secrets needs updating.
We need to make sure this doesn't run In PRs then (by default secrets are not available in PRs, unless you're explicitly using it in the yml)

jennmueng:
The auth token we currently have in secrets is my own, but only with read access, so I think it should just work for the new test org. Hmm I do realize the possible security implication of running it in PRs, we can disable e2e tests for PRs then.

bruno-garcia:
> . Hmm I do realize the possible security implication of running it in PRs, we can disable e2e tests for PRs then.

yeah please, disable it for PRs

PR #1103, opened 2020-09-30
fix: xcode 12 compatibility

kasinskas:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

According to https://github.com/facebook/react-native/issues/29633#issuecomment-694187116 - "React pod is in fact really only an umbrella dependency for pure JS applications to depend on, whereas the native APIs that these libraries rely on actually reside in the React-Core pod." This PR updates the dependency from `React` to `React-Core`.

Xcode 12 fails to build if a module depends on React instead of React-Core. 
More info: https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

Use this branch to install with an app running on Xcode 12 or Xcode11. Both versions will work.

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/RNSentry.podspec
+++ b/RNSentry.podspec
@@ -16,7 +16,7 @@ Pod::Spec.new do |s|
 
   s.preserve_paths = '*.js'
 
-  s.dependency 'React'
+  s.dependency 'React-Core'
   s.dependency 'Sentry', '~> 5.2.0'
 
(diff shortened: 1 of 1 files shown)


bruno-garcia:
Thanks for this @kasinskas

PR #1131, opened 2020-10-13
feat: SDK Major Bumps

jennmueng:
Incorporate android v3.0.0 and cocoa v6.0.3.

### Notable Changes:
- build(android): (Breaking) Android package name changed from `io.sentry.RNSentryPackage` to `io.sentry.react.RNSentryPackage`, which requires users on React Native <0.60.0 to have to update the imported package name in their `MainApplication.java`.
- feat(ios): Migrate to using `captureEnvelope` on iOS/MacOS, allowing transactions to be sent.
- feat/fix: Support envelopes with type other than `event`, and fix `event` type being set on every envelope.
- fix(android): As `enableAutoSessionTracking` is now `true` by default, allow user to pass `false` to disable it.
- feat(android): Add `enableNdkScopeSync` property to `ReactNativeOptions`.
- feat(android): Pass `attachStacktrace` option property down to android SDK.

### Migration Docs:
Linked PR to `sentry-docs` to add migration docs from v1.x to v2.x of the React Native SDK:
https://github.com/getsentry/sentry-docs/pull/2528

### How this was tested:

Updated the wrapper's `sendEvent` test to reflect iOS now using `captureEnvelope`. Everything else tested manually with the sample app, and is working. Otherwise see e2e tests.


android/build.gradle
```diff
@@ -26,5 +26,5 @@ android {
 
 dependencies {
     implementation 'com.facebook.react:react-native:+'
-    api 'io.sentry:sentry-android:2.3.1'
+    api 'io.sentry:sentry-android:3.0.0'
```
marandaneto, line 29:
```suggestion
    api 'io.sentry:sentry-android:3.1.0'
```

android/src/main/java/io/sentry/RNSentryModule.java
```diff
@@ -92,12 +92,18 @@ public void startWithOptions(final ReadableMap rnOptions, Promise promise) {
...
+            if (rnOptions.hasKey("enableAutoSessionTracking")) {
+                options.setEnableSessionTracking(rnOptions.getBoolean("enableAutoSessionTracking"));
             }
             if (rnOptions.hasKey("sessionTrackingIntervalMillis")) {
                 options.setSessionTrackingIntervalMillis(rnOptions.getInt("sessionTrackingIntervalMillis"));
             }
+            if (rnOptions.hasKey("enableNdkScopeSync")) {
+                options.setEnableScopeSync(rnOptions.getBoolean("enableNdkScopeSync"));
+            }
+            if (rnOptions.hasKey("attachStacktrace")) {
```
marandaneto, line 104: do we already offer `attachThreads`? I'd suggest adding it as well.
it's now disabled by defaut

android/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,5 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="io.sentry">
+    package="io.sentry.react">
     <uses-permission android:name="android.permission.INTERNET" />
```
marandaneto, line 3: this is already merged automatically by `sentry-android-core` btw but not related to this PR

ios/RNSentry.m
```diff
@@ -119,29 +121,33 @@ + (BOOL)requiresMainQueueSetup {
...
-            // captureEvent queues the event for submission, so storing to disk happens asynchronously
-            // We need to make sure the event is written to disk before resolving the promise.
-            // This could be replaced by SentrySDK.flush() when available.
-            [[[[SentrySDK currentHub] getClient] fileManager] storeEvent:sentryEvent];
-        } else {
-            [SentrySDK captureEvent:sentryEvent];
-        }
-#endif
+    if ([NSJSONSerialization isValidJSONObject:envelopeDict]) {
+        SentrySdkInfo *sdkInfo = [[SentrySdkInfo alloc] initWithDict: envelopeDict[@"header"]];
```
philipphofmann, line 129: `l`: Code formatting
```suggestion
        SentrySdkInfo *sdkInfo = [[SentrySdkInfo alloc] initWithDict:envelopeDict[@"header"]];
```

ios/RNSentry.m
```diff
@@ -119,29 +121,33 @@ + (BOOL)requiresMainQueueSetup {
...
+        SentryId *eventId = [[SentryId alloc] initWithUUIDString:envelopeDict[@"header"][@"event_id"]];
+        SentryEnvelopeHeader *envelopeHeader = [[SentryEnvelopeHeader alloc] initWithId:eventId andSdkInfo:sdkInfo];
+
+        NSData *envelopeItemData = [NSJSONSerialization dataWithJSONObject:envelopeDict[@"payload"] options:0 error:nil];
```
philipphofmann, line 133: `m`: I know this was already there, but maybe it is good to not ignore the error here. I guess if this happens there's not much we can do and the envelope anyway can't be sent. So we could do something like this:
```
NSError *error;
[NSJSONSerialization dataWithJSONObject:envelopeDict[@"payload"] options:0 error:&error];

if (nil != error) {
  reject(@"SentryReactNative", @"Cannot serialize event", nil);
}
```

philipphofmann (commented):
Really awesome seeing this going to our customers soon.

android/src/main/java/io/sentry/RNSentryModule.java
jennmueng: We actually do, but it's right below that line and not an if statement like everything else. Let me refactor that.

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+- build(android): Changes android package name from `io.sentry.RNSentryPackage` to `io.sentry.react.RNSentryPackage` (Breaking). #1131
+- fix: As auto session tracking is now on by default, allow user to pass `false` to disable it. #1131
+- build: Bump `sentry-android` to 3.0.0. #1131
```
marandaneto, line 7:
```suggestion
- build: Bump `sentry-android` to 3.1.0. #1131
```

marandaneto (approved):
LGTM

CHANGELOG.md
jennmueng: Good catch, thanks!

philipphofmann (approved):
LGTM for the Cocoa part.

PR #1137, opened 2020-10-20
fix: Move exceptions to threads field on Android

jennmueng:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Add a if block in the Android side of the native bridge in the case that if an event has exceptions (`exception` field) (where the JS stack trace would be) we would copy it over to `threads` instead.

The thread's stracktrace was taking precedence over the JS stack trace on the dashboard, even though the JS stack trace wasn't being overwritten.

Called every type of error on the sample app, ensured that only happens to JS errors and not native iOS/Android errors.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/android/src/main/java/io/sentry/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/RNSentryModule.java
@@ -144,9 +144,9 @@ public class RNSentryModule extends ReactContextBaseJavaModule {
                     }
                 }
 
-                // If there are exceptions, move it to threads.
+                // If there are exceptions and the event originated from js, move it to threads.
                 // This is due to `threads` taking precedent over `exception` on the dashboard, hiding the JS Stack Trace.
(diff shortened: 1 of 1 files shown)


bruno-garcia:
Sounds like a hack. The server is supposed to handle a stack trace coming in the exception and in the threads panel already. We had this problem with Android at first. Seems that the server would match by thread-id and show the exception stack trace on the threads panel (instead of the thread's stack trace). That's a vague recollection that someone else will know better but I think this is a server concern and not something each SDK should be factoring in.

@marandaneto @jan-auer @matejminar may know more about this

HazAT:
We should fix this in the UI, talked to @priscilawebdev https://app.asana.com/0/1182975495223897/1198885927872565

PR #1230, opened 2020-12-06
feat: Auto performance tracing with XHR/fetch, and routing instrumentation

jennmueng:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Note: Needs https://github.com/getsentry/sentry-javascript/pull/3144 to be released over in js before this can be merged.

See `App.tsx` in the sample as an example of how to initialize auto performance tracing.

Screen Shot 2021-01-10 at 9 49 33 PM

### Routing

The routing Instrumentation has to be different than how it's currently implemented for browser given how routing in React Native is handled different and is not global. We would need to create an instance where a route event can be emitted by a navigation router and be listened to by our integration.

#### Base/Manual Routing Instrumentation

_"Manual-Auto"_

Create an instance of `RoutingInstrumentation`, pass it to the tracing integration, and then call `onRouteWillChange` every time the route changes with the desired transaction context. This sets an idle transaction on the scope.

```typescript
// Construct a new instrumentation instance. This is needed to communicate between the integration and React
const routingInstrumentation = new Sentry.RoutingInstrumentation();

Sentry.init({
  ...
  integrations: [
      // Pass instrumentation to be used as `routingInstrumentation`
      routingInstrumentation,
      ...
    }),
  ],
})

const App = () => {
  // Call this before the route changes
  routingInstrumentation.onRouteWillChange(transactionContext)
};
```

#### React Navigation V5 Instrumentation

Below is an example of how to use the current implementation:
```typescript
// Construct a new instrumentation instance. This is needed to communicate between the integration and React
const reactNavigationV5Instrumentation = new Sentry.ReactNavigationV5Instrumentation();

Sentry.init({
  ...
  integrations: [
      // Pass instrumentation to be used as `routingInstrumentation`
      routingInstrumentation: reactNavigationV5Instrumentation,
      ...
    }),
  ],
})

const App = () => {
  // Create a ref for the navigation container
  const navigation = React.createRef<NavigationContainerRef>();

  // Register the navigation container with the instrumentation
  React.useEffect(() => {
    reactNavigationV5Instrumentation.registerNavigationContainer(navigation);
  }, []);

  return (
    // Connect the ref to the navigation container
    <NavigationContainer ref={navigation}>
      ...
    </NavigationContainer>
  );
};
```

Preferably, all the react component part can just be done in a custom hook

#### React Navigation V4 Instrumentation

Works almost exactly like V5 on the surface, although entirely different underneath.

```typescript
// Construct a new instrumentation instance. This is needed to communicate between the integration and React
const reactNavigationV4Instrumentation = new Sentry.ReactNavigationV4Instrumentation();

Sentry.init({
  ...
  integrations: [
      // Pass instrumentation to be used as `routingInstrumentation`
      routingInstrumentation: reactNavigationV4Instrumentation,
      ...
    }),
  ],
})

const App = () => {
  // Create a ref for the navigation container
  const appContainer = React.createRef();

  // Register the navigation container with the instrumentation
  React.useEffect(() => {
    reactNavigationV4Instrumentation.registerAppContainer(appContainer);
  }, []);

  return (
    // Connect the ref to the navigation container
    <AppContainer ref={appContainer} />
  );
};
```

### Idle Transaction
The idle transaction is carried over from the browser implementation of tracing. It needs to be exposed upstream from `@sentry/tracing`. An idle transaction should be started and set on the scope every time a route changes using a routing instrumentation.

### Fetch/XHR
Fetch/XHR instrumentation is carried over from the browser implementation, and it also needs to be exposed.

### React Profiler
Works out of the box.

https://sentry.io/organizations/sentry-sdks/discover/sentry-react-native:157803b6f5b94f91b5da33014911f181/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&query=&sort=-timestamp&statsPeriod=24h&widths=-1&widths=-1&widths=-1&widths=-1&widths=-1

### Need to expose from @sentry/tracing

Some things need to be exposed upstream from `@sentry/tracing`:
```
IdleTransaction,
startIdleTransaction,
registerRequestInstrumentation,
RequestInstrumentationOptions,
defaultRequestInstrumentationOptions,
```

## Remaining things to do/investigate
- [x] Manual Routing Instrumentation
- [x] React Navigation V5
- [x] React Navigation V4
- [x] Fetch/XHR request instrumentation
- [x] Test Everything on iOS Sim
- [x] Test Everything on Android Sim
- [x] Clean up code and comment

Tested locally on iOS simulator with new sample app.

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/tracing/reactnativetracing.ts
+++ b/src/js/tracing/reactnativetracing.ts
@@ -41,6 +41,16 @@ export interface ReactNativeTracingOptions
    * There is no routing instrumentation if nothing is passed.
    */
   routingInstrumentation?: RoutingInstrumentation;
+
+  /**
+   * beforeNavigate is called before a navigation transaction is created and allows users to modify transaction
+   * context data, or drop the transaction entirely (by setting `sampled = false` in the context).
+   *
+   * @param context: The context data which will be passed to `startTransaction` by default
+   *
+   * @returns A (potentially) modified context object, with `sampled = false` if the transaction should be dropped.
+   */
+  beforeNavigate?(context: TransactionContext): TransactionContext;
 }
 
 const defaultReactNativeTracingOptions: ReactNativeTracingOptions = {
@@ -132,11 +142,25 @@ export class ReactNativeTracing implements Integration {
     }
 
     // eslint-disable-next-line @typescript-eslint/unbound-method
-    const { idleTimeout, maxTransactionDuration } = this.options;
+    const {
+      beforeNavigate,
+      idleTimeout,
+      maxTransactionDuration,
+    } = this.options;
 
-    if (context.sampled === false) {
+    const expandedContext = {
+      ...context,
+      trimEnd: true,
+    };
+
+    const modifiedContext =
+      typeof beforeNavigate === "function"
+        ? beforeNavigate(expandedContext)
+        : expandedContext;
+
+    if (modifiedContext.sampled === false) {
       logger.log(
-        `[ReactNativeTracing] Will not send ${context.op} transaction because of beforeNavigate.`
+        `[ReactNativeTracing] Will not send ${context.op} transaction.`
       );
     }
 
@@ -152,7 +176,6 @@ export class ReactNativeTracing implements Integration {
     );
     idleTransaction.registerBeforeFinishCallback(
       (transaction, endTimestamp) => {
-        // this._metrics.addPerformanceEntries(transaction);
         adjustTransactionDuration(
           secToMs(maxTransactionDuration),
           transaction,
--- a/src/js/tracing/reactnativetracing.ts
+++ b/src/js/tracing/reactnativetracing.ts
@@ -43,6 +43,14 @@ export interface ReactNativeTracingOptions
   routingInstrumentation?: RoutingInstrumentationInstance;
 
   /**
+   * Does not sample transactions that are from routes that have been seen any more and don't have any spans.
+   * This removes a lot of the clutter as most back navigation transactions are now ignored.
+   *
+   * Default: true
+   */
+  ignoreEmptyBackNavigationTransactions?: boolean;
+
+  /**
    * beforeNavigate is called before a navigation transaction is created and allows users to modify transaction
    * context data, or drop the transaction entirely (by setting `sampled = false` in the context).
    *
@@ -57,6 +65,7 @@ const defaultReactNativeTracingOptions: ReactNativeTracingOptions = {
   ...defaultRequestInstrumentationOptions,
   idleTimeout: 1000,
   maxTransactionDuration: 600,
+  ignoreEmptyBackNavigationTransactions: true,
 };
 
 /**
@@ -119,6 +128,14 @@ export class ReactNativeTracing implements Integration {
       traceXHR,
       tracingOrigins,
       shouldCreateSpanForRequest,
+    });
+
+    addGlobalEventProcessor((event) => {
+      // eslint-disable-next-line no-empty
+      if (event.type === "transaction") {
+      }
+
+      return event;
     });
   }
 
@@ -183,6 +200,21 @@ export class ReactNativeTracing implements Integration {
         );
       }
     );
+
+    if (this.options.ignoreEmptyBackNavigationTransactions) {
+      idleTransaction.registerBeforeFinishCallback((transaction) => {
+        if (
+          transaction.data["routing.route.hasBeenSeen"] &&
+          (!transaction.spanRecorder ||
+            transaction.spanRecorder.spans.filter(
+              (span) => span.spanId !== transaction.spanId
+            ).length === 0)
+        ) {
+          // Route has been seen before and has no child spans.
+          transaction.sampled = false;
+        }
+      });
+    }
 
     return idleTransaction as TransactionType;
   }
--- a/src/js/tracing/reactnativetracing.ts
+++ b/src/js/tracing/reactnativetracing.ts
@@ -14,7 +14,7 @@ import {
 import { logger } from "@sentry/utils";
 
 import { RoutingInstrumentationInstance } from "../tracing/routingInstrumentation";
-import { adjustTransactionDuration, secToMs } from "./utils";
+import { adjustTransactionDuration } from "./utils";
 
 export interface ReactNativeTracingOptions
   extends RequestInstrumentationOptions {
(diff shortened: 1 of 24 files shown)


src/js/tracing/reactnavigation.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  /**
+   * Function to be called on every React-Navigation action dispatch
+   */
+  private _onDispatch(dispatchProp: DispatchProp): void {
+    const action = dispatchProp.data?.action;
+    if (action) {
+      const { type, payload } = action;
+      if (type === "NAVIGATE" && payload) {
```
giautm, line 53: Should add `PUSH` & `REPLACE` actions?

jennmueng: Yes we should account for those too

giautm:
Hi @jennmueng, do you need any help to finish this PR? I'm so awaiting it's go live

jennmueng:
> Hi @jennmueng, do you need any help to finish this PR? I'm so awaiting it's go live

Nothing that would need help, just need time to finish everything and test everything

jennmueng:
Updated the v5 instrumentation to handle cases where a router with action types other than the default stack ones is used. It now creates an idle transaction on **every** successful navigation action dispatch, but it does not include route context just yet. It will wait for the navigation state to change, and then update the transaction with the route context. If the state change doesn't happen within 200ms, the transaction is then discarded.

It provides the option of setting a `shouldAttachTransaction` method that would be called after each state change to determine whether to discard this transaction or keep it. By default it sets a transaction on every new route that isn't a repeat of the last route.

jennmueng:
Update: Working perfectly and ready for iOS, however I've run into an issue on Android where the current SDK (`3.2.0`) does not fully support transaction types. Currently going to try a workaround.

jennmueng:
Temporary Hack to use the js `FetchTransport` to send transactions on Android for now. This will not be GA and instead just for an alpha. We will need to wait for transactions on FileObserver envelope support on Android

jennmueng:
Now with instrumentation for React-Navigation V4. It one works differently than the V5 one on the inside, although extremely similar on the outside. This time we patch the router on the `AppContainer`.

jennmueng:
Blocked by https://github.com/getsentry/sentry-javascript/pull/3144 over in JS as everyone there is out of office.

bruno-garcia:
Maybe worth resolving conflict with `master` so it's ready to merge

jennmueng:
> Maybe worth resolving conflict with `master` so it's ready to merge

Sure, but it would also need a bump to the @sentry/javascript dependencies too before we can merge it btw.

jennmueng:
Just a note: The CI is failing due to things from @sentry/tracing not being exported yet. Need to wait for a new release over at JS before these will pass.

janicduplessis:
Probably out of scope for this PR, but are there plans to look into also instrumenting the native side? I'm pretty sure RN has hooks for this. This way it would be possible to visualize the full rendering, including react js render, RN bridging, layout and native render.

bruno-garcia:
> Probably out of scope for this PR, but are there plans to look into also instrumenting the native side? I'm pretty sure RN has hooks for this. This way it would be possible to visualize the full rendering, including react js render, RN bridging, layout and native render.

I dunno if @jennmueng already looked at it, but could you point us to the hooks we can use for that?

We're adding performance API to the native layers next, but we don't have a story for propagating transaction through the bridge yet.

I'm hoping that at least a span for the bridge call will be possible even without any support from the native SDKs.

jennmueng:
@janicduplessis Yes we're definitely looking into instrumenting React Native way more in the near future. We would really appreciate if you know any of these hooks that you mentioned to see if we can instrument something there.

janicduplessis:
On Android there's this gist that shows how to use the ReactMarker API https://gist.github.com/axemclion/f01cbb32fcb5b14bf6b4fd4594192825#file-perflogger-java, which should have some interesting info. Not sure about what is available on iOS. Are you on RN contributor discord? There are probably people there that can help too.

bruno-garcia:
@janicduplessis Thanks for the link! Not sure about @jennmueng, I'm not on that discord but I'm on the `#react-native` channel of [Sentry's Discord server](http://discord.gg/sentry), would you like to join and we discuss this there?

jennmueng:
Note: Can use `tracesSampler` instead of the `shouldSendTransaction`

src/js/transports/native.ts
```diff
@@ -8,10 +10,23 @@ export class NativeTransport implements Transport {
   /** A simple buffer holding all requests. */
   protected readonly _buffer: PromiseBuffer<Response> = new PromiseBuffer(30);
 
+  /** Hack that's used to send transactions on Android */
+  private _fetchTransport: Transport;
```
marandaneto, line 14: you can now experiment with https://github.com/getsentry/sentry-java/releases/tag/4.0.0-alpha.3
that should resolve the problem and the hack could be removed

jennmueng: Ok thanks, I'll give it a try and I think we can have an alpha ready by like Monday.

src/js/tracing/utils.ts
```diff
@@ -0,0 +1,34 @@
+import { IdleTransaction, SpanStatus } from "@sentry/tracing";
+
+/**
+ * Converts from milliseconds to seconds
+ * @param time time in ms
+ */
+export function msToSec(time: number): number {
+  return time / 1000;
+}
```
HazAT, line 9: We don't use this function, or do we?

src/js/tracing/reactnativetracing.ts
```diff
@@ -0,0 +1,221 @@
...
+      `[ReactNativeTracing] Starting ${context.op} transaction on scope`
+    );
+    idleTransaction.registerBeforeFinishCallback(
+      (transaction, endTimestamp) => {
+        adjustTransactionDuration(
+          secToMs(maxTransactionDuration),
+          transaction,
+          endTimestamp
+        );
+      }
```
HazAT, line 201: Here when calling `secToMs` inline, you can just do it in the `adjustTransactionDuration` internally. So you don't need to import `secToMs` here.

HazAT (approved):
Few nits

src/js/tracing/utils.ts
jennmueng: No we don't, it was copied over from js at first, let me remove it.

Issue #1248, opened 2020-12-17
Disable Native SDK initialization from React Native and use an existing instance


pioneer32:
**Platform:**
- [X]  iOS
- [ ]  Android

**SDK:**
- [X]  `@sentry/react-native` (>= 1.0.0)
- [X]  `sentry-cocoa` (6.0.9)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:**
- `@sentry/react-native` 2.1.0
- `sentry-cocoa` 6.0.9

**`react-native` version:** 0.60.5 (also tried to upgrade to 0.62.2 - the same outcome as with 0.60.5)

**Are you using Expo?**
- [ ]  Yes
- [X]  No

**Are you using sentry.io or on-premise?**
- [X]  sentry.io (SaaS)
- [ ]  on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

When the native one is initialised first and JS one is the second:
https://sentry.io/organizations/vladislav-churakov/issues/2096897776/?project=1834395&query=is%3Aunresolved

When no native one is used, one JS one:
https://sentry.io/organizations/vladislav-churakov/issues/2094670206/?project=1834395&query=is%3Aunresolved

**Configuration:**

(`@sentry/react-native`)
```
      Sentry.init({
        dsn: 'https://....ingest.sentry.io/...'
        environment: Config.RELEASE_TYPE, // a string that can only be "test" "preprod" "production"
        debug: true, // Changing this has no effect on what we see in Sentry
      });
```
(`sentry-cocoa` in AppDelegate.swift)
```
  func application( ... didFinishLaunchingWithOptions ... ) -> Bool {
      SentrySDK.start { options in
        options.dsn = "https://....ingest.sentry.io/..."
        options.environment = SettingsHelper.sharedInstance.releaseType // a string that can only be "test" "preprod" "production"
        options.debug = true // Changing this has no effect on what we see in Sentry
      }

```

---
**I have following issue:**

Given there's an iOS native app that uses sentry-cocoa and all the native stacktraces are nicely symbolised with some functionality done with react-native (a "hybrid" app).

When `SentrySDK.start` is called in `AppDelegate.swift` the first (we need this as we need to see native crashes) and then `Sentry.init` is called in JS as react-native starts , the JS stacktrace looks very bad in Sentry:

![image](https://user-images.githubusercontent.com/4023517/102555935-a48cf600-412c-11eb-8f0b-f60be07ba37b.png)

[...]

PR #1259, opened 2020-12-22
feat: Allow disabling native SDK initialization but still use it

jennmueng:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Allows the user to disable the initialization of the Native SDK in the case that they already have one initialized themselves. We do this by adding the `shouldInitializeNativeSdk` key to `ReactNativeOptions`. This is not breaking as it will default to `true`.

User has an app that is extensively native iOS and the React Native section is only run way later: https://github.com/getsentry/sentry-react-native/issues/1248

Wrote a new test in `wrapper.ts`.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Should add documentation on this option.

### DIFF
--- a/src/js/options.ts
+++ b/src/js/options.ts
@@ -20,6 +20,12 @@ export interface ReactNativeOptions extends BrowserOptions {
    */
   enableNativeCrashHandling?: boolean;
 
+  /** Initializes the native SDK on init.
+   * Set this to `false` if you have an existing native SDK and don't want to re-initialize.
+   * @default true
+   */
+  shouldInitializeNativeSdk?: boolean;
+
   /** Maximum time to wait to drain the request queue, before the process is allowed to exit. */
   shutdownTimeout?: number;
 
--- a/src/js/options.ts
+++ b/src/js/options.ts
@@ -20,8 +20,15 @@ export interface ReactNativeOptions extends BrowserOptions {
    */
   enableNativeCrashHandling?: boolean;
 
-  /** Initializes the native SDK on init.
+  /**
+   * Initializes the native SDK on init.
    * Set this to `false` if you have an existing native SDK and don't want to re-initialize.
+   *
+   * NOTE: Be careful and only use this if you know what you are doing.
+   * If you use this flag, make sure a native SDK is running before the JS Engine initializes or events might not be captured.
+   * Also, make sure the DSN on both the React Native side and the native side are the same one.
+   * We strongly recommend checking the documentation if you need to use this.
+   *
    * @default true
    */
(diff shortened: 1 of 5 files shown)


src/js/options.ts
```diff
@@ -20,6 +20,12 @@ export interface ReactNativeOptions extends BrowserOptions {
    */
   enableNativeCrashHandling?: boolean;
 
+  /** Initializes the native SDK on init.
+   * Set this to `false` if you have an existing native SDK and don't want to re-initialize.
+   * @default true
+   */
+  shouldInitializeNativeSdk?: boolean;
```
marandaneto, line 27: what are the edge cases of such flag?

eg 1. RN SDK captures an event, Native SDKs are not inited.
2. RN sets up X DSN but events go to Y DSN etc...

not against it, but we'll need to figure out such edge cases and write them down.

jennmueng: Yes I agree, should mention any possible side effects and edge cases, and also need to document this clearly. Thanks for the catch!

pioneer32: Being the author of the initial feature request, I'd like to add my 2 cents: Having the only source for DSN/environment etc would be a better option, so that I don't need to worry about that in the JS side, since it'd've been already fully initialised and configured on the native side.

marandaneto: @pioneer32 that would be the perfect solution, agree, but I'd say we'd go the other way around.
we'd sync the options from RN -> Native as we already do, which is let's say ~90+% of the use cases.
you could open an issue as a feature request and let's see if it gets upvoted, but I'd bet on the case described above.

bruno-garcia (approved):
LGTM

src/js/options.ts
pioneer32: @marandaneto Thank you for clarification! My guess is `shouldInitializeNativeSdk` sort of going to disable calling https://github.com/getsentry/sentry-react-native/blob/97b16b72cc2970a9de89ff2903064a19879d2b8b/ios/RNSentry.m#L67 somehow (I might be wrong though). However, being keen to see what it'd look like, I commented out that line intentionally, and it seems to work as expected, almost :) FYI The problem I encountered was: all the tags/breadcrumbs/etc set to the native client are NOT copied into JS client before sending an error from JS.

bruno-garcia:
For reference, this came up in Flutter too already: https://github.com/getsentry/sentry-dart/discussions/255#discussioncomment-262444

Issue #1292, opened 2021-01-18
SDK doesn't start session on first app launch on iOS


philipphofmann:
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [x]  iOS
- [ ]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 2.1.0

**`react-native` version:** 
https://github.com/getsentry/sentry-react-native/blob/ccf1fe10d608a4497bf6a5974ac6131e787f6cd3/sample/package.json#L20-L24

**Configuration:**

https://github.com/getsentry/sentry-react-native/blob/ccf1fe10d608a4497bf6a5974ac6131e787f6cd3/sample/src/App.tsx#L18-L36

---
**I have following issue:**

The sentry-cocoa SDK doesn't start a session on the first launch of the app, because [`SentrySessionTracker.didBecomeActive`](https://github.com/getsentry/sentry-cocoa/blob/master/Sources/Sentry/SentrySessionTracker.m#L116) doesn't fire. If you put the app in the background and open it again `didBecomeActive` fires. I'm not sure if we need to fix this in sentry-cocoa, in sentry-react-native, or both.

**Steps to reproduce:**
* Start sample app
* Put the app in background for 6 seconds
* Launch app again

**Actual result:**

The SDK doesn't send a session.

**Expected result:**

The SDK should send a healthy session.

This is related to https://github.com/getsentry/sentry-dart/issues/271

marandaneto:
we've fixed it on Flutter and it worked pretty well, I'd just replicate the same solution.

See: https://github.com/getsentry/sentry-dart/pull/274/files

@bruno-garcia we need to prioritize this.

bruno-garcia:
@marandaneto right! thanks

PR #1308, opened 2021-01-27
fix: Handle auto session tracking start on iOS

jennmueng:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Posts a `SentryHybridSdkDidBecomeActive` notification once if the app is active, on iOS SDK init to start a session on iOS.

https://github.com/getsentry/sentry-react-native/issues/1292

Tested on sample app and verified that sessions get sent on initial app launch.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/sample/ios/Podfile.lock
+++ b/sample/ios/Podfile.lock
@@ -1,6 +1,6 @@
 PODS:
   - boost-for-react-native (1.63.0)
-  - CocoaAsyncSocket (7.6.4)
+  - CocoaAsyncSocket (7.6.5)
   - CocoaLibEvent (1.0.0)
   - DoubleConversion (1.1.6)
   - FBLazyVector (0.63.3)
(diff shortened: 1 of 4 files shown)


marandaneto (approved):
looks good, thanks!

sample/ios/Podfile.lock
```diff
@@ -455,7 +455,7 @@ EXTERNAL SOURCES:
 
 SPEC CHECKSUMS:
   boost-for-react-native: 39c7adb57c4e60d6c5479dd8623128eb5b3f0f2c
-  CocoaAsyncSocket: 694058e7c0ed05a9e217d1b3c7ded962f4180845
+  CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
```
marandaneto, line 458: wondering why this has changed

sample/ios/Podfile.lock
```diff
@@ -495,11 +495,11 @@ SPEC CHECKSUMS:
...
-  Sentry: 5b741b3c932fa293c28ba6a6f510b72bd63c9993
+  RNSentry: b894f70e20c276e175d8083aa424d2b5fd2d4128
+  Sentry: 9d055e2de30a77685e86b219acf02e59b82091fc
   Yoga: 7d13633d129fd179e01b8953d38d47be90db185a
   YogaKit: f782866e155069a2cca2517aafea43200b01fd5a
 
 PODFILE CHECKSUM: 9d75e167307300a63add2b53a03e1d19bb02cd35
 
-COCOAPODS: 1.10.1
+COCOAPODS: 1.9.3
```
marandaneto, line 505: I believe this should be rollback?

sample/ios/Podfile.lock
jennmueng: When I bump `sentry-cocoa` I have to delete the whole Podfile.lock because otherwise pod install doesn't work. At this point maybe we should consider stopping committing lock files?

marandaneto: yeah its also from the sample, so not important, I'd remove, we do that on dart/flutter https://github.com/getsentry/sentry-dart/blob/main/.gitignore#L22-L23

Issue #1312, opened 2021-01-31
Routing instrumentation should use `beforeNavigate` like on JS.


jennmueng:
Currently in `2.2.0-beta.0` the React Navigation routing instrumentations take in a `shouldSendTransaction` instead of using `beforeNavigate` like on `@sentry/tracing`. This API is inconsistent and potentially confusing to Users, so we need to find a way to make this consistent.

PR #1311, opened 2021-01-31
build(js): Bump js dependencies to 6.0.3

jennmueng:
## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Includes the transaction/span update methods: https://github.com/getsentry/sentry-javascript/pull/3192 that are required to use `beforeNavigate` in ReactNativeTracing.

## :green_heart: How did you test it?
Ran on simulators

## :pencil: Checklist

- [x] I reviewed submitted code
- [ ] I added tests to verify changes
- [x] All tests passing
- [x] No breaking changes

PR #1313, opened 2021-01-31
feat: Use beforeNavigate in routing instrumentation to match behavior on JS

jennmueng:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

**NOTE: This needs #1311 to be merged first!**

Uses the `beforeNavigate` option that is passed to `ReactNativeTracing` to have the same behavior as on JS.
_(This removes the deprecated feature of returning undefined in the type, although handles it just in case users are used to the old behavior.)_

We will now also export a convenience type `ReactNavigationTransactionContext extends TransactionContext` for users who use the React Navigation instrumentations:
```js
beforeNavigate: (context: Sentry.ReactNavigationTransactionContext) => {
  // context.data.route, context.data.previousRoute, context.tags is typed for convenience

  return context;
}
```

This also does not implement any context-change-wiping that was discussed with @lobsterkatie in https://github.com/getsentry/sentry-javascript/pull/3192 as I believe there needs to be more discussion internally about which properties should not be changed by the user.

Addresses #1312 

Updated tests for React Navigation V4 and V5. Tested on simulators running both instrumentations.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes. **NOTE:** This breaks users who used `shouldSendTransaction` on 2.2.0-beta.0. But it was undocumented in anticipation for this change so there should not be any or at worst very few.


jennmueng:
Also note, tests will fail until this is rebased on top of #1311

HazAT (approved):

PR #1352, opened 2021-02-18
fix: Removing Flipper from sample app and bump Appium to 1.20.2

jennmueng:
This PR fixes the failing e2e tests we've been having on iOS.

We never used [Flipper](https://fbflipper.com) in our sample app, so I removed it as it was causing problems with our e2e test build. It was there in the first place as it was included by default in React Native templates.

Bumps `appium` to 1.20.2,
Bumps `wd` to 1.14.0

#skip-changelog


bruno-garcia (commented):
Lots of deleted code/dependencies and a green CI. 💯

PR #1356, opened 2021-02-19
meta: Disable changelog requirement on Danger for PRs created by bots

jennmueng:
This way we stop getting spammed by failed CI on every dependabot PR due to no changelog.

Reference: https://danger.systems/js/reference.html#GitHubUser

#skip-changelog


bruno-garcia (approved):
Probably want to port this to other repos eventually.

PR #1365, opened 2021-02-26
fix(performance): Handle edge cases in React Navigation routing instrumentation

### LINKED ISSUE #1361: React-Native performance traces not being uploaded to Sentry
### Environment

Using Sentry.io on React-Native.

Running React-Native on Android device only (ios not used).

```
"@sentry/react-native": "^2.2.0",
"react": "16.13.1",
"react-native": "^0.63.4",
"@react-navigation/drawer": "^5.8.2",
"@react-navigation/native": "^5.5.1",
"@react-navigation/stack": "^5.5.1",
```

### Steps to Reproduce

Instructions from documentation were followed;

Setup:
```typescript
// index.ts

import * as Sentry from '@sentry/react-native';
import { SENTRY_DSN, SENTRY_ENV } from '@env';

export const routingInstrumentation = new Sentry.ReactNavigationV5Instrumentation();

Sentry.init({
  dsn: SENTRY_DSN,
  tracesSampleRate: 1.0,
  enableAutoSessionTracking: true,
  environment: SENTRY_ENV,
  integrations: [
      tracingOrigins: ['localhost', 'some.other.url', /^\//],
      routingInstrumentation
    })
  ],
  debug: true,
  logLevel: 3
});

// ...

```
Attempt to send a performance trace.
```typescript
// App.ts

// ...
const App = () => {
  const transaction = Sentry.startTransaction({
    name: 'App Component',
    sampled: true
  });
  transaction.sampled = true;
  const span = transaction.startChild({ op: 'functionX' }); // This function returns a Span
  span.sampled = true;
  
  setTimeout(() => {
    span.finish(); // Remember that only finished spans will be sent with the transaction
    transaction.finish();
  }, 500);

  return (
 // ...
```
Metro bundler logs from React-Native on App startup:
```
Sentry Logger [Log]: Integration installed: ReactNativeErrorHandlers
Sentry Logger [Log]: Integration installed: Release
Sentry Logger [Log]: Integration installed: InboundFilters
Sentry Logger [Log]: Integration installed: FunctionToString
Sentry Logger [Log]: Integration installed: Breadcrumbs
Sentry Logger [Log]: Integration installed: LinkedErrors
Sentry Logger [Log]: Integration installed: UserAgent
Sentry Logger [Log]: Integration installed: DebugSymbolicator
Sentry Logger [Log]: Integration installed: RewriteFrames
Sentry Logger [Log]: Integration installed: DeviceContext
Sentry Logger [Log]: Setting idle transaction on scope. Span ID: a5a7ff17b3c5b929
Sentry Logger [Log]: Starting heartbeat
Sentry Logger [Log]: pinging Heartbeat -> current counter: 0
Sentry Logger [Log]: [Tracing] starting navigation transaction - Route Change
Sentry Logger [Log]: [ReactNativeTracing] Starting navigation transaction "Route Change" on scope
Sentry Logger [Log]: Integration installed: ReactNativeTracing
Sentry Logger [Log]: [Tracing] finishing IdleTransaction 2021-02-23T20:25:50.545Z navigation
Sentry Logger [Log]: [Tracing] flushing IdleTransaction
Sentry Logger [Log]: [Tracing] Discarding transaction because its trace was not chosen to be sampled.
Sentry Logger [Log]: [Tracing] starting undefined transaction - App Component
```

### Expected Result

Performance traces should be uploaded and visible on Sengry.io dashboard.

### Actual Result

No performance traces appear, logs suggest that none are uploaded.

jennmueng:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

In an attempt to find causes of routing transactions not being sampled. I ended up finding multiple edge cases for routing transactions to not be sent. Below are the list and explanations for what I found. Sorry in advance that this description is especially long:

1. If the Instrumentation gets the navigation container registered before the integration’s `setupOnce` is called. So it tries to update the state of the initial route’s transaction which does not yet exist. When the integration is setup and it registers the routing instrumentation, which triggers the `_onDispatch` to be called and creates the initial blank transaction, the state is never set and the timeout to not sample is never cancelled, so the transaction ends up not being sampled.
    - **Status**: Done
    - Fix:
      - Add a `_initialStateHandled` boolean.
        - On routing instrumentation register, if the flag is false, will start a blank transaction. If an app/navigation container has already been registered, it populates with the state right away and sets it to true.
        - On app/navigation container register, if it is false and a routing instrumentation has been registered, it will populate the latest transaction and sets it to true.
2. If the fast reload feature on RN is used, and the file that the instrumentation+sdk is initialized in is reloaded, it creates a new instance of the instrumentation, so now you have multiple trying to create transactions on the same scope. As only one routing instrumentation can be registered to the `ReactNativeTracing` integration, this means the subsequent ones don't get registered. The initial one also loses the original reference to the navigation container given the method that we set it.
    - **Status**: Done
    - Fix:
        - Set a global (`__sentry_rn_v4_registered`, `__sentry_rn_v5_registered`) to ensure only one is registered at a time.
        - Don't use the passed React.useRef reference object, instead just use a direct reference to the navigator. This will not be un-set when the component containing the ref unmounts.

3. If they re-register the navigation container multiple times (such as forgetting to pass the `[]` as the second parameter to `React.useEffect` like in this [code snippet we have on the docs](https://docs.sentry.io/platforms/react-native/performance/included-instrumentation/#react-navigation-v5))
   - **Status**: Fixed by case 2's fix.
 
4. (V5) React Navigation V5 does not actually set the ref on mount if linking is enabled, will need to move the `registerNavigationContainer` call to `onReady` as per https://github.com/react-navigation/react-navigation/issues/8445#issuecomment-789126307.
   -  **Status**:  This has been done in the sample, and will need to be done in the docs as well: https://github.com/getsentry/sentry-docs/issues/3183

Fixes #1361 along with others who have reported issues to me.

New unit tests for both V4 and V5 that test:
1. Accepts a app/navigation container ref **object** (`{ current: NavigationContainer }`)
2. Accepts a app/navigation container direct reference.
3. Does not register a navigation container if there already exists one (`__sentry_rn_v4_registered` or `__sentry_rn_v5_registered` is true).
4. Works if the app/navigation container is registered before the tracing integration is setup.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/sample/src/App.tsx
+++ b/sample/src/App.tsx
@@ -60,15 +60,19 @@ Sentry.init({
 const Stack = createStackNavigator();
 
 const App = () => {
-  const navigation = React.createRef<NavigationContainerRef>();
+  const navigation = React.useRef<NavigationContainerRef>();
 
-  React.useEffect(() => {
-    reactNavigationV5Instrumentation.registerNavigationContainer(navigation);
-  }, []);
+  React.useEffect(() => {}, []);
 
   return (
     <Provider store={store}>
-      <NavigationContainer ref={navigation}>
+      <NavigationContainer
+        ref={navigation}
+        onReady={() => {
+          reactNavigationV5Instrumentation.registerNavigationContainer(
+            navigation,
+          );
+        }}>
         <Stack.Navigator>
           <Stack.Screen name="Home" component={HomeScreen} />
           <Stack.Screen name="Tracker" component={TrackerScreen} />
--- a/sample/src/App.tsx
+++ b/sample/src/App.tsx
@@ -62,8 +62,6 @@ const Stack = createStackNavigator();
 const App = () => {
   const navigation = React.useRef<NavigationContainerRef>();
 
-  React.useEffect(() => {}, []);
-
   return (
     <Provider store={store}>
       <NavigationContainer
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- fix(performance): Handle edge cases in React Navigation routing instrumentation.
+- fix(performance): Handle edge cases in React Navigation routing instrumentation. #1365
 
 ## 2.2.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fix(performance): Handle edge cases in React Navigation routing instrumentation.
+
 ## 2.2.2
 
 - fix: Fix unhandled promise rejections not being tracked #1367
--- a/src/js/tracing/reactnavigationv4.ts
+++ b/src/js/tracing/reactnavigationv4.ts
@@ -1,5 +1,5 @@
 import { Transaction } from "@sentry/types";
-import { logger } from "@sentry/utils";
+import { getGlobalObject, logger } from "@sentry/utils";
 
 import { BeforeNavigate } from "./reactnativetracing";
 import {
@@ -37,7 +37,7 @@ export interface AppContainerInstance {
 }
 
 interface AppContainerRef {
-  current?: AppContainerInstance | null;
+  current: AppContainerInstance | null;
 }
 
 /**
@@ -47,7 +47,7 @@ interface AppContainerRef {
 class ReactNavigationV4Instrumentation extends RoutingInstrumentation {
   static instrumentationName: string = "react-navigation-v4";
 
-  private _appContainerRef: AppContainerRef = { current: null };
+  private _appContainerRef: AppContainerInstance | null = null;
 
   private readonly _maxRecentRouteLen: number = 200;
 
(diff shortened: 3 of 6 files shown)


jennmueng:
Update: There is also a potential issue when you have both `ref` and `linking` set in React Navigation V5: https://github.com/react-navigation/react-navigation/issues/8445

bruno-garcia (approved):
RSLGTM!

sample/src/App.tsx
```diff
@@ -60,15 +60,19 @@ Sentry.init({
 const Stack = createStackNavigator();
 
 const App = () => {
-  const navigation = React.createRef<NavigationContainerRef>();
+  const navigation = React.useRef<NavigationContainerRef>();
 
-  React.useEffect(() => {
-    reactNavigationV5Instrumentation.registerNavigationContainer(navigation);
-  }, []);
+  React.useEffect(() => {}, []);
```
jennmueng, line 65: Need to remove this dead code block

PR #1370, opened 2021-03-05
feat(performance): Option to set route change timeout in routing instrumentation

### LINKED ISSUE #1369: ReactNavigationV5Instrumentation un-usable due to unconfigurable timeout
cc @jennmueng

Further to my now closed issue: https://github.com/getsentry/sentry-react-native/issues/1361 I've figured out why the route transition traces are not working.

[This non-configurable timeout](https://github.com/getsentry/sentry-react-native/blob/master/src/js/tracing/reactnavigationv5.ts#L27) `STATE_CHANGE_TIMEOUT_DURATION ` means that I can't get traces on my route transitions.

Monkey patching the sentry node module with a larger timeout results in the traces getting sent.

Ironically I'm trying to use sentry to diagnose and trace my slow route transitions on old devices and this very timeout prevents me from doing that because they are too slow!

This should be configurable via an options object I think.

Let me know if I've misunderstood something - thanks.

jennmueng:

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Adds the ability to set the timeout duration where a routing transaction is considered "discarded". The default timeout duration is now set to 1000ms.

This stems from an issue of users having unperformant navigation routes that take longer than the old default of 200ms getting their valid navigation transactions discarded. It makes sense as you're supposed to measure the performance of these route changes.

Resolves #1369 and #1361 

Adds test for both V4 and V5.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - fix(performance): Handle edge cases in React Navigation routing instrumentation. #1365
 - build(android): Bump sentry-android to 4.3.0 #1373
+- feat(performance): Option to set route change timeout in routing instrumentation #1370
 
 ## 2.2.2
 
--- a/sample/src/App.tsx
+++ b/sample/src/App.tsx
@@ -20,7 +20,11 @@ import {store} from './reduxApp';
 import {version as packageVersion} from '../../package.json';
 import {SENTRY_INTERNAL_DSN} from './dsn';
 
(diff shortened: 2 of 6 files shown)


bruno-garcia (approved):
Would be nice to see the usage from the sample how to propagate an override to the value

chrisfinch:
Nice, thanks @jennmueng 

👍  from me

PR #1387, opened 2021-03-11
fix(performance): remove @sentry/wizard dependency

doronpr:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x ] Refactoring

## :scroll: Description
fix(performance): remove @sentry/wizard dependencyfix(performance): remove @sentry/wizard dependency
Note: this *could* break projects that required `@sentry/wizard` or `@sentry/cli` and got them via this project (transitive dependency). These projects should add  `@sentry/wizard` or `@sentry/cli` as direct dependencies.

@sentry/wizard is redundant and currently brings in @sentri/cli which takes ~37mb (see below)
image

Tested in our (Wix) app.

## :pencil: Checklist

- [x ] I reviewed submitted code
- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
I recommend to bump up the major as this could be considered as a breaking change (see description)
Merge :) 🙏

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- fix(performance): remove @sentry/wizard dependency
 
 ## 2.3.0
 
--- a/package.json
+++ b/package.json
@@ -47,8 +47,7 @@
     "@sentry/react": "6.2.1",
     "@sentry/tracing": "6.2.1",
     "@sentry/types": "6.2.1",
-    "@sentry/utils": "6.2.1",
-    "@sentry/wizard": "^1.2.2"
+    "@sentry/utils": "6.2.1"
   },
   "devDependencies": {
     "@sentry-internal/eslint-config-sdk": "6.2.1",
@@ -72,8 +71,8 @@
   },
   "rnpm": {
     "commands": {
-      "postlink": "node node_modules/@sentry/wizard/dist/bin.js -i reactNative -p ios android",
-      "postunlink": "node node_modules/@sentry/wizard/dist/bin.js -i reactNative -p ios android --uninstall"
+      "postlink": "npx @sentry/wizard -i reactNative -p ios android",
+      "postunlink": "npx @sentry/wizard -i reactNative -p ios android --uninstall"
     },
     "android": {
       "packageInstance": "new RNSentryPackage()"
--- a/react-native.config.js
+++ b/react-native.config.js
@@ -10,9 +10,9 @@ module.exports = {
     },
     hooks: {
       postlink:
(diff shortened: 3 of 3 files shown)


doronpr:
Build failed on `caused by: sentry reported an error: Invalid token header. No credentials provided. (http status: 401)`
But I suspect it's not related to my change. 🤔

jennmueng:
The wizard dependency was added for the sake of ease of setup. Although I agree that this is used once, and you could just run it with `npx`, it reduces a step for people preferring to use `yarn`. The wizard itself is pretty small (1.4MB for me), so it shouldn't be an issue in my opinion.

However, that the `@sentry/cli` dependency would still be required for most people as they would need it to upload their sourcemaps via our automatic source map scripts, and you would need it anyways even if you upload them manually via script, so I don't think it makes sense to remove it.

I assume you guys are trimming your dependency tree over at @wix? 

_Note that the failing build is fine as it doesn't run on PRs not made by members of the @getsentry org._

doronpr:
Thanks for the detailed answer!

Regarding `yarn` it can be solved w/ `yarn dlx`. If you are open to it I can update the PR.
Devs normally don't use `@sentry/cli` in day to day scenario therefore I do think it makes sense for it to be installed on demand. And if one do need it, I think it's a bit enigmatic to relay on `sentry-react-native` - what if I want a specific version of the `cli`?  Doesn't look like this project offers any **direct** usage of the `cli` except for the one off usage of the wizard.
Anyway, not sure if I convinced you but your call obviously :)
Of course we don't include the `cli` in the app we ship, this issue is about hurting DX (developer experience).

doronpr:
> > Thanks for the detailed answer!
> > Regarding `yarn` it can be solved w/ `yarn dlx`. If you are open to it I can update the PR.
> > Devs normally don't use `@sentry/cli` in day to day scenario therefore I do think it makes sense for it to be installed on demand. And if one do need it, I think it's a bit enigmatic to relay on `sentry-react-native` - what if I want a specific version of the `cli`? Doesn't look like this project offers any **direct** usage of the `cli` except for the one off usage of the wizard.
> > Anyway, not sure if I convinced you but your call obviously :)
> > Of course we don't include the `cli` in the app we ship, this issue is about hurting DX (developer experience).
> 
> Regarding `@sentry/cli`, it is used in the background every time you build the app on both iOS/Android to auto-upload sourcemaps. I think it makes the developer experience better not requiring you to separately install it, and get sourcemaps and dsyms automatically.

Hmm, yes uploading sourcemaps is done via @sentry/cli but I'm not sure it justifies having it as a dependency as it's not an ability that this project provides (nor is it documented here as far as I can see).
Anyway we can agree to disagree ✌️ 😊 
I'm closing the PR

PR #1388, opened 2021-03-12
Prevent native crash by choosing release info from Sentry.init before…

ajp8164:
… native build settings (plist/gradle)

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ X] Refactoring

## :scroll: Description
Reverse order of preference for selection of build version and build number. Prefer the values passed into Sentry.init() over those set in Info.plist and/or build.gradle.

When using environment variables in Info.plist and build.gradle Sentry attempts to use the env variable placeholder text before it checks to see if Sentry.init() sets `release` and `dist`. Using the placeholder text raises a native exception resulting in a hard app crash. The native module for iOS is not rejecting when the when it fails to read from a dict. This change allows the "stronger" binding (via source) to be preferred over the externally defined values in build settings files.

Execution in a production app.

## :pencil: Checklist

- [ X] I reviewed submitted code
- [ X] All tests passing
- [ X] No breaking changes

### DIFF
--- a/src/js/integrations/release.ts
+++ b/src/js/integrations/release.ts
@@ -24,16 +24,6 @@ export class Release implements Integration {
         return event;
       }
 
-      try {
-        const release = await NATIVE.fetchRelease();
-        if (release) {
-          event.release = `${release.id}@${release.version}+${release.build}`;
-          event.dist = `${release.build}`;
-        }
-      } catch (_Oo) {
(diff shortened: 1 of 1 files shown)


jennmueng (approved):
This makes sense to me, good for catching this!

ajp8164:
Awesome! Thanks!

PR #1406, opened 2021-03-18
feat: Add onReady callback that gets called after Native SDK initialization

### LINKED ISSUE #1234: Missing symbolication for hermesFatalErrorHandler
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [ ]  iOS
- [x]  Android

**SDK:**

**SDK version:** 2.0.2

**`react-native` version:** 0.63.4

I am using 'io.sentry:sentry-android-gradle-plugin:1.7.36' plugin with this configuration:
 * uploadNativeSymbols true
 * includeNativeSources true

**Are you using Expo?**
- [ ]  Yes
- [x]  No

- [x]  sentry.io (SaaS)
- [ ]  on-premise

https://sentry.io/organizations/vincent-toupet/issues/2029861519/?project=5410474&query=is%3Aunresolved

---
**I have following issue:**

I have several fatal issues linked to (or in reference with) Hermes (facebook::hermes::detail::hermesFatalErrorHandle). Symbolication is missing even though I am using the Android gradle plugin to upload native symbols and native sources.

There is no way for me to debug this and traces are not symbolicated.

Additionally, would there be a way to add few lines of the logcat to the breadcrumbs? I haven't found anything within the documentation to do so.

jennmueng:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Adds a callback `onReady` to be passed as an option to `Sentry.init`. After the native SDK initialization step is finished, this callback is called with an object with property `nativeDidInitialize`. This value is `true` when the native SDK was initialized by the SDK, `false` otherwise.

In these cases:
- `enableNative: false` = onReady called with `{ nativeDidInitialize: false }`.
- `autoInitializeNativeSdk: false` = onReady called with `{ nativeDidInitialize: false }`.
- Error happens when initializing native SDK but JS layer is fine = onReady called with `{ nativeDidInitialize: false }`.
- Error happens when initializing JS layer SDK = onReady **not** called.

### How I got here
Originally was just going to be a callback after the native SDK initialized. But I figured having it still be called with `false` so you know it was not initialized is way more useful.

Resolves #1234

Added tests and tested with sample app.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Document this option

### DIFF
--- a/src/js/sdk.ts
+++ b/src/js/sdk.ts
@@ -29,12 +29,7 @@ const DEFAULT_OPTIONS: ReactNativeOptions = {
 /**
  * Inits the SDK
  */
-export function init(
-  passedOptions: ReactNativeOptions = {
-    enableNative: true,
-    enableNativeCrashHandling: true,
-  }
-): void {
+export function init(passedOptions: ReactNativeOptions): void {
   const reactNativeHub = new Hub(undefined, new ReactNativeScope());
   makeMain(reactNativeHub);
 
--- a/test/backend.test.ts
+++ b/test/backend.test.ts
@@ -93,22 +93,22 @@ describe("Tests ReactNativeBackend", () => {
     });
   });
 
-  describe("onNativeReady", () => {
-    test("calls onNativeReady callback with true if Native SDK is initialized", () => {
+  describe("onReady", () => {
+    test("calls onReady callback with true if Native SDK is initialized", () => {
       new ReactNativeBackend({
         dsn: EXAMPLE_DSN,
         enableNative: true,
-        onNativeReady: ({ nativeDidInitialize }) => {
+        onReady: ({ nativeDidInitialize }) => {
           expect(nativeDidInitialize).toBe(true);
         },
       });
     });
 
-    test("calls onNativeReady callback with false if Native SDK was not initialized", () => {
+    test("calls onReady callback with false if Native SDK was not initialized", () => {
       new ReactNativeBackend({
         dsn: EXAMPLE_DSN,
         enableNative: false,
-        onNativeReady: ({ nativeDidInitialize }) => {
+        onReady: ({ nativeDidInitialize }) => {
           expect(nativeDidInitialize).toBe(true);
         },
       });
--- a/test/backend.test.ts
+++ b/test/backend.test.ts
@@ -100,8 +100,8 @@ describe("Tests ReactNativeBackend", () => {
       new ReactNativeBackend({
         dsn: EXAMPLE_DSN,
         enableNative: true,
-        onReady: ({ nativeDidInitialize }) => {
-          expect(nativeDidInitialize).toBe(true);
+        onReady: ({ didCallNativeInit }) => {
+          expect(didCallNativeInit).toBe(true);
 
           done();
         },
@@ -112,8 +112,8 @@ describe("Tests ReactNativeBackend", () => {
       new ReactNativeBackend({
         dsn: EXAMPLE_DSN,
         enableNative: false,
-        onReady: ({ nativeDidInitialize }) => {
-          expect(nativeDidInitialize).toBe(false);
+        onReady: ({ didCallNativeInit }) => {
+          expect(didCallNativeInit).toBe(false);
 
           done();
         },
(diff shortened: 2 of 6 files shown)


test/backend.test.ts
```diff
@@ -87,6 +93,28 @@ describe("Tests ReactNativeBackend", () => {
...
+        },
+      });
+    });
+
+    test("calls onReady callback with false if Native SDK was not initialized", () => {
+      new ReactNativeBackend({
+        dsn: EXAMPLE_DSN,
+        enableNative: false,
+        onReady: ({ nativeDidInitialize }) => {
+          expect(nativeDidInitialize).toBe(true);
```
philipphofmann, line 112: `l`: Shouldn't this be `false`?

test/backend.test.ts
```diff
@@ -87,6 +93,28 @@ describe("Tests ReactNativeBackend", () => {
     });
   });
 
+  describe("onReady", () => {
```
philipphofmann, line 96: `l`: Can we also add a test for when initializing the native SDK fails.

src/js/sdk.ts
```diff
@@ -29,12 +29,7 @@ const DEFAULT_OPTIONS: ReactNativeOptions = {
 /**
  * Inits the SDK
  */
-export function init(
-  passedOptions: ReactNativeOptions = {
-    enableNative: true,
-    enableNativeCrashHandling: true,
```
philipphofmann, line 35: `m`: Why did we remove this?

philipphofmann (commented):
I don't fully understand why we need that callback. Shouldn't the SDK init be synchronous and you can call capture errors right after the SDK init?

jennmueng:
> I don't fully understand why we need that callback. Shouldn't the SDK init be synchronous and you can call capture errors right after the SDK init?

Although it's synchronous on the JS layer, due to native bridge calls being asynchronous, initing the native layer SDK is asynchronous.

src/js/sdk.ts
jennmueng: It was redundant given we merge it with `DEFAULT_OPTIONS` https://github.com/getsentry/sentry-react-native/blob/939e067be94cf3951855b05e3948a542c4b2e49e/src/js/sdk.ts#L22-L27

test/backend.test.ts
jennmueng: Huh yes this should be false, my mistake....although why did the tests pass then hmm

jennmueng: Forgot that jest does not wait for callbacks, need to use the `done` parameter for it to wait.

bruno-garcia (commented):
`didInitialize` sounds a bit misleading here. If I understood correctly what we're doing here is making sure the bridge called to native didn't throw an error. It could be that `init` was called into the Android SDK which resulted in a NoOp for some reason, or NDK/sentry-native didn't enable) and the callback would return 'didInitialize=true'.

Is this really what we need to communicate here? Perhaps rewording the property is a better approach? Maybe we check `NATIVE.isEnabled` on top to make sure the native layer is enabled? Or a different name for the property in the callback object?

jennmueng:
> `didInitialize` sounds a bit misleading here. If I understood correctly what we're doing here is making sure the bridge called to native didn't throw an error. It could be that `init` was called into the Android SDK which resulted in a NoOp for some reason, or NDK/sentry-native didn't enable) and the callback would return 'didInitialize=true'.
> 
> Is this really what we need to communicate here? Perhaps rewording the property is a better approach? Maybe we check `NATIVE.isEnabled` on top to make sure the native layer is enabled? Or a different name for the property in the callback object?

Hmm yes I agree this callback is more of a "Yes we called init on the native SDK". Hmm I don't have any idea what to name it then, any suggestions?

(e2e test failures should be unrelated, re-running to make sure)

bruno-garcia:
> > `didInitialize` sounds a bit misleading here. If I understood correctly what we're doing here is making sure the bridge called to native didn't throw an error. It could be that `init` was called into the Android SDK which resulted in a NoOp for some reason, or NDK/sentry-native didn't enable) and the callback would return 'didInitialize=true'.
> > Is this really what we need to communicate here? Perhaps rewording the property is a better approach? Maybe we check `NATIVE.isEnabled` on top to make sure the native layer is enabled? Or a different name for the property in the callback object?
> 
> Hmm yes I agree this callback is more of a "Yes we called init on the native SDK". Hmm I don't have any idea what to name it then, any suggestions?
> 
> (e2e test failures should be unrelated, re-running to make sure)

I see you already named `didCallNativeInit` which is much nicer! thanks!

bruno-garcia (approved):

PR #1453, opened 2021-04-15
fix: Type navigation container ref arguments as any to avoid Typescript errors

### LINKED ISSUE #1445: Passing the NavigationContainerRef to registerNavigationContainer raises a TS error
### Environment

How do you use Sentry?
using Sentry SaaS (sentry.io)

Which SDK and version?
`"@sentry/react-native": "^2.4.0"`
`"typescript": "^4.2.3"`

### Steps to Reproduce

Following the steps for instrumenting react navigation (https://docs.sentry.io/platforms/react-native/performance/included-instrumentation/#enable-routing-instrumentation)

setup:

image

image

image

### Expected Result

The instrumentation and setup would work with no TS errors

### Actual Result

I get a TS error when passing the `NavigationContainerRef` to `registerNavigationContainer`, I believe that it might be because the recent versions of TS are more strict?

```
Argument of type 'RefObject<NavigationContainerRef>' is not assignable to parameter of type 'NavigationContainerV5Ref | NavigationContainerV5'.
  Type 'RefObject<NavigationContainerRef>' is not assignable to type 'NavigationContainerV5Ref'.
    Types of property 'current' are incompatible.
      Type 'NavigationContainerRef | null' is not assignable to type 'NavigationContainerV5 | null'.
        Type 'NavigationContainerRef' is not assignable to type 'NavigationContainerV5 | null'.
          Type 'NavigationContainerRef' is not assignable to type 'NavigationContainerV5'.
            Types of property 'addListener' are incompatible.
              Type '<EventName extends keyof NavigationContainerEventMap>(type: EventName, callback: EventListenerCallback<NavigationContainerEventMap, EventName>) => () => void' is not assignable to type '(type: string, listener: () => void) => void'.
                Types of parameters 'type' and 'type' are incompatible.
                  Type 'string' is not assignable to type 'keyof NavigationContainerEventMap'.
```

jennmueng:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

When making integrations for other libraries, to avoid a dependency of the library we usually just define the types inline, however doing so means we have to manually keep them up to date otherwise a new update on the other library could cause new type errors, so I decided to just allow `any` types for the React Navigation container. We check the passed ref and warn the user if it is invalid, so it should be OK to accept an `any` type.

Resolves #1445

Eslint on CI to make sure the types work, and e2e tests to make sure nothing is broken.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fix: Type navigation container ref arguments as any to avoid Typescript errors #1453
+
 ## 2.4.0
 
 - fix: Don't call `NATIVE.fetchRelease` if release and dist already exists on the event #1388
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- fix: Type navigation container ref arguments as any to avoid Typescript errors #1453
+- fix: Type navigation container ref arguments as any to avoid TypeScript errors #1453
 
 ## 2.4.0
 
--- a/src/js/tracing/reactnavigationv4.ts
+++ b/src/js/tracing/reactnavigationv4.ts
@@ -112,9 +112,7 @@ class ReactNavigationV4Instrumentation extends RoutingInstrumentation {
    * Pass the ref to the app container to register it to the instrumentation
    * @param appContainerRef Ref to an `AppContainer`
    */
-  public registerAppContainer(
-    appContainerRef: AppContainerRef | AppContainerInstance
-  ): void {
+  public registerAppContainer(appContainerRef: any): void {
     const _global = getGlobalObject<{ __sentry_rn_v4_registered?: boolean }>();
 
     /* We prevent duplicate routing instrumentation to be initialized on fast refreshes
@@ -145,7 +143,7 @@ class ReactNavigationV4Instrumentation extends RoutingInstrumentation {
 
         _global.__sentry_rn_v4_registered = true;
       } else {
-        logger.log(
+        logger.warn(
           "[ReactNavigationV4Instrumentation] Received invalid app container ref!"
         );
       }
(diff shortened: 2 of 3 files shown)


marandaneto (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fix: Type navigation container ref arguments as any to avoid Typescript errors #1453
```
bruno-garcia, line 5: nit:
```suggestion
- fix: Type navigation container ref arguments as any to avoid TypeScript errors #1453
```

src/js/tracing/reactnavigationv4.ts
```diff
@@ -112,9 +108,8 @@ class ReactNavigationV4Instrumentation extends RoutingInstrumentation {
    * Pass the ref to the app container to register it to the instrumentation
    * @param appContainerRef Ref to an `AppContainer`
    */
-  public registerAppContainer(
-    appContainerRef: AppContainerRef | AppContainerInstance
-  ): void {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public registerAppContainer(appContainerRef: any): void {
```
bruno-garcia, line 112: The list of possible types is just not known? 
There isn't a base/common type we could add?

jennmueng: As it's a reference to the React Navigation component, the type is pretty complex union of multiple types which is why I felt like it would be tedious to copy over and define inline: https://github.com/react-navigation/react-navigation/blob/ae484782a6ff25ad2864693dfc4694e2b50cf524/packages/core/src/types.tsx#L468-L488

bruno-garcia (approved):

PR #1457, opened 2021-04-20
feat: Close method to fully disable the SDK on all layers

### LINKED ISSUE #1329: Ability to disable completely the SDK
Pass the call to Android and iOS to ensure all layers get closed.

Similar to https://github.com/getsentry/sentry-dart/issues/291

Depends on: 
https://github.com/getsentry/sentry-java/issues/1225
https://github.com/getsentry/sentry-cocoa/issues/934

jennmueng:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Exposes a new `close` method that closes the SDK. This includes the JavaScript layer and calls down to the native SDK layer.

Although this uses the latest Cocoa SDK `7.0.0`, it still uses Android `4.4.0-alpha.2`.

Resolves #1329 

Added a unit test for the wrapper, and a JS-layer integration test on the e2e test. The test only tests that the eventId is not set by beforeSend. Cannot do a full e2e test there is no way to determine how long to wait for an event to not show up on the dashboard.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/android/src/main/java/io/sentry/react/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/react/RNSentryModule.java
@@ -331,4 +331,10 @@ public class RNSentryModule extends ReactContextBaseJavaModule {
             scope.setTag(key, value);
         });
     }
+
+    @ReactMethod
+    public void closeNativeSdk(Promise promise) {
+      Sentry.close();
+      promise.resolve(true);
+    }
 }
--- a/ios/RNSentry.m
+++ b/ios/RNSentry.m
@@ -17,102 +17,91 @@
 @end
 
 @implementation RNSentry {
-  bool sentHybridSdkDidBecomeActive;
+   bool sentHybridSdkDidBecomeActive;
 }
 
-- (dispatch_queue_t)methodQueue {
-  return dispatch_get_main_queue();
+
+
+- (dispatch_queue_t)methodQueue
+{
+    return dispatch_get_main_queue();
 }
 
 + (BOOL)requiresMainQueueSetup {
-  return YES;
+    return YES;
 }
 
 RCT_EXPORT_MODULE()
 
-- (NSDictionary<NSString *, id> *)constantsToExport {
-  return @{@"nativeClientAvailable" : @YES, @"nativeTransport" : @YES};
+- (NSDictionary<NSString *, id> *)constantsToExport
+{
+    return @{@"nativeClientAvailable": @YES, @"nativeTransport": @YES};
 }
 
-RCT_EXPORT_METHOD(startWithOptions
-                  : (NSDictionary *_Nonnull)options resolve
-                  : (RCTPromiseResolveBlock)resolve rejecter
-                  : (RCTPromiseRejectBlock)reject) {
-  NSError *error = nil;
+RCT_EXPORT_METHOD(startWithOptions:(NSDictionary *_Nonnull)options
+                  resolve:(RCTPromiseResolveBlock)resolve
+                  rejecter:(RCTPromiseRejectBlock)reject)
+{
+    NSError *error = nil;
 
-  SentryBeforeSendEventCallback beforeSend =
-      ^SentryEvent *(SentryEvent *event) {
-    // We don't want to send an event after startup that came from a Unhandled
-    // JS Exception of react native Because we sent it already before the app
-    // crashed.
-    if (nil != event.exceptions.firstObject.type &&
-        [event.exceptions.firstObject.type
-            rangeOfString:@"Unhandled JS Exception"]
-                .location != NSNotFound) {
-      NSLog(@"Unhandled JS Exception");
-      return nil;
+    SentryBeforeSendEventCallback beforeSend = ^SentryEvent*(SentryEvent *event) {
+        // We don't want to send an event after startup that came from a Unhandled JS Exception of react native
+        // Because we sent it already before the app crashed.
+        if (nil != event.exceptions.firstObject.type &&
+            [event.exceptions.firstObject.type rangeOfString:@"Unhandled JS Exception"].location != NSNotFound) {
+            NSLog(@"Unhandled JS Exception");
+            return nil;
+        }
+
+        // set the event.origin tag to be ios if the event originated from the sentry-cocoa sdk.
+        if (event.sdk && [event.sdk[@"name"] isEqualToString:@"sentry.cocoa"]) {
+            NSMutableDictionary *newTags = [NSMutableDictionary new];
+            [newTags addEntriesFromDictionary:event.tags];
+            [newTags setValue:@"ios" forKey:@"event.origin"];
+            event.tags = newTags;
+        }
+
+        return event;
+    };
+
+    [options setValue:beforeSend forKey:@"beforeSend"];
+
+    SentryOptions *sentryOptions = [[SentryOptions alloc] initWithDict:options didFailWithError:&error];
+    if (error) {
+        reject(@"SentryReactNative", error.localizedDescription, error);
+        return;
+    }
+    [SentrySDK startWithOptionsObject:sentryOptions];
+
+    // If the app is active/in foreground, and we have not sent the SentryHybridSdkDidBecomeActive notification, send it.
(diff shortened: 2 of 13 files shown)


sample/src/App.tsx
```diff
@@ -29,7 +29,6 @@ const reactNavigationV5Instrumentation = new Sentry.ReactNavigationV5Instrumenta
 Sentry.init({
   // Replace the example DSN below with your own DSN:
   dsn: SENTRY_INTERNAL_DSN,
-  debug: true,
```
jennmueng, line 32: Note: If `close` is called and debug mode is true, then sending any events afterwards would always show an error that the SDK is disabled. Do we want this?

jennmueng: @marandaneto What is the behavior on flutter? cc @bruno-garcia

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -3,18 +3,28 @@
...
 import com.facebook.react.bridge.ReactContextBaseJavaModule;
 import com.facebook.react.bridge.ReactMethod;
+import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.ReadableMapKeySetIterator;
-import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.WritableMap;
 import com.facebook.react.module.annotations.ReactModule;
-
+import io.sentry.Breadcrumb;
```
marandaneto, line 16: it's hard to review this class because everything changed, could you roll back the formatting? what actually changed?

ios/RNSentry.m
```diff
@@ -9,261 +9,277 @@
 #import <Sentry/Sentry.h>
 
 @implementation RNSentry {
-   bool sentHybridSdkDidBecomeActive;
+  bool sentHybridSdkDidBecomeActive;
```
marandaneto, line 12: same here, formatting

sample/src/App.tsx
marandaneto: on Java and Flutter do nothing because if you call close, the Hub bound to the Sentry static class is NoOp

src/js/sdk.ts
```diff
@@ -118,3 +118,18 @@ export function nativeCrash(): void {
...
+ */
+export async function close(): Promise<void> {
+  try {
+    const client = getCurrentHub().getClient<ReactNativeClient>();
+
+    if (client) {
+      await client.close();
+    }
+  } catch (e) {
+    throw new SentryError("Failed to close the SDK");
```
bruno-garcia, line 133: We shouldn't throw here though. Lets just log with Error level but we shouldn't be potentially crashing the app because the SDK failed to close

ios/RNSentry.m
```diff
@@ -266,4 +266,10 @@ + (BOOL)requiresMainQueueSetup {
     [SentrySDK crash];
 }
 
+RCT_EXPORT_METHOD(closeNativeSdk
+                  : (RCTPromiseResolveBlock)resolve rejecter
+                  : (RCTPromiseRejectBlock)reject) {
+  [[SentrySDK currentHub] getClient].options.enabled = NO;
```
marandaneto, line 272: see this https://github.com/getsentry/sentry-dart/pull/424/files#r617398293
using latest Cocoa beta

jennmueng: Made it a separate PR #1459

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -72,6 +72,10 @@ public String getName() {
     @ReactMethod
     public void startWithOptions(final ReadableMap rnOptions, Promise promise) {
         SentryAndroid.init(this.getReactApplicationContext(), options -> {
+            if (rnOptions.hasKey("debug") && rnOptions.getBoolean("debug")) {
+                options.setDebug(true);
+                logger.setLevel(Level.INFO);
```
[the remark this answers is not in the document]
marandaneto: yep, if they want to change the level in the native SDK, they can set the `diagnosticLevel` as well.
but `logger` here is specifically about `RNSentryModule` and not Sentry's SDK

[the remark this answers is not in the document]
marandaneto: https://github.com/getsentry/sentry-dart/blob/main/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt#L91-L96

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -214,19 +210,90 @@ private static PackageInfo getPackageInfo(Context ctx) {
...
+      logger.info(String.format("Native Integrations '%s'",
+                                options.getIntegrations().toString()));
+      sentryOptions = options;
+    });
+
+    promise.resolve(true);
+  }
+
+  @ReactMethod
+  public void setLogLevel(int level) {
```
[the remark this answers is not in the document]
marandaneto: yep, and propagate it to the Native SDK https://github.com/getsentry/sentry-dart/blob/main/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt#L91-L96

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -214,19 +210,90 @@ private static PackageInfo getPackageInfo(Context ctx) {
...
+    throw new RuntimeException(
+        "TEST - Sentry Client Crash (only works in release mode)");
+  }
+
+  @ReactMethod
+  public void fetchRelease(Promise promise) {
+    WritableMap release = Arguments.createMap();
+    release.putString("id", packageInfo.packageName);
+    release.putString("version", packageInfo.versionName);
+    release.putString("build", String.valueOf(packageInfo.versionCode));
```
marandaneto, line 239: versionCode is deprecated depending OS version, see https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/ContextUtils.java#L30-L47

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -214,19 +210,90 @@ private static PackageInfo getPackageInfo(Context ctx) {
...
+  @ReactMethod
+  public void captureEnvelope(String envelope, Promise promise) {
+    try {
+      File installation =
+          new File(sentryOptions.getOutboxPath(), UUID.randomUUID().toString());
+      try (FileOutputStream out = new FileOutputStream(installation)) {
+        out.write(envelope.getBytes(Charset.forName("UTF-8")));
+      }
+    } catch (Exception e) {
+      logger.info("Error reading envelope");
```
marandaneto, line 252: maybe this should be more than just `info`

android/src/main/java/io/sentry/react/RNSentryModule.java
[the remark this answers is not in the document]
jennmueng: Yes as @marandaneto this is only to enable the logs from the `RNSentryModule`, let's add `diagnosticLevel` in a separate PR.

bruno-garcia (approved):
Thanks!

Overtorment:
this is not released yet, correct?
is that gonna be the recommended way to disable Sentry in case the user opts out of all analytics/tracking in the app?

bruno-garcia:
> this is not released yet, correct?
> is that gonna be the recommended way to disable Sentry in case the user opts out of all analytics/tracking in the app?

It's released as `2.5.0-beta.1`. For now this is the way to go.

We plan to add a proper `user consent` API if you must be able to "toggle" through approved and non-approved.
Currently calling `Init` then `Close` then `Init` isn't guaranteed to work completely (`Init` is designed to be called only once).

But to simply turn off Sentry upon user request, yes `Close` will do that for you

PR #1460, opened 2021-04-22
ci: Add CodeQL scanning

rhcarvalho:
Similar to what we've added to other repos in the org.

#skip-changelog

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'java', 'javascript' ]
+        language: [ 'javascript' ]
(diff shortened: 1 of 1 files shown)


rhcarvalho:
Build error for the Java part in https://github.com/getsentry/sentry-react-native/pull/1460/checks?check_run_id=2413005852

@jennmueng could you please have a look and help me update the step to build the java code in this repo?

Reference https://docs.github.com/en/code-security/secure-coding/configuring-the-codeql-workflow-for-compiled-languages#java

jennmueng:
@rhcarvalho Looks like it's missing Android libraries. Usually you would need to build the Android code not from the SDK's android folder but the React Native app's android folder, in this case `./sample/android`. However does CodeQL work with Java code for Android?

rhcarvalho:
Thanks, @jennmueng!

> However does CodeQL work with Java code for Android?

I don't know, so I'll just disable it for now.

jennmueng (approved):

PR #1514, opened 2021-05-12
Native + JS Startup Time Measurement (Do Not Merge)

jennmueng:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This is just a quick proof of concept on which points RN would use to measure startup time...**The first two timestamp points will need to be retrieved from the Native SDK and not like what is here currently**

Measures the startup time in a transaction. Creates a parent transaction that contains the app's full startup time, with two child spans, the first being the native layer startup time, and the second the js layer startup time.

Screen Shot 2021-05-12 at 8 19 07 PM

The timestamp points used are the constructor of the RNSentry module as the native layer started point, and the global closure of the JS code as the js layer started point. The first pre-start timestamp is currently a temporary native layer started - 2000ms.

This proof of concept code temporarily lies in the backend class, where it shouldn't be and will be moved somewhere more suitable. It also does not handle the case of a refreshed js bundle causing a new js "startup" 

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/android/src/main/java/io/sentry/react/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/react/RNSentryModule.java
@@ -51,9 +51,13 @@ public class RNSentryModule extends ReactContextBaseJavaModule {
     private static PackageInfo packageInfo;
     private SentryOptions sentryOptions;
 
+  private long nativeStartedTime;
+
     public RNSentryModule(ReactApplicationContext reactContext) {
         super(reactContext);
         RNSentryModule.packageInfo = getPackageInfo(reactContext);
+
+    nativeStartedTime = System.currentTimeMillis();
     }
 
     @Override
@@ -166,6 +170,16 @@ public class RNSentryModule extends ReactContextBaseJavaModule {
 
         promise.resolve(true);
     }
+
+  @ReactMethod
+  public void getNativeStartupTimestamps(Promise promise) {
+    WritableMap response = Arguments.createMap();
+
+    response.putDouble("nativeStartedTime", (double)nativeStartedTime);
+    response.putDouble("nativeStartTime", (double)(nativeStartedTime - 2000));
+
+    promise.resolve(response);
+  }
 
     @ReactMethod
     public void crash() {
--- a/src/js/backend.ts
+++ b/src/js/backend.ts
@@ -1,13 +1,16 @@
 import { BrowserBackend } from "@sentry/browser/dist/backend";
-import { BaseBackend, NoopTransport } from "@sentry/core";
+import { BaseBackend, getCurrentHub, NoopTransport } from "@sentry/core";
 import { BrowserOptions, Transports } from "@sentry/react";
 import { Event, EventHint, Severity, Transport } from "@sentry/types";
+import { timestampInSeconds } from "@sentry/utils";
 // @ts-ignore LogBox introduced in RN 0.63
 import { Alert, LogBox, YellowBox } from "react-native";
 
 import { ReactNativeOptions } from "./options";
 import { NativeTransport } from "./transports/native";
 import { NATIVE } from "./wrapper";
+
+const jsStartedTimestampSeconds = timestampInSeconds();
 
 /** The Sentry ReactNative SDK Backend. */
 export class ReactNativeBackend extends BaseBackend<BrowserOptions> {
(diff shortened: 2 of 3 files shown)


android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -51,9 +51,13 @@
     private static PackageInfo packageInfo;
     private SentryOptions sentryOptions;
 
+  private long nativeStartedTime;
+
     public RNSentryModule(ReactApplicationContext reactContext) {
         super(reactContext);
         RNSentryModule.packageInfo = getPackageInfo(reactContext);
+
+    nativeStartedTime = System.currentTimeMillis();
```
marandaneto, line 60: that's too late, maybe you should read the app start up from the Android SDK, but its not merged yet

marandaneto: this also does not account for warm/cold start up

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -167,6 +171,16 @@ public void startWithOptions(final ReadableMap rnOptions, Promise promise) {
         promise.resolve(true);
     }
 
+  @ReactMethod
+  public void getNativeStartupTimestamps(Promise promise) {
+    WritableMap response = Arguments.createMap();
+
+    response.putDouble("nativeStartedTime", (double)nativeStartedTime);
+    response.putDouble("nativeStartTime", (double)(nativeStartedTime - 2000));
```
marandaneto, line 179: magic number?

src/js/backend.ts
```diff
@@ -1,14 +1,17 @@
...
 import { Event, EventHint, Severity, Transport } from "@sentry/types";
+import { timestampInSeconds } from "@sentry/utils";
 // @ts-ignore LogBox introduced in RN 0.63
 import { Alert, LogBox, YellowBox } from "react-native";
 
 import { ReactNativeOptions } from "./options";
 import { NativeTransport } from "./transports/native";
 import { NATIVE } from "./wrapper";
 
+const jsStartedTimestampSeconds = timestampInSeconds();
```
marandaneto, line 13: nativeStartedTime is in order of millis, this one in order of seconds, can we get millis precision?

android/src/main/java/io/sentry/react/RNSentryModule.java
jennmueng: Oh this is just a proof of concept of how JS would handle it, these numbers would be taken from the native SDK

src/js/backend.ts
```diff
@@ -115,6 +118,38 @@ export class ReactNativeBackend extends BaseBackend<BrowserOptions> {
...
     this._options.onReady?.({ didCallNativeInit });
+
+    if (didCallNativeInit) {
+      const {
+        nativeStartTime,
+        nativeStartedTime,
+      } = await NATIVE.getNativeStartupTimestamps();
+
+      const nativeStartTimeSeconds = nativeStartTime / 1000;
+      const nativeStartedTimeSeconds = nativeStartedTime / 1000;
```
marandaneto, line 129: we want millis precision

src/js/backend.ts
```diff
@@ -115,6 +118,38 @@ export class ReactNativeBackend extends BaseBackend<BrowserOptions> {
...
+        op: "startup",
+        startTimestamp: nativeStartTimeSeconds,
+        sampled: true,
+      });
+
+      t.startChild({
+        description: "Native App Start",
+        op: "native-startup",
+        startTimestamp: nativeStartTimeSeconds,
+      }).finish(nativeStartedTimeSeconds);
```
marandaneto, line 143: is there a finish method that takes an interval in JS? I guess its not a documented API and something we'd like to have too

src/js/backend.ts
jennmueng: The timestamps in @sentry/tracing use seconds with 3 decimals for millis, `timestampInSeconds` returns such

src/js/backend.ts
```diff
@@ -115,6 +118,38 @@ export class ReactNativeBackend extends BaseBackend<BrowserOptions> {
...
+    if (didCallNativeInit) {
+      const {
+        nativeStartTime,
+        nativeStartedTime,
+      } = await NATIVE.getNativeStartupTimestamps();
+
+      const nativeStartTimeSeconds = nativeStartTime / 1000;
+      const nativeStartedTimeSeconds = nativeStartedTime / 1000;
+
+      const t = getCurrentHub().startTransaction({
```
marandaneto, line 131: these are metrics, not a transaction, so ideally we'd not start transaction and spans, but rather sent along with the App's transactions

android/src/main/java/io/sentry/react/RNSentryModule.java
bruno-garcia: Just a PoC, data would come from native

PR #1529, opened 2021-05-17
fix: Use the latest outbox path from hub options instead of private options

### LINKED ISSUE #1516: autoInitializeNativeSdk is broken on Android
### Environment

How do you use Sentry?

Sentry SaaS (sentry.io)

Which SDK and version?

@sentry/react-native@2.4.2

### Steps to Reproduce

We initialize Android sentry SDK in MainApplication to properly catch early native errors, because of this we set `autoInitializeNativeSdk` to `false` on JS side.

### Expected Result

Events properly sent to sentry server.

### Actual Result

No event sent to server. Error from console:
```
react-native-sentry: Error reading envelope
```

It is because of [RNSentryModule.captureEnvelope](https://github.com/getsentry/sentry-react-native/blob/2.4.2/android/src/main/java/io/sentry/react/RNSentryModule.java#L191) trying to get path from private `sentryOptions` variable, that can be settled only from `startWithOptions` which skipped because of `autoInitializeNativeSdk`.

### Suggestion fix

May be we can factor out native initialization code to singleton that can be accessed from native application side, so application will call it before RN module initialized?

jennmueng:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Gets the latest outbox path from the hub through `HubAdapter` instead of a private variable like before. This fixes the issue of not having the correct outbox path or not having it entirely if the user decides to init the SDK elsewhere and disabling auto init using `autoInitNativeSdk: false`

Fixes #1516 

Android Sample + e2e tests (waiting)

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Maybe release this fix before the beta as `2.4.3`.

### DIFF
--- a/android/src/main/java/io/sentry/react/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/react/RNSentryModule.java
@@ -33,6 +33,7 @@ import io.sentry.android.core.NdkIntegration;
 import io.sentry.android.core.SentryAndroid;
 import io.sentry.Sentry;
 import io.sentry.Breadcrumb;
+import io.sentry.HubAdapter;
 import io.sentry.Integration;
 import io.sentry.SentryLevel;
 import io.sentry.SentryOptions;
@@ -49,7 +50,6 @@ public class RNSentryModule extends ReactContextBaseJavaModule {
     final static Logger logger = Logger.getLogger("react-native-sentry");
 
     private static PackageInfo packageInfo;
-    private SentryOptions sentryOptions;
 
     public RNSentryModule(ReactApplicationContext reactContext) {
         super(reactContext);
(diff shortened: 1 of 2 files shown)


android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -184,7 +183,8 @@ public void fetchRelease(Promise promise) {
     @ReactMethod
     public void captureEnvelope(String envelope, Promise promise) {
         try {
-            File installation = new File(sentryOptions.getOutboxPath(), UUID.randomUUID().toString());
+      String outboxPath = HubAdapter.getInstance().getOptions().getOutboxPath();
```
marandaneto, line 186:
```suggestion
      final String outboxPath = HubAdapter.getInstance().getOptions().getOutboxPath();
      if (outboxPath == null) {
          logger.severe("outboxPath isn't set.");
      }
```

marandaneto: Let's log if outboxPath isn't set yet, that means SDK is yet not inited, the logger message could be improved.

marandaneto (approved):

PR #1542, opened 2021-05-25
feat: Track stalls in the JavaScript event loop as measurements

jennmueng:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

### Important TL;DR

- Tracks stalls in the JavaScript event loop as measurements attached to transactions.
- Is **enabled by default** as a default integration. Can be disabled by disabling the default integration.
- Will actually work on other Javascript platforms other than React Native. We should consider expending it if it turns out to work well.
- Monkey patches the `hub.startTransaction` and `transaction.finish` along with routing idleTransactions through `ReactNativeTracing` to instrument every transaction.
- Supports the `trimEnd` functionality where the end timestamp of the transaction is set to the last finished span.
- Does not support `startTimetamp` and `endTimestamp` being custom set.
- If you want to see it in action running on JS, try logging `measurements` in the if statement on line 32 here when you run the jest tests: https://github.com/getsentry/sentry-react-native/blob/0d573be96eb6865db6fcaaf4c0a4554d54d29247/test/integrations/stalltracking.test.ts#L6-L40

### How it works

Measures the stall time in the JS event loop by setting a timeout and seeing how long it strays from the expected time that it's supposed to run. By default, the "`minimumStallThreshold `" is set to `50ms`, with each timeout running every  `50ms` which allows a margin of error of `minimumStallThreshold  = 50ms` before the straying is considered a stall. The stall time is then calculated by taking the total difference from the time the timeout is supposed to be run — `totalTimeTaken - 50ms`.

This is inspired by the `JSEventLoopWatchdog` [used internally in React Native]( https://github.com/facebook/react-native/blob/006f5afe120c290a37cf6ff896748fbc062bf7ed/Libraries/Interaction/JSEventLoopWatchdog.js).

#### Spans

We monkey patch the finish method of spans that have been added to the `spanRecorder`. Every time a span inside a transaction is finished, we log the last end timestamp in the chance that `trimEnd` is used and the transaction's end timestamp is set to the last span's end timestamp.

### Example Measurement
```json
{
	"measurements": {
	  "stall_count": { 
			"value": 2 
		},
	  "stall_total_time": { 
			"value": 247.875 
		}, 
	  "stall_longest_time": { 
			"value": 175.125 
		},
	}
}
```

### Why we don't use this method to measure slow and frozen JS frames:
We would need to set the timeout to 16.67ms (for 60fps) and log any iterations that take longer than that. However, in practice when doing so it ends up being incredibly flakey, with the majority of the event loop iterations taking longer than the 16.67ms needed for 60fps and it ends up logging almost all frames as slow. This is not accurate to what the fps performance monitor is logging when using RN in dev mode.

Tracking JS event loop stalls should be correlated to JS Frame drops. Although we would prefer to track frame drops to align the behavior with the native SDKs, this is the only way to track a measurement that represents app slowdowns on the JS layer that I can find.

This method is inspired by how [React Native measures stalls internally.]( https://github.com/facebook/react-native/blob/b176617d1e0cb7834698f147db4522290098ea00/Libraries/Interaction/JSEventLoopWatchdog.js)

**Extensive** unit tests, tested on Android and iOS sample app.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## Next Steps

We also discussed logging each stall as a child span, which I believe would be massively helpful in pinpointing exactly when slowdowns happened in your app.


src/js/measurements.ts
```diff
@@ -0,0 +1,143 @@
...
+    if (busyTime >= 80) {
+      const stallTime = busyTime - 80;
+      this._stallCount += 1;
+      this._totalStallTime += stallTime;
+      this._longestStall = Math.max(this._longestStall, stallTime);
+    }
+
+    this._lastInterval = now;
+    this._timeout = setTimeout(this._iteration.bind(this), 40);
+  }
```
jennmueng, line 66: Measures the stall duration via setting a timeout to run, in this case, in 40ms, and then measures if the actual "busy time" is greater than 80ms. These numbers are just hard coded and arbitrary for now.

In theory we could measure slow and frozen frames using this by running the timeout every 16.667ms (60FPS / 1000ms) given we determine what duration constitutes "slow" frames and "frozen" frames. We should also note that running the timer every 16.667ms could have an effect on the performance of the app itself.

src/js/measurements.ts
```diff
@@ -0,0 +1,143 @@
...
+    ]);
+
+    const stalls = new Stalls();
+    stalls.start();
+
+    // eslint-disable-next-line @typescript-eslint/unbound-method
+    const originalFinish = transaction.finish;
+
+    transaction.finish = (endTimestamp: number | undefined) => {
+      if (
```
jennmueng, line 125: If a transaction has a custom set `endTimestamp` we do not add the stall measurements. Should do the same with and check if `startTimestamp` was custom set.

src/js/measurements.ts
```diff
@@ -0,0 +1,143 @@
...
+
+      /*
+        Overwrites the transaction start and finish to start and finish stall tracking.
+        Preferably instead of overwriting add a callback method for this in the Transaction itself.
+      */
+      const _startTransaction = _patchStartTransaction(
+        originalStartTransaction
+      );
+
+      carrier.__SENTRY__.extensions.startTransaction = _startTransaction;
```
jennmueng, line 88: Here, we patch `hub.startTransaction()`, which is also aliased as `Sentry.startTransaction()`.

src/js/measurements.ts
```diff
@@ -0,0 +1,159 @@
...
+    transactionContext: TransactionContext,
+    customSamplingContext?: CustomSamplingContext
+  ): Transaction {
+    const transaction: Transaction = originalStartTransaction.apply(this, [
+      transactionContext,
+      customSamplingContext,
+    ]);
+
+    const stalls = new Stalls();
+    stalls.start();
```
bruno-garcia, line 129: If we get multiple concurrent transactions each one will have its own tracker, right>
So multiple jobs being queued for execution to measure latency?

jennmueng: Hmm yes, I realized that can have an effect on performance. If we decide to have one global one, it should not be always tracking and only when some transaction is running, and is closed when all the transactions are finished. Should I go with that?

jennmueng:
As discussed on the grooming, this will be updated to send

```json
{
	"measurements": {
	  "stall_count": { 
			"value": 2 
		},
	  "stall_total_time": { 
			"value": 247.875 
		}, 
	  "stall_longest_time": { 
			"value": 175.125 
		},
	}
}

```

src/js/measurements.ts
bruno-garcia: It would be awesome to have this. If it doesn't degrade performance in any obvious way it sounds good. It must have an opt-out though so folks can turn this off if it becomes an issue

jennmueng: Added this feature.

jennmueng:
Requested review from @getsentry/team-webplatform due to how heavy JS this PR is along with being applicable to the rest of the JS universe.

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,328 @@
+/* eslint-disable max-lines */
```
AbhiPrasad, line 1: This is a nice approach - will do a more detailed review in a bit. What do you think we need to do to move this into sentry-javascript.

Also the monkey patching is not ideal, we should update tracing to make it easier to patch rather than patch the idle transaction this way tbh.

jennmueng: Yeah I strongly agree the monkey patching is bad, but we reckoned that we're going to release this as a prerelease to see how it goes, if it's actually effective, and if it's valuable to customers, if it turns out well we can then move it over to the javascript repo and solidify it by updating tracing to actually support hooks to the start and finish methods.

AbhiPrasad: Sounds like a plan 👍

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,328 @@
...
+      this._runningTransactions.some((t) => t.spanId === transaction.spanId)
+    ) {
+      logger.error(
+        "[StallTracking] Tried to start stall tracking on a transaction already being tracked. Measurements might be lost."
+      );
+
+      return () => {
+        // noop
+        return null;
+      };
```
AbhiPrasad, line 90:
```suggestion
      // noop
      return () => null;
```

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,328 @@
...
+        statsAtTimestampPerTransaction.set(
+          spanEndTimestamp,
+          this._getCurrentStats(transactionId)
+        );
+      }
+
+      // We delete the first span (earliest timestamp) if there are greater than N span stats in the map.
+      if (statsAtTimestampPerTransaction.size > MAX_SPAN_STATS_LEN) {
+        const [key] = statsAtTimestampPerTransaction.keys();
+        statsAtTimestampPerTransaction.delete(key);
```
AbhiPrasad, line 217: Nice, forgot that maps keep insertion order

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,328 @@
...
+
+    if (statsForTransaction) {
+      const exactAtTimestamp = statsForTransaction.get(endTimestamp);
+
+      if (exactAtTimestamp) {
+        return exactAtTimestamp;
+      }
+
+      const statsWithinMargin: [number, StallMeasurements][] = [];
+      for (const [timestamp, stats] of statsForTransaction.entries()) {
```
AbhiPrasad, line 240: Can we use an `array.reduce` here?

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,328 @@
...
+          Math.max(
+            this._longestStallsByTransaction.get(transaction.spanId) ?? 0,
+            stallTime
+          )
+        );
+      });
+    }
+
+    this._lastInterval = now;
+    this._timeout = setTimeout(this._iteration.bind(this), timeoutDuration);
```
AbhiPrasad, line 326: Should we have a `_isTracking` check here?

test/integrations/stalltracking.test.ts
```diff
@@ -0,0 +1,333 @@
...
+    });
+    transaction.initSpanRecorder();
+
+    const finishTracking = stallTracking.registerTransactionStart(transaction);
+
+    // We create a very **nasty** object
+    const nastyObject: { value: string[] } = { value: [] };
+    for (let i = 0; i < 100000; i++) {
+      nastyObject.value.push("nasty");
+    }
```
AbhiPrasad, line 20:
```suggestion
    const expensiveObject: { value: string[] } = { value: Array(100000).fill("expensive") };
```

test/integrations/stalltracking.test.ts
```diff
@@ -0,0 +1,333 @@
...
+      name: "Test Transaction",
+    });
+
+    const finishTracking = stallTracking.registerTransactionStart(transaction);
+
+    const measurements = finishTracking();
+
+    expect(measurements).not.toBe(null);
+
+    // If the stall tracking does not correctly stop, the process will keep running. We detect this by passing --detectOpenHandles to jest.
```
AbhiPrasad, line 56: Can we use jest's `done()` callback here?

AbhiPrasad (changes requested):
Give me some more time to think about this implementation, but on a high level I like the direction.

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,328 @@
...
+}
+
+/** Margin of error of 20ms */
+const MARGIN_OF_ERROR_SECONDS = 0.02;
+/** Max number of span stats stored */
+const MAX_SPAN_STATS_LEN = 20;
+
+/**
+ * Ensures the timestamp is in seconds. This is because the endTimestamp
+ * passed to .finish() could be seconds or milliseconds.
```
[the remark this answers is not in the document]
jennmueng: From the user point of view don't we support users passing either ms or sec or did I misunderstand?

src/js/integrations/stalltracking.ts
jennmueng: Haha yeah when I was about to write this and go through the keys I was like wait, they're already in insertion order lol

src/js/integrations/stalltracking.ts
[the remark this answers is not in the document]
AbhiPrasad: Will defer to @rhcarvalho for the most correct answer here

package.json
```diff
@@ -14,7 +14,7 @@
     "build": "tsc -p tsconfig.build.json",
     "build:watch": "tsc -p tsconfig.build.json -w --preserveWatchOutput",
     "clean": "rimraf dist coverage",
-    "test": "jest",
+    "test": "jest --detectOpenHandles",
```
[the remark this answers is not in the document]
jennmueng: Yeah and it's hanging the CI right now so it's not good. I'm in favor of removing it, but maybe we can just expose `_timeout` and `_isTracking` to jest test that the timeout iterations have stopped?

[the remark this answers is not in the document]
jennmueng: Or with an `_isTracking` check in `iteration` just check for `isTracking: false`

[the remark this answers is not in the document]
AbhiPrasad: I'm a +1 on `_isTracking` check

test/integrations/stalltracking.test.ts
jennmueng: Oh yeah I somehow forgot all about `done()` 😅

src/js/index.ts
```diff
@@ -39,6 +39,10 @@ export {
 // aka. this has side effects
 import "@sentry/tracing";
 
+// Add the React Native SDK's own tracing extensions, this needs to happen AFTER @sentry/tracing's
+import { _addTracingExtensions } from "./measurements";
+_addTracingExtensions();
```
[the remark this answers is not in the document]
jennmueng: Hmm I'm not too familiar with side effects (or tree shaking for that matter), is this only a web/webpack thing?

jennmueng:
After some deliberating. I will remove the transaction `endTimestamp` support and reduce the span finish tracking to just the last one which will greatly reduce complexity.

src/js/index.ts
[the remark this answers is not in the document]
AbhiPrasad: Yeah it's a webpack thing - no idea if it applies in React native land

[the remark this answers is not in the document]
jennmueng: If that's the case, then we're good as React Native doesn't use webpack but rather their own bundler—Metro

jennmueng:
- Removed the complex support for `endTimestamp` and instead we just support `trimEnd` being used. Although due to the different behavior regarding unfinished spans, we handle idleTransactions and normal transactions differently. Please double check the logic I used for this here:
https://github.com/getsentry/sentry-react-native/blob/9ddd3e9caaa7ab4d9ee668f4e4364104f23a58a5/src/js/integrations/stalltracking.ts#L149-L187

- Updated tests to test the latest implementation along with adding new tests:
  - Multiple stalls are counted for
  - Idletransaction with unfinished spans
  - Unfinished spans in idle transactions
  - Only measures stalls within the transaction duration when `trimEnd` is used

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,336 @@
...
+  ): StallMeasurements | null {
+    const endTimestamp = passedEndTimestamp ?? transaction.endTimestamp;
+
+    const finishedSpanCount = transaction.spanRecorder
+      ? transaction.spanRecorder.spans.reduce(
+          (count, s) =>
+            s !== transaction && s.endTimestamp ? count + 1 : count,
+          0
+        )
+      : 0;
```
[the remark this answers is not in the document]
AbhiPrasad: Also means we can remove the ternary on line 159.

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,336 @@
...
+  }
+
+  /**
+   * Iteration of the stall tracking interval. Measures how long the timer strayed from its expected time of running, and how
+   * long the stall is for.
+   */
+  private _iteration(): void {
+    const now = timestampInSeconds() * 1000;
+    const totalTimeTaken = now - this._lastIntervalMs;
+    const timeoutDuration = this._acceptableBusyTime / 2;
```
AbhiPrasad, line 312: Can we leave a comment why timeoutDuration is half of the acceptable busy time? I kind of understand why, but I'm sure future readers would appreciate the context.

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,336 @@
...
+  /**
+   * Clears all the collected stats
+   */
+  private _reset(): void {
+    this._stallCount = 0;
+    this._totalStallTime = 0;
+    this._lastIntervalMs = 0;
+    this._longestStallsByTransaction = new Map();
+    this._runningTransactions = new Set();
+    this._statsAtTimestamp = new Map();
```
AbhiPrasad, line 302: let's try to avoid creating new objects.

```suggestion
    this._longestStallsByTransaction.clear();
    this._runningTransactions.clear();
    this._statsAtTimestamp.clear();
```

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,336 @@
...
+      this._stallCount += 1;
+      this._totalStallTime += stallTime;
+
+      this._runningTransactions.forEach((transaction) => {
+        this._longestStallsByTransaction.set(
+          transaction,
+          Math.max(
+            this._longestStallsByTransaction.get(transaction) ?? 0,
+            stallTime
+          )
```
AbhiPrasad, line 325: nit: Can we extract this into a variable?

AbhiPrasad (commented):
gimme another day to reflect on this, but tbh we can probably ship it

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,336 @@
...
+          statsOnStart.stall_total_time.value,
+      },
+      stall_longest_time: statsOnFinish.stall_longest_time,
+    };
+  }
+
+  /**
+   * Logs the finish time of the span for use in `trimEnd: true` transactions.
+   */
+  private _logSpanFinish(
```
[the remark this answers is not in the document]
jennmueng: `_markSpanFinish` it is

src/js/integrations/stalltracking.ts
jennmueng: Hmm the original code from Facebook actually used 1/5th the "acceptable busy time", the number is pretty arbitrary and up to how often we want to run the iterations and what we consider a stall.

I came up with a better solution is to have `minimumStallThreshold` instead of `acceptableBusyTime`. `timeoutDuration` would instead be a fixed `50ms` (or whatever we want to set it to, while keeping in mind the performance impact of having a lower number, and the lower detection sensitivity of the higher number), and a stall would now instead be logged on `totalTimeTaken  > 50ms + minimumStallThreshold`. How does that sound?

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,335 @@
...
+   */
+  public registerTransactionStart(
+    transaction: Transaction
+  ): (endTimestamp?: number) => StallMeasurements | null {
+    if (this._runningTransactions.has(transaction)) {
+      logger.error(
+        "[StallTracking] Tried to start stall tracking on a transaction already being tracked. Measurements might be lost."
+      );
+
+      // noop
```
AbhiPrasad, line 83: nit: I would just remove, doesn't add much as a comment
```suggestion
```

src/js/integrations/stalltracking.ts
AbhiPrasad: Yeah I like the `minimumStallThreshold` idea, plz include a comment in the code talking about how we switched from `acceptableBusyTime` -> `minimumStallThreshold` so we preserve this context.

src/js/sdk.ts
```diff
@@ -77,6 +79,9 @@ export function init(passedOptions: ReactNativeOptions): void {
     if (options.enableNative) {
       options.defaultIntegrations.push(new DeviceContext());
     }
+    if (options.enableStallTracking) {
+      options.defaultIntegrations.push(new StallTracking());
```
AbhiPrasad, line 83: Should we only enable if tracing is active?

src/js/sdk.ts
```diff
@@ -24,6 +25,7 @@ const DEFAULT_OPTIONS: ReactNativeOptions = {
   enableNativeCrashHandling: true,
   enableNativeNagger: true,
   autoInitializeNativeSdk: true,
+  enableStallTracking: true,
```
AbhiPrasad, line 28: One thought, do we want to make it not default enabled for this release, ask users to try it out manually, and then enable by default in the release after that?

AbhiPrasad (approved):
Spent the weekend thinking about it, and ok let's ship it. I would get a mobile team +1 just to confirm before you merge though. Also please make sure to address the default integration thing as well.

src/js/sdk.ts
jennmueng: @bruno-garcia What do you say. Although this would be released in a beta, if we just enable it, it might work because the people who would be trying this out would be doing so just for it anyways? I'm fine with either.

jennmueng:
Changed so tracing needs to be enabled for the `StallTracking` integration to be added along with tests for it.

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,344 @@
...
+  /** Total amount of time of all stalls that occurred during the current tracking session */
+  private _totalStallTime: number = 0;
+  /** Total number of stalls that occurred during the current tracking session */
+  private _stallCount: number = 0;
+
+  /** The last timestamp the iteration ran in milliseconds */
+  private _lastIntervalMs: number = 0;
+  private _timeout: ReturnType<typeof setTimeout> | null = null;
+
+  private _longestStallsByTransaction: Map<Transaction, number> = new Map();
```
bruno-garcia, line 56: Can we use `WeakMap`? We don't need to have a strong reference here and it could leak if for some reason `finish` isn't called.
This suggestion applies to any map of <Transaction, X>

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,344 @@
...
+  private _statsAtTimestamp: Map<
+    Transaction,
+    { timestamp: number; stats: StallMeasurements }
+  > = new Map();
+  private _runningTransactions: Set<Transaction> = new Set();
```
bruno-garcia, line 61: We could use `_longestStallsByTransaction` for all operations below and get spare this extra Set

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,344 @@
...
+      This is not safe in the case that something changes upstream, but if we're planning to move this over to @sentry/javascript anyways,
+      we can have this temporarily for now.
+    */
+    const isIdleTransaction = "activities" in transaction;
+
+    let statsOnFinish: StallMeasurements | undefined;
+    if (endTimestamp && isIdleTransaction) {
+      /*
+        There is different behavior regarding child spans in a normal transaction and an idle transaction. In normal transactions,
+        the child spans that aren't finished will be dumped, while in an idle transaction they're cancelled and finished.
```
bruno-garcia, line 158: I think this is just the current state of affairs but we should have a chat with the team and align one way or the other.

Doesn't mean this blocks the PR, but it's something we need to address.

src/js/integrations/stalltracking.ts
jennmueng: Replaced with a max running transaction count as discussed due to WeakMap not allowing us to iterate through the keys or with a way to know how many transactions we are tracking.

src/js/integrations/stalltracking.ts
```diff
@@ -0,0 +1,393 @@
...
+  }
+
+  /**
+   * Deletes leaked transactions (Earliest transactions when we have more than MAX_RUNNING_TRANSACTIONS transactions.)
+   */
+  private _flushLeakedTransactions(): void {
+    if (this._statsByTransaction.size > MAX_RUNNING_TRANSACTIONS) {
+      const leakedTransactions = Array.from(
+        this._statsByTransaction.keys()
+      ).slice(0, this._statsByTransaction.size - MAX_RUNNING_TRANSACTIONS);
```
AbhiPrasad, line 386: Feel like this having this logic means that we should probably be using an LRU cache instead of a Map.

But I guess it's fine to keep it as it is. An explicit for loop will be more efficient than slice + forEach here.

```ts
let counter = 0;
const transactions = this._statsByTransaction.keys();
const len = this._statsByTransaction.size - MAX_RUNNING_TRANSACTIONS
for (let t of transactions) {
    if (counter > len) break;
    counter++;
    this._statsByTransaction.delete(t);
}
```

bruno-garcia (approved):
Can we see this flying?

AbhiPrasad (approved):

Issue #1229, opened 2020-12-04
Allow custom release for source map upload scripts


Francesco-Voto:
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [x]  iOS
- [x]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 2.0.2

**`react-native` version:** 0.62.3

**Are you using Expo?**
- [ ]  Yes
- [x]  No

**I have following issue:**

Hello,
should it be possibile to create a different (custom) naming convention for release when source maps are created and uploaded?

Currently we found that release is created as following:
project_name@release+build_number 

We would like to customise it.

Thank you

jennmueng:
We have this feature on our backlog to be able to customize the release on the source map upload script, I'll label this issue as a Feature Request and will post any updates when we implement this.

But for now, you can manually upload source maps with a custom `release` name and `dist` with the cli command here: https://docs.sentry.io/platforms/react-native/sourcemaps/. Note that the `release` and `dist` on the source map and an event have to match **_exactly_** for the source maps to be correctly detected.

jennmueng:
Marked this `up-for-grabs`, if anyone is interested I would start here: 

https://github.com/getsentry/sentry-cli/blob/10dada5914eac25c2c233af6028524c7813057a2/src/commands/react_native_xcode.rs#L271-L281

and maybe add an argument `--release`

retyui:
Have the same problem

I check the source code of `sentry.gradle` + `react_native_xcode.rs` and find that `releaseName` value hardcoded and no way to change it :


**Android:**
https://github.com/getsentry/sentry-react-native/blob/2612b4dafc08c034a12dd6bc38f7148a3004cbf5/sentry.gradle#L168

**iOS:**
https://github.com/getsentry/sentry-cli/blob/d63a0b2ce9c033f5bc1c35bc1405d7719e29e969/src/commands/react_native_xcode.rs#L277-L289

---


I suggest to create env or config support

1) [Android] Using `sentryCli` config

```gradle
project.ext.sentryCli = [
    logLevel: "debug",
    release: '1.2.3',                                                                 // specific value
    getRelease: variant => "${variant.getVersionName()}"    // or function that return value
]
``` 

2) [iOS + Android] Env variable :

```sh
[...]

PR #1548, opened 2021-05-29
[Feat] Allow to define a custom release name for source-maps upload

retyui:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Fix an issue: https://github.com/getsentry/sentry-react-native/issues/1229

iOS cli PR: https://github.com/getsentry/sentry-cli/pull/967

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Fix: Allow custom release for source map upload scripts (#1229)
+- Feat: Allow custom release for source map upload scripts (#1548)
 
 ## 3.0.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix: Allow custom release for source map upload scripts (#1229)
+
 ## 3.0.3
 
 - Fix: Set Java 8 for source and target compatibility if not using AGP >= 4.2.x (#1763)
--- a/sentry.gradle
+++ b/sentry.gradle
@@ -164,8 +164,10 @@ def extractReleasesInfo() {
     android.applicationVariants.each { variant ->
 
         variant.outputs.each { output ->
-            def versionCode = output.getVersionCode()
-            def releaseName = "${variant.getApplicationId()}@${variant.getVersionName()}+${versionCode}"
+            def defaultVersionCode = output.getVersionCode()
+            def versionCode = System.getenv("SENTRY_DIST") ?: defaultVersionCode
+            def defaultReleaseName = "${variant.getApplicationId()}@${variant.getVersionName()}+${versionCode}"
+            def releaseName = System.getenv("SENTRY_RELEASE") ?: defaultReleaseName
             def variantName = variant.getName()
             def outputName = output.getName()
             if (releases[variantName] == null) {


sentry.gradle
```diff
@@ -165,7 +165,8 @@ def extractReleasesInfo() {
 
         variant.outputs.each { output ->
             def versionCode = output.getVersionCode()
```
jennmueng, line 167:
```suggestion
            def defaultVersionCode = output.getVersionCode()
            def versionCode = System.getenv("SENTRY_DIST") ?: defaultVersionCode
```
We would need the `dist` to be configurable as well.

sentry.gradle
```diff
@@ -165,7 +165,8 @@ def extractReleasesInfo() {
...
-            def releaseName = "${variant.getApplicationId()}@${variant.getVersionName()}+${versionCode}"
+            def defaultReleaseName = "${variant.getApplicationId()}@${variant.getVersionName()}+${versionCode}"
+            def releaseName = System.getenv("SENTRY_RELEASE_NAME") ?: defaultReleaseName
```
jennmueng, line 169:
```suggestion
            def releaseName = System.getenv("SENTRY_RELEASE") ?: defaultReleaseName
```
Let's call it `SENTRY_RELEASE`.

jennmueng (commented):
Thank you so much for this as we haven't had the chance to get to it yet. We would need the `dist` to be configurable as well, and let's call it `SENTRY_RELEASE` and `SENTRY_DIST`. Then we would be good to go.

cc @bruno-garcia

sentry.gradle
retyui: done

sentry.gradle
retyui: done

retyui:
end-to-end tests failed (

bruno-garcia:
Also keep in mind, particularly for the docs, that `SENTRY_RELEASE` should take the format of `package@version`.
Sentry releases are global for the whole org and the product understand the release format as such and helps clarify things in the UI, such as Release Health.

retyui:
> Also keep in mind, particularly for the docs, that SENTRY_RELEASE should take the format of package@version.

Is it a strict rule?  
We don't use `Multiple Bundle IDs` and for us, format `1.2.3` better instead of `my.app.com@1.2.3`

jennmueng:
> > Also keep in mind, particularly for the docs, that SENTRY_RELEASE should take the format of package@version.
> 
> Is it a strict rule?
> We don't use `Multiple Bundle IDs` and for us, format `1.2.3` better instead of `my.app.com@1.2.3`

No it's not strict.

retyui:
So, could you please merge this PR? @jennmueng @bruno-garcia

jennmueng:
@retyui We're currently preoccupied with a lot of things on our hands right now, we'll need to run this locally and test it out before we can mere.

sentry.gradle
```diff
@@ -164,8 +164,10 @@ def extractReleasesInfo() {
     android.applicationVariants.each { variant ->
 
         variant.outputs.each { output ->
-            def versionCode = output.getVersionCode()
-            def releaseName = "${variant.getApplicationId()}@${variant.getVersionName()}+${versionCode}"
+            def defaultVersionCode = output.getVersionCode()
```
bruno-garcia, line 167: I'm a bit worried this change has no test to validate it.
Is it possible to add some tests even if just an integration test?

retyui: I don't have experience enough to write tests for Gradle scripts

jennmueng: Hmm we can take it from here ourselves then, it is next on our goals anyway to improve auto source map uploading by completely rewriting it and having it use the release and dist from your JS code without needing to pass an environment variable.

retyui: The main problem that the sentry CLI will be invoked by Gradle and no way to modify \ change argument with what it was executed

retyui:
iOS already merged https://github.com/getsentry/sentry-cli/pull/967

What do I need to do to speed up a review?

retyui:
сс @jennmueng @bruno-garcia

jennmueng:
We will get to this as soon as we can. We know this exists but there are more important matters at hand. Please stop pinging us.

retyui:
another polite reminder

polite reminder gif

sentry.gradle
marandaneto: since release and dist can be set at runtime, but the Gradle script is executed at build time, the chances to get a release and dist misaligned are, even more, I'd say that we have to figure out how to allow custom dist and release to be set on JS but syncing it down to the native layers, but being sure that source maps are also being uploaded with these new values, the fix isn't as simple as this PR though, it'd require more changes and I'm unsure how to do it before a deeper investigation.

retyui: I think synchronization source maps and setting values in the runtime is a developer's care. So it isn't a blocker from a point of view

This ability already implemented for iOS: https://github.com/getsentry/sentry-cli/pull/967

and I still waiting for these changes for the Android Gradle script

bruno-garcia:
I'm fine with the change. It's up to @marandaneto and @jennmueng to approve this though.

@retyui could you please add an entry to the changelog?

bruno-garcia (approved):

retyui:
> @retyui could you please add an entry to the changelog?

no problem

sentry.gradle
marandaneto: if iOS also has the ability to do so, I'm fine with the changes, can you add an entry to changelog @retyui? thanks

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix: Allow custom release for source map upload scripts (#1229)
```
marandaneto, line 5:
```suggestion
- Feat: Allow custom release for source map upload scripts (#1548)
```

marandaneto (approved):

marandaneto:
updated from `Fix` to `Feat` since its a customization rather than a bug

PR #1598, opened 2021-06-11
fix: Fix duplicate breadcrumbs on Android

### LINKED ISSUE #1409: Duplicate Breadcrumbs
### Environment

How do you use Sentry?
sentry.io

Which SDK and version?
- react native 0.61.5
- @sentry/react-native: ^2.3.0

### Steps to Reproduce
```javascript
const dsn = Config.SENTRY_DSN;
const debug = false;
const release = `${DeviceInfo.getBundleId()}@${versionName}@${versionCode}+codepush:${codepushDist}`;
const dist = `${versionName}.${codepushDist}`;
const environment = isProductionEnv ? 'production' : 'staging';
const sampleRate = debug ? 1 : 0.5;
const tracesSampleRate = 0.2;
const maxBreadcrumbs = 150; 

    Sentry.init({
        dsn,
        debug,
        release,
        dist,
        environment,
        sampleRate,
        tracesSampleRate,
        maxBreadcrumbs,
    });
```

Set those initialization and surfing on my app normally until I get some errors.

### Expected Result

breadcrumb data is not logged twice.

### Actual Result
Looking at Breadcrumb section, I found that each breadcrumb data logged twice. `xhr` breadcrumb is logged by default by Sentry SDK. `btn_clicked` and `navigate` is manually logged, I mean, my code that logged it. Both of them logged twice. 

image

this is how I add a breadcrumb.
```
export function recordBreadcrumb(key, message, level = Sentry.Severity.Info) {
    Sentry.addBreadcrumb({
        category: key,
        message,
        level,
    });
}
...
function logButtonClicked(buttonOrFunctionName, value = {}){
   ...
   CrashLoggingUtils.recordBreadcrumb('btn_clicked', buttonOrFunctionName);
}
```

I've ensured that `logButtonClicked()` only called once a time.

jennmueng:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Clears breadcrumbs on the event before sending across the bridge on Android when `mechanism.handled` is true. See the comment for long explanation.

Fixes #1409 

Tested and reproduced via sample app the issue a couple times before the fix, and a couple times after to confirm that it no longer occurs. Some unit tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/wrapper.ts
+++ b/src/js/wrapper.ts
@@ -40,15 +40,27 @@ export const NATIVE = {
       sdk: event.sdk,
     };
 
-    const payload = {
-      ...event,
-      message: {
-        message: event.message,
-      },
-    };
-
     if (NATIVE.platform === "android") {
       const headerString = JSON.stringify(header);
+
+      /*
+        We do this to avoid duplicate breadcrumbs on Android as sentry-android applies the breadcrumbs
+        from the native scope onto every envelope sent through it. This scope will contain the breadcrumbs
+        sent through the scope sync feature. This causes duplicate breadcrumbs.
+        We then remove the breadcrumbs here but only when mechanism.handled is true. If it is handled == false,
+        this is a signal that the app would crash and android would lose the breadcrumbs by the time the app is restarted to read
+        the envelope.
+      */
+      if (event.exception?.values?.[0]?.mechanism?.handled) {
+        event.breadcrumbs = [];
+      }
+
+      const payload = {
+        ...event,
+        message: {
+          message: event.message,
+        },
+      };
 
       const payloadString = JSON.stringify(payload);
       let length = payloadString.length;
@@ -72,6 +84,13 @@ export const NATIVE = {
       return RNSentry.captureEnvelope(envelopeString);
     }
 
+    const payload = {
+      ...event,
+      message: {
+        message: event.message,
+      },
+    };
+
     // Serialize and remove any instances that will crash the native bridge such as Spans
     const serializedPayload = JSON.parse(JSON.stringify(payload));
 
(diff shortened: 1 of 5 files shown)


src/js/wrapper.ts
```diff
@@ -40,16 +40,28 @@ export const NATIVE = {
...
 
+      /*
+        We do this to avoid duplicate breadcrumbs on Android as sentry-android applies the breadcrumbs
+        from the native scope onto every envelope sent through it. This scope will contain the breadcrumbs
+        sent through the scope sync feature. This causes duplicate breadcrumbs.
+        We then remove the breadcrumbs here but only when mechanism.handled is true. If it is handled == false,
+        this is a signal that the app would crash and android would lose the breadcrumbs by the time the app is restarted to read
+        the envelope.
+      */
+      if (event.exception?.values?.[0]?.mechanism?.handled) {
```
bruno-garcia, line 54: Do we need to check if scope sync is enabled? And the native layer is enabled etc? IIRC we have something like `enableNative` here

bruno-garcia (approved):
Just a note on the guard code but LGTM

src/js/wrapper.ts
marandaneto: yep we agreed on that too

jennmueng: @bruno-garcia @marandaneto  `enableNative` is already checked beforehand and scope sync is always enabled, we haven't added an option to disable it yet.

src/js/wrapper.ts
```diff
@@ -40,16 +40,28 @@ export const NATIVE = {
...
-      message: {
-        message: event.message,
-      },
-    };
-
     if (NATIVE.platform === "android") {
       const headerString = JSON.stringify(header);
 
+      /*
+        We do this to avoid duplicate breadcrumbs on Android as sentry-android applies the breadcrumbs
```
philipphofmann, line 47: Nice comment 🥇

marandaneto (approved):

Issue #1701, opened 2021-07-28
Release 3.0.0


bruno-garcia:
* Align `event.origin` and `event.environment`
* Introduce `Sentry.initWith` that takes the root component
* Add tracing integration OOTB
* Auto detect routing library (React Navigation v4 or v5 and in Q3, React Native Navigation (wix))

Bardiamist:
Is migration guide exist? Now I tried to update to v3 and can't build app.

```
/Users/bardiamist/Library/Developer/Xcode/DerivedData/proj-gcfwctharjklzxamsqckslyxgkbd/Build/Intermediates.noindex/proj.build/Debug-iphonesimulator/proj.build/Script-00DD1BFF1BD5951E006B06BC.sh: line 4: ../node_modules/@sentry/cli/bin/sentry-cli: No such file or directory
Command PhaseScriptExecution failed with a nonzero exit code
```

jennmueng:
@Bardiamist Try a full node_modules reinstall, that error looks to be a missing sentry cli. There shouldn't be changes needed for 3.0.0 apart from different `event.origin` values.

Bardiamist:
> @Bardiamist Try a full node_modules reinstall, that error looks to be a missing sentry cli. There shouldn't be changes needed for 3.0.0 apart from different `event.origin` values.

Unfortunately not helped

christophemenager:
I do have the same error @jennmueng.
It appears that in 3.0.0 there is no cli folder in the /node_modules/@sentry, so the error looks logical
How could we fix this ?

benlondon:
>  I do have the same error @jennmueng.
It appears that in 3.0.0 there is no cli folder in the /node_modules/@sentry, so the error looks logical
How could we fix this ?

While not the best long term solution, adding an explicit dependency on the CLI worked for me:

```
"dependencies": {
  "@sentry/react-native": "^3.0.0",
  "@sentry/cli": "^1.68.0",
...
```

nomi9995:
> > I do have the same error @jennmueng.
> > It appears that in 3.0.0 there is no cli folder in the /node_modules/@sentry, so the error looks logical
> > How could we fix this ?
> 
> While not the best long term solution, adding an explicit dependency on the CLI worked for me:
> 
> ```
> "dependencies": {
>   "@sentry/react-native": "^3.0.0",
>   "@sentry/cli": "^1.68.0",
> ...
> ```

yes it is working

CesarLanderos:
[...]

PR #1704, opened 2021-07-30
feat: App start measurement, initial ui.load transaction, and auto performance enable flag.

jennmueng:

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Screen Shot 2021-07-30 at 2 56 50 PM

- Auto performance monitoring (`ReactNativeTracing` integration) will now be enabled by default **when tracing is enabled with either `tracesSampleRate` or `tracesSampler`**.
  - Disable with `enableAutoPerformanceTracking: false`.
  - Configure by passing `ReactNativeTracing` to integrations.
- `enableStallTracking` now depends on `enableAutoPerformanceTracking`.
- App start span and measurements will be added to the first route transaction along with changing the operation to `ui.load`.  If no routing instrumentation is set then a transaction is started (default).
- Bumps Sentry Cocoa SDK to `7.2.0-beta.8`
- Bumps Sentry Android SDK to `5.1.0-beta.9`

Working towards #1701 

Added unit tests for the app start instrumentation and the option flags.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -175,14 +175,32 @@ public void crash() {
...
         release.putString("build", String.valueOf(packageInfo.versionCode));
         promise.resolve(release);
     }
 
+    @ReactMethod
+    public void fetchNativeAppStart(Promise promise) {
+        final AppStartState appStartInstance = AppStartState.getInstance();
+        final Date appStartTime = appStartInstance.getAppStartTime();
+
+        final double appStartTimestamp = (double) appStartTime.getTime();
```
bruno-garcia, line 191: We need to keep in mind auto performance monitoring might be disabled or failed to run i the native layer. What's the return type here for `getTime`? Should we cast to `Double` instead assuming this can be null? I'm afraid it'll either throw on some invalid cast or give a `0` which could either break the math or be misleading

bruno-garcia (commented):
There are some merge conflicts now

android/src/main/java/io/sentry/react/RNSentryModule.java
jennmueng: `getTime` returns a `long`. So I will add the null check for `appStartTime` which can be null

jennmueng:
Weird, iOS sample app builds fine locally.. will look into this in the morning

bruno-garcia (approved):
:shipit:

jennmueng:
@bruno-garcia We need to wait for https://github.com/getsentry/sentry-cocoa/pull/1251 to be in a release, I misunderstood and thought it was in `beta.8` of cocoa.

PR #1711, opened 2021-08-04
feat: Native slow/frozen frames measurements

jennmueng:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


bruno-garcia:
> However, we then run into a problem: `transaction.finish` is not async and needs to return the event id. To make the async call across the native bridge, we start the async call on `transaction.finish`, then we actually await for it in an event processor which **can** be async. We then set the measurements on the transaction event there. This is very obviously a hack but it this is to allow us to get this feature up for EA.

I didn't go through thoroughly enough to answer my own question but I'll drop here for the record:

Are we waiting forever in case an exception is thrown mid-flight and the promise isn't ever resolved?

 
> On iOS we fetch the native frames from the cocoa SDK, however on Android we instrument the frames ourselves in the native bridge module.


> Currently native frames is controlled by `enableNativeFramesTracking` passed as an option to the `ReactNativeTracing` integration. Currently inconsistent with `enableStallTracking` which is passed to the init call. 

Sounds like we should keep this consistent and flatten them out. Lets chat this offline

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -116,6 +121,22 @@ public void initNativeSdk(final ReadableMap rnOptions, Promise promise) {
...
+                    }
+
+                    // Only add the current activity to frames metrics tracking if auto performance is on.
+                    Activity currentActivity = getCurrentActivity();
+
+                    if (currentActivity != null) {
+                        RNSentryModule.frameMetricsAggregator.add(currentActivity);
+                    }
+                } else {
+                    this.disableNativeFramesTracking();
```
AbhiPrasad, line 137: Shouldnt this also run if `rnOptions.hasKey("enableAutoPerformanceTracking"))` is false?

sample/ios/Podfile
```diff
@@ -10,7 +10,7 @@ target 'sample' do
   use_react_native!(
     :path => config["reactNativePath"],
     # to enable hermes on iOS, change `false` to `true` and then install pods
-  	:hermes_enabled => true
+  	:hermes_enabled => false
```
AbhiPrasad, line 13: Do we need this change?

src/js/tracing/nativeframes.ts
```diff
@@ -0,0 +1,263 @@
...
+  };
+  frames_slow: {
+    value: number;
+  };
+  frames_frozen: {
+    value: number;
+  };
+}
+
+const MARGIN_OF_ERROR_SECONDS = 0.05;
```
AbhiPrasad, line 21: Can we leave a comment why this is 0.05?

src/js/tracing/nativeframes.ts
```diff
@@ -0,0 +1,263 @@
...
+  frames_total: {
+    value: number;
+  };
+  frames_slow: {
+    value: number;
+  };
+  frames_frozen: {
+    value: number;
+  };
+}
```
AbhiPrasad, line 19:
```suggestion
type FramesMeasurements = Record<'frames_total' | 'frames_slow' | 'frames_frozen', { value: number }>;
```

let's also not extend from the `Measurements` type, it allows for arbitrary keys which we don't want here.

src/js/tracing/nativeframes.ts
```diff
@@ -0,0 +1,263 @@
...
+export class NativeFramesInstrumentation {
+  private _finishFrames: Map<
+    string,
+    { timestamp: number; nativeFrames: NativeFramesResponse | null }
+  > = new Map();
+  private _framesListeners: Map<string, () => void> = new Map();
+  private _lastSpanFinishFrames?: {
+    timestamp: number;
+    nativeFrames: NativeFramesResponse;
+  };
```
AbhiPrasad, line 35: Would appreciate some comments about what these data structures are, so it's easier for future readers/users.

src/js/tracing/nativeframes.ts
```diff
@@ -0,0 +1,263 @@
...
+
+  public constructor(
+    addGlobalEventProcessor: (e: EventProcessor) => void,
```
AbhiPrasad, line 38: Feels weird that this isn't an integration, but I understand why.

src/js/tracing/nativeframes.ts
```diff
@@ -0,0 +1,263 @@
...
+      if (doesExist()) {
+        return this._processEvent(event);
+      }
+
+      return event;
+    });
+  }
+
+  /**
+   * To be called when a transaction is started
```
AbhiPrasad, line 55: Can we make this comment better indicate why this should be used?

src/js/tracing/nativeframes.ts
```diff
@@ -0,0 +1,263 @@
...
+   */
+  public onTransactionStart(transaction: Transaction): void {
+    void NATIVE.fetchNativeFrames().then((framesMetrics) => {
+      if (framesMetrics) {
+        transaction.setData("__startFrames", framesMetrics);
```
AbhiPrasad, line 60: Why the `__` prefix?

src/js/tracing/nativeframes.ts
```diff
@@ -0,0 +1,263 @@
...
+    });
+
+    instrumentChildSpanFinish(transaction, this._onSpanFinish.bind(this));
+  }
+
+  /**
+   * To be called when a transaction is finished
+   */
+  public onTransactionFinish(transaction: Transaction): void {
+    void this._fetchFramesForTransaction(transaction);
```
AbhiPrasad, line 71: nit: `_fetchFramesForTransaction` is only used once, let's just move it's logic into `onTransactionFinish` and `rm -rf` the function call.

src/js/tracing/nativeframes.ts
```diff
@@ -0,0 +1,263 @@
...
+   * To be called when a transaction is started
+   */
+  public onTransactionStart(transaction: Transaction): void {
+    void NATIVE.fetchNativeFrames().then((framesMetrics) => {
+      if (framesMetrics) {
+        transaction.setData("__startFrames", framesMetrics);
+      }
+    });
+
+    instrumentChildSpanFinish(transaction, this._onSpanFinish.bind(this));
```
AbhiPrasad, line 64: Why make this a util instead of a class method? Nobody else uses this.

src/js/tracing/nativeframes.ts
```diff
@@ -0,0 +1,263 @@
...
+   */
+  public onTransactionFinish(transaction: Transaction): void {
+    void this._fetchFramesForTransaction(transaction);
+  }
+
+  /**
+   * Called on a span finish to fetch native frames to support transactions with trimEnd.
+   */
+  private _onSpanFinish(_: Span, endTimestamp?: number): void {
+    if (!endTimestamp) {
```
AbhiPrasad, line 78: If lack of an `endTimestamp` means that logic should be run, can we just change so that `_onSpanFinish` is only called if an end timestamp doesn't exist (move the logic outside of this func?)

src/js/tracing/nativeframes.ts
```diff
@@ -0,0 +1,263 @@
...
+    if (this._finishFrames.has(traceId)) {
+      return this._prepareMeasurements(traceId, finalEndTimestamp, startFrames);
+    }
+
+    return new Promise((resolve) => {
+      const timeout = setTimeout(() => {
+        this._framesListeners.delete(traceId);
+
+        resolve(null);
+      }, 2000);
```
AbhiPrasad, line 109: Should this be configurable?

src/js/tracing/nativeframes.ts
```diff
@@ -0,0 +1,263 @@
...
+
+/**
+ * Instrumentation to add native slow/frozen frames measurements onto transactions.
+ */
+export class NativeFramesInstrumentation {
+  private _finishFrames: Map<
+    string,
+    { timestamp: number; nativeFrames: NativeFramesResponse | null }
+  > = new Map();
+  private _framesListeners: Map<string, () => void> = new Map();
```
AbhiPrasad, line 31: Can any of these be a `WeakMap`

AbhiPrasad (commented):
Got half way, will review the rest tomorrow. Feel free to tell me no and push back on some of my questions/changes, have much less context on this than y'all.

Let me also take some time to think about the async problem, I agree that this feels hacky. A related aside: The tricky part of accessing the same event multiple times through the sdk's prepare event pipeline why we should maybe revamp integrations in general to rely on lifecycle hooks in the SDK (like how span processing works in Otel sdks)

sample/ios/Podfile
jennmueng: Nope, this got accidentally committed, i was using a debugger and it doesn't work with iOS hermes. Will revert.

src/js/tracing/nativeframes.ts
jennmueng: `__startFrames` is just to be used by this instrumentation and removed on the transaction finish. It's to make it clear that this is internal I guess.

src/js/tracing/nativeframes.ts
jennmueng: Ah, so the stall tracking instrumentation uses something similar, so made this a util with the plan to refactor the stall tracking instrumentation to use this util.

src/js/tracing/nativeframes.ts
jennmueng: Did this so `onTransactionFinish` doesn't need to return a promise to avoid needing a `void` when we call it in the `ReactNativeTracing`.

src/js/tracing/nativeframes.ts
jennmueng: I don't think it should, this is just a very generous arbitrary timeout to avoid leaking.

src/js/tracing/nativeframes.ts
jennmueng: Nope, I tried to but we need to access these maps from inside the event processor, where we won't have a reference to the original the transaction rather just the traceId.

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -116,6 +121,20 @@ public void initNativeSdk(final ReadableMap rnOptions, Promise promise) {
             if (rnOptions.hasKey("sendDefaultPii")) {
                 options.setSendDefaultPii(rnOptions.getBoolean("sendDefaultPii"));
             }
+            if (rnOptions.hasKey("enableAutoPerformanceTracking") && rnOptions.getBoolean("enableAutoPerformanceTracking")) {
+                if (RNSentryModule.frameMetricsAggregator == null) {
+                    RNSentryModule.frameMetricsAggregator = new FrameMetricsAggregator();
```
AbhiPrasad, line 126: Should we always instantiate a new `frameMetricsAggregator` when we are going through the init code path?

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -116,6 +121,20 @@ public void initNativeSdk(final ReadableMap rnOptions, Promise promise) {
...
+                }
+
+                // Only add the current activity to frames metrics tracking if auto performance is on.
```
AbhiPrasad, line 129: nit: This comment doesn't add any new information, this should be obvious from the boolean check just 5,6 lines above.

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -206,6 +225,54 @@ public void fetchNativeAppStart(Promise promise) {
...
+
+                if (framesRates != null) {
+                    final SparseIntArray totalIndexArray = framesRates[FrameMetricsAggregator.TOTAL_INDEX];
+                    if (totalIndexArray != null) {
+                        for (int i = 0; i < totalIndexArray.size(); i++) {
+                            int frameTime = totalIndexArray.keyAt(i);
+                            int numFrames = totalIndexArray.valueAt(i);
+                            totalFrames += numFrames;
+                            // hard coded values, its also in the official android docs and frame metrics API
+                            if (frameTime > 700) {
```
AbhiPrasad, line 251: let's extract these numbers (700, 16) into constants, and then put the comments about these values there (It'll mean the git blame if these change are easier to parse as well)

ios/RNSentry.m
```diff
@@ -127,12 +128,33 @@ + (BOOL)requiresMainQueueSetup {
...
+        if (frames == nil) {
+            resolve(nil);
+        }
+
+        resolve(@{
+            @"totalFrames": [NSNumber numberWithLong:frames.total],
+            @"frozenFrames": [NSNumber numberWithLong:frames.frozen],
+            @"slowFrames": [NSNumber numberWithLong:frames.slow],
+        });
+    }
```
AbhiPrasad, line 155: Let's flip the conditional 

```suggestion
    if (PrivateSentrySDKOnly.isFramesTrackingRunning) {
        SentryScreenFrames *frames = PrivateSentrySDKOnly.currentScreenFrames;

        if (frames == nil) {
            resolve(nil);
        }

        resolve(@{
            @"totalFrames": [NSNumber numberWithLong:frames.total],
            @"frozenFrames": [NSNumber numberWithLong:frames.frozen],
            @"slowFrames": [NSNumber numberWithLong:frames.slow],
        });
    } else {
      resolve(nil);
    }
```

src/js/tracing/nativeframes.ts
AbhiPrasad: Ah hmmm, thought that this was kept and not removed on finish. Feels weird to use the event as a carrier for temporary data. Is there no way around this?

@bruno-garcia what do you think?

src/js/tracing/nativeframes.ts
AbhiPrasad: Ah gotcha ok yeah let's keep it like this then.

src/js/tracing/nativeframes.ts
AbhiPrasad: Doing the `transaction.data.__startFrames` getter feels like a code smell to me.

src/js/tracing/reactnativetracing.ts
```diff
@@ -85,6 +92,7 @@ const defaultReactNativeTracingOptions: ReactNativeTracingOptions = {
   ignoreEmptyBackNavigationTransactions: true,
   beforeNavigate: (context) => context,
   enableAppStartTracking: true,
+  enableNativeFramesTracking: true,
```
AbhiPrasad, line 95: Shouldn't this be false for now?

src/js/tracing/reactnativetracing.ts
```diff
@@ -129,12 +139,30 @@ export class ReactNativeTracing implements Integration {
...
 
     this._getCurrentHub = getCurrentHub;
 
     void this._instrumentAppStart();
 
+    if (enableNativeFramesTracking) {
+      this.nativeFramesInstrumentation = new NativeFramesInstrumentation(
+        addGlobalEventProcessor,
+        () => {
+          const self = getCurrentHub().getIntegration(ReactNativeTracing);
```
AbhiPrasad, line 153: Can we just pass a boolean instead of a function?

test/tracing/nativeframes.test.ts
```diff
@@ -0,0 +1,356 @@
...
+            expect(event).toBeDefined();
+
+            if (event) {
+              expect(event.measurements).toBeUndefined();
+
+              expect(event.contexts?.trace?.data).toBeDefined();
+
+              if (event.contexts?.trace?.data) {
+                expect(
+                  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
AbhiPrasad, line 343: We should just blanket disable this in tests, but we leave that for another PR

AbhiPrasad (commented):
Ok nice tests, I'll come back to give ✅ after you've synced about stuff

romtsn (approved):

src/js/tracing/reactnativetracing.ts
jennmueng: Actually, now that I gave it a second thought shouldn't the check that the integration exists on the hub (and subsequently that it has this native frames instrumentation) be run on every event processor?

src/js/tracing/reactnativetracing.ts
jennmueng: We want to have everything enabled by default (when tracing is enabled) and only blanket disabled by a `enablePerformanceMonitoring: false` cc @bruno-garcia 

Will remove these individual options in a separate PR though.

src/js/tracing/nativeframes.ts
jennmueng: We can just store it in a map then.

jennmueng: @AbhiPrasad Ok I've given it a go and using a map means we could have a leak, the transaction could be started but never finished or lost and we would still store the start frames. A weakmap can't be used due to us not having a reference to the transaction inside the event processor that needs to access the start frames. I guess using the event as a carrier as is would work best as it covers this edge case?

android/src/main/java/io/sentry/react/RNSentryModule.java
jennmueng: I guess we should. Originally had it this way as RN could call init multiple times if the JS bundle gets reloaded. However it makes sense to start fresh with the new aggregator + activity and let the garbage collector get the previous one.

src/js/tracing/nativeframes.ts
AbhiPrasad: Ah hmm. I mean it's not the end of the world to keep it on the event, especially if the alternative has possible problems. Let's just use the transaction data then.

src/js/tracing/reactnativetracing.ts
jennmueng: @AbhiPrasad [All the integrations on the JS repo that set an event processor](https://github.com/getsentry/sentry-javascript/tree/master/packages/integrations/src) seem to check on every event, I assumed it was to support multiple hubs right?

AbhiPrasad: Ah good point. Didn't realize this, let's keep the check on every event processor run then.

bruno-garcia (approved):
Lets ship this! :shipit:

PR #1751, opened 2021-08-31
fix: Move @sentry/wizard dependency to devDependency

jamonholmgren:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

This moves the dependency `@sentry/wizard` to a `devDependency` of `sentry-react-native`.

We noticed that `@sentry/wizard` was bringing in `lodash` as a dependency to our React Native project. Since the wizard is only intended to be a development tool, it seemed appropriate to make it a `devDependency` instead.

I didn't test it, given the nature of the PR. If you want me to spin up the sample app I can, but it doesn't really make a lot of sense to do that.

## :pencil: Checklist

- [ ] I added tests to verify changes (n/a)
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Discuss ... and merge if cool!

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,13 +47,13 @@
     "@sentry/react": "6.12.0-beta.2",
     "@sentry/tracing": "6.12.0-beta.2",
     "@sentry/types": "6.12.0-beta.2",
-    "@sentry/utils": "6.12.0-beta.2",
-    "@sentry/wizard": "^1.2.2"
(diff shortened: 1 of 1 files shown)


jennmueng (approved):
As we only use the wizard in the install steps, along with using `npx` in our installation docs, it makes sense to move the wizard to devDependencies. Thank you!

PR #1774, opened 2021-09-15
feat: Routing instrumentation for React Native Navigation

jennmueng:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


src/js/tracing/reactnativenavigation.ts
```diff
@@ -0,0 +1,208 @@
...
+type ComponentType =
+  | "Component"
+  | "TopBarTitle"
+  | "TopBarBackground"
+  | "TopBarButton";
+
+export interface ComponentWillAppearEvent extends ComponentEvent {
+  componentName: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  passProps?: Record<any, any>;
```
AbhiPrasad, line 33: can we type as `unknown`?

src/js/tracing/reactnativenavigation.ts
```diff
@@ -0,0 +1,208 @@
...
+  events: () => EventsRegistry;
+}
+
+/**
+ * Instrumentation for React Native Navigation. See docs or sample app for usage.
+ *
+ * How this works:
+ * - `_onCommand` is called every time a commands happens and sets an IdleTransaction on the scope without any route context.
+ * - `_onComponentWillAppear` is then called AFTER the state change happens due to a dispatch and sets the route context onto the active transaction.
+ * - If `_onComponentWillAppear` isn't called within `STATE_CHANGE_TIMEOUT_DURATION` of the dispatch, then the transaction is not sampled and finished.
```
AbhiPrasad, line 61: Where is `STATE_CHANGE_TIMEOUT_DURATION` defined? Also where does this logic happen?

src/js/tracing/reactnativenavigation.ts
```diff
@@ -0,0 +1,208 @@
...
+      .events()
+      .registerComponentWillAppearListener(
+        this._onComponentWillAppear.bind(this)
+      );
+  }
+
+  /**
+   * To be called when a navigation command is dispatched
+   */
+  private _onCommand(): void {
```
AbhiPrasad, line 113: Ahh this feels like a https://en.wikipedia.org/wiki/Finite-state_machine, but it's probably too much effort to refactor the logic into a state machine.

AbhiPrasad (commented):
This is a real quick review 👀, will circle back to give my final ✅ tomorrow.

src/js/tracing/reactnativenavigation.ts
jennmueng: should be `options.routeChangeTimeoutMs`

src/js/tracing/reactnativenavigation.ts
```diff
@@ -0,0 +1,208 @@
...
+  private _latestTransaction?: TransactionType;
+  private _recentComponentIds: string[] = [];
+  private _stateChangeTimeout?: number | undefined;
+
+  public constructor(
+    /** The react native navigation `NavigationDelegate`. This is usually the import named `Navigation`. */
+    navigation: NavigationDelegate,
+    options: Partial<ReactNativeNavigationOptions> = {}
+  ) {
+    super();
```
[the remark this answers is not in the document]
jennmueng: Yeah, the base `RoutingInstrumentation` doesn't have any options

src/js/tracing/reactnativenavigation.ts
```diff
@@ -0,0 +1,206 @@
...
+
+      this._latestTransaction.updateWithContext(finalContext);
+
+      this._prevComponentEvent = event;
+    } else {
+      this._discardLatestTransaction();
+    }
+  }
+
+  /** Cancels the latest transaction so it does not get sent to Sentry. */
```
AbhiPrasad, line 188: Why did we decide this is the behaviour vs. just finishing the transaction with a different status? Won’t the transaction being cancelled still be valuable to know?

AbhiPrasad: Got it, makes sense to me. Should we create a span measuring the time between `_onCommand` and `_onComponentWillAppear`?

Also, can we add your comment this as a comment in the `ReactNativeNavigationInstrumentation` docstring?

src/js/tracing/reactnativenavigation.ts
```diff
@@ -0,0 +1,206 @@
...
+    super.registerRoutingInstrumentation(listener, beforeNavigate);
+
+    this._navigation
+      .events()
+      .registerCommandListener(this._onCommand.bind(this));
+
+    this._navigation
+      .events()
+      .registerComponentWillAppearListener(
+        this._onComponentWillAppear.bind(this)
```
AbhiPrasad, line 104: Why component `componentWillAppear` over `componentDidAppear`?

src/js/tracing/reactnativenavigation.ts
```diff
@@ -0,0 +1,206 @@
...
+  componentName: string;
+  passProps?: Record<string | number | symbol, unknown>;
+  componentType: ComponentType;
+}
+
+export interface EventSubscription {
+  remove(): void;
+}
+
+export interface EventsRegistry {
```
AbhiPrasad, line 40: Can we hook into https://wix.github.io/react-native-navigation/api/events#onapplaunched somehow?

AbhiPrasad (approved):
Small comments, but we can address them in later PRs

src/js/tracing/reactnativenavigation.ts
AbhiPrasad: also, do you think it makes sense to emit breadcrumbs for each command?

src/js/tracing/reactnativenavigation.ts
jennmueng: We want the earliest confirmation that the screen actually did change but I don't think it really makes a difference. If we want to add a span to measuring the time like you suggested in the other comment, I'd change this to instrument `componentDidAppear`.

src/js/tracing/reactnativenavigation.ts
jennmueng: I agree with both, lets add the both the span and the breadcrumbs in a separate PR so we can add that to all the other routing instrumentations as well.

Hellhiem:
@jennmueng Have you tested this with TypeScript ? When I pass `Navigation` from `react-native-navigation` there are multiple incompatible types issues

Issue #1864, opened 2021-11-01
tracesSampler becomes NSNull in iOS and the app cannot be started


kitoko552:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
"@sentry/react-native": "^3.1.1"

### Steps to Reproduce

1. Sentry.init with tracesSampler and enable [the automatic instrumentation of performance monitoring](https://docs.sentry.io/platforms/react-native/performance/instrumentation/automatic-instrumentation/).

```js
Sentry.init({
  dsn: Config.sentry.dns,
  tracesSampler: samplingContext => {
    return 0.2
  },
  integrations: [
    new Sentry.ReactNativeTracing({
      routingInstrumentation,
      enableAppStartTracking: false,
    }),
  ],
})
```

```js
const navigationRef = createNavigationContainerRef()
const RootContents = React.memo(() => {
  return (
    <NavigationContainer
      ref={navigationRef}
      onReady={() => {
        routingInstrumentation.registerNavigationContainer(navigationRef)
      }}
      onStateChange={async state => {
        ...
      }}
    >
      ...
    </NavigationContainer>
  )
})
...
export default Sentry.wrap(Root) // Root uses RootContents
```

2. Build the app on iOS

3. The app crashes at [here](https://github.com/getsentry/sentry-cocoa/blob/81d635f41bc164c3b5708256988c5924a28379e7/Sources/Sentry/SentryTracesSampler.m#L34) because `_options.tracesSampler` is NSNull

### Expected Result

No Crashing and the app can be started correctly.

### Actual Result

The stack trace is here.

```
  * thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x1e3dfa8e8)
    frame #0: 0x00000001e3dfa8e8 CoreFoundation`__kCFNull
  * frame #1: 0x0000000101b62134 {our_app_name}`-[SentryTracesSampler sample:](self=0x000000028187f9e0, _cmd="sample:", context=0x0000000281f03c00) at SentryTracesSampler.m:34:38
[...]

PR #1872, opened 2021-11-05
fix(ios): tracesSampler becomes NSNull in iOS and the app cannot be started

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

Closes https://github.com/getsentry/sentry-react-native/issues/1864

running with sampler

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/ios/RNSentry.m
+++ b/ios/RNSentry.m
@@ -65,7 +65,7 @@ RCT_EXPORT_METHOD(initNativeSdk:(NSDictionary *_Nonnull)options
     NSMutableDictionary * mutableOptions =[options mutableCopy];
     [mutableOptions setValue:beforeSend forKey:@"beforeSend"];
 
-    // remove performance traces sample rate and traces sampler since we don't want to syncronize these configuration
+    // remove performance traces sample rate and traces sampler since we don't want to synchronize these configurations
     // to the Native SDKs.
     // The user could tho initialize the SDK manually and set themselves.
     [mutableOptions removeObjectForKey:@"tracesSampleRate"];
--- a/ios/RNSentry.m
+++ b/ios/RNSentry.m
@@ -64,6 +64,12 @@ RCT_EXPORT_METHOD(initNativeSdk:(NSDictionary *_Nonnull)options
 
     [options setValue:beforeSend forKey:@"beforeSend"];
 
+    // remove performance traces sample rate and traces sampler since we don't want to syncronize these configuration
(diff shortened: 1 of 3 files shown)


ios/RNSentry.m
```diff
@@ -62,7 +62,14 @@ + (BOOL)requiresMainQueueSetup {
         return event;
     };
 
-    [options setValue:beforeSend forKey:@"beforeSend"];
+    NSMutableDictionary * mutableOptions =[options mutableCopy];
+    [mutableOptions setValue:beforeSend forKey:@"beforeSend"];
```
philipphofmann, line 66: `l`:
```suggestion
    NSMutableDictionary * mutableOptions = [options mutableCopy];
    mutableOptions[@"beforeSend"] = beforeSend;
```

ios/RNSentry.m
```diff
@@ -62,7 +62,14 @@ + (BOOL)requiresMainQueueSetup {
...
+
+    // remove performance traces sample rate and traces sampler since we don't want to syncronize these configuration
```
philipphofmann, line 68:
```suggestion
    // remove performance traces sample rate and traces sampler since we don't want to synchronize these configurations
```

philipphofmann (approved):
LGTM. I think we could add a layer of abstraction to add tests in Swift for the whole bridge. I can help you with that. Of course not in this PR. What do you think?

ios/RNSentry.m
marandaneto: I'd rather keep as it is and change in another PR, there are more places in this class doing the same way

ios/RNSentry.m
marandaneto: good catch, thanks

marandaneto:
> LGTM. I think we could add a layer of abstraction to add tests in Swift for the whole bridge. I can help you with that. Of course not in this PR. What do you think?

yep, I intend to do that after cleaning up all the JS issues first, also for Android, will hook you up when we are there, thanks.

philipphofmann:
> > LGTM. I think we could add a layer of abstraction to add tests in Swift for the whole bridge. I can help you with that. Of course not in this PR. What do you think?
> 
> yep, I intend to do that after cleaning up all the JS issues first, also for Android, will hook you up when we are there, thanks.

Agree, I created an issue so we don't forget about it.

PR #1886, opened 2021-11-12
fix: Warn when promise rejections won't be caught

### LINKED ISSUE #1077: Unhandled promises are not logged at all (iOS and Android)
**Platform:**
- [x]  iOS

**SDK:**

**SDK version:** "@sentry/react-native": "^1.7.2",

**`react-native` version:** "react-native": "^0.63.2",

**Are you using Expo?**
- [ ]  Yes
- [x]  No

- [x]  sentry.io (SaaS)
- [ ]  on-premise

No issue was created

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  // no extra options
});
```

---
**I have following issue:**

Normal exceptions are logged to Sentry.io just fine, but unhandled promises are not logged at all.  My understanding is that this should happen by default.  I do get a yellow box showing the unhandled exception but it seems that Sentry never sees it.

This repros on both Android and iOS.

**Steps to reproduce:**
1. Install sentry with minimal config
2. Generate an unhandled promise rejection

**Actual result:**

Yellow box appears but no error logged to sentry.io

image

**Expected result:**

Unhandled promise rejections are logged to sentry.io

jennmueng:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This fix will appear to not work in the original sample app in this repo as our SDK will use the promise library from the root folder instead of the one inside the sample folder that React Native uses. Thus the end to end tests have been updated to use a packaged version of the SDK via `yalc`. This way the promise rejection handler will be tested.

e2e test workflow will run the script `sample/scripts/prepareConfigsForTesting.sh` which will swap out the babel config and the metro config to the testing one which will use the packaged SDK instead.

Extra
- Removes Flipper from Android sample
- Also the behavior of the promise rejection handler will follow the behavior of the default react native handler rather than just calling. a console.warn like before.

Fixes #1077

End to end tests will now package the SDK and test the unhandled promise instrumentation.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Merge docs PR https://github.com/getsentry/sentry-docs/pull/4392


src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -51,66 +51,79 @@ export class ReactNativeErrorHandlers implements Integration {
...
-        if ("_onHandle" in _global.Promise && "_onReject" in _global.Promise) {
-          _global.Promise._onHandle = _onHandle;
-          _global.Promise._onReject = _onReject;
-        } else if ("_Y" in _global.Promise && "_Z" in _global.Promise) {
-          _global.Promise._Y = _onHandle;
-          _global.Promise._Z = _onReject;
-        }
+      if (Promise !== _global.Promise) {
+        logger.warn(
+          "Unhandled promise rejections will not be caught by Sentry. Read about how to fix this on our troubleshooting page."
```
marandaneto, line 117: link to the docs once merged

jennmueng: @marandaneto Do we want to directly have a link to the docs in the warning message?

jennmueng:
For some reason the android emulator for the e2e test isn't working, says app never started. Works locally... going to keep debugging this

jennmueng:
Accidentally hit close

sample/android/app/src/main/java/io/sentry/sample/MainApplication.java
```diff
@@ -25,7 +25,19 @@ public boolean getUseDeveloperSupport() {
...
+      for (ReactPackage pkg : packages) {
+        if (pkg instanceof RNSentryPackage) {
+          packageExists = true;
+          break;
+        }
+      }
+
+      if (!packageExists) {
+        packages.add(new RNSentryPackage());
+      }
```
marandaneto, line 39: why is this necessary?

sample/ios/Podfile
```diff
@@ -13,7 +13,9 @@ target 'sample' do
   	:hermes_enabled => true
   )
 
-  pod 'RNSentry', :path => '../../RNSentry.podspec'
+  if !ENV['TEST']
+    pod 'RNSentry', :path => '../../RNSentry.podspec'
+  end
```
marandaneto, line 18: why do we need that? does that mean that the `test` version don't include RNSentry?

sample/src/App.tsx
```diff
@@ -60,8 +60,8 @@ Sentry.init({
   tracesSampleRate: 1.0,
   // Sets the `release` and `dist` on Sentry events. Make sure this matches EXACTLY with the values on your sourcemaps
   // otherwise they will not work.
-  release: packageVersion,
-  dist: `${packageVersion}.0`,
+  release: '1.2.3',
+  dist: `1.2.3.0`,
```
marandaneto, line 64: should we revert this change?

src/js/integrations/reactnativeerrorhandlers.ts
marandaneto: since it mentions the troubleshooting page its fine the way how it is

src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -51,66 +51,79 @@ export class ReactNativeErrorHandlers implements Integration {
...
+        onHandled: (_id: string) => {},
+      };
+
+      try {
+        // Here we try to use React Native's original promise rejection handler.
+        // eslint-disable-next-line @typescript-eslint/no-var-requires,@typescript-eslint/no-unsafe-member-access
+        promiseRejectionTrackingOptions = require("react-native/Libraries/promiseRejectionTrackingOptions")
+          .default;
+      } catch (e) {
+        // do nothing, as the user may be running an older version of RN without promiseRejectionTrackingOptions
```
marandaneto, line 83: is it possible to know if the reason is that there is no `promiseRejectionTrackingOptions` at all? wondering if we should log a warn in any other case

marandaneto:
@AbhiPrasad would you mind reviewing this one too? that's a field I don't master, thanks.

sample/src/App.tsx
```diff
@@ -17,7 +17,7 @@ import EndToEndTestsScreen from './screens/EndToEndTestsScreen';
 import ReduxScreen from './screens/ReduxScreen';
 
 import {store} from './reduxApp';
-import {version as packageVersion} from '../../package.json';
+// import {version as packageVersion} from '../../package.json';
```
bruno-garcia, line 20:
```suggestion
```

sample/android/app/src/main/java/io/sentry/sample/MainApplication.java
jennmueng: When the SDK is installed via linking to the parent folder for development we need to add the package manually with ` packages.add(new RNSentryPackage())`, however when the SDK is packaged and installed for the e2e tests React Native auto-adds the package already on Android, if the line is there then the app crashes at runtime for duplicate packages.

sample/ios/Podfile
jennmueng: The test version has the pod auto-added by React Native, while we're developing using the SDK in the parent folder we need to add it manually.

sample/src/App.tsx
jennmueng: We can no longer use the version from the parent folder when the SDK is packaged and the linking to the parent folder is removed so thus why this change here

src/js/integrations/reactnativeerrorhandlers.ts
jennmueng: Hmm I don't think we would need to warn as if the `promiseRejectionTrackingOptions` doesn't exist for any reason the log message in debug mode for the developer just won't be as pretty, all that the `promiseRejectionTrackingOptions` from RN does is just prettify the log messages

.github/workflows/e2e.yml
```diff
@@ -31,12 +31,22 @@ jobs:
...
+      - name: Build SDK
+        run: yarn build
+      - name: Package SDK
+        run: yalc publish
+      - name: Prepare sample for testing
+        working-directory: ./sample
+        run: sh ./scripts/prepareConfigsForTesting.sh
+      - name: Install SDK in sample
+        working-directory: ./sample
+        run: yalc add @sentry/react-native
```
AbhiPrasad, line 44: Think its unavoidable now, but lets please split up the CI changes from the implementation so that things are easier to revert/`git blame` in the future.

sample/android/app/src/main/java/io/sentry/sample/MainApplication.java
```diff
@@ -25,7 +25,19 @@ public boolean getUseDeveloperSupport() {
     protected List<ReactPackage> getPackages() {
       @SuppressWarnings("UnnecessaryLocalVariable")
       List<ReactPackage> packages = new PackageList(this).getPackages();
-      packages.add(new RNSentryPackage());
+
+      boolean packageExists = false;
+      for (ReactPackage pkg : packages) {
+        if (pkg instanceof RNSentryPackage) {
+          packageExists = true;
```
AbhiPrasad, line 32: can we just do an early return here instead of the mutable flag?

src/js/integrations/reactnativeerrorhandlers.ts
AbhiPrasad, line 83: Let's please extract setting `promiseRejectionTrackingOptions` into a function and then always return a const. The require check is an implementation detail that `_handleUnhandledRejections` shouldn't have to worry about imo.

AbhiPrasad (changes requested):
awesome, sorry for the late review. I'll make sure the ✅ comes back way faster :)))

jennmueng:
Thanks @AbhiPrasad! Just got back from thanksgiving break and addressed your review. Next time will make separate PRs for CI changes.

marandaneto (approved):

src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -51,66 +57,88 @@ export class ReactNativeErrorHandlers implements Integration {
...
-      /* eslint-disable
-        @typescript-eslint/no-var-requires,
-        import/no-extraneous-dependencies,
-        @typescript-eslint/no-explicit-any,
-        @typescript-eslint/no-unsafe-member-access
-      */
+    try {
+      // Here we try to use React Native's original promise rejection handler.
+      // eslint-disable-next-line @typescript-eslint/no-var-requires,@typescript-eslint/no-unsafe-member-access
+      promiseRejectionTrackingOptions = require("react-native/Libraries/promiseRejectionTrackingOptions")
```
AbhiPrasad, line 112: Can’t we just return here?

AbhiPrasad (approved):

.github/workflows/e2e.yml
marandaneto: agreed, but splitting up would break the e2e since the CI changes are needed, let's keep this in mind for the next PR but I'd rather merge it for now since it's an important fix.

marandaneto:
thanks for reviewing @AbhiPrasad

PR #1923, opened 2021-12-02
fix: Fix dynamic require for promise options ignoring try catch

### LINKED ISSUE #1922: Error: Unable to resolve module `react-native/Libraries/promiseRejectionTrackingOptions`
react native：0.61.4
sentry：3.2.4

error: bundling failed: Error: Unable to resolve module `react-native/Libraries/promiseRejectionTrackingOptions` from `node_modules/@sentry/react-native/dist/js/integrations/reactnativeerrorhandlers.js`: react-native/Libraries/promiseRejectionTrackingOptions could not be found within the project.

jennmueng:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

Fixes a fatal bug caused by the SDK trying to make an inline dynamic require to a file that does not exist on older RN versions. Although the require is inside a try-catch block, the metro bundler's dependency magic, in trying to optimize dependencies, ignores the try catch statement.

Metro added the optional dependency feature in version 0.59, so having a dynamic require statement inside a try catch is not feasible for users on prior versions, **however** we can get around metro's dependency magic by defining the name of the dependency as a string variable, and instead, we pass it into `require.call`.

Fixes #1922 

Tested on an RN app running `react-native: 0.63.1` and `metro: 0.58.0` (before the optional dependency feature was added in 0.59).

E2E tests passing

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fix: Fix dynamic require for promise options bypassing try catch block and crashing apps #1923
+
 ## 3.2.4
 
 - fix: Warn when promise rejections won't be caught #1886
--- a/src/js/integrations/reactnativeerrorhandlers.ts
+++ b/src/js/integrations/reactnativeerrorhandlers.ts
@@ -95,11 +95,12 @@ export class ReactNativeErrorHandlers implements Integration {
    * Gets the promise rejection handlers, tries to get React Native's default one but otherwise will default to console.logging unhandled rejections.
    */
   private _getPromiseRejectionTrackingOptions(): PromiseRejectionTrackingOptions {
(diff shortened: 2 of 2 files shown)


AbhiPrasad (approved):
Approving to unblock so we can cut a patch.

Do we not run the e2e tests against a matrix of older react native versions?

jennmueng:
> Approving to unblock so we can cut a patch.
> 
> Do we not run the e2e tests against a matrix of older react native versions?

Thanks! No we do not, been mentioned here and there but we never got to it.

marandaneto:
> Approving to unblock so we can cut a patch.
> 
> Do we not run the e2e tests against a matrix of older react native versions?

we discussed that yesterday, adding to the backlog https://github.com/getsentry/sentry-react-native/issues/1924

Issue #1925, opened 2021-12-03
Deprecate options.initialScope and remove in the next major version


marandaneto:
reason is https://github.com/getsentry/sentry-react-native/issues/1661#issuecomment-984685325

PR #1963, opened 2021-12-14
Deprecate initialScope in favor of configureScope

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
Deprecate initialScope in favor of configureScope

Closes https://github.com/getsentry/sentry-react-native/issues/1925

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
Remove in the next major version

### DIFF
--- a/src/js/options.ts
+++ b/src/js/options.ts
@@ -1,5 +1,6 @@
 import { BrowserOptions } from "@sentry/react";
 import { ProfilerProps } from "@sentry/react/dist/profiler";
+import { CaptureContext } from "@sentry/types/dist/scope";
 
 import { TouchEventBoundaryProps } from "./touchevents";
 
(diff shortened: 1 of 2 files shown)


src/js/options.ts
```diff
@@ -79,7 +80,13 @@ export interface ReactNativeOptions extends BrowserOptions {
...
    * @default true
    * */
-   enableOutOfMemoryTracking?: boolean;
+  enableOutOfMemoryTracking?: boolean;
+
+  /**
+   * Set data to the inital scope
+   * @deprecated Use `Sentry.configureScope(...)`
+   */
+  initialScope?: CaptureContext;
```
marandaneto, line 89: `initialScope` is inherited from JS options

marandaneto: @AbhiPrasad or @jennmueng do you know if there's a better way to deprecate a feature that comes from the JS SDK and is available to the RN SDK, but we don't want to expose it to the RN SDK, so we rather than override the attribute and annotate it with @deprecated is that fine? thanks

AbhiPrasad: I think annotating with `@deprecated` is the best we got, do we also want to filter out the option before it gets passed down?

marandaneto: no, not really, it still works, so just deprecated is fine for now, thanks @AbhiPrasad

jennmueng (approved):

Issue #1968, opened 2021-12-10
React native app just warm start


dangdo1:
_Platform:_
- [x]  Android -> 29, 30, 31
- [x]  React-Native -> 0.66.3
- `    "@sentry/react-native": "^3.2.6",
    "@sentry/tracing": "^6.15.0",
    "@sentry/wizard": "^1.2.15",` 
---------------------------------

> @Override
>   public synchronized void onActivityCreated(
>       final @NonNull Activity activity, final @Nullable Bundle savedInstanceState) {
>     setColdStart(savedInstanceState);
>     addBreadcrumb(activity, "created");
>     startTracing(activity);
>     firstActivityCreated = true;
>   }

  

>  LOG  Sentry Logger [Log]: [Measurements] Adding measurements to transaction {
>   "app_start_warm": {
>     "value": 1232.434
>   }
> }

  
  **savedInstanceState alway null and my app just warm start**

marandaneto:
@dangdo1 its indeed an issue, not sure if on the RN SDK or Android SDK yet, will debug, thanks for the repro.

marandaneto:
@dangdo1 I've dug into it and found out the root cause, App start was always reported as Warm start but the numbers were correct anyway.
PR is in place https://github.com/getsentry/sentry-react-native/pull/1985

PR #1970, opened 2021-12-16
Fix App start on Android

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

Closes https://github.com/getsentry/sentry-react-native/issues/1968

## :pencil: Checklist

- [ ] All tests passing
- [X] No breaking changes


android/build.gradle
```diff
@@ -24,5 +24,5 @@ android {
 
 dependencies {
     implementation 'com.facebook.react:react-native:+'
-    api 'io.sentry:sentry-android:5.4.3'
+    api 'io.sentry:sentry-android:5.5.1-SNAPSHOT'
```
marandaneto, line 27: pending on releasing 5.5.1, using mavenLocal for now

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -198,25 +198,24 @@ public void fetchNativeRelease(Promise promise) {
     @ReactMethod
     public void fetchNativeAppStart(Promise promise) {
         final AppStartState appStartInstance = AppStartState.getInstance();
-        final Date appStartTime = appStartInstance.getAppStartTime();
+        final Long appStartInterval = appStartInstance.getAppStartInterval();
```
marandaneto, line 201: `getAppStartInterval` only returns a value if the App start is known already

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -198,25 +198,24 @@ public void fetchNativeRelease(Promise promise) {
...
 
-        if (appStartTime == null) {
+        if (appStartInterval == null && didFetchAppStart) {
```
marandaneto, line 203: no need to be sending the app start `map` back to JS every time if already sent, this should be done on iOS too

src/js/tracing/reactnativetracing.ts
```diff
@@ -383,6 +379,15 @@ export class ReactNativeTracing implements Integration {
...
 
     idleTransaction.registerBeforeFinishCallback((transaction) => {
+      // the problem is that the App start cold/warm by this point is unknown yet, the JS bundle is loaded
+      // before the very first Activity, which is when the Android SDK detects the type of the App start.
+      if (this.options.enableAppStartTracking) {
+        logger.log(
+          `[ReactNativeTracing] thats a test`
+        );
+        void this._instrumentAppStart();
+      }
```
marandaneto, line 389: moved from `setupOnce` to here, since its not known yet the App start, and here it is

PR #1982, opened 2021-12-20
fix: Fix types for allowing non-serializeable values

mrousavy:
While this is perfectly valid code:

```ts
    Sentry.setContext('test', {
      values: [{ hi: 'test' }, { test: 'hi' }]
    })
```

it doesn't serialize the values in the dashboard:

Screenshot 2021-12-20 at 13 41 28

This PR fixes this by marking the types to only be serializeable values (string, number, boolean) to forbid Arrays and Objects. The user then has to serialize them himself.

### DIFF
--- a/src/js/scope.ts
+++ b/src/js/scope.ts
@@ -3,6 +3,9 @@ import { Breadcrumb, User } from "@sentry/types";
 
 import { NATIVE } from "./wrapper";
 
+type SerializeableValue = string | number | boolean
+type SerializeableMap = { [key: string]: SerializeableValue }
+
 /**
  * Extends the scope methods to set scope on the Native SDKs
  */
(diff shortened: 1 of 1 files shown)


marandaneto:
@mrousavy thanks for that, this will likely be a breaking change tho.
Did you experience that on Android or iOS? it could be a serialization bug, needs a bit of investigation.

marandaneto:
> @mrousavy thanks for that, this will likely be a breaking change tho. Did you experience that on Android or iOS? it could be a serialization bug, needs a bit of investigation.

@jennmueng can you take a look at it? thanks

Krisztiaan:
This does not look like it'd be caused by the *current* client-side code, but more likely an issue on the processing side. Client currently calls `JSON.stringify`, which should handle this kind of object.

May be closable?

marandaneto:
> This does not look like it'd be caused by the _current_ client-side code, but more likely an issue on the processing side. Client currently calls `JSON.stringify`, which should handle this kind of object.
> 
> May be closable?

Apparently not since logging the beforeSend event turns out to: https://github.com/getsentry/sentry-react-native/issues/2154

```
beforeSend: (e) => {
    console.log('Event beforeSend:', e);
    return e;
}
```

marandaneto:
Closed because of https://github.com/getsentry/sentry-react-native/issues/2154#issuecomment-1073678440
We'll discuss internally if we should increase that or not
For use, increase the value of https://docs.sentry.io/platforms/react-native/configuration/options/#normalize-depth

PR #1983, opened 2021-12-20
fix: Do not report empty measurements

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: Do not report empty measurements

We're sending zeroed slow/frozen/total measurements

Running it

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes


romtsn (approved):

jennmueng (approved):
This PR looks good to me but from the SDK side shouldn't we still be reporting the measurements even if they are all zero and let the backend determine if the data is significant or not?

marandaneto:
> This PR looks good to me but from the SDK side shouldn't we still be reporting the measurements even if they are all zero and let the backend determine if the data is significant or not?

Zeroed measurements are useless, we'd be sending data that would be discarded/not counted anyway, Android/Cocoa also don't send it.

PR #1836, opened 2021-10-14
build(sample): Bump sample app to RN 0.66.4

jennmueng:
## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Bumps sample app to the latest RN 0.66.4 following the upgrade diff from https://react-native-community.github.io/upgrade-helper/

## :bulb: Motivation and Context

Keeping the sample app up to date helps us catch potential bugs.

## :green_heart: How did you test it?
Ran both sample apps locally + See e2e on CI.

## :pencil: Checklist

- [x] I reviewed submitted code
- [ ] I added tests to verify changes
- [ ] All tests passing
- [x] No breaking changes

#skip-changelog

PR #1984, opened 2021-12-20
fix: Polyfill the promise library to permanently fix unhandled rejections

### LINKED ISSUE #1942: Unhandled promise rejections are not logged to sentry
My feature request is connected to the bug https://github.com/getsentry/sentry-react-native/issues/1077. 

I've tried the [solution from docs](https://docs.sentry.io/platforms/react-native/troubleshooting/#unhandled-promise-rejections), but it didn't work out for me. So I found another solution with polyfilling existing Promise in React Native.

I've used [this package](https://github.com/iyegoroff/react-native-promise-rejection-utils)
```
import { setUnhandledPromiseRejectionTracker } from 'react-native-promise-rejection-utils';
import { getCurrentHub } from '@sentry/core';
setUnhandledPromiseRejectionTracker((id, error) => {
  getCurrentHub().captureException(error, {
    data: { id },
    originalException: error,
  });
});
```

Inside the `react-native-promise-rejection-utils` package polyfills the existing promise. You can check it out [here](https://github.com/iyegoroff/react-native-promise-rejection-utils/blob/master/index.js) 

So I propose changing [the method](https://github.com/getsentry/sentry-react-native/blob/master/src/js/integrations/reactnativeerrorhandlers.ts#L64), removing the check for the promise version and using the package `react-native-promise-rejection-utils` or polyfill directly without relaying on this package.

I think this solution is better in many ways:
1. It doesn't require user to do or check anything. It just works
2. `_checkPromiseVersion` can be removed

jennmueng:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Instead of warning the user and requiring them to set the Promise library resolution, we instead polyfill the global promise library to ensure we can install our handler.

Adds a new option `patchGlobalPromise` that can be passed to either `Sentry.init` or the `ReactNativeErrorHandlers` integration.

There is the expectation that a third party SDK shouldn't have the side effect of polyfilling a commonly used global instance, but as RN has not changed the functionality of its promises for quite a while we should be fine. If you use a polyfill such as core-js for example, such as adding `Promise.allSettled`, they would have to use the polyfill after Sentry init. We can also add back in the Promise warning along with a boolean toggle for people who do not want to polyfill.

Resolves #1942 

**NOTE**
Polyfilling like this also fixes unhandled promise rejections in Hermes, which was what has been blocking #1836.

e2e tests test unhandled promise rejections.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## Next Steps:
Rewrite https://docs.sentry.io/platforms/react-native/troubleshooting/#unhandled-promise-rejections

-> https://github.com/getsentry/sentry-docs/pull/4525

### DIFF
--- a/src/js/integrations/reactnativeerrorhandlers.ts
+++ b/src/js/integrations/reactnativeerrorhandlers.ts
@@ -1,7 +1,7 @@
 import { eventFromException } from "@sentry/browser";
 import { getCurrentHub } from "@sentry/core";
 import { Integration, Severity } from "@sentry/types";
-import { addExceptionMechanism, getGlobalObject, logger } from "@sentry/utils";
+import { addExceptionMechanism, logger } from "@sentry/utils";
 
 import { ReactNativeClient } from "../client";
 
--- a/src/js/integrations/reactnativeerrorhandlers.ts
+++ b/src/js/integrations/reactnativeerrorhandlers.ts
@@ -62,7 +62,7 @@ export class ReactNativeErrorHandlers implements Integration {
         this._polyfillPromise();
       }
 
-      // this._attachUnhandledRejectionHandler();
+      this._attachUnhandledRejectionHandler();
       this._checkPromiseAndWarn();
     }
   }
@@ -92,9 +92,8 @@ export class ReactNativeErrorHandlers implements Integration {
     /* eslint-enable import/no-extraneous-dependencies,@typescript-eslint/no-var-requires */
   }
   /**
-   *
+   * Attach the unhandled rejection handler
    */
-  // @ts-ignore temp
   private _attachUnhandledRejectionHandler(): void {
     const tracking: {
       disable: () => void;
--- a/src/js/integrations/reactnativeerrorhandlers.ts
+++ b/src/js/integrations/reactnativeerrorhandlers.ts
@@ -101,9 +101,23 @@ export class ReactNativeErrorHandlers implements Integration {
       // eslint-disable-next-line import/no-extraneous-dependencies,@typescript-eslint/no-var-requires
     } = require("promise/setimmediate/rejection-tracking");
 
-    const promiseRejectionTrackingOptions = this._getPromiseRejectionTrackingOptions();
+    const promiseRejectionTrackingOptions: PromiseRejectionTrackingOptions = {
+      onUnhandled: (id, rejection = {}) => {
+        // eslint-disable-next-line no-console
+        console.warn(
+          `Possible Unhandled Promise Rejection (id: ${id}):\n${rejection}`
+        );
+      },
+      onHandled: (id) => {
+        // eslint-disable-next-line no-console
+        console.warn(
+          `Promise Rejection Handled (id: ${id})\n` +
+            "This means you can ignore any previous messages of the form " +
+            `"Possible Unhandled Promise Rejection (id: ${id}):"`
+        );
+      },
+    };
 
-    tracking.disable();
     tracking.enable({
       allRejections: true,
       onUnhandled: (id: string, error: Error) => {
@@ -144,27 +158,6 @@ export class ReactNativeErrorHandlers implements Integration {
         "Unhandled promise rejections will not be caught by Sentry. Read about how to fix this on our troubleshooting page."
       );
     }
-  }
-  /**
-   * Gets the promise rejection handlers, tries to get React Native's default one but otherwise will default to console.logging unhandled rejections.
-   */
-  private _getPromiseRejectionTrackingOptions(): PromiseRejectionTrackingOptions {
-    return {
-      onUnhandled: (id, rejection = {}) => {
-        // eslint-disable-next-line no-console
-        console.warn(
-          `Possible Unhandled Promise Rejection (id: ${id}):\n${rejection}`
-        );
-      },
-      onHandled: (id) => {
-        // eslint-disable-next-line no-console
-        console.warn(
-          `Promise Rejection Handled (id: ${id})\n` +
-            "This means you can ignore any previous messages of the form " +
-            `"Possible Unhandled Promise Rejection (id: ${id}):"`
-        );
-      },
-    };
   }
   /**
    * Handle errors
--- a/src/js/integrations/reactnativeerrorhandlers.ts
+++ b/src/js/integrations/reactnativeerrorhandlers.ts
@@ -58,33 +58,28 @@ export class ReactNativeErrorHandlers implements Integration {
     if (this._options.onunhandledrejection) {
       /*
         In newer RN versions >=0.63, the global promise is not the same reference as the one imported from the promise library.
-        This is due to a version mismatch between promise versions. The version will need to be fixed with a package resolution.
-        We first run a check and show a warning if needed.
+        This is due to a version mismatch between promise versions.
       */
       this._polyfillPromise();
-      this._checkPromiseVersion();
     }
   }
   /**
-   *
+   * Polyfill the global promise instance with one we can be sure that we can attach the tracking to.
    */
   private _polyfillPromise(): void {
+    /* eslint-disable import/no-extraneous-dependencies,@typescript-eslint/no-var-requires*/
     const {
       polyfillGlobal,
-      // eslint-disable-next-line @typescript-eslint/no-var-requires
     } = require("react-native/Libraries/Utilities/PolyfillFunctions");
-    // eslint-disable-next-line import/no-extraneous-dependencies,@typescript-eslint/no-var-requires
     const Promise = require("promise/setimmediate/es6-extensions");
-    // eslint-disable-next-line import/no-extraneous-dependencies
+
     require("promise/setimmediate/done");
-    // eslint-disable-next-line import/no-extraneous-dependencies
     require("promise/setimmediate/finally");
 
     polyfillGlobal("Promise", () => {
       const tracking: {
         disable: () => void;
         enable: (arg: unknown) => void;
-        // eslint-disable-next-line @typescript-eslint/no-var-requires,import/no-extraneous-dependencies
       } = require("promise/setimmediate/rejection-tracking");
 
       const promiseRejectionTrackingOptions = this._getPromiseRejectionTrackingOptions();
@@ -108,6 +103,7 @@ export class ReactNativeErrorHandlers implements Integration {
       });
 
       return Promise;
+      /* eslint-enable import/no-extraneous-dependencies,@typescript-eslint/no-var-requires */
     });
   }
   /**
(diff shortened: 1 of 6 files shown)


src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -58,15 +58,28 @@ export class ReactNativeErrorHandlers implements Integration {
...
+   */
+  private _polyfillPromise(): void {
+    /* eslint-disable import/no-extraneous-dependencies,@typescript-eslint/no-var-requires*/
+    const {
+      polyfillGlobal,
+    } = require("react-native/Libraries/Utilities/PolyfillFunctions");
+    const Promise = require("promise/setimmediate/es6-extensions");
 
+    require("promise/setimmediate/done");
+    require("promise/setimmediate/finally");
```
AbhiPrasad, line 77: I’m assuming these imports patch the promise global as a side effect? Can we leave a comment to make that clear?

src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -58,15 +58,28 @@ export class ReactNativeErrorHandlers implements Integration {
...
+  }
+  /**
+   * Polyfill the global promise instance with one we can be sure that we can attach the tracking to.
+   */
+  private _polyfillPromise(): void {
+    /* eslint-disable import/no-extraneous-dependencies,@typescript-eslint/no-var-requires*/
+    const {
+      polyfillGlobal,
+    } = require("react-native/Libraries/Utilities/PolyfillFunctions");
+    const Promise = require("promise/setimmediate/es6-extensions");
```
AbhiPrasad, line 74: Do we have to try catch this? What about checking for `global.Promise`?

src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -58,15 +58,28 @@ export class ReactNativeErrorHandlers implements Integration {
...
 
+    require("promise/setimmediate/done");
+    require("promise/setimmediate/finally");
+
+    polyfillGlobal("Promise", () => {
       const tracking: {
```
AbhiPrasad, line 80: Is it fine that the require of `promise/setimmediate/es6-extensions` happens outside the function closure?

Are we in danger of other people mutating/patching the promise export?

src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -58,15 +58,28 @@ export class ReactNativeErrorHandlers implements Integration {
...
-        This is due to a version mismatch between promise versions. The version will need to be fixed with a package resolution.
-        We first run a check and show a warning if needed.
+        This is due to a version mismatch between promise versions.
       */
-      this._checkPromiseVersion();
+      this._polyfillPromise();
+    }
+  }
+  /**
+   * Polyfill the global promise instance with one we can be sure that we can attach the tracking to.
```
AbhiPrasad, line 67: Would love some code comments that essentially explain how we got here. The different promise patching strats we used, and why is is the best one.

will help the forensics in the future lol 🕵️‍♂️

AbhiPrasad (commented):
Lemme dig into this so I understand a little better before ✅ - tho scream at me on discord if you feel like this is prio and I’ll focus on it.

src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -58,15 +58,28 @@ export class ReactNativeErrorHandlers implements Integration {
...
+   */
+  private _polyfillPromise(): void {
```
marandaneto, line 69: Should we use `require.call(null, moduleName).default` instead?
Similar to https://github.com/getsentry/sentry-react-native/commit/600dcb358b54e413f8b952d4d14063f2bc943834

marandaneto: We just need to be sure that it works on previous versions of RN.

jennmueng: @marandaneto No we don't as 
- `react-native/Libraries/Utilities/PolyfillFunctions` is available [since 0.55](https://github.com/facebook/react-native/commit/f7f5dc66493ad25a85927a9503728b0491c8aab9#diff-82642ae61cae901638245d169cbe2590ff3ecb02e393cba41f07ab0208d475df)
- `promise/setimmediate/**` will always be available as the promise library has been a dependency of RN since [before 0.56.0](https://github.com/facebook/react-native/blob/3c6a916536da67a4591f71883723c67a0c7be9d7/package.json#L185) (our minimum required RN version for this SDK)


Thanks for the double check.

jennmueng: Also the require.call hack seemed to have unusual behavior with release/debug builds since last time so we should try to avoid it in future.

marandaneto:
@bruno-garcia after discussing with @jennmueng, We've agreed to add a flag to the Options, that is enabled by default, but is possible to disable patching the Global promise, which removes the side effect of poly filling the Global promise during SDK init, obviously catching unhandled promises won't work then.

Something like `patchGlobalPromise` or `polyfillGlobalPromise`.

This flag would also go to the docs, likely rewriting this section https://docs.sentry.io/platforms/react-native/troubleshooting/#unhandled-promise-rejections

src/js/integrations/reactnativeerrorhandlers.ts
jennmueng: try catching require statements don't work in older versions of React Native and the hack I tried using last time had undefined behavior so I want to avoid doing so now. These dependencies should be present since our minimum required Rn version 0.56.

src/js/integrations/reactnativeerrorhandlers.ts
jennmueng: Sure, will also mention that this mimics how RN does it exactly

bruno-garcia:
> @bruno-garcia after discussing with @jennmueng, We've agreed to add a flag to the Options, that is enabled by default, but is possible to disable patching the Global promise, which removes the side effect of poly filling the Global promise during SDK init, obviously catching unhandled promises won't work then.
> 
> Something like `patchGlobalPromise` or `polyfillGlobalPromise`.
> 
> This flag would also go to the docs, likely rewriting this section [docs.sentry.io/platforms/react-native/troubleshooting/#unhandled-promise-rejections](https://docs.sentry.io/platforms/react-native/troubleshooting/#unhandled-promise-rejections)

It's a very good idea to have an opt-out!

src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -135,6 +145,27 @@ export class ReactNativeErrorHandlers implements Integration {
       );
     }
   }
+  /**
+   * Gets the promise rejection handlers, tries to get React Native's default one but otherwise will default to console.logging unhandled rejections.
```
AbhiPrasad, line 149: > tries to get React Native’s default one

Is the doc string still true for this function?

src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -56,68 +58,76 @@ export class ReactNativeErrorHandlers implements Integration {
...
+   * Attach the unhandled rejection handler
+   */
+  private _attachUnhandledRejectionHandler(): void {
+    const tracking: {
+      disable: () => void;
+      enable: (arg: unknown) => void;
+      // eslint-disable-next-line import/no-extraneous-dependencies,@typescript-eslint/no-var-requires
+    } = require("promise/setimmediate/rejection-tracking");
+
+    const promiseRejectionTrackingOptions = this._getPromiseRejectionTrackingOptions();
```
AbhiPrasad, line 104: Can we just in-line the stuff from this function if we are just going to call the func? (We don’t return the function pointers in any way)

src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -56,68 +58,76 @@ export class ReactNativeErrorHandlers implements Integration {
...
+
+    tracking.disable();
+    tracking.enable({
```
AbhiPrasad, line 107: While we here, can we leave a comment explaining why we have the disable then enable?

AbhiPrasad (approved):
Ok I like the global flag and general approach, but I’ll keep reading up in case we get more problems with this again.

also the react native code base at https://github.com/facebook/react-native is quite the adventure to peruse 😁

AbhiPrasad:
Also, another thing I wanna make sure is fine - what happens if somebody else polyfills the global promise? Is there a danger of our implementations conflicting? Is this a pattern used by other libraries?

src/js/integrations/reactnativeerrorhandlers.ts
jennmueng: Hmm, looks like it just sets the properties on `Promise` to null before setting it again anyways in enable. So it's not needed. It's something we've kept here since the start. I'll remove it.

jennmueng:
> Also, another thing I wanna make sure is fine - what happens if somebody else polyfills the global promise? Is there a danger of our implementations conflicting? Is this a pattern used by other libraries?

@AbhiPrasad Yep, the possible conflict if the user or other libraries are polyfilling the global promise instance like us is what kept me off of this solution originally. I think having the option to disable our polyfill while also being able to manually ensure the promise instances are the same is good enough to get promise rejection tracking working on the latest RN.

AbhiPrasad (approved):
Happy new year! Let’s 🚢

marandaneto (approved):

SimenB:
Not sure if this is the correct place, but a new version of `promise` has been released which includes support for `Promise.allSettled`. However, updating to it causes a bunch of errors (`TypeError: Promise.allSettled is not a function.`). I've tried setting `patchGlobalPromise: false` which makes no difference. Rolling back to 8.1.0 works

An update has also landed within RN, so this will hit everybody somewhat soonish: https://github.com/facebook/react-native/pull/34544

krystofwoldrich:
> Not sure if this is the correct place, but a new version of `promise` has been released which includes support for `Promise.allSettled`. However, updating to it causes a bunch of errors (`TypeError: Promise.allSettled is not a function.`). I've tried setting `patchGlobalPromise: false` which makes no difference. Rolling back to 8.1.0 works
> 
> An update has also landed within RN, so this will hit everybody somewhat soonish: [facebook/react-native#34544](https://github.com/facebook/react-native/pull/34544)

Thank you for the heads up. We will definitely take a look on that.

SimenB:
I'm using expo, might have something to do with it?

marandaneto:
@SimenB please raise a new issue with more context, this issue is already closed, a minimal reproducible example would help as well.

marandaneto:
Since you are using `expo`, also double-check with them in case there are conflicts.

Issue #1999, opened 2022-01-07
Crash in fetchNativeDeviceContexts


philipphofmann:
### Environment

React-Native 3.2.10

### Steps to Reproduce

Came from a customer. TLDR; the error happens when `enableNative: true`

> 1. We can run our project with Sentry but with "enableNative: false" which we think will disable the automatic catching of exceptions and crashes on native side, correct?
> 2. We can see the issue on the iPhone but we are finding it difficult to solve it. We need guidance to resolve the issue (The issue occurs only when we run the project with enableNative: true otherwise its working fine we can track logs/messages in Sentry)
> 3. After adding debug: true and enableNative: true we get the below error (It is the same error attached in previous reply as an image):

Sentry Logger [Warn]: Unable to symbolicate stack trace: Bundle was not loaded from the packager

2022-01-05 19:03:35.157 [fatal][tid:main] Exception '*** +[NSJSONSerialization dataWithJSONObject:options:error:]: value parameter is nil' was thrown while invoking fetchNativeDeviceContexts on target RNSentry with params (

### Expected Result

No crash.

### Actual Result

Crashes with 
```
+[NSJSONSerialization dataWithJSONObject:options:error:]: value parameter is nil' was thrown while invoking fetchNativeDeviceContexts on target RNSentry with params (10716,10717)

callstack: (
0 CoreFoundation 0x00000001a151f93c 5BBDEA97-01D2-30D8-8123-43118E96A409 + 1161532
1 libobjc.A.dylib 0x00000001b52bc480 objc_exception_throw + 56
2 Foundation 0x00000001a2750134 1C20A210-3AFB-3E1A-AE51-DF3DECADD302 + 401716
3 CustomerApp 0x0000000102e7af0c __47-[RNSentry fetchNativeDeviceContexts:rejecter:]_block_invoke + 172
4 CustomerApp 0x00000001030eb980 -[SentryHub configureScope:] + 116
5 CustomerApp 0x0000000103102cc8 +[SentrySDK configureScope:] + 100
6 CustomerApp 0x0000000102e7ae30 -[RNSentry fetchNativeDeviceContexts:rejecter:] + 208
7 CoreFoundation 0x00000001a15240e0 5BBDEA97-01D2-30D8-8123-43118E96A409 + 1179872
8 CoreFoundation 0x00000001a1405e00 5BBDEA97-01D2-30D8-8123-43118E96A409 + 7680
[...]

PR #2002, opened 2022-01-12
fix: fetchNativeDeviceContexts returns an empty Array if no Device Context available

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: fetchNativeDeviceContexts returns an empty Array if no Device Context available

Avoid crash https://github.com/getsentry/sentry-react-native/issues/1999

Running it without initing the SDK

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## :crystal_ball: Next steps
Investigate why `fetchNativeDeviceContexts` returned null in the first place.

### DIFF
--- a/ios/RNSentry.m
+++ b/ios/RNSentry.m
@@ -140,19 +140,23 @@ RCT_EXPORT_METHOD(fetchNativeDeviceContexts:(RCTPromiseResolveBlock)resolve
                   rejecter:(RCTPromiseRejectBlock)reject)
 {
     NSLog(@"Bridge call to: deviceContexts");
+    NSMutableDictionary<NSString *, id> *contexts = [NSMutableDictionary new];
     // Temp work around until sorted out this API in sentry-cocoa.
     // TODO: If the callback isnt' executed the promise wouldn't be resolved.
     [SentrySDK configureScope:^(SentryScope * _Nonnull scope) {
         NSDictionary<NSString *, id> *serializedScope = [scope serialize];
         // Scope serializes as 'context' instead of 'contexts' as it does for the event.
-        NSDictionary<NSString *, id> *contexts = [serializedScope valueForKey:@"context"];
+        NSDictionary<NSString *, id> *tempContexts = [serializedScope valueForKey:@"context"];
+        if (tempContexts != nil) {
+            [contexts addEntriesFromDictionary:tempContexts];
+        }
 #if DEBUG
         NSData *data = [NSJSONSerialization dataWithJSONObject:contexts options:0 error:nil];
         NSString *debugContext = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
         NSLog(@"Contexts: %@", debugContext);
 #endif
-        resolve(contexts);
     }];
+    resolve(contexts);
 }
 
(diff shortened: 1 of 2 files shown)


ios/RNSentry.m
```diff
@@ -140,19 +140,23 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
                   rejecter:(RCTPromiseRejectBlock)reject)
 {
     NSLog(@"Bridge call to: deviceContexts");
+    NSMutableDictionary<NSString *, id> *contexts = [NSMutableDictionary new];
```
philipphofmann, line 143: `l`: If you do it like this and the comment below, you would avoid allocating an unneeded NSMutableDictionary .
```suggestion
    NSDictionary<NSString *, id> *contexts = nil;
```

ios/RNSentry.m
```diff
@@ -140,19 +140,23 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
...
     // Temp work around until sorted out this API in sentry-cocoa.
     // TODO: If the callback isnt' executed the promise wouldn't be resolved.
     [SentrySDK configureScope:^(SentryScope * _Nonnull scope) {
         NSDictionary<NSString *, id> *serializedScope = [scope serialize];
         // Scope serializes as 'context' instead of 'contexts' as it does for the event.
-        NSDictionary<NSString *, id> *contexts = [serializedScope valueForKey:@"context"];
+        NSDictionary<NSString *, id> *tempContexts = [serializedScope valueForKey:@"context"];
+        if (tempContexts != nil) {
+            [contexts addEntriesFromDictionary:tempContexts];
+        }
```
philipphofmann, line 152: `l`:
```suggestion
        contexts = [serializedScope valueForKey:@"context"];
        if (contexts == nil) {
            context = [NSDictionary new];
        }
```

ios/RNSentry.m
```diff
@@ -140,19 +140,23 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
...
 #if DEBUG
         NSData *data = [NSJSONSerialization dataWithJSONObject:contexts options:0 error:nil];
         NSString *debugContext = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
         NSLog(@"Contexts: %@", debugContext);
 #endif
-        resolve(contexts);
     }];
+    resolve(contexts);
```
philipphofmann, line 159: `m`: Did you move this out of the callback on purpose?

philipphofmann (commented):
One comment about the callback.

ios/RNSentry.m
marandaneto: yes, otherwise the Promise never gets resolved, the `configureScope` callback isn't called if the SDK isn't init'ed.

ios/RNSentry.m
marandaneto: if I do that, I need to check the `contexts` nullability outside of the callback too, due to https://github.com/getsentry/sentry-react-native/pull/2002#discussion_r783057923

philipphofmann (approved):
LGTM 🚀

PR #2014, opened 2022-01-20
fix: serialize circular inputs without throwing

### LINKED ISSUE #1477: react-navigation non-serializable state recursive call to `dropUndefinedKeys`
### Environment

`sentry.io`
`"@sentry/react-native": "^2.4.0"`
`react-navigation`

### Steps to Reproduce

1. Pass a circular object as a `param` to a navigation call

### Expected Result

Navigation successful, something's logged in Sentry

### Actual Result

RangeError, call stack size exceeded in `dropUndefinedKeys`

Krisztiaan:

- [x] Bugfix

## :scroll: Description
Added a `replacer` for some `JSON.serialize` calls, to avoid unprotected throws due to circular constructs.

Solves a potential crash when data from userland is circular.
Resolves #1477 

Added unit test

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
🚢 
If you could also advise where to put the replacer creator function, that'd be nice.


marandaneto:
@Krisztiaan thanks for doing this.
I will let @jennmueng and @AbhiPrasad review this since they own the JS know-how.

Krisztiaan:
> @Krisztiaan thanks for doing this.
> I will let @jennmueng and @AbhiPrasad review this since they own the JS know-how.

No problems, this is very close to what I ended up doing before we pass anything to sentry in our project.

marandaneto:
@Krisztiaan btw:

> /home/runner/work/sentry-react-native/sentry-react-native/test/wrapper.test.ts
>   576:0  error  Parsing error: Declaration or statement expected

AbhiPrasad:
We have the `walk` utility in the `@sentry/utils` package which does similar, we can probably just use that instead of a custom replacer: https://github.com/getsentry/sentry-javascript/blob/fcd97c44b6e9f38c6aa9abeee06368c88ab08ea5/packages/utils/src/object.ts#L312

See how it's used here: https://github.com/getsentry/sentry-javascript/blob/fcd97c44b6e9f38c6aa9abeee06368c88ab08ea5/packages/utils/src/object.ts#L360-L379

marandaneto:
> We have the `walk` utility in the `@sentry/utils` package which does similar, we can probably just use that instead of a custom replacer: [getsentry/sentry-javascript@`fcd97c4`/packages/utils/src/object.ts#L312](https://github.com/getsentry/sentry-javascript/blob/fcd97c44b6e9f38c6aa9abeee06368c88ab08ea5/packages/utils/src/object.ts#L312)
> 
> See how it's used here: [getsentry/sentry-javascript@`fcd97c4`/packages/utils/src/object.ts#L360-L379](https://github.com/getsentry/sentry-javascript/blob/fcd97c44b6e9f38c6aa9abeee06368c88ab08ea5/packages/utils/src/object.ts#L360-L379)

@Krisztiaan would you like to update this PR to use the Sentry's utils method?

Krisztiaan:
Thanks, I'll do a little bit of PR housekeeping this weekend.

Krisztiaan:
@marandaneto the change is done, also saw a place to bring in `normalize` :)

AbhiPrasad (approved):

bruno-garcia:
Lining seems to have Failed:

Ideally we'd automate this as CI failures due to linting can be frustrating. But could you please fix this one?

/home/runner/work/sentry-react-native/sentry-react-native/src/js/wrapper.ts
9
  2:1  error  Run autofix to sort these imports!  simple-import-sort/sort
10

11
/home/runner/work/sentry-react-native/sentry-react-native/test/wrapper.test.ts
12
  576:0  error  Parsing error: Declaration or statement expected
13

14
✖ 2 problems (2 errors, 0 warnings)
15
  1 error and 0 warnings potentially fixable with the `--fix` option.

marandaneto:
@Krisztiaan `lint` is still unhappy.

> 11
> /home/runner/work/sentry-react-native/sentry-react-native/test/wrapper.test.ts
> 12
> 576:0 error Parsing error: Declaration or statement expected
> 13

Krisztiaan:
@marandaneto tests are failing with `walk`, due to different parsing logic than what's expected here.
image

marandaneto:
> @marandaneto tests are failing with `walk`, due to different parsing logic than what's expected here. (`serializeValue`) Pushed so you can see the fail cases too. Advise on how you'd want to reconcile, please. Either test expectations are now wrong, or `serializeValue` does too much in `walk` to use here.
> 
> For a more thorough example, something like [`cycle.js`](https://github.com/douglascrockford/JSON-js/blob/master/cycle.js) could be used to decycle objects, so a replacer would not be needed here.
> 
> image

mmm, Maybe @AbhiPrasad or @jennmueng can help here.

Krisztiaan:
Do let me know what's up. Imo this could be somewhat important to deal with in a timely manner, as it can cause unexpected crashes, related to a crash detection lib, and that is not that cool. However easy to fix.

marandaneto:
> Do let me know what's up. Imo this could be somewhat important to deal with in a timely manner, as it can cause unexpected crashes, related to a crash detection lib, and that is not that cool. However easy to fix.

@AbhiPrasad @bruno-garcia @jennmueng ping, in case you all can help us to unblock this.

src/js/wrapper.ts
```diff
@@ -144,7 +144,7 @@ export const NATIVE: SentryNativeWrapper = {
       };
 
       // Serialize and remove any instances that will crash the native bridge such as Spans
-      const serializedPayload = JSON.parse(JSON.stringify(payload));
+      const serializedPayload = normalize(payload);
```
marandaneto, line 147:
```suggestion
      const serializedPayload = JSON.parse(JSON.stringify(payload, walk));
```

@Krisztiaan I believe this would fix the broken tests, we can fix the normalization in another PR.

marandaneto: Oh `normalize` actually calls `JSON.parse(JSON.stringify(...))` so not really.

Krisztiaan:
This is not a `react-navigation` only issue. Pass a circular object to any of these calls, or just try the added tests from this PR without the functional modifications of this PR to see `sentry` breaking with `JSON.stringify` failing.

> So your changes won't fix the original issue anyway.

According to the tests, this is a fix, but `walk` causes side-effects that you'd need to make a decision on, which version to use, or `walk` to be discarded, and another way of de-cycling used.

Issue #2052, opened 2022-02-10
e2e tests broken on iOS


jennmueng:
Our e2e tests are broken on iOS. Still works on Android. The tests also only fail on CI, locally they work just fine. 

The appium driver for some reason cannot detect the buttons on the screen. No changes were made to the e2e tests between the last working version and the latest, and there seemed to be no new version of Appium. Only thing that changed was [Github Actions updated the default XCode version from 13.1 to 13.2.1](https://github.com/actions/virtual-environments/commit/3351872cbebeb034a8b8708f53d2f454ae4c5529#diff-e68d4f6e862f03aa8800c0d5de1c1072dd2d01553ca0986d818f012c932a25a4R12)

Avenues I've tested:
- Attached XCode debugger and made sure that the accessibilityId was set just fine.
- Made sure I am using latest XCode and latest iOS simulator locally to match the CI.
- Took a screenshot on CI when running the tests to make sure the screen actually rendered and the app did not crash
- Downgraded Appium/XCUITest Driver (ran into other issues)

jennmueng:
@marandaneto @bruno-garcia Closing this as it seemed to have fixed itself?

PR #2053, opened 2022-02-11
fix: Discard prior transactions on react navigation dispatch

### LINKED ISSUE #2051: Users see transactions name `Route Change` in Sentry.
Users see transactions name `Route Change` in Sentry.

It seems this is a template here that should be replaced with the right route name:

https://github.com/getsentry/sentry-react-native/blob/4fd13137fe71984d1166c33198076da9db28f158/src/js/tracing/utils.ts#L14-L19

Router integration used: React Navigation

jennmueng:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

When multiple navigation actions are dispatched before the `routeChangeTimeoutMs`, and the state does not change occur for a prior action, the transaction is not cancelled and sent with the blank "Route Change" transaction context. This change makes sure any navigation transaction that had a successful navigation state change is removed from the instrumentation (it'll still be present on the scope until finished) and any new navigation action dispatch will **cancel** any prior waiting transactions to ensure they won't be sent when they don't have a successful state change.

Fixes #2051 

Added a jest test where multiple navigation actions were dispatched in a row

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
As we did not recieve a reproduction, need to ask customer to see if bug is fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- fix: Discard prior transactions on react navigation dispatch #2053
+
 ## 3.2.13
 
 - fix(deps): Add `@sentry/wizard` back in as a dependency to avoid missing dependency when running react-native link. #2015
--- a/src/js/tracing/reactnavigation.ts
+++ b/src/js/tracing/reactnavigation.ts
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
 import { Transaction as TransactionType } from "@sentry/types";
 import { getGlobalObject, logger } from "@sentry/utils";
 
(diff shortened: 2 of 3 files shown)


bruno-garcia (approved):
The e2e for iOS broken is expected due to #2052 
But Android should be green but seems to be flaky due to:

```
Error: The template is not valid. .github/workflows/e2e.yml (Line: 34, Col: 16): hashFiles('**/yarn.lock') couldn't finish within 120 seconds.
```

bruno-garcia:
image
We're good!

Issue #950, opened 2020-07-03
Touch Events displayName not working


jbolotin:
**OS:**
- [ ]  Windows
- [X]  MacOS
- [ ]  Linux

**Platform:**
- [ ]  iOS
- [X]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 0.0.0
1.5.0
**`react-native` version:** 0.0.0
0.62.2
**Are you using Expo?**
- [ ]  Yes
- [X]  No

**Are you using sentry.io or on-premise?**
- [X]  sentry.io (SaaS)
- [ ]  on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

[link](https://bit.ly/2NSWxCH)

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  dsn,
  integrations: [new ExtraErrorData()],
  attachStacktrace: true,
});
```

or

(`react-native-sentry`)
```
Sentry.config(
  'https://...@sentry.io/...'
  // other options
 ).install();
```

---
**I have following issue:**

I'm following the [new docs](https://docs.sentry.io/platforms/react-native/touchevents/) for Touch Events, and I think that this is a fantastic new feature. Unfortunately, when I set `displayName` as a property on a component that contains a `TouchableOpacity`, it only works on some of the components. Most of the components it just shows `Image` or `Svg` even if I set a `displayName` on the component that has the `TouchableOpacity` in it

**Steps to reproduce:**
- Set `displayName` on functional component
- Click it to trigger touch event
- Trigger error elsewhere

**Actual result:**

Some touch events have the correct name, but most do not

**Expected result:**

All touch events have name set correctly for components where it is set

jennmueng:
@jbolotin So the way react native handles touch events is that it gradually works up the component tree. Our SDK follows that and if it sees a component with a `displayName` it will log that. What I think is happening in your case is that `Image` and `Svg` are also using the `displayName` and there's no way for it to tell whether that's a `displayName` you've set or it's something else. You can avoid this by including it in `ignoredDisplayNames` in the options for the `TouchEventBoundary` here: https://docs.sentry.io/platforms/react-native/touchevents/#ignoreddisplaynames. Let me know if this works for you.

[...]

PR #2068, opened 2022-02-17
feat: Touch events now track components with `sentry-label`

### LINKED ISSUE #1913: Use `accessibilityLabel` for identifying touch events
Hi!

I am not sure if this is doable, but properly tracking which components have been touched is a bit cumbersome as you need to set `displayName` for the component, so:
- it needs to be a component
- you need separate component for every touchable in the view to differentiate them, ie, you cant directly use
  ```tsx
  <View>
    <Button />
    <Button />
  </View>
   ```
  and know which button has been pressed.

Similarly to some testing libraries, it would be quite useful to enrich the event with the `accessibilityLabel` prop on the touched component (and its parents?). It would give much more useful information in traces and would immediately benefit every accessible app.

jennmueng:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Support tracking touch events on components with the prop `sentry-label`. 

Note that this does not throw a typescript error as ["it is not a valid JS Identifier"](https://www.typescriptlang.org/docs/handbook/jsx.html#attribute-type-checking)

It will fall back to other properties, with the hierarchy `sentry-label` > `accessibilityLabel` > `displayName`.

Also, removed the tracking of `name` on components without a displayName, found that doing so just caused components to be logged twice.

Resolves #1913 
Likely #950 too

Tested on sample app, and added unit tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


marandaneto:
@AbhiPrasad or @bruno-garcia would you be able to review this? Looks good to me but I don't know much about the component tree and its properties, thanks.

AbhiPrasad (approved):

bruno-garcia:
Thanks @jennmueng 
Does this require a migration guide update? Seems like it does need a docs entry though?

jennmueng:
@bruno-garcia Yep, working on the docs PR

Issue #1362, opened 2021-02-24
Support `enableNativeCrashHandling` for iOS


marandaneto:
Right now `enableNativeCrashHandling` only works for Android.

marandaneto:
related to https://github.com/getsentry/sentry-dart/issues/317

PR #2101, opened 2022-03-03
feat: Support enableNativeCrashHandling for iOS

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
feat: Support enableNativeCrashHandling for iOS

Closes https://github.com/getsentry/sentry-react-native/issues/1362

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/ios/RNSentry.m
+++ b/ios/RNSentry.m
@@ -77,6 +77,15 @@ RCT_EXPORT_METHOD(initNativeSdk:(NSDictionary *_Nonnull)options
         reject(@"SentryReactNative", error.localizedDescription, error);
         return;
     }
+    
+    BOOL enableNativeCrashHandling = (BOOL)[mutableOptions valueForKey:@"enableNativeCrashHandling"];
+    
+    if (!enableNativeCrashHandling) {
+        NSMutableArray *integrations = sentryOptions.integrations.mutableCopy;
+        [integrations removeObject:@"SentryCrashIntegration"];
+        sentryOptions.integrations = integrations;
+    }
+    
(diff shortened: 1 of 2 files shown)


ios/RNSentry.m
```diff
@@ -77,6 +77,15 @@ + (BOOL)requiresMainQueueSetup {
         reject(@"SentryReactNative", error.localizedDescription, error);
         return;
     }
+    
+    BOOL enableNativeCrashHandling = (BOOL)[mutableOptions valueForKey:@"enableNativeCrashHandling"];
```
bruno-garcia, line 81: Is this going to default to `@NO` if there's no value in the key?

bruno-garcia (commented):
Is this something we can test with the e2e tests?

ios/RNSentry.m
marandaneto: No, there's always a value, https://github.com/getsentry/sentry-react-native/blob/main/src/js/sdk.tsx#L29

marandaneto: I've added a safeguard since one could forcefully `enableNativeCrashHandling: null,`

philipphofmann (approved):
LGTM 🚀

marandaneto:
> Is this something we can test with the e2e tests?

Unless we have different e2e setups, if we remove this integration, the app won't crash, so technically possible.
Since this is only about removing an integration, unit tests would be enough.

PR #2120, opened 2022-03-08
Fix sample app values and logs

marandaneto:
_#skip-changelog_

### DIFF
--- a/sample/src/App.tsx
+++ b/sample/src/App.tsx
@@ -53,6 +53,7 @@ Sentry.init({
       },
     }),
   ],
+  enableAutoSessionTracking: true,
   // For testing, session close when 5 seconds (instead of the default 30) in the background.
   sessionTrackingIntervalMillis: 5000,
   // This will capture ALL TRACES and likely use up all your quota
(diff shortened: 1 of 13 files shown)


sample/src/App.tsx
```diff
@@ -36,11 +36,11 @@ Sentry.init({
   onReady: ({didCallNativeInit}) => {
     console.log('onReady called with didCallNativeInit:', didCallNativeInit);
   },
-  maxBreadcrumbs: 150, // Extend from the default 100 breadcrumbs.
```
marandaneto, line 39: JS does not allow more than 100 crumbs, there's an absolut number.
So either we remove it or just set it to 100.

sample/src/App.tsx
```diff
@@ -36,11 +36,11 @@ Sentry.init({
...
   integrations: [
     new Sentry.ReactNativeTracing({
-      idleTimeout: 5000, // This is the default timeout
-      routingInstrumentation: reactNavigationV5Instrumentation,
+      // The time to wait in ms until the transaction will be finished, For testing, default is 1000 ms
+      idleTimeout: 5000,
```
marandaneto, line 42: Default is 1000 and not 5000.

sample/src/App.tsx
```diff
@@ -52,7 +52,6 @@ Sentry.init({
       },
     }),
   ],
-  enableAutoSessionTracking: true,
```
marandaneto, line 55: It's enabled by default.

jennmueng (approved):

PR #2220, opened 2022-04-26
fix: replace git.io links with redirect targets

asottile-sentry:
see: https://github.blog/changelog/2022-04-25-git-io-deprecation/

Committed via https://github.com/asottile/all-repos

#skip-changelog


marandaneto (approved):
Thanks @asottile-sentry

Issue #1813, opened 2021-10-06
does this library support react-native-macos?


thefinnomenon:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
@sentry/react-native 3.1.1

### Steps to Reproduce

1. `yarn add @sentry/react-native`
2. `npx @sentry/wizard -i reactNative -p ios`
3. `pod install`
4. `react native macos`

### Expected Result

The Mac app builds and reports errors to Sentry.

### Actual Result
```
/Users/finn/Projects/react-native-universal-monorepo/packages/macos/node_modules/@sentry/react-native/ios/RNSentry.m:75:44: error: no known instance method for selector 'applicationState'
    if ([[UIApplication sharedApplication] applicationState] == UIApplicationStateActive && !sentHybridSdkDidBecomeActive && sentryOptions.enableAutoSessionTracking) {
                                           ^~~~~~~~~~~~~~~~
In module 'AppKit' imported from /Users/finn/Projects/react-native-universal-monorepo/packages/macos/macos/Pods/Headers/Public/React-Core/React/RCTUIKit.h:124:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/AppKit.framework/Headers/NSApplication.h:137:12: note: receiver is instance of class declared here
@interface NSApplication : NSResponder <NSUserInterfaceValidations, NSMenuItemValidation, NSAccessibilityElement, NSAccessibility>
           ^
/Users/finn/Projects/react-native-universal-monorepo/packages/macos/node_modules/@sentry/react-native/ios/RNSentry.m:75:65: error: use of undeclared identifier 'UIApplicationStateActive'
    if ([[UIApplication sharedApplication] applicationState] == UIApplicationStateActive && !sentHybridSdkDidBecomeActive && sentryOptions.enableAutoSessionTracking) {
                                                                ^
/Users/finn/Projects/react-native-universal-monorepo/packages/macos/node_modules/@sentry/react-native/ios/RNSentry.m:75:44: error: no known instance method for selector 'applicationState'
    if ([[UIApplication sharedApplication] applicationState] == UIApplicationStateActive && !sentHybridSdkDidBecomeActive && sentryOptions.enableAutoSessionTracking) {
                                           ^~~~~~~~~~~~~~~~
[...]

PR #2240, opened 2022-05-04
MacOS support

ospfranco:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

By conditionally compiling some of the more advanced functionality the SDK now compiles on react-native-macos applications. Don't know how the commented functionality works, XCode was simply throwing errors on the methods, which I guess are not exposed/defined based on platform checks too.

I need react-native-macos support for my applications. The commented functionality (performance tracking) is not vital for my use case, for now getting the errors is better than nothing.

I built my own [app](https://ospfranco.com/sol) with support and the events are being received on Sentry. Not sure how to add tests for the macOS target though, since I didn't change any functionality not sure if this is necessary.

## :pencil: Checklist

- [] I added tests to verify changes
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Please review, let me know if you **need** me to add tests (and if so how/which). If everything is working a prompt release would be much appreciatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.4.3
+
+- feat: Support macOS (#2240) by @ospfranco
+
 ## 3.4.2
 
 - fix: Fix cold start appearing again after js bundle reload on Android. #2229
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 3.4.3
+## Unreleased
 
 - feat: Support macOS (#2240) by @ospfranco
 
--- a/ios/RNSentry.m
+++ b/ios/RNSentry.m
@@ -33,14 +33,6 @@ static bool didFetchAppStart;
     return YES;
 }
 
-+ (NSNotificationName)didBecomeActiveNotificationName {
-#if TARGET_OS_IPHONE || TARGET_OS_MACCATALYST
-    return UIApplicationDidBecomeActiveNotification;
-#else
-    return NSApplicationDidBecomeActiveNotification;
-#endif
-}
-
 RCT_EXPORT_MODULE()
 
 - (NSDictionary<NSString *, id> *)constantsToExport
@@ -107,24 +99,20 @@ RCT_EXPORT_METHOD(initNativeSdk:(NSDictionary *_Nonnull)options
 
     [SentrySDK startWithOptionsObject:sentryOptions];
 
-#if TARGET_OS_IPHONE
+#if TARGET_OS_IPHONE || TARGET_OS_MACCATALYST
+    BOOL appIsActive = [[UIApplication sharedApplication] applicationState] == UIApplicationStateActive;
+#else
+    BOOL appIsActive = [[NSApplication sharedApplication] isActive];
+#endif
+
     // If the app is active/in foreground, and we have not sent the SentryHybridSdkDidBecomeActive notification, send it.
-    if ([[UIApplication sharedApplication] applicationState] == UIApplicationStateActive && !sentHybridSdkDidBecomeActive && sentryOptions.enableAutoSessionTracking) {
+    if (appIsActive && !sentHybridSdkDidBecomeActive && sentryOptions.enableAutoSessionTracking) {
         [[NSNotificationCenter defaultCenter]
             postNotificationName:@"SentryHybridSdkDidBecomeActive"
             object:nil];
 
         sentHybridSdkDidBecomeActive = true;
     }
-#else
-    if([[NSApplication sharedApplication] isActive] && !sentHybridSdkDidBecomeActive && sentryOptions.enableAutoSessionTracking) {
-        [[NSNotificationCenter defaultCenter]
-         postNotificationName:@"SentryHybridSdkDidBecomeActive"
-         object:nil];
-
-        sentHybridSdkDidBecomeActive = true;
-    }
-#endif
 
 
 
--- a/ios/RNSentry.m
+++ b/ios/RNSentry.m
@@ -85,7 +85,6 @@ RCT_EXPORT_METHOD(initNativeSdk:(NSDictionary *_Nonnull)options
         }
     }
 
-
     // Enable the App start and Frames tracking measurements
     if ([mutableOptions valueForKey:@"enableAutoPerformanceTracking"] != nil) {
         BOOL enableAutoPerformanceTracking = (BOOL)[mutableOptions valueForKey:@"enableAutoPerformanceTracking"];
--- a/ios/RNSentry.m
+++ b/ios/RNSentry.m
@@ -33,6 +33,14 @@ static bool didFetchAppStart;
     return YES;
 }
 
++ (NSNotificationName)didBecomeActiveNotificationName {
+#if TARGET_OS_IPHONE || TARGET_OS_MACCATALYST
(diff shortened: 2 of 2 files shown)


ios/RNSentry.m
```diff
@@ -85,16 +85,19 @@ + (BOOL)requiresMainQueueSetup {
...
 
+#if TARGET_OS_IPHONE
     // Enable the App start and Frames tracking measurements
     if ([mutableOptions valueForKey:@"enableAutoPerformanceTracking"] != nil) {
         BOOL enableAutoPerformanceTracking = (BOOL)[mutableOptions valueForKey:@"enableAutoPerformanceTracking"];
 
         PrivateSentrySDKOnly.appStartMeasurementHybridSDKMode = enableAutoPerformanceTracking;
         PrivateSentrySDKOnly.framesTrackingMeasurementHybridSDKMode = enableAutoPerformanceTracking;
     }
+#endif
```
marandaneto, line 96: I think only `framesTrackingMeasurementHybridSDKMode` needs to be protected here, so I'd rather do what the Flutter SDK does, https://github.com/getsentry/sentry-dart/blob/main/flutter/ios/Classes/SentryFlutterPluginApple.swift#L146-L148=

ios/RNSentry.m
```diff
@@ -215,6 +227,7 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
                   rejecter:(RCTPromiseRejectBlock)reject)
 {
 
+#if TARGET_OS_IPHONE
```
marandaneto, line 230: That also works for `macCatalyst`, so I'd rather add that too.
https://github.com/getsentry/sentry-dart/blob/main/flutter/ios/Classes/SentryFlutterPluginApple.swift#L368-L383=

marandaneto:
The Flutter SDK does the same thing, https://github.com/getsentry/sentry-dart/blob/main/flutter/ios/Classes/SentryFlutterPluginApple.swift
With `macCatalyst` support as well, can we do that in this PR?
I'd also avoid code duplication.

ospfranco:
Sure, thanks for the review, I'm not very experienced in the native space and also no idea about other sentry features, but the snippets in the FlutterSDK do make a lot of sense :)

marandaneto:
> Sure, thanks for the review, I'm not very experienced in the native space and also no idea about other sentry features, but the snippets in the FlutterSDK do make a lot of sense :)

Happy to guide :) thanks for the PR.

ios/RNSentry.m
```diff
@@ -85,16 +85,19 @@ + (BOOL)requiresMainQueueSetup {
...
         BOOL enableAutoPerformanceTracking = (BOOL)[mutableOptions valueForKey:@"enableAutoPerformanceTracking"];
 
         PrivateSentrySDKOnly.appStartMeasurementHybridSDKMode = enableAutoPerformanceTracking;
         PrivateSentrySDKOnly.framesTrackingMeasurementHybridSDKMode = enableAutoPerformanceTracking;
     }
+#endif
 
     [SentrySDK startWithOptionsObject:sentryOptions];
 
+#if TARGET_OS_IPHONE
```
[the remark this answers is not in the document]
ospfranco: This one lost me. The notification name would work with listeners, but this code looks like it checks for this condition once. Also, the `!sentHybridSdkDidBecomeActive` wouldn't work either.

A bit more guidance on what is the expected resulting code would be great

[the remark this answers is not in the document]
ospfranco: https://github.com/getsentry/sentry-react-native/pull/2240/commits/66605b9c8295d10d51255e82c3692bc4a421facb

Does this work?

ios/RNSentry.m
```diff
@@ -85,18 +85,28 @@ + (BOOL)requiresMainQueueSetup {
...
+#endif
     }
 
+
     [SentrySDK startWithOptionsObject:sentryOptions];
 
+#if TARGET_OS_IPHONE || TARGET_OS_MACCATALYST
+    BOOL appIsActive = [[UIApplication sharedApplication] applicationState] == UIApplicationStateActive;
+#else
+    BOOL appIsActive = [[NSApplication sharedApplication] isActive];
```
marandaneto, line 105: I'm not entirely sure this is the equivalent of `UIApplicationStateActive` for `NSApplication`.
`UIApplication` has 3 states, while `isActive` returns a boolean.
on Flutter, we register a listener, https://github.com/getsentry/sentry-dart/blob/main/flutter/ios/Classes/SentryFlutterPluginApple.swift#L35-L54=
and flip a flag if its active `didReceiveDidBecomeActiveNotification`.
I don't recall the reason, maybe this is because `NSApplication` does not have the equivalent of `applicationState` for `UIApplication`

marandaneto: @philipphofmann or @brustolin ideas?

ospfranco: I think the application state on iOS has to do more with which app is on the forefront, whereas on macOS apps can still be active but not key. In either case this is only the app start notification (from what I can see) so it shouldn't really change too much as long as the notification that SDK has been initialized is dispatched.

philipphofmann: Using `isActive` should be fine here, as we only post it when initializing the SDK. We use the `SentryHybridSdkDidBecomeActive`  for OOM, which doesn't work for macOS and for sessions.

philipphofmann (commented):
Thanks for helping us out on this @ospfranco.

ios/RNSentry.m
```diff
@@ -85,18 +85,28 @@ + (BOOL)requiresMainQueueSetup {
         }
     }
 
+
```
marandaneto, line 88:
```suggestion
```

ios/RNSentry.m
```diff
@@ -85,18 +85,28 @@ + (BOOL)requiresMainQueueSetup {
...
     if ([mutableOptions valueForKey:@"enableAutoPerformanceTracking"] != nil) {
         BOOL enableAutoPerformanceTracking = (BOOL)[mutableOptions valueForKey:@"enableAutoPerformanceTracking"];
 
         PrivateSentrySDKOnly.appStartMeasurementHybridSDKMode = enableAutoPerformanceTracking;
+#if TARGET_OS_IPHONE || TARGET_OS_MACCATALYST
         PrivateSentrySDKOnly.framesTrackingMeasurementHybridSDKMode = enableAutoPerformanceTracking;
+#endif
     }
 
+
```
marandaneto, line 99:
```suggestion
```

marandaneto:
@ospfranco Please add an entry to the `CHANGELOG`, if CI is happy, I guess we are happy.

> Not sure how to add tests for the macOS target though

We have E2E tests running on iOS, so it should be fine as it is.
We don't have a macOS pipeline on CI, but that does not need to be addressed on this PR, rather https://github.com/getsentry/sentry-react-native/issues/1813

ospfranco:
@marandaneto done, someone with access needs to approve the workflow to run the tests

brustolin:
It looks good to me.

ospfranco:
Tests are not happy, but doesn't look like it is due to my changes.

marandaneto (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.4.3
```
marandaneto, line 3:
```suggestion
## Unreleased
```

PR #2250, opened 2022-05-20
Bump Sentry JavaScript 7

lucas-zimerman:

- [ ] Bugfix
- [X] New feature
- [X] Enhancement
- [X] Refactoring

## :scroll: Description

- Bump Sentry JavaScript to version 7.
- Add support for Mac M1 on the Sample (arm64 simulator)
- Fix Break changes.
- Add option to debug the sample app on iOS
- Merged the backend on the client code.
- Wrapper now sends envelopes and not events.

For now I just run the sample and waited for an event from the native SDK.

https://sentry.io/organizations/sentry-sdks/discover/sentry-react-native:d1f62b5dbf5c4c8daf8f8bb737e2680f/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&query=&sort=-timestamp&statsPeriod=24h&yAxis=count%28%29

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

- [x]  Fix tests.
- [x]  Fix Break changes that may not be implemented correctly.
- [x]  Refactor how to correctly consume Envelopes to be sent to the Native layer.
- [x]  Fix Scope Sync.
- [x]  Validate if we'll use the flag `allowSyntheticDefaultImports`
- [x]  Check if the Client/Transport layer are working properly.
- [x] Fix Sdk version/name on CaptureEnvelope
- [x]  Validate if remaining integrations are working properly.


lucas-zimerman:
@marandaneto  Should we drop the support of Severity in favor of SeverityLevel, or keep both?

marandaneto:
@AbhiPrasad and @lobsterkatie Would you be so kind to do a quick/high-level review in this PR since you all have worked on the JS 7 major release? otherwise, I'd need to learn all the breaking changes just for the review.
Much appreciated.

AbhiPrasad:
Please drop support of Severity in favor of SeverityLevel!

cc @getsentry/team-web-sdk-frontend in general, please lets get more eyes on this!

AbhiPrasad:
Also - now that the official release has been cut, we now have a proper final CHANGELOG: https://github.com/getsentry/sentry-javascript/releases/tag/7.0.0

kopax-polyconseil:
Also, it seems the entrypoint is still using `dist` for react-native:

```
 Type 'import("/home/dka/workspace/github.com/pass-culture/pass-culture-app-native/node_modules/@sentry/types/types/scope").Scope' is not assignable to type 'import("/home/dka/workspace/github.com/pass-culture/pass-culture-app-native/node_modules/@sentry/react-native/node_modules/@sentry/types/dist/scope").Scope'.
```

Is this related to `@sentry/react-native` or `@sentry/types` `package.json` ?

src/js/sdk.tsx
```diff
@@ -65,10 +69,10 @@ export function init(passedOptions: ReactNativeOptions): void {
     ];
 
     if (__DEV__) {
-      options.defaultIntegrations.push(new DebugSymbolicator());
+      passedOptions.defaultIntegrations.push(new DebugSymbolicator());
```
lucas-zimerman, line 72: @marandaneto  or @AbhiPrasad should we push those integrations to the defaultIntegration or the integration itself from options? (options no longer has a parameter called defaultIntegration.

marandaneto:
> Android e2e seems broken, but not due to this PR, see a test that ran on the release of 3.4.3: [getsentry/sentry-react-native/runs/6703000870?check_suite_focus=true](https://github.com/getsentry/sentry-react-native/runs/6703000870?check_suite_focus=true)
> 
> EDIT: The Code for consuming Envelopes seems broken on Android, I'll give it a look on it.

The e2e test is on iOS, it's an issue with the GH action environments, it breaks sometimes and it fixes by itself, it's strange, and never found out the reason.

src/js/client.ts
```diff
@@ -1,33 +1,75 @@
...
+    options.transport = this._chooseTransport;
+    options.stackParser = defaultStackParser;
+
+    this._browserClient = new BrowserClient({
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      transport: this._chooseTransport,
+      stackParser: defaultStackParser,
+      integrations: defaultIntegrations,
+    });
+  }
```
AbhiPrasad, line 51:
```suggestion
  public constructor(options: ReactNativeClientOptions) {
    options.transport = this._chooseTransport.bind(this);
    
    super(options);
 
    // This is a workaround for now using fetch on RN, this is a known issue in react-native and only generates a warning
    // YellowBox deprecated and replaced with with LogBox in RN 0.63
    if (LogBox) {
      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
      LogBox.ignoreLogs(['Require cycle:']);
    } else {
      // eslint-disable-next-line deprecation/deprecation
      YellowBox.ignoreWarnings(['Require cycle:']);
    }

    void this._initNativeSdk();
   
    this._browserClient = new BrowserClient({
      transport: options.transport,
      stackParser: options.stackParser || defaultStackParser,
      integrations: [],
    });
  }
```

src/js/client.ts
```diff
@@ -1,33 +1,75 @@
...
+
+  /**
+   * @inheritDoc
+   */
+  public eventFromException(_exception: unknown, _hint?: EventHint): PromiseLike<Event> {
+    return this._browserClient.eventFromException(_exception, _hint);
```
AbhiPrasad, line 57: You might be able to ignore my suggestion from above, and remove these overloaded methods by just extended the `BrowserClient` instead of the `BaseClient`.

src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -201,12 +200,12 @@ export class ReactNativeErrorHandlers implements Integration {
 
         const options = client.getOptions();
 
-        const event = await eventFromException(error, {
-          originalException: error,
-        }, options.attachStacktrace);
+        const event = await client.eventFromException(error, { // FIX REACT NATIVE CLIENT.
```
AbhiPrasad, line 203: We can export these methods again from the SDK if that makes it easier for you.

src/js/sdk.tsx
AbhiPrasad: let's push this onto integrations option key and stay away from mutating default integrations.

AbhiPrasad (commented):
1st pass

lucas-zimerman:
Ok I spotted the problem with the Java SDK.

https://github1s.com/getsentry/sentry-java/blob/HEAD/sentry/src/main/java/io/sentry/protocol/Message.java
and
https://github1s.com/getsentry/sentry-javascript/blob/HEAD/packages/types/src/event.ts#L19-L20
The Message parameter from both SDKs are now incompatible (Android/JavaScript), the same issue does not happen with cocoa.

src/js/client.ts
lucas-zimerman: I made a minor change by changing
```
    options.transport = this._chooseTransport.bind(this);
    
    super(options);
```
to
```
    super(options);
    
    options.transport = this._chooseTransport.bind(this);
```
because `super` must be called before accessing `this` in the constructor of a derived class.

src/js/tracing/nativeframes.ts
```diff
@@ -150,12 +150,15 @@ export class NativeFramesInstrumentation {
     const measurements = {
       frames_total: {
         value: finalFinishFrames.totalFrames - startFrames.totalFrames,
+        unit: ''
```
marandaneto, line 153: Can we just ignore them or its mandatory now?

src/js/wrapper.ts
```diff
@@ -66,105 +68,87 @@ interface SentryNativeWrapper {
...
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-      envelopeWasSent = await RNSentry.captureEnvelope({
-        header,
-        payload: serializedPayload,
-      });
-    }
+      for (const envelopeItems of envelope[1]) {
+        if (envelopeItems[0].type == 'event' || envelopeItems[0].type == 'transaction') {
+          envelopeItems[1] = this._processLevels(envelopeItems[1] as Event);
+        }
```
marandaneto, line 139: Maybe we extract this to a method since it's repeating for Android and iOS.

src/js/tracing/nativeframes.ts
lucas-zimerman: yes, the unit is now required:
https://github.com/getsentry/sentry-javascript/blob/838dcfb8808b9b2a851b08dd4e942e48cbb40e50/packages/types/src/transaction.ts#L127

src/js/client.ts
lforst: Hi, also chiming in. This seems a bit fishy to me. Does this actually work if you set `options.transport` after the `super()` call? The base client uses that option within it's constructor. So only setting it afterwards might be too late...

As far as I can see, `_chooseTransport` is a pure function. This means we could very easily extract it out of this class (simply move it out of its body), and set `options.transport` before the `super()` call without accessing `this`. 

Would this work you think?

marandaneto (approved):

Issue #2317, opened 2022-06-27
Native frames always timeout after upgrade to 4.0.1


wbinarytree:
### Environment

How do you use Sentry?

self-hosted

Which SDK and version?

sentry-react-native 4.0.1

### Steps to Reproduce

Setup a Sentry with frame tracking like this:

```
      Sentry.init({
        dsn: dsn,
        debug: __DEV__,
        enableNativeCrashHandling: restricted,
        environment: env
        release: releaseName,
        dist: getBuildNumber(),
        integrations: [
            new Sentry.ReactNativeTracing({
            routingInstrumentation: routingInstrumentation,
            enableNativeFramesTracking: true,
            enableStallTracking: true,
          }),

```

### Expected Result

Frames are correctly reported. 

### Actual Result

Frames are not reported, always occurs error like:

```
[NativeFrames] Native frames timed out for OwnerApi: transaction 
```

Proposed solution. 

I dig into the code and see that actually the default integration will added regardless of user integration in this line:

https://github.com/getsentry/sentry-react-native/blob/main/src/js/sdk.tsx#L105

And eventually when we set up integrations here : 
https://github.com/getsentry/sentry-javascript/blob/master/packages/core/src/integration.ts#L61

The default `ReactNativeTracing` integration will override user setup `ReactNativeTracing`

marandaneto:
@wbinarytree thanks, I'll check this out.

wbinarytree:
@marandaneto Do you prefer I submit a PR for that?

marandaneto:
> Fix: SDK overwrites the user defined ReactNativeTracing #2319

Thanks, I've opened a PR already since I needed to investigate it.

wbinarytree:
@marandaneto Btw do we have a plan when we update the sentry cocoa to 7.8.1. It seems have a critical fix for frame tracking.

PR #2319, opened 2022-06-28
Fix: SDK overwrites the user defined ReactNativeTracing

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
SDK overwrites the user-defined ReactNativeTracing and because of that, slow and frozen frames are not reported.

Fixes https://github.com/getsentry/sentry-react-native/issues/2317
Bug introduced by https://github.com/getsentry/sentry-react-native/pull/2250 likely because `ReactNativeClientOptions` does not have a `defaultIntegrations` anymore.

Running and unit test

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Fix: SDK overwrites the user defined ReactNativeTracing ([#2313](https://github.com/getsentry/sentry-react-native/pull/2313))
+- Fix: SDK overwrites the user defined ReactNativeTracing ([#2319](https://github.com/getsentry/sentry-react-native/pull/2319))
 
 ## 4.0.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,6 @@ By bumping Sentry Javascript, new breaking changes were introduced, to know more
 ## 4.0.0-beta.4
 
 - Bump Sentry Cocoa 7.17.0 ([#2300](https://github.com/getsentry/sentry-react-native/pull/2300))
-
   - [changelog](https://github.com/getsentry/sentry-cocoa/blob/7.17.0/CHANGELOG.md)
   - [diff](https://github.com/getsentry/sentry-cocoa/compare/7.16.1...7.17.0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Fix: SDK overwrites the user defined ReactNativeTracing ([#2313](https://github.com/getsentry/sentry-react-native/pull/2313))
+
 ## 4.0.2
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -31,6 +35,7 @@ By bumping Sentry Javascript, new breaking changes were introduced, to know more
 ## 4.0.0-beta.4
 
 - Bump Sentry Cocoa 7.17.0 ([#2300](https://github.com/getsentry/sentry-react-native/pull/2300))
+
```
marandaneto, line 38:
```suggestion
```

src/js/sdk.tsx
```diff
@@ -50,59 +50,64 @@ export function init(passedOptions: ReactNativeOptions): void {
     stackParser: stackParserFromStackParserOptions(passedOptions.stackParser || defaultStackParser)
   };
 
+  function addIntegration(integration: Integration): void {
```
marandaneto, line 53: Checking each integration has runtime cost but the list of integrations should max just a few items.

marandaneto:
@Lms24 would you be able to give me some feedback?
I just want to be sure that I'm not doing the wrong thing.
So the merging and filterDuplicates have changed and now this is necessary.

src/js/sdk.tsx
bruno-garcia: We recently talked about this in dotnet. This could technically be a breaking change if one depends o n adding the same integration twice. E.g FileWatcher("path1") and another for path2.

Not a risk here?

marandaneto: No, because this is only for adding our own integrations, but actually Sentry JS SDK already deduplicates the integration list, that's under the RN SDK control tho.

test/sdk.test.ts
```diff
@@ -93,6 +97,20 @@ describe('Tests the SDK functionality', () => {
...
       });
+
+      it('Do not overwrite user defined integrations', () => {
+        const nav = new ReactNavigationInstrumentation();
+        const tracing = new ReactNativeTracing({ routingInstrumentation: nav });
+        init({
+          tracesSampleRate: 0.5,
+          enableAutoPerformanceTracking: true,
+          integrations: [tracing],
+        });
```
Lms24, line 108: Do you think it makes sense to test a similar scenario where `defaultIntegrations: [tracing]` is specified in the init options? Just to make sure that in this case, `tracing` isn't overridden either.

CHANGELOG.md
```diff
@@ -40,6 +41,7 @@ By bumping Sentry Javascript, new breaking changes were introduced, to know more
...
```
marandaneto, line 44:
```suggestion
```

test/sdk.test.ts
marandaneto: Yep, makes sense, Done.

PR #2328, opened 2022-06-29
Bump GH actions

marandaneto:
_#skip-changelog_


bruno-garcia (approved):
There's an automation for this. @philipphofmann added on cocoa. Would be nice to get that everywhere

marandaneto:
Yes, the downside is that it opens 20 PRs at once and you can't merge if the branch isn't up to date, so you have to "update branch" 1 by 1 and this takes more time than doing yourself, unfortunately.

philipphofmann:
Maybe you should disable the branch protection rule that the branch needs to be up to date. This can be crazy annoying.

Issue #2267, opened 2022-06-07
Autolinking broken in 0.69


oblador:
Due to https://github.com/react-native-community/cli/pull/1537 some things in your `react-native.config.js` have been deprecated and removed, thus failing autolinking schema validation (for example `ios.sharedLibraries`). 

### Steps to Reproduce

1. Create a vanilla RN project with latest 0.69 RC with `npx react-native init Example --version 0.69.0-rc.6`. 
2. Add this package.
3. Observe it not being autolinked.

oblador:
@marandaneto no it's unrelated, it's because of the linked PR.

renchap:
I can confirm this.

After upgrading to the latest RN 0.69-rc6 version and running `pod update`, I get:
```
Downloading dependencies
Installing DoubleConversion 1.1.6
Installing FBLazyVector 0.69.0-rc.6 (was 0.68.2)
Installing FBReactNativeSpec 0.69.0-rc.6 (was 0.68.2)
Installing Flipper-DoubleConversion 3.2.0.1 (was 3.2.0)
Installing Flipper-Glog 0.5.0.5 (was 0.5.0.4)
Installing RCTRequired 0.69.0-rc.6 (was 0.68.2)
Installing RCTTypeSafety 0.69.0-rc.6 (was 0.68.2)
Installing RNReanimated 2.8.0
Installing React 0.69.0-rc.6 (was 0.68.2)
Installing React-Codegen 0.69.0-rc.6 (was 0.68.2)
Installing React-Core 0.69.0-rc.6 (was 0.68.2)
Installing React-CoreModules 0.69.0-rc.6 (was 0.68.2)
Installing React-RCTActionSheet 0.69.0-rc.6 (was 0.68.2)
Installing React-RCTAnimation 0.69.0-rc.6 (was 0.68.2)
Installing React-RCTBlob 0.69.0-rc.6 (was 0.68.2)
Installing React-RCTImage 0.69.0-rc.6 (was 0.68.2)
Installing React-RCTLinking 0.69.0-rc.6 (was 0.68.2)
Installing React-RCTNetwork 0.69.0-rc.6 (was 0.68.2)
Installing React-RCTSettings 0.69.0-rc.6 (was 0.68.2)
Installing React-RCTText 0.69.0-rc.6 (was 0.68.2)
Installing React-RCTVibration 0.69.0-rc.6 (was 0.68.2)
Installing React-bridging (0.69.0-rc.6)
[...]

PR #2332, opened 2022-06-30
Fix: Auto linking for RN >= 0.69

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Remove the removed config from the config file.
This is a breaking change.

Fixes https://github.com/getsentry/sentry-react-native/issues/2267

Running it

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
Patch https://github.com/getsentry/sentry-wizard


marandaneto:
@bruno-garcia What if we say that the RN SDK 5.x requires RN version >= 0.69?
We could technically drop stuff like https://github.com/getsentry/sentry-react-native/pull/2015/
It'd be easier to maintain as well since we don't need to test stuff on older versions, we'd need to patch bug fixes on v4.x when needed though.
Ideally not the best solution for the user, but this is a breaking change and auto-linking won't work for either before or after 0.69.

bruno-garcia:
> 

Sounds like it's it's time. Can we spit out build warnings/errors in case the expected version isn't match? My biggest worry is of users getting unrelated error messages, being frustrated and/or raising issues with support. 

If we can break at build time, with:

Sentry 5.0 on older RN: `The Sentry SDK version 5.0.0 and above requires React Native 0.69 or higher`
Sentry 4.x on older RN: `The Sentry SDK version 4 doesn't support React Native 0.69. Please upgrade to Sentry 5.0.0 or higher`

marandaneto:
TODO: `yarn install` after sentry-wizard gets released to update the yarn.lock file.

yarn.lock
```diff
@@ -1255,12 +1257,12 @@
...
-  version "1.2.17"
-  resolved "https://registry.yarnpkg.com/@sentry/wizard/-/wizard-1.2.17.tgz#c3247b47129d002cfa45d7a2048d13dc6513457b"
-  integrity sha512-wXzjOZVDzh+1MhA9tpZXCKNTDusL2Dc298KGQkEkNefbW+OlYbsWwF7GAihLKUjYOSnKKZFzLIwbD+gxwxWxlw==
+"@sentry/wizard@2.0.0":
+  version "2.0.0"
+  resolved "https://registry.yarnpkg.com/@sentry/wizard/-/wizard-2.0.0.tgz#94bd4d0eeaae67b9577bd1fe2688b2be9d1f77e9"
+  integrity sha512-v6rOlJE0bOGSuekd+LTitofObIXKxtrQW3pXG2mwFVPQ6wezpFpUd8D/5hLJntLSvon/vFkQu43WvU+0VNBR3Q==
   dependencies:
-    "@sentry/cli" "^1.52.4"
+    "@sentry/cli" "^1.72.0"
```
bitsandfoxes, line 1265: is there a specific reason you stick to 1.72.0?

yarn.lock
```diff
@@ -1173,16 +1173,18 @@
     proxy-from-env "^1.1.0"
     which "^2.0.2"
 
-"@sentry/cli@^1.52.4":
-  version "1.57.0"
-  resolved "https://registry.yarnpkg.com/@sentry/cli/-/cli-1.57.0.tgz#87237ec6b77cb7582f018b580e9bb9481a809baa"
-  integrity sha512-lgfSQMRVcf00IkHkmqY6L9pv1uOEKDKbEjFQu2qstjg3kycswo1prz60xORJVpfr4qXDzqbSsAKEBp4VKG+ilw==
+"@sentry/cli@^1.72.0":
```
bitsandfoxes, line 1176: Are you sticking with `1.72.0` because of the wizard dependency?

bitsandfoxes (approved):

yarn.lock
marandaneto: Yeah, that's the min. version of the CLI for the newer wizard, but the `^` means anything above it, it's not 1.72 only.

SiSa68:
With this version, the source map is not uploaded automatically for my application (with react-native 69.4 and in debug mode)
Is this a known issue?

Acetyld:
Any idea on release date?

PR #2339, opened 2022-07-04
Fix yarn lock

marandaneto:
Fix https://github.com/getsentry/sentry-react-native/pull/2336/files

_#skip-changelog_


bitsandfoxes (approved):
🔥

PR #2355, opened 2022-07-07
chore: sample react-native updater

vaind:
resolves https://github.com/getsentry/team-mobile/issues/20#issuecomment-1162982754

Note: I've added `npx -y syncyarnlock --save --keepPrefix` to update `sample/package.json` - not tested in CI though - should I make a custom setup in my fork or do we merge this and fix afterwards if it doesn't work on `main`?

### DIFF
--- a/.github/workflows/update-deps.yml
+++ b/.github/workflows/update-deps.yml
@@ -57,3 +57,11 @@ jobs:
       name: CLI
     secrets:
       api_token: ${{ secrets.CI_DEPLOY_KEY }}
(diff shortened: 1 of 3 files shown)


vaind:
> @vaind I'd prefer that you test before because this would block new releases in case I have to fix something meanwhile.

tested in my repo, here's how the PR would look like: https://github.com/vaind/sentry-react-native/pull/1

marandaneto (approved):

PR #2389, opened 2022-07-22
v5.0.0 release

marandaneto:
PR for v5.0.0

_#skip-changelog_


krystofwoldrich:
I've tried RN 0.70 with the latest alpha release and everything seems to be working.

To my surprise activating the new TurboModules didn't break the build. I haven't find anywhere if you can mix NativeModules and TurboModules, but it would make sense that you can otherwise the transition to the new architecture would be too painful.

I also tried the new Fabric renderer and errors being reported as expected.

krystofwoldrich:
The perf ios test is failing because of the cocoa pods codesigning from Xcode 14
- https://github.com/facebook/react-native/issues/34673

krystofwoldrich:
Can I get a final review on this PR? 

It's a lot of changes and the individual changes have been reviewed already. We just need to skim it, just in case something slipped thru.

krystofwoldrich (approved):
LGTM 🚀 

But I'll wait for more people to skim thru this before merging. (All individual changes have been already reviewed.)

This will be merged without squash as we would lose too much information about the changes here.

.github/workflows/codegen.yml
```diff
@@ -0,0 +1,37 @@
...
+      - main
+      - release/**
+  pull_request:
+
+jobs:
+  codegen:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform: ["android"] # "ios" will be added after codegen is fixed
```
kahest, line 15: what is the context of this fix? is this blocked?

src/js/version.ts
```diff
@@ -1,3 +1,3 @@
 export const SDK_PACKAGE_NAME = 'npm:@sentry/react-native';
 export const SDK_NAME = 'sentry.javascript.react-native';
-export const SDK_VERSION = '4.15.0';
+export const SDK_VERSION = '5.0.0-rc.1';
```
kahest, line 3: Should be `5.0.0` now?

.github/workflows/codegen.yml
krystofwoldrich: Not blocked anymore, the fix was released in [RN 0.70.4](https://github.com/facebook/react-native/releases/tag/v0.70.4).

src/js/version.ts
krystofwoldrich: The release CI pipeline will update the version.

chrisedington:
Looking forward to this for 0.69 :)

Issue #2397, opened 2022-07-27
[BUG]: @sentry/react-native is ignoring `options.transport` function given in Sentry.init


zivl:
### Environment
Node version: 16

Which SDK and version?
`@sentry/react-native@4.1.2`

### Steps to Reproduce

```typescript
import * as Sentry from '@sentry/react-native';
import sentryTestkit from 'sentry-testkit';

it('should init sentry', () => {
  const { sentryTransport } = sentryTestkit();

  Sentry.init({
    dsn: 'https://dummy_dsn_with_valid_format@example.org/123',
    enableNative: false,
    transport: sentryTransport,
  });
});
```

### Expected Result

According to [Sentry docs](https://docs.sentry.io/platforms/react-native/configuration/options/#transport) + [type declaration here](https://github.com/getsentry/sentry-react-native/blob/c905df1f9cecfc490d72a23182286408c6d57aa3/src/js/options.ts#L108) the `transport` option should be supported to be able to set custom transport layer.

### Actual Result

The transport layer is set and then overriding the `options.transport` without prior checking of maybe a `transport` was given in `options` [(see code here)](https://github.com/getsentry/sentry-react-native/blob/c905df1f9cecfc490d72a23182286408c6d57aa3/src/js/client.ts#L35)

#### Reference
For transparency, this issue originated in `sentry-testkit`'s issue: https://github.com/wix/sentry-testkit/issues/113

PR #2398, opened 2022-07-27
should not ignore `options.transport` function provided in `Sentry.init(...)`

zivl:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
I've updated the ReactNativeClient constructor to first check if `options.transport` is provided and if so - then it takes it and calls to `super(options)`. If `options.transport` not provided then keep the previous logic and conduct react-native transport layer and then calls to `super(options)`.

this resolves issue #2397 

I've added a test which gets the given custom transport layer and then runs and validates the customized `send` and `flush` functions.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/test/client.test.ts
+++ b/test/client.test.ts
@@ -114,8 +114,9 @@ describe('Tests ReactNativeClient', () => {
         dsn: EXAMPLE_DSN,
         transport: myCustomTransportFn
       } as ReactNativeClientOptions);
-
+      // eslint-disable-next-line @typescript-eslint/unbound-method
       expect(client.getTransport()?.flush).toBe(myFlush);
+      // eslint-disable-next-line @typescript-eslint/unbound-method
       expect(client.getTransport()?.send).toBe(mySend);
     });
   });
--- a/test/client.test.ts
+++ b/test/client.test.ts
@@ -102,10 +102,12 @@ describe('Tests ReactNativeClient', () => {
     });
     
     test('use custom transport function', async () => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const mySend = (request: Envelope) => Promise.resolve();
+      const myFlush = (timeout?: number) => Promise.resolve(Boolean(timeout));
       const myCustomTransportFn = (): Transport => ({
-        // eslint-disable-next-line @typescript-eslint/no-unused-vars
-        send: (request: Envelope) => Promise.resolve(),
-        flush: (timeout?: number) => Promise.resolve(Boolean(timeout))
+        send: mySend,
+        flush: myFlush
       });
       const client = new ReactNativeClient({
         ...DEFAULT_OPTIONS,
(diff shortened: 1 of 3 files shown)


test/client.test.ts
```diff
@@ -99,6 +100,22 @@ describe('Tests ReactNativeClient', () => {
...
+        flush: (timeout?: number) => Promise.resolve(Boolean(timeout))
+      });
+      const client = new ReactNativeClient({
+        ...DEFAULT_OPTIONS,
+        dsn: EXAMPLE_DSN,
+        transport: myCustomTransportFn
+      } as ReactNativeClientOptions);
+
+      await expect(client.getTransport()?.flush(1)).resolves.toBe(true);
+      await expect(client.getTransport()?.send({} as Envelope)).resolves.toBeUndefined();
```
liamjones, line 117: If you pulled the Transport send/flush functions to variables you could have an even stronger check of: `expect(client.getTransport()?.flush).toBe(myFlushFunction)`

zivl: yeah I guess that's another option, shall I change it or it is not critical?

liamjones: 🤷 I defer to the Sentry maintainers on that one 😄

marandaneto: Yep I'd prefer the stronger check too.

marandaneto:
@zivl please add the changelog entry.
Thank you for fixing this.

test/client.test.ts
zivl: done

zivl:
@marandaneto change requests are done. please review again

marandaneto:
> @marandaneto change requests are done. please review again

Awesome.
Double-check the linter.

```
/home/runner/work/sentry-react-native/sentry-react-native/test/client.test.ts
  118:14  error  Avoid referencing unbound methods which may cause unintentional scoping of `this`  @typescript-eslint/unbound-method
  11[9](https://github.com/getsentry/sentry-react-native/runs/7539677797?check_suite_focus=true#step:5:10):14  error  Avoid referencing unbound methods which may cause unintentional scoping of `this`  @typescript-eslint/unbound-method
```

When fixing this, we are ready to LGTM.

zivl:
@marandaneto I can't get out of it without really put a `eslint-ignore...` comment as this is a dummy implementation for the sake of the test. it is also stated [here](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/unbound-method.md#when-not-to-use-it) that it is better to just add an ignore comment, can you go with that? or have any other solution?

marandaneto:
> @marandaneto I can't get out of it without really put a `eslint-ignore...` comment as this is a dummy implementation for the sake of the test. it is also stated [here](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/unbound-method.md?rgh-link-date=2022-07-27T14%3A41%3A58Z#when-not-to-use-it) that it is better to just add an ignore comment, can you go with that? or have any other solution?

We can ignore it, no problem.

marandaneto (approved):

zivl:
@marandaneto, @bruno-garcia can you pls run workflows and merge?

Issue #1327, opened 2021-02-02
Add Attachment support


bruno-garcia:
The native SDKs (sentry-android, sentry-cocoa) added support to Attachments.

The [spec is in the develop docs](https://develop.sentry.dev/sdk/features/#attachments).

Thumbs up if this is the feature you've been waiting for.

renchap:
In addition to this, it would be awesome to be able to run a callback when an error happen and generate attachments from this callback.

I have multiple states in the app (Redux, async storage…) and I would like to dump them in attachments when an issue occurs in the app to be able to understand the error if it comes from a bad state.
Currently I could add them to the event, but it can quickly go over the 200kb limit.

marandaneto:
This is blocked by https://github.com/getsentry/sentry-javascript/issues/4240

The RN SDK uses the JS SDK for building an event and the current version does not support the Envelope feature which is required for sending attachments.

To make it work we'd need to override many of the JS behavior and later replace them once the Envelope feature is landed on the linked PR natively on the JS SDK.

The best is to postpone and implement once #4240 is done, also considering attaching attachments to hard crashes on the Native layer.

Points to consider:

- Use the envelope feature from #4240
- Attach attachments on hard crashes, that requires changes on Sentry-Android, Sentry-Native, and Sentry-Cocoa since we don't sync down the attachments to the Native layers (e.g. KSCrash and Sentry-Native)
- We could technically support an array of bytes and file paths since RN does not support files natively, people most of the time use 3rd party plugins to do so.

mmerickel:
We got hit by lack of support for this feature and the addGlobalEventProcessor approach that's documented for the javascript SDK was not working for us. We'd see 201 Accepted requests for the attachment on the server but they did not get attached to the event. Ultimately we ended up writing a bridge module and using the sentry-cocoa and sentry-android native APIs to send events with attachments for the limited case where we needed an attachment. Would really love support for attachments to get added to sentry-react-native using the `scope.addAttachment` features available in the native sdks.

hdsenevi:
[...]

PR #2463, opened 2022-09-08
(feat) Add Attachments

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Add attachments support for RN. Supported attachments are string (will be send as is) and Uint8Array. File paths are not supported as RN doesn't have build-in FS. Support for certain libs used in RN can be added later if necessary.

- https://github.com/getsentry/sentry-react-native/issues/1327

Using sample app and unit tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


src/js/transports/TextEncoder.ts
```diff
@@ -0,0 +1,13 @@
+import { TextEncoderInternal } from '@sentry/types';
+import { Buffer } from 'buffer';
+
+export const makeUtf8TextEncoder = (): TextEncoderInternal => {
+  const textEncoder = {
+    encode: (text: string) => {
+      const bytes = new Uint8Array(Buffer.from(text, 'utf8'));
```
marandaneto, line 7: Let's try to avoid deps and vendor the class from `buffer` if possible.
https://develop.sentry.dev/sdk/philosophy/#dependencies-cost

src/js/wrapper.ts
```diff
@@ -115,46 +117,80 @@ export const NATIVE: SentryNativeWrapper = {
...
+            serializablePayload = JSON.parse(JSON.stringify(processedPayload));
+            break;
+          }
+          case 'attachment': {
+            serializablePayload = typeof itemPayload === 'string'
+              ? Buffer.from(itemPayload, 'utf8').toString('base64')
+              : undefined;
+            serializablePayload = itemPayload instanceof Uint8Array
+              ? Buffer.from(itemPayload.buffer).toString('base64')
+              : undefined;
```
marandaneto, line 176: I'd not use base64 but pass the array of bytes as it is.

ios/RNSentry.m
```diff
@@ -280,20 +280,28 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
...
         SentryEnvelopeHeader *envelopeHeader = [[SentryEnvelopeHeader alloc] initWithId:eventId sdkInfo:sdkInfo traceContext:traceContext];
 
-        NSError *error;
-        NSData *envelopeItemData = [NSJSONSerialization dataWithJSONObject:envelopeDict[@"payload"] options:0 error:&error];
-        if (nil != error) {
-            reject(@"SentryReactNative", @"Cannot serialize event", error);
-        }
-
-        NSString *itemType = envelopeDict[@"payload"][@"type"];
+        NSString *itemType = envelopeDict[@"item"][@"header"][@"type"];
```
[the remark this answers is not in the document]
krystofwoldrich: Sound good, at first I didn't want to change the native implementations in RN. But this is much cleaner.

src/js/vendor/buffer/utf8ToBytes.ts
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable */
+
+/**
+ * Convert a string to a byte array
+ *
+ * This is a utf8ToBytes function from the buffer module (with added types)
+ * https://github.com/feross/buffer/blob/795bbb5bda1b39f1370ebd784bea6107b087e3a7/index.js#L1956
+ *
+ * License: MIT (https://github.com/feross/buffer)
+ */
```
krystofwoldrich, line 10: @marandaneto Is the docs for the external lib okay like this?

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -302,24 +312,15 @@ public void captureEnvelope(String envelope, Promise promise) {
...
 
-    @ReactMethod
-    public void getStringBytesLength(String payload, Promise promise) {
-        try {
-            promise.resolve(payload.getBytes("UTF-8").length);
-        } catch (UnsupportedEncodingException e) {
-            promise.reject(e);
-        }
-    }
-
```
krystofwoldrich, line 322: I've removed this method since the raw bytes of the envelope are assembled in js, the bytes length have to be correct in the js.

ios/RNSentry.m
```diff
@@ -270,47 +270,34 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
...
+    }
 
-        #if DEBUG
+    #if DEBUG
+        [PrivateSentrySDKOnly captureEnvelope:envelope];
+    #else
+        if (options[@'store']) {
+            // Storing to disk happens asynchronously with captureEnvelope
+            [PrivateSentrySDKOnly storeEnvelope:envelope];
+        } else {
```
krystofwoldrich, line 296: I wasn't sure how to remove this if statement. It stores the envelope for fatal errors. Or is it okay for now?

android/src/main/java/io/sentry/react/RNSentryModule.java
marandaneto: You need to guarantee that this works tho, this was a bug multiple times in multiple places.

ios/RNSentry.m
marandaneto: for iOS, this is needed, iOS sends events async and if the App crashes right away, we only have time to write to the disk.

src/js/vendor/buffer/utf8ToBytes.ts
marandaneto: I believe so, please double check either with @bruno-garcia or the open source team, they have more experience with vendoring and licenses.

ios/RNSentry.m
krystofwoldrich: ok, thanks.

android/src/main/java/io/sentry/react/RNSentryModule.java
krystofwoldrich: Is anywhere issue, where are some steps how to reproduce the issue, so I can double check if it works now.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Add attachments support #2463
```
marandaneto, line 5:
```suggestion
### Features

- Add attachments support #2463
```

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -292,7 +293,16 @@ public void fetchNativeFrames(Promise promise) {
     }
 
     @ReactMethod
-    public void captureEnvelope(String envelope, Promise promise) {
+    public void captureEnvelope(ReadableArray rawBytes, ReadableMap options, Promise promise) {
+        if (!options.getBoolean("store")) {
+            logger.warning("Envelopes are always stored.");
+        }
```
marandaneto, line 299: If this is intentional, we don't need to log it.

android/src/main/java/io/sentry/react/RNSentryModule.java
marandaneto: You can use git blame and track when and why the code was added, https://github.com/getsentry/sentry-react-native/pull/773 has more context.

ios/RNSentry.m
```diff
@@ -270,47 +270,34 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
...
-        SentryEnvelopeItemHeader *envelopeItemHeader = [[SentryEnvelopeItemHeader alloc] initWithType:itemType length:envelopeItemData.length];
-        SentryEnvelopeItem *envelopeItem = [[SentryEnvelopeItem alloc] initWithHeader:envelopeItemHeader data:envelopeItemData];
+    NSMutableData *data = [[NSMutableData alloc] initWithCapacity: [bytes count]];
+    for(NSNumber *number in bytes) {
+        char byte = [number charValue];
+        [data appendBytes: &byte length: 1];
+    }
 
-        SentryEnvelope *envelope = [[SentryEnvelope alloc] initWithHeader:envelopeHeader singleItem:envelopeItem];
+    SentryEnvelope *envelope = [PrivateSentrySDKOnly envelopeWithData:data];
```
marandaneto, line 284: Can we validate that the envelope contains the `traceContext` when is a transaction? just need to be sure that we are not losing data during the deserialization.

src/js/vendor/buffer/utf8ToBytes.ts
bruno-garcia: Important to check the license: 
https://github.com/feross/buffer/blob/795bbb5bda1b39f1370ebd784bea6107b087e3a7/LICENSE

with MIT we're good. But you need to add the license to the header of the file. Or the whole file to a folder for example

src/js/wrapper.ts
```diff
@@ -87,77 +90,40 @@ export const NATIVE: SentryNativeWrapper = {
...
-
-        envelopeItemsBuilder += `\n${itemHeader}\n${itemPayload}`;
+      let bytesPayload: number[] = [];
+      if (typeof itemPayload === 'string') {
+        bytesPayload = utf8ToBytes(itemPayload);
+      } else if (itemPayload instanceof Uint8Array) {
+        bytesPayload = [...itemPayload];
+      } else {
+        bytesPayload = utf8ToBytes(JSON.stringify(itemPayload));
+        isFatalCrash = 'level' in itemPayload && itemPayload.level === 'fatal';
```
marandaneto, line 112: Let's create a new issue to improve this later, so we check the `mechanism.handled` instead, this logic could be extracted and also used for the Android breadcrumbs check, `wasExceptionHandled`.

test/wrapper.test.ts
```diff
@@ -342,15 +239,19 @@ describe('Tests Native Wrapper', () => {
...
       }
-      expect(RNSentry.getStringBytesLength).not.toBeCalled();
       expect(RNSentry.captureEnvelope).not.toBeCalled();
     });
+    test('Encloses message to an object on Android', async () => {
+      // TODO:
+    });
+    test('Do not introduce empty breadcrumbs if undefined before on Android', async () => {
+      // TODO:
+    });
```
marandaneto, line 249: Are we addressing this in this PR? otherwise I'd remove it.

test/wrapper.test.ts
```diff
@@ -342,15 +239,19 @@ describe('Tests Native Wrapper', () => {
...
     test('Clears breadcrumbs on Android if mechanism.handled is true', async () => {
       NATIVE.platform = 'android';
 
       const event: Event = {
         event_id: 'event0',
-        message: 'test',
```
marandaneto, line 353: Any reason to remove the `message`?

test/wrapper.test.ts
```diff
@@ -202,64 +191,22 @@ describe('Tests Native Wrapper', () => {
...
-          message: event.message,
-          instance: {
-            value: 0,
-          },
-        },
-      });
+      expect(RNSentry.captureEnvelope).toBeCalledWith(
+        utf8ToBytes('{"event_id":"event0","sent_at":"123"}\n'
+          + '{"type":"event","content_type":"application/json","length":87}\n'
+          + '{"event_id":"event0","message":"test","sdk":{"name":"test-sdk-name","version":"2.1.3"}}\n'
```
marandaneto, line 197: Can we assign the value to a string and do proper indentation to make the tests readable?

marandaneto:
@krystofwoldrich left a few comments, just a few things to address and ready to go, thanks for doing this.

android/src/main/java/io/sentry/react/RNSentryModule.java
krystofwoldrich: okay, I'll remove it. I just wanted it to be clear that it doesn't have any effect on android behaviour.

test/wrapper.test.ts
krystofwoldrich: Ohh, thanks, good catch, I definitely want to implement these.

test/wrapper.test.ts
krystofwoldrich: I was also thinking about that, but when assigned to template string we cant keep the indentation. So it also looks a bit strange.

test/wrapper.test.ts
krystofwoldrich: Yes, I want to separate test of the breadcrumbs behaviour and the message replacement (that's why I had the empty test prepared.). So the test are actually testing what they say they do.

src/js/vendor/buffer/utf8ToBytes.ts
krystofwoldrich: Okay, cool. Will do.

android/src/main/java/io/sentry/react/RNSentryModule.java
krystofwoldrich: Thanks, will do next time. 

I've added a test for the byte size of the payload. There should be no trouble since we assemble the raw bytes in JS now and then pass them to the native layers.

src/js/wrapper.ts
krystofwoldrich: - https://github.com/getsentry/sentry-react-native/issues/2470

ios/RNSentry.m
krystofwoldrich: Screenshot 2022-09-15 at 10 55 23

Looks good.

marandaneto (approved):
Well done @krystofwoldrich

PR #2472, opened 2022-09-15
Fix deps changelog

marandaneto:
_#skip-changelog_


krystofwoldrich (approved):
Was there a reason why the dep-bot add it to Features?

marandaneto:
> Was there a reason why the dep-bot add it to Features?

yes, that was the old behavior before upgrading the bot.

PR #2475, opened 2022-09-15
Fix flaky ios e2e test simulator application not found

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/scripts/get-xcode-version.sh
+++ b/scripts/get-xcode-version.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Matches the version number from the output of xcodebuild -version
+# Example: 13.4.1
+xcodeBin=/usr/bin/xcodebuild
+
+xcodeVersionOutput=$($xcodeBin -version)
+xcodeVersion=$(echo "$xcodeVersionOutput" | perl -pe '($_)=/([0-9]+([.][0-9]+)+)/')
+
+echo "$xcodeVersion"


krystofwoldrich:
I've run the branch with the fix 5 times and it didn't failed. Nevertheless I don't know if it's just a flake or the fix helped.

I though we could run some basic statistic to see if there will be any change after this PR.

This picture shows e2e runs over past two weeks.

Screenshot 2022-09-19 at 11 17 10

We can watch the success/failure ration over time and see if it changes. 

Note: the graph doesn't include outliers (all runs above 250 mins.)

marandaneto:
@krystofwoldrich the current state is that the iOS test fails, can we check what's failing?

krystofwoldrich:
@marandaneto thank I see, looks like I didn't find the expected element.

I will add wait for element instead of the random driver sleep.

krystofwoldrich:
also with the new xcode I could not run e2e on m1, i'm trying to figure that out
- https://github.com/appium/appium/issues/17497

krystofwoldrich:
Just for reference unity ios workflow
https://github.com/getsentry/sentry-unity/blob/main/.github/workflows/ci.yml

krystofwoldrich:
I think changing to v14 should be okay since v11 is not maintained anymore and libs are dropping support. 

https://github.com/nodejs/release#release-schedule

PR #2477, opened 2022-09-15
fix(sample) Run sample Xcode project on M1

krystofwoldrich:
Without these changes the sample project doesn't run on M1 macs.

@marandaneto Could you try on your Intel macbook if the project builds for the simulator?

#skip-changelog


marandaneto:
@krystofwoldrich let CI run and then we see :D

krystofwoldrich:
Either we are super unlucky and the test timeouted 3x in a row, or there is something wrong with the build even when it seems to succeed.

marandaneto:
> Either we are super unlucky and the test timeouted 3x in a row, or there is something wrong with the build even when it seems to succeed.

I don't have an M1 so I can't test it, the current `main` branch is green, so this change isn't needed for Intel processors.
Should we still merge this or not? if CI passes.

krystofwoldrich:
> > Either we are super unlucky and the test timeouted 3x in a row, or there is something wrong with the build even when it seems to succeed.
> 
> I don't have an M1 so I can't test it, the current `main` branch is green, so this change isn't needed for Intel processors. Should we still merge this or not? if CI passes.

If the ci passes I would merge it, otherwise anyone with the m1, who pulls the project will get failing build. Ideally the ci passes and we should try to build it on intel machine.

marandaneto (approved):

Issue #2470, opened 2022-09-14
Stop using `event.level == 'fatal'` to determinate if app's going to crash


krystofwoldrich:
### Description

As @marandaneto mentioned in https://github.com/getsentry/sentry-react-native/pull/2463/files#r970694168 we are using `event.level === 'fatal'` to determinate if app is going to crash. Based on this we are storing envelopes on iOS.

As user can alter `event.level` this is not 100 % reliable. We should use `mechanism.handled` to decide if envelope will be stored.

Once changed the same logic can be used to decide if `breadcrumbs` on Android will be removed.

PR #2480, opened 2022-09-15
fix: Determinate hard crash based on all exceptions like native sdks

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

Use the same logic as android sdk to determinate hard crash.

fix: https://github.com/getsentry/sentry-react-native/issues/2470

Unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


src/js/wrapper.ts
```diff
@@ -109,7 +104,7 @@ export const NATIVE: SentryNativeWrapper = {
         bytesPayload = [...itemPayload];
       } else {
         bytesPayload = utf8ToBytes(JSON.stringify(itemPayload));
-        isFatalCrash = 'level' in itemPayload && itemPayload.level === 'fatal';
+        hardCrashed = (isEvent(itemPayload) && isHardCrash(itemPayload)) || hardCrashed;
```
marandaneto, line 107: Why do we need the `isEvent` and the `|| hardCrashed`? I'm a bit confused.

krystofwoldrich: So I was thinking either:
- make `isHardCrash` accept all the payloads and internally for not Event payloads resolve `false`
- or make `isHardCrash` accept only Event because it's the only payload holding error info, but then I need to first check the type of the payload before passing it forward.

- one resolves on cleaner use at this line
- the other in cleaner function

krystofwoldrich: And the `|| hardCrashed` is because there can be multiple items in the envelope but only one is a  hard crash.

src/js/misc.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Extracts the hard crash information from the event exceptions.
+ * No exceptions or undefined handled are not hard crashes.
+ */
+export function isHardCrash(event: Event): boolean {
+  let isHandled = true;
+
+  for (const exception of event.exception?.values ?? []) {
+    if (!(exception.mechanism?.handled !== false)) {
+      isHandled = false;
```
marandaneto, line 21: You can return `false` early here.

src/js/misc.ts
```diff
@@ -0,0 +1,26 @@
+import {
+  Event,
+} from '@sentry/types';
+
+/**
+ * Recognizes an Event based on 'exception' property.
+ */
+export function isEvent(event: unknown): event is Event {
+  return typeof event === 'object' && event !== null && 'exception' in event;
+}
```
marandaneto, line 10: Some other parts of the code do `itemHeader.type == 'event` because an event from `captureMessage` does not have an `exception` but still is an event.

src/js/misc.ts
krystofwoldrich: Thanks, yes and safe time going thru the rest of the exceptions.

src/js/misc.ts
krystofwoldrich: Ohh, sorry I missed that. In that case I will make the function `isHardCrash` to accept all the payload items (and solve the existence of exception there) and delete the isEvent.

src/js/wrapper.ts
```diff
@@ -109,7 +104,7 @@ export const NATIVE: SentryNativeWrapper = {
         bytesPayload = [...itemPayload];
       } else {
         bytesPayload = utf8ToBytes(JSON.stringify(itemPayload));
-        isFatalCrash = 'level' in itemPayload && itemPayload.level === 'fatal';
+        hardCrashed = isHardCrash(itemPayload) || hardCrashed;
```
marandaneto, line 107: Instead of doing `|| hardCrashed` I'd rather not check the next events if `hardCrashed = true` so you spare the computation of the next items when at least 1 is `hardCrashed=true`.

src/js/wrapper.ts
marandaneto: See https://github.com/getsentry/sentry-react-native/pull/2480/files#r974163085

src/js/wrapper.ts
krystofwoldrich: So is it okay to discard the item after the hardCrash? 

In other word stop processing after first hard crash and send it to the native layer?

marandaneto (approved):
Thanks @krystofwoldrich

Issue #2471, opened 2022-09-14
Single time unit and postfix for `ReactNativeTracingOptions`


krystofwoldrich:
### Description

At the moment `ReactNativeTracingOptions` has multiple options to set some sort of time range. These options are in different time unit and they don't have post fix which would suggest the unit. 

```ts
@deprecated
idleTimeout: number;
@deprecated
maxTransactionDuration: number;

idleTimeoutMs: number;
maxTransactionDurationMs: number;
```

To not introduce breaking change, when the deprecated options are used they will set their new counterparts which will be used further in the code.

PR #2481, opened 2022-09-15
Add new react native tracing options based on JS sdk

krystofwoldrich:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

Fix: https://github.com/getsentry/sentry-react-native/issues/2471

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


src/js/tracing/reactnativetracing.ts
```diff
@@ -392,7 +416,7 @@ export class ReactNativeTracing implements Integration {
     idleTransaction.registerBeforeFinishCallback(
       (transaction, endTimestamp) => {
         adjustTransactionDuration(
-          maxTransactionDuration,
+          finalTimeoutMs,
```
krystofwoldrich, line 419: This function needs to be fixed for ms.

src/js/tracing/reactnativetracing.ts
```diff
@@ -132,6 +145,17 @@ export class ReactNativeTracing implements Integration {
     this.options = {
       ...defaultReactNativeTracingOptions,
       ...options,
+      finalTimeoutMs: options.finalTimeoutMs
+        // eslint-disable-next-line deprecation/deprecation
+        ?? (typeof options.maxTransactionDuration === 'number'
```
marandaneto, line 150: why do we need to check if its a `number`?

krystofwoldrich: because of the recalculation from s to ms.

krystofwoldrich:
Sorry for the force push, but I started this on the main branch and I didn't want to introduce other changes in 5 besides the options change.

src/js/tracing/reactnativetracing.ts
marandaneto: `maxTransactionDuration` always has a default value anyway, why would that be undefined?

krystofwoldrich: because here we are calculation with user input and that can be undefined.

Issue #2462, opened 2022-09-08
Integrations init function doesn't receive default integrations


krystofwoldrich:
**OS:**
- [x]  Windows
- [x]  MacOS
- [x]  Linux

**Platform:**
- [x]  iOS
- [x]  Android

**SDK version:** 4.3.0

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  dsn: 'xxx',
  beforeSend: (event, hint) => {
    console.log('beforeSend', event, hint);
    return event;
  },
  debug: true,
  environment: 'development',
  integrations: (integrations) => {
    return integrations.filter((integration) => {
      return integration.name !== 'Dedupe';
    });
  },
});
```

---
**I have following issue:**

The integrations received by the function are empty unless `defaultIntegrations` are supplied in the init.

**Steps to reproduce:**
- Create empty RN project
- Add Sentry
- Use the same init

**Actual result:**

`integrations` = `[]`

**Expected result:**

`integrations` = `[ all default installed integrations from the following list ]`

![Screenshot 2022-09-08 at 11 03 57](https://user-images.githubusercontent.com/31292499/189081961-376af197-0532-4a18-b58e-731ec52aaf52.png)

PR #2494, opened 2022-09-23
fix(init): Integrations factory gets default integrations

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
I mainly wrote tests to lock in the behaviour of `integrations` and `defaultIntegrations` attributes of the Sentry.init function.

Fix: https://github.com/getsentry/sentry-react-native/issues/2462

Sample app, unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/sdk.tsx
+++ b/src/js/sdk.tsx
@@ -1,7 +1,11 @@
 import { getIntegrationsToSetup, initAndBind, setExtra } from '@sentry/core';
 import { Hub, makeMain } from '@sentry/hub';
 import { RewriteFrames } from '@sentry/integrations';
-import { defaultStackParser, getCurrentHub } from '@sentry/react';
+import {
+  defaultIntegrations as reactDefaultIntegrations,
+  defaultStackParser,
+  getCurrentHub,
+} from '@sentry/react';
 import { Integration, StackFrame, UserFeedback } from '@sentry/types';
 import { getGlobalObject, logger, stackParserFromStackParserOptions } from '@sentry/utils';
 import * as React from 'react';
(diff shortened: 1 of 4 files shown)


src/js/sdk.tsx
```diff
@@ -54,40 +58,35 @@ export function init(passedOptions: ReactNativeOptions): void {
...
-    integrations: getIntegrationsToSetup(passedOptions),
+    integrations: [],
     stackParser: stackParserFromStackParserOptions(passedOptions.stackParser || defaultStackParser)
   };
 
-  function addIntegration(integration: Integration): void {
-    if (options.integrations.filter((i) => i.name == integration.name).length === 0) {
-      options.integrations.push(integration);
-    }
-  }
```
marandaneto, line 65: Adding this prevented adding duplicated integrations, I suspect this would introduce the bug again.
Here is the PR that added it, https://github.com/getsentry/sentry-react-native/pull/2319
Maybe try to reproduce the bug and see if still happens after this fix.

krystofwoldrich: Yes, I've seen it, but I think this function was only necessary because the `getIntegrationsToSetup` was called with wrong data too early.

There is this test for the bug and it passes.
https://github.com/getsentry/sentry-react-native/blob/15c1141e7e7fdce2874f0f9df6c5c0e7d4e67d32/test/sdk.test.ts#L112

marandaneto: All good then, nice finding.

marandaneto (approved):

PR #2496, opened 2022-09-26
feat(sdk): Add ClientReports

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

If clientReport exists it will be send with the next envelope.

Closes: https://github.com/getsentry/sentry-react-native/issues/2034

Unit tests, sample app, checked the client dropped in the graph

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Add ClientReports ([#2496](https://github.com/getsentry/sentry-react-native/pull/2496))
 
+### Sentry Self-hosted Compatibility
+
+- Starting with version `4.6.0` of the `@sentry/react-native` package, [Sentry's self hosted version >= v21.9.0](https://github.com/getsentry/self-hosted/releases) is required or you have to manually disable sending client reports via the `sendClientReports` option. This only applies to self-hosted Sentry. If you are using [sentry.io](https://sentry.io), no action is needed.
+
 ## 4.5.0
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add ClientReports ([#2496](https://github.com/getsentry/sentry-react-native/pull/2496))
+
 ## 4.5.0
 
 ### Features
--- a/src/js/client.ts
+++ b/src/js/client.ts
@@ -117,7 +117,6 @@ export class ReactNativeClient extends BaseClient<ReactNativeClientOptions> {
       },
     );
     this._sendEnvelope(envelope);
-    super._sendEnvelope(envelope);
   }
 
   /**
@@ -163,7 +162,7 @@ export class ReactNativeClient extends BaseClient<ReactNativeClientOptions> {
   /**
    *
    */
-  private _attachClientReportTo(envelope: ClientReportEnvelope): Envelope {
+  private _attachClientReportTo(envelope: ClientReportEnvelope): void {
     const outcomes = this._clearOutcomes();
 
     if (outcomes.length > 0) {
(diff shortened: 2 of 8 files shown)


src/js/sdk.tsx
```diff
@@ -41,6 +41,7 @@ const DEFAULT_OPTIONS: ReactNativeOptions = {
   transportOptions: {
     textEncoder: makeUtf8TextEncoder(),
   },
+  sendClientReports: true,
```
marandaneto, line 44: Is it disabled by default on JS?

krystofwoldrich: no

src/js/client.ts
```diff
@@ -144,4 +158,23 @@ export class ReactNativeClient extends BaseClient<ReactNativeClientOptions> {
...
+    if (outcomes.length > 0) {
+      const clientReportItem: ClientReportItem = [
+        { type: 'client_report' },
+        {
+          timestamp: dateTimestampInSeconds(),
+          discarded_events: outcomes,
+        },
+      ];
+
+      envelope[items].push(clientReportItem);
```
marandaneto, line 177: What if the envelope already has a `clientReportItem` item? is it 2? should we merge them?

krystofwoldrich: It can happen, but if the server can handle it, why should we? 

Maybe to save a bit of space? Would that be a reason?

marandaneto: It's cheaper to process on the server, has fewer data to transfer over the wire, and space in the disk, I recall folks from the ingest team also preferring a single client report per envelope whenever possible.

src/js/sdk.tsx
marandaneto: So then this line does nothing?

src/js/client.ts
krystofwoldrich: Okay that's what I thought.

src/js/sdk.tsx
krystofwoldrich: actually, we are not using the js default so it has to be there.
I got confused for a moment

src/js/client.ts
```diff
@@ -116,8 +126,38 @@ export class ReactNativeClient extends BaseClient<ReactNativeClientOptions> {
...
+
+    let shouldClearOutcomesBuffer = true;
+    if (this._transport && this._dsn) {
+      this._transport.send(envelope)
+        .then(null, reason => {
+          logger.error('Error while sending event:', reason);
+          if (reason instanceof SentryError) { // SentryError is thrown by SyncPromise
+            shouldClearOutcomesBuffer = false;
+            // If this is called asynchronously we want the _outcomesBuffer to be cleared
+          }
```
marandaneto, line 147: I'd have a logger call to the if and else branches.

marandaneto (approved):
Thanks @krystofwoldrich 
LGTM.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add ClientReports ([#2496](https://github.com/getsentry/sentry-react-native/pull/2496))
```
marandaneto, line 7: I'd also add a note about the breaking change with self hosted versions, See the one on Android changelog file.

### Sentry Self-hosted Compatibility

- Starting with version `6.0.0` of the `sentry` package, [Sentry's self hosted version >= v21.9.0](https://github.com/getsentry/self-hosted/releases) is required or you have to manually disable sending client reports via the `sendClientReports` option. This only applies to self-hosted Sentry. If you are using [sentry.io](https://sentry.io), no action is needed.

src/js/sdk.tsx
marandaneto, line 44: We have to sync this down to Android and iOS.

Issue #2499, opened 2022-09-27
beforeBreadcrumb, withScope callbacks do not have their Exceptions caught


krystofwoldrich:
this resolves in the app crashing

- https://github.com/getsentry/team-mobile/issues/21

JS sdk doesn't solve this because the website doesn't crash like RN app does

List of callback:
- beforeSend - safe from JS sdk
-  beforeBreadcrumb  - not safe - JS sdk won't cahnge it (it doesn't crash a page)
- withScope - not safe - JS sdk won't cahnge it (it doesn't crash a page)
-   onReady - not safe
- integrations - not safe
- initialScope - deprecated not safe 

What should we do with:
- transport - note safe -> create default?
- tracesSampler  - not safe -> default to 0?

marandaneto:
@krystofwoldrich I suspect that we've missed one of them:

configureScope:

```
                Sentry.configureScope((scope: Scope) => {
                  ...
                });
```

Or is it safe? [I guess not](https://github.com/getsentry/sentry-javascript/blob/f5cbc4c79c8757ebc2110aca4f4c01c2344a06b0/packages/hub/src/exports.ts#L67-L73).

krystofwoldrich:
@marandaneto Yes, we missed this one. I'll add it.

PR #2502, opened 2022-09-28
fix(callback): Callback given to Sentry are execute safely

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

These callback would crash the whole app if error is not caught.

I've thought about creating one safe() method that could wrap any callback, but for different callback the behaviour changes. Sometimes it makes sense to return undefined, sometimes defaults...

Closes: https://github.com/getsentry/sentry-react-native/issues/2499

Sample app.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Fixes
+- SDK Gracefully downgrades when callback throws an error ([#2502](https://github.com/getsentry/sentry-react-native/pull/2502))
```
marandaneto, line 6:
```suggestion
### Fixes
- SDK Gracefully downgrades when callback throws an error ([#2502](https://github.com/getsentry/sentry-react-native/pull/2502))

```

src/js/client.ts
```diff
@@ -167,12 +167,13 @@ export class ReactNativeClient extends BaseClient<ReactNativeClientOptions> {
...
-
-      this._options.onReady?.({ didCallNativeInit: false });
-
-      return;
+    } finally {
+      try {
+        this._options.onReady?.({ didCallNativeInit });
+      } catch (error) {
+        logger.error('The OnReady callback threw an error: ', error);
+      }
```
marandaneto, line 175: The logic here changed or am I missing something? at least `didCallNativeInit` isn't always `false` as before.

src/js/sdk.tsx
```diff
@@ -232,4 +236,28 @@ export async function close(): Promise<void> {
...
+ *
+ * This is essentially a convenience function for:
+ *
+ *     pushScope();
+ *     callback();
+ *     popScope();
+ *
+ * @param callback that will be enclosed into push/popScope.
+ */
+export function withScope(callback: (scope: Scope) => void): ReturnType<Hub['withScope']> {
```
marandaneto, line 254: We can add a test for this one.

src/js/client.ts
marandaneto: yeah, I guess its cleaner and has the same effect, happy with it.

marandaneto:
So the `safe` methods are unit tested but not the functions that use them, if you remove the `safeFactory` wrapper when setting the before breadcrumb, tests are still green but the behavior is broken, I'd suggest also adding a test for the specific callbacks, just that they don't throw.

marandaneto (approved):
Awesome @krystofwoldrich

Issue #1919, opened 2021-11-30
nested breadcrumb data on android is not treated correctly 


adbl:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
@sentry/react-native 3.2.3

### Steps to Reproduce

Call `Sentry.addBreadcrumb(...)` in key places during app execution with nested data on Android before some `reportError` happens. `normalizeDepth` is not configured and so should be the default: `3`?

### Expected Result

- To see the nested breadcrumb data displayed like it is done for iOS events, as a tree.
- To be able to use advanced data scrubbing such as `[Mask] [Anything] from [$breadcrumb.data.**.access-token]` to filter out sensitive values in the nested breadcrumb data, which is working for iOS events.

### Actual Result

- The nested breadcrumb data objects are displayed like stringified JSON, the UI doesn't show them as a tree.
- Advanced data scrub rules do not match the nested keys and thus sensitive values remain.

![image](https://user-images.githubusercontent.com/177880/144144516-d8130e73-d132-4859-8cf0-d9aedc8fc412.png)

The raw event indicates that the second level breadcrumb data objects are in fact serialized json strings:
```
         {
            "category" : "navigation",
            "timestamp" : 1637957388.671,
            "message" : "to: HealthQuestions-lINe0T4bK19QRqu6xATkl\nfrom: HealthQuestions-dwXLN-pRYJfyx-tk7OZb1",
            "data" : {
               "params" : "{\"analyticsParams\":{\"questionName\":\"WEIGHT\"},\"questionTag\":28}"
            },
            "type" : "navigation",
            "level" : "info"
         },
```
compared with iOS raw event data:
```
         {
            "data" : {
               "params" : {
                  "questionTag" : 28,
                  "analyticsParams" : {
                     "questionName" : "WEIGHT"
                  }
               }
            },
            "category" : "navigation",
            "level" : "info",
            "message" : "to: HealthQuestions-OAPrwTv68Z-GMpjTVcWe2\nfrom: HealthQuestions-eUubRHrlS0tumJjSwKDS4",
            "type" : "navigation",
            "timestamp" : 1638307295.225
         },
```

We have also found an odd example (our sentry issue is `2825368904` and event  is `3cd99275e3d645ea962d186440f3e648`) where the breadcrumbs have been duplicated and half of them are displayed correctly and half are not, which may be a clue (and probably related to https://github.com/getsentry/sentry-react-native/issues/1409):
```
[...]

PR #2519, opened 2022-10-05
Nested breadcrumb data on android was not treated correctly

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Reason is here https://github.com/getsentry/sentry-react-native/issues/1919#issuecomment-1268547391

Fixes https://github.com/getsentry/sentry-react-native/issues/1919

Running on Android and iOS

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/android/src/main/java/io/sentry/react/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/react/RNSentryModule.java
@@ -362,7 +362,7 @@ public class RNSentryModule extends ReactContextBaseJavaModule {
                         String key = it.nextKey();
                         String value = otherUserKeys.getString(key);
 
-                        // data is ConcurrentHashMap and can't have null values
+                        // other is ConcurrentHashMap and can't have null values
                         if (value != null) {
                             otherUserKeysMap.put(key, value);
                         }
--- a/android/src/main/java/io/sentry/react/RNSentryModule.java
+++ b/android/src/main/java/io/sentry/react/RNSentryModule.java
@@ -414,13 +414,9 @@ public class RNSentryModule extends ReactContextBaseJavaModule {
             }
 
             if (breadcrumb.hasKey("data")) {
-                ReadableMap data = breadcrumb.getMap("data");
-                ReadableMapKeySetIterator it = data.keySetIterator();
-                while (it.hasNextKey()) {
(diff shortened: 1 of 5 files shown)


android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -414,13 +417,13 @@ public void addBreadcrumb(final ReadableMap breadcrumb) {
...
-                while (it.hasNextKey()) {
-                    String key = it.nextKey();
-                    String value = data.getString(key);
-
-                    breadcrumbInstance.setData(key, value);
+                final ReadableMap data = breadcrumb.getMap("data");
+                for(final Map.Entry<String, Object> entry : data.toHashMap().entrySet()) {
+                    final Object value = entry.getValue();
+                    // data is ConcurrentHashMap and can't have null values
+                    if (value != null) {
```
krystofwoldrich, line 424: It looks good, I'm just wander, it would work the same if we would add the if value != null to the while look, right?

marandaneto: nope, `ReadableMapKeySetIterator` does not have a method that returns an `Object`, and previously it was always stringified so it was always `getString()` called.

krystofwoldrich: Ohh I see. Thanks.

krystofwoldrich (approved):
All good. :) 

I would maybe add a test, just to lock in the behavior, that the native methods are called with the correct normalized data object.

marandaneto:
> All good. :)
> 
> I would maybe add a test, just to lock in the behavior, that the native methods are called with the correct normalized data object.

Tests already exist and they are green because it's not a breaking change, what would you suggest?
https://github.com/getsentry/sentry-react-native/blob/e99d97badabb31ec095bfd33f9379cc8a52bae97/test/wrapper.test.ts#L420-L437
https://github.com/getsentry/sentry-react-native/blob/e99d97badabb31ec095bfd33f9379cc8a52bae97/test/wrapper.test.ts#L148-L150
Maybe adding more nested example?

PR #2566, opened 2022-10-24
rnn: make onComponentWillAppear protected

doronpr:
This is needed in order to extend this class and allow for event data enrichment

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
rnn: make onComponentWillAppear protected

This is needed in order to extend this class and allow for event data enrichment

Locally

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/tracing/reactnativenavigation.ts
+++ b/src/js/tracing/reactnativenavigation.ts
@@ -63,8 +63,8 @@ export interface NavigationDelegate {
  *
  * How this works:
  * - `_onCommand` is called every time a commands happens and sets an IdleTransaction on the scope without any route context.
- * - `onComponentWillAppear` is then called AFTER the state change happens due to a dispatch and sets the route context onto the active transaction.
- * - If `onComponentWillAppear` isn't called within `options.routeChangeTimeoutMs` of the dispatch, then the transaction is not sampled and finished.
+ * - `_onComponentWillAppear` is then called AFTER the state change happens due to a dispatch and sets the route context onto the active transaction.
+ * - If `_onComponentWillAppear` isn't called within `options.routeChangeTimeoutMs` of the dispatch, then the transaction is not sampled and finished.
  */
 export class ReactNativeNavigationInstrumentation extends InternalRoutingInstrumentation {
   public static instrumentationName: string = 'react-native-navigation';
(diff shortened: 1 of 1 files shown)


krystofwoldrich:
@doronpr Thank you for the PR, but before we proceed could you add an example of the data enrichment that this is needed for? Something that is not possible with the current [`beforeNavigate`](https://github.com/getsentry/sentry-react-native/blob/a516e6369e85ebcebea7677f611dabe52cde04cf/src/js/tracing/reactnativenavigation.ts#L202) hook that is called in the `_onComponentWillAppear` method?

doronpr:
> @doronpr Thank you for the PR, but before we proceed could you add an example of the data enrichment that this is needed for? Something that is not possible with the current [`beforeNavigate`](https://github.com/getsentry/sentry-react-native/blob/a516e6369e85ebcebea7677f611dabe52cde04cf/src/js/tracing/reactnativenavigation.ts#L202) hook that is called in the `_onComponentWillAppear` method?
> 
> Also, there are some lint issues that would need to be fixed.

lint fixed :)


> @doronpr Thank you for the PR, but before we proceed could you add an example of the data enrichment that this is needed for? Something that is not possible with the current [`beforeNavigate`](https://github.com/getsentry/sentry-react-native/blob/a516e6369e85ebcebea7677f611dabe52cde04cf/src/js/tracing/reactnativenavigation.ts#L202) hook that is called in the `_onComponentWillAppear` method?
> 
> Also, there are some lint issues that would need to be fixed.

Hey @krystofwoldrich, the problem with `beforeNavigate` is that it's being passed into `registerRoutingInstrumentation` which is called [internally](https://github.com/getsentry/sentry-react-native/blob/83be16fd96630c3e567aca2bb773d04ad5b08824/src/js/tracing/reactnativetracing.ts#L188) by Sentry so not sure how I can override it. But I just realized I can override `registerRoutingInstrumentation` and pass an enriched `beforeNavigate`. So closing the PR, thanks!

Just to answer your Q - we identified some issues when a cyclic object is passed in the props.
Not sure if these issues are related to Sentry but removing these props seem to help.

Issue #2131, opened 2022-03-10
`captureMessage` on Android using `attachStacktrace=true` does not send the stacktrace


marandaneto:
The JS SDK appends the stack trace to a [top-level property](https://develop.sentry.dev/sdk/event-payloads/stacktrace/) instead of the `threads` interface.
It works on iOS because the event is sent as it is.
It fails on Android because of the round trip serialization, since this property does not exist on Android, it's lost.

* Android V6 keeps the unknown fields, which would solve the problem (Still alpha tho).
* The JS SDK could adapt it to use the `threads` interface (Gonna be done on V7, See the [issue](https://github.com/getsentry/sentry-javascript/issues/2605)).
* The RN SDK could process the event and move the top-level property within the `threads` interface.

The 3rd option right now seems faster/easier.

marandaneto:
The JS SDK already did the changes on [V7](https://github.com/getsentry/sentry-go/blob/c40f30a89cdda4cb59239f2f77466d3d4219de61/client.go#L434-L452)
[The issue](https://github.com/getsentry/sentry-javascript/issues/2605) is still open tho

marandaneto:
We've decided to go either with upgrading the Android SDK to v6 or JS SDK to v7, the one that comes first.

marandaneto:
Blocked by https://github.com/getsentry/sentry-javascript/issues/5879

Issue #2461, opened 2022-09-08
`attachStacktrace` enabled by default on the JS layer


marandaneto:
### Description

Android and iOS have it enabled by default.
Its disabled by default on JS and RN inherits this option from JS when events are captured in the JS layer.
Double-check if this [issue](https://github.com/getsentry/sentry-react-native/issues/2131) is fixed as well.
Amend the docs https://github.com/getsentry/sentry-docs/pull/2815/files

krystofwoldrich:
https://github.com/getsentry/sentry-react-native/issues/2131 is fixed, but the stack trace is sent in the `exception` with the message. Correctly it should be in the `threads`

marandaneto:
@krystofwoldrich can we somehow monkeypatch the [fix](https://github.com/getsentry/sentry-javascript/pull/6038) directly into RN?
Looking at the [v8 milestone on JS](https://github.com/getsentry/sentry-javascript/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+milestone%3A8.0.0), not sure about the timeline, might be a bit of stuff and this is going to be after our v5, so we'd need to do v6 to add this breaking change.

krystofwoldrich:
For sure we can fix in in RN and remove the fix after upgrading to v8.

@marandaneto When we release RN v5, won't we have to release RN v6 anyway because of changes in JS .v8

marandaneto:
mmm, your call then @krystofwoldrich

krystofwoldrich:
Patching in would be quite complicated since it's not only about patching `eventFromMessage` but also the `Event` interface is missing the `threads` property.

Or maybe we could add `threads` to `Event` in JS SDK v7. 
- https://github.com/getsentry/sentry-javascript/pull/6038#r1047070949

marandaneto:
Adding threads would not be breaking, just extending the protocol, good idea.

PR #2577, opened 2022-10-27
fix(event): Message event contains stacktrace if `attachStacktrace` o…

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Theoretically only passing the option to the browser client should solve the problem. But for some unexplainable reason, the frames are always an empty array, if the synthetic error is not passed from the RN JS code.

```js
function eventFromString(
  stackParser,
  input,
  syntheticException,
  attachStacktrace,
) {
  const event = {
    message: input,
  };
  let frames;
  if (attachStacktrace && syntheticException) {
    frames = parseStackFrames(stackParser, syntheticException); // 29 frames parsed
    if (frames.length) {
      console.log('frames:', frames, frames.length); //frames: Array(29) 29 Array(29)
      event.exception = {
        values: [{ value: input, stacktrace: { frames } }],
      };
    }
  }

  console.log('event with stack', event, frames);//event with stack Object Array(29)
  //but event.exception.values[0].stacktrace.frames is Array(0)
}
```
Code from JS SDK.
https://github.com/getsentry/sentry-javascript/blob/e1d43ebb34086605aefb9243fce03fc97425b4be/packages/browser/src/eventbuilder.ts#L276

Fixes: https://github.com/getsentry/sentry-react-native/issues/2131

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/src/js/sdk.tsx
+++ b/src/js/sdk.tsx
@@ -284,8 +284,8 @@ export function withScope(callback: (scope: Scope) => void): ReturnType<Hub['wit
  * @param Severity Define the level of the message.
  * @returns The generated eventId.
  */
-export function captureMessage(message: string, level?: SeverityLevel): void {
-  getCurrentHub().getClient<ReactNativeClient>()?.captureMessage(
+export function captureMessage(message: string, level?: SeverityLevel): string | undefined {
+  return getCurrentHub().getClient<ReactNativeClient>()?.captureMessage(
     message,
     level,
     {
--- a/src/js/sdk.tsx
+++ b/src/js/sdk.tsx
@@ -5,7 +5,7 @@ import {
   defaultStackParser,
   getCurrentHub,
 } from '@sentry/react';
-import { Integration, Scope, SeverityLevel, StackFrame, UserFeedback } from '@sentry/types';
+import { Integration, Scope, StackFrame, UserFeedback } from '@sentry/types';
 import { logger, stackParserFromStackParserOptions } from '@sentry/utils';
 import * as React from 'react';
 
@@ -275,21 +275,4 @@ export function withScope(callback: (scope: Scope) => void): ReturnType<Hub['wit
     }
   };
   getCurrentHub().configureScope(safeCallback);
-}
-
-/**
- * Captures a message event and sends it to Sentry.
- *
- * @param message The message to send to Sentry.
- * @param Severity Define the level of the message.
- * @returns The generated eventId.
- */
-export function captureMessage(message: string, level?: SeverityLevel): string | undefined {
-  return getCurrentHub().getClient<ReactNativeClient>()?.captureMessage(
-    message,
-    level,
-    {
-      syntheticException: new Error(message),
-    },
-  );
 }
--- a/src/js/sdk.tsx
+++ b/src/js/sdk.tsx
@@ -5,7 +5,7 @@ import {
   defaultStackParser,
   getCurrentHub,
 } from '@sentry/react';
-import { Integration, Scope, StackFrame, UserFeedback } from '@sentry/types';
+import { Integration, Scope, SeverityLevel, StackFrame, UserFeedback } from '@sentry/types';
 import { logger, stackParserFromStackParserOptions } from '@sentry/utils';
 import * as React from 'react';
 
@@ -275,4 +275,21 @@ export function withScope(callback: (scope: Scope) => void): ReturnType<Hub['wit
     }
   };
   getCurrentHub().configureScope(safeCallback);
+}
+
+/**
+ * Captures a message event and sends it to Sentry.
+ *
+ * @param message The message to send to Sentry.
+ * @param Severity Define the level of the message.
+ * @returns The generated eventId.
+ */
+export function captureMessage(message: string, level?: SeverityLevel): void {
+  getCurrentHub().getClient<ReactNativeClient>()?.captureMessage(
+    message,
+    level,
+    {
+      syntheticException: new Error(message),
+    },
+  );
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## Fixes
+
+- Message event can have attached stacktrace ([#2577](https://github.com/getsentry/sentry-react-native/pull/2577))
+
 ### Dependencies
 
 - Bump Android SDK from v6.5.0 to v6.6.0 ([#2572](https://github.com/getsentry/sentry-react-native/pull/2572))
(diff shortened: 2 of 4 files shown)


src/js/sdk.tsx
```diff
@@ -276,3 +276,20 @@ export function withScope(callback: (scope: Scope) => void): ReturnType<Hub['wit
...
+
+/**
+ * Captures a message event and sends it to Sentry.
+ *
+ * @param message The message to send to Sentry.
+ * @param Severity Define the level of the message.
+ * @returns The generated eventId.
+ */
+export function captureMessage(message: string, level?: SeverityLevel): string | undefined {
+  return getCurrentHub().getClient<ReactNativeClient>()?.captureMessage(
```
krystofwoldrich, line 288: When I make this change the `frames` start to disappear and I don't know why.

```suggestion
  return getCurrentHub().captureMessage(
```

AbhiPrasad: is it because the synthetic exception gets re-written?

krystofwoldrich: The `synthetic exception` gets all the way to the [`eventFromString`](https://github.com/getsentry/sentry-javascript/blob/e1d43ebb34086605aefb9243fce03fc97425b4be/packages/browser/src/eventbuilder.ts#L276), but there the `frames` are missing in the `stacktrace` even tho that are assigned to it.

marandaneto:
@krystofwoldrich I lack context here, didn't agree on fixing this on JS? https://github.com/getsentry/sentry-javascript/pull/6038

krystofwoldrich:
@marandaneto 

Yes, we will fix the JS part, but besides that, there was an option missing in [RN](https://github.com/getsentry/sentry-react-native/pull/2577/files#diff-7fa63c87605654a66f1a1b0a5463c2b05bc0591ebda5f2e6ed77ea3fbb809674R75).

But after adding [this option](https://github.com/getsentry/sentry-react-native/pull/2577/files#diff-7fa63c87605654a66f1a1b0a5463c2b05bc0591ebda5f2e6ed77ea3fbb809674R75) the stack trace is still missing when `captureMessage` is called thru the hub. The example code from the description shows what goes wrong. But I don't know why.

marandaneto:
> @marandaneto
> 
> Yes, we will fix the JS part, but besides that, there was an option missing in [RN](https://github.com/getsentry/sentry-react-native/pull/2577/files#diff-7fa63c87605654a66f1a1b0a5463c2b05bc0591ebda5f2e6ed77ea3fbb809674R75).
> 
> But after adding [this option](https://github.com/getsentry/sentry-react-native/pull/2577/files#diff-7fa63c87605654a66f1a1b0a5463c2b05bc0591ebda5f2e6ed77ea3fbb809674R75) the stack trace is still missing when `captureMessage` is called thru the hub. The example code from the description shows what goes wrong. But I don't know why.

Gotcha, I'd expect this change to be part of https://github.com/getsentry/sentry-react-native/issues/2461 and not https://github.com/getsentry/sentry-react-native/issues/2131
Why do we have its own `captureMessage`? isn't already defined in the JS SDK?

src/js/sdk.tsx
marandaneto: IDK if its the case, but in some languages/platforms, when you do `Error(message)`, the stack trace isn't available.
You'd need to `throw Error(message)` and only then access the stack trace from the thrown exception.

krystofwoldrich: The stack is present in any case it just disappears later in the [`eventFromString`](https://github.com/getsentry/sentry-javascript/blob/e1d43ebb34086605aefb9243fce03fc97425b4be/packages/browser/src/eventbuilder.ts#L276)

AbhiPrasad:
This is the weirdest bug I've seen in a loooong time.

As discussed with @krystofwoldrich we probably want to address the bug like so:
1) merge this PR (or if we split this up, it's parts) to unblock users and get it working
2) validate that this is not happening for regular browser JS
  - if it is happening, we need to investigate further on the JS team
3) create a minimal reproduction for react native that just shows the broken functionality and file a ticket with React Native directly to see what is happening

marandaneto:
> This is the weirdest bug I've seen in a loooong time.
> 
> As discussed with @krystofwoldrich we probably want to address the bug like so:
> 
> 1. merge this PR (or if we split this up, it's parts) to unblock users and get it working
> 2. validate that this is not happening for regular browser JS
> 
> * if it is happening, we need to investigate further on the JS team
> 
> 3. create a minimal reproduction for react native that just shows the broken functionality and file a ticket with React Native directly to see what is happening

Ok, that makes sense, I just lacked the context around this and the ingestion.

src/js/sdk.tsx
marandaneto: @krystofwoldrich can you check if this has something to do with `DebugSymbolicator`?
Does it happen in release and debug mode? or only one of them?

krystofwoldrich: @marandaneto Wow, thank you, DebugSymbolicator is causing it. I haven't looked at why, yet.

krystofwoldrich:
Thank you @AbhiPrasad and @marandaneto 

So the bug is caused by [`DebugSymbolicator`](https://github.com/getsentry/sentry-react-native/blob/f916f2b3afbb9884f9037b8486539ca6cc8b327c/src/js/integrations/debugsymbolicator.ts#L34-L35)

So I think this can be reviewed and merged since in production the stack frames will be sent correctly.

marandaneto (approved):

marandaneto:
> Thank you @AbhiPrasad and @marandaneto
> 
> So the bug is caused by [`DebugSymbolicator`](https://github.com/getsentry/sentry-react-native/blob/f916f2b3afbb9884f9037b8486539ca6cc8b327c/src/js/integrations/debugsymbolicator.ts#L34-L35)
> 
> So I think this can be reviewed and merged since in production the stack frames will be sent correctly.
> 
> For #2461 I will create a new PR that will merge to `5.0.0` branch.

This still depends on the JS SDK bump, after the other PR, right?

krystofwoldrich:
> > Thank you @AbhiPrasad and @marandaneto
> > So the bug is caused by [`DebugSymbolicator`](https://github.com/getsentry/sentry-react-native/blob/f916f2b3afbb9884f9037b8486539ca6cc8b327c/src/js/integrations/debugsymbolicator.ts#L34-L35)
> > So I think this can be reviewed and merged since in production the stack frames will be sent correctly.
> > For #2461 I will create a new PR that will merge to `5.0.0` branch.
> 
> This still depends on the JS SDK bump, after the other PR, right?

The JS SDK bump will move the stack traces from exceptions to threads, but from `sentry.io` user perspective it will start working right away after merging this.

Issue #2497, opened 2022-09-26
BufferSize TransportOptions doesn't work


krystofwoldrich:
NativeTransport uses hardcoded 30 items in promise buffer 

https://github.com/getsentry/sentry-react-native/blob/1e21438517050133d603247baeff05a464c76407/src/js/transports/native.ts#L11

Cocoa SDK uses different name for the bufferSize

https://github.com/getsentry/sentry-cocoa/blob/172c95a0f5823a6a1c37cd1b2d348ed047ba67b4/Sources/Sentry/SentryOptions.m#L47

The naming across the sdks should mathch.

krystofwoldrich:
@marandaneto And maxQueueSize is option of the sdk or transport?

philipphofmann:
It's a public option on [Java](https://github.com/getsentry/sentry-java/blob/24d5c4f3c7acddcc1075f105a221488abf657f0d/sentry/src/main/java/io/sentry/SentryOptions.java#L1154-L1163) and most other SDKs.

krystofwoldrich:
Okay, if I'm looking correct Transport gets the same SentryOptions.

In JS there is different interface only for TransportOptions, where `bufferSize` is defined.

krystofwoldrich:
Deprecate the current unused option and used a new one named similar to the native mobile skds. Pass the options to java and ios. Keep default 30.

PR #2578, opened 2022-10-27
feat(transport): Add buffer size option to native transport

krystofwoldrich:

- [x] Bugfix
- [x] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description
iOS Doesn't have an option for maxQueueSize, therefore the option is only passed to Java and JS layers.

Fixes: https://github.com/getsentry/sentry-react-native/issues/2497

Sample app.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Features
+- Add buffer size option to native transport ([#2578](https://github.com/getsentry/sentry-react-native/pull/2578))
```
marandaneto, line 6: H: Always jump a line :)

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -130,6 +130,12 @@ public void initNativeSdk(final ReadableMap rnOptions, Promise promise) {
             if (rnOptions.hasKey("sendDefaultPii")) {
                 options.setSendDefaultPii(rnOptions.getBoolean("sendDefaultPii"));
             }
+            if (rnOptions.hasKey("transportOptions")) {
+                final ReadableMap transportOptions = rnOptions.getMap("transportOptions");
+                if (transportOptions.hasKey("bufferSize")) {
+                    options.setMaxQueueSize(rnOptions.getInt("bufferSize"));
+                }
+            }
```
marandaneto, line 138: H: This should be called `maxQueueSize` and it should be a high-level property, not under `transportOptions`.

src/js/sdk.tsx
```diff
@@ -40,6 +40,7 @@ const DEFAULT_OPTIONS: ReactNativeOptions = {
   enableOutOfMemoryTracking: true,
   patchGlobalPromise: true,
   transportOptions: {
+    bufferSize: DEFAULT_BUFFER_SIZE,
```
marandaneto, line 43: It's ok to call it `bufferSize` here since it inherits from JS.
But for RN Apps, lets stick to `maxQueueSize` high level property, so one only does `options.maxQueueSize = X`.

marandaneto: Relates to https://github.com/getsentry/sentry-react-native/pull/2578/files#r1007056274

krystofwoldrich: I can add `maxQueueSize` but then we will be advertising two places where people can set the `bufferSize`/`maxQueueSize`, because as you said we inherit it from JS. That could be super confusing. Especially given only one option will work.

I would say if we want to match the mobile SDKs that is fine, but then we should break the inheritance of the options from the JS SDK so we can remove `bufferSize` from the `transportOptions`.

CHANGELOG.md
```diff
@@ -10,13 +14,6 @@
 
 ## 4.8.0
 
-### Various fixes & improvements
-
-- fix(attachments): Maximum call stack exceeded error for large attachments (#2579) by @old
-- fix(event): Message event contains stacktrace if `attachStacktrace` o… (#2577) by @krystofwoldrich
-- chore: update scripts/update-cocoa.sh to 7.29.0 (#2571) by @github-actions
-- chore: update scripts/update-android.sh to 6.6.0 (#2572) by @github-actions
```
marandaneto, line 18: Why is this removed?

src/js/options.ts
```diff
@@ -121,18 +121,30 @@ export interface BaseReactNativeOptions {
...
+
+  /**
+   * The max queue size for capping the number of envelopes waiting to be sent by Transport.
+   */
+  maxQueueSize?: number;
+}
+
+export interface ReactNativeTransportOptions extends BrowserTransportOptions {
+  /**
+   * @deprecated use `maxQueueSize` in te root of the SDK options.
```
marandaneto, line 133:
```suggestion
   * @deprecated use `maxQueueSize` in the root of the SDK options.
```

src/js/options.ts
```diff
@@ -121,18 +121,30 @@ export interface BaseReactNativeOptions {
...
+   *
+   * Setting this option does not have any effect.
+  */
```
marandaneto, line 136: H: Not sure if this comment makes sense because the code does:

> const maxQueueSize = passedOptions.maxQueueSize
>     // eslint-disable-next-line deprecation/deprecation
>     ?? passedOptions.transportOptions?.bufferSize
>     ?? DEFAULT_OPTIONS.maxQueueSize;

So it does have an effect, right?

src/js/options.ts
```diff
@@ -121,17 +121,29 @@ export interface BaseReactNativeOptions {
...
+   * The max queue size for capping the number of envelopes waiting to be sent by Transport.
+   */
+  maxQueueSize?: number;
+}
+
+export interface ReactNativeTransportOptions extends BrowserTransportOptions {
+  /**
+   * @deprecated use `maxQueueSize` in the root of the SDK options.
+  */
+  bufferSize?: number;
```
marandaneto, line 135: By doing this, if I do:
```
Sentry.init({
	transport(transportOptions) {
	    transportOptions.bufferSize = 100;
	    // ...
	  },
}
```

Do I see in the auto-complete that `bufferSize` is deprecated? 
Should the linter work?
it's not working for me at least.

krystofwoldrich: I think it is some ts and lint caching problem, it works for me after reloading my VSC window.

Screenshot 2022-11-04 at 14 28 05

Try that I hope it will help.

marandaneto: No need to block this, just wanted to be sure that it works.

PR #2579, opened 2022-10-27
fix: maximum call stack exceeded error

old:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fixed `Sentry Logger [error]: Error while sending event: RangeError: Maximum call stack size exceeded` error that occurs when adding large attachments to sentry capture

fixes `Sentry Logger [error]: Error while sending event: RangeError: Maximum call stack size exceeded` error which happens when the sentry payload bytes exceeds the number of arguments the `Array.prototype.push` method can support

it is recommended to only use the spread operator for smaller arrays. So when adding a large attachment such as a json file that contains a lot of data this will result in the `maximum call stack size exceeded error` which drops the sentry capture 

> The spread syntax fails to work with larger Arrays. When testing with 100,000 elements, the spread syntax generated a RangeError with the message: Maximum call stack size exceeded. I manually adjusted the number in the tool to identify the actual number of elements that caused this error. I found that number of elements to be 63,653. It’s unclear to me at this time why this number of elements causes this error. I don’t know if it’s my specific environment, specific code, or something else. I believe it’s beyond the scope of this article though

sample app and in my project that uses this repo

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
   - [changelog](https://github.com/getsentry/sentry-cocoa/blob/master/CHANGELOG.md#7290)
   - [diff](https://github.com/getsentry/sentry-cocoa/compare/7.28.0...7.29.0)
 
+### Fixes
+
+- Fixed maximum call stack exceeded error resulting from large payloads ([#2579](https://github.com/getsentry/sentry-react-native/pull/2579))
+
 ## 4.7.1
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## Fixes
 
 - Message event can have attached stacktrace ([#2577](https://github.com/getsentry/sentry-react-native/pull/2577))
+- Fixed maximum call stack exceeded error resulting from large payloads ([#2579](https://github.com/getsentry/sentry-react-native/pull/2579))
 
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


krystofwoldrich:
Hi, thank you for the fix. 
I've read the article https://javascript.plainenglish.io/efficiently-merging-arrays-in-javascript-32993788a8b2

Should we maybe consider `concat` as it offers the best performance or combination?
```js
envelopeBytes = envelopeBytes.concat(bytesPayload);
```

Could you also add a changelog?

old:
> Hi, thank you for the fix. I've read the article https://javascript.plainenglish.io/efficiently-merging-arrays-in-javascript-32993788a8b2
> 
> Should we maybe consider `concat` as it offers the best performance or combination?
> 
> ```js
> envelopeBytes = envelopeBytes.concat(bytesPayload);
> ```
> 
> Could you also add a changelog?

I originally tried .concat but it resulted in this error `entry Logger [error]: Error while sending event: TypeError: "envelopeBytes" is read-only`

krystofwoldrich:
> > Hi, thank you for the fix. I've read the article https://javascript.plainenglish.io/efficiently-merging-arrays-in-javascript-32993788a8b2
> > Should we maybe consider `concat` as it offers the best performance or combination?
> > ```js
> > envelopeBytes = envelopeBytes.concat(bytesPayload);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Could you also add a changelog?
> 
> I originally tried .concat but it resulted in this error `entry Logger [error]: Error while sending event: TypeError: "envelopeBytes" is read-only`. I don't think performance here will be an issue.

Yes that make sense, if we would use concat we would need to change the `envelopeBytes` to `let`.

marandaneto:
@krystofwoldrich this seems like something we have to merge and cut a fix asap.

krystofwoldrich:
> @krystofwoldrich this seems like something we have to merge and cut a fix asap.

Yes, definitely. I will release the fix today.

krystofwoldrich (approved):

Issue #2580, opened 2022-10-27
Send modules along with SentryEvent


krystofwoldrich:
This is RN specific issue of https://github.com/getsentry/team-mobile/issues/54

RN apps are bundled by Metro therefore we will have to create the list of modules during a build.

krystofwoldrich:
I assume we want only runtime modules, that made it to the application bundle. 

Therefore I would make use of a source map that Metro bundler produces and iterate over the bundled files searching for all `package.json` files.

Similarly to how NodeJS SDK collects the modules information https://github.com/getsentry/sentry-javascript/blob/5d76ba54a6a7e84aa5de91496697fae3060f4248/packages/node/src/integrations/modules.ts#L17

krystofwoldrich:
I've researched all the possible ways of hooking into the Mero bundler, so we avoid running the bundling process twice, bud sadly no luck.

`collectModules` is our method that searches folders and collects modules from `package.json` files.

- `resolveRequest` function fails if custom is supplied ([issue](https://github.com/facebook/metro/issues/891))
-  `createModuleIdFactory` can be used to collect modules 
```js
    createModuleIdFactory: () => {
      const factory = createModuleIdFactory();

      return path => {
        collectModules(path);
        return factory(path);
      };
    },
  ```
- `postProcessBundleSourcemap` not implemented ([issue](https://github.com/facebook/metro/issues/400)) :(
- `processModuleFilter` can be used to collect modules and to modify the code in theory, but any modification of the module results in corupted bundle
```js
    processModuleFilter: (module) => {
      if (module.path.includes('sentry-react-native/dist/js/integrations/modules.json')) {
        module.output[0].data.code = module.output[0].data.code.replace('{}', globalModules);
        console.log('Sentry: Modules collected.');
      }
      return true;
    },
```


Using Assets.
Another option would be to save modules to assets. But there I'm not sure yet if we can copy the asset with the modules after the bundle is done since by default assets are copied before the bundle run.

PR #2606, opened 2022-11-08
Add collect modules

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
iOS will ship with a bash script that is called after RN Bundle phase.
Android has an extra task in the Gradle plugin that collects the modules.

Closes https://github.com/getsentry/sentry-react-native/issues/2580

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes


sentry.gradle
```diff
@@ -153,6 +154,32 @@ gradle.projectsEvaluated {
...
+          def modulesTask = tasks.create(nameModulesTask, Exec) {
+              description = "collect javascript modules from bundle source map"
+              group = 'sentry.io'
+
+              workingDir reactRoot
+
+              def collectModulesScript = project.hasProperty('sentry.collectModulesScript')
+                ? file(project.getProperty('sentry.collectModulesScript')).getAbsolutePath()
+                : "$reactRoot/node_modules/@sentry/react-native/dist/js/tools/collectModules.js"
+              def modulesOutput = "$reactRoot/android/app/src/main/assets/modules.json"
```
krystofwoldrich, line 166: @markushi Is there a cleaner way of adding a file to the app asset than creating it in `android/app/src/main/assets`?

CHANGELOG.md
```diff
@@ -2,9 +2,13 @@
 
 ## Unreleased
 
+To collect JS dependencies on iOS add `../node_modules/@sentry/react-native/scripts/collect-modules.sh` at the end of
+the `Bundle React Native code and images` build phase. The collection only works on Release builds.
```
marandaneto, line 6: This should be a subitem of:

> JS Runtime dependencies are sent in Events

Also, I'd give more instructions, only the name of the section in the build phase isn't enough, file name, expected result etc would be better, or write in the docs and link the docs here, example: https://docs.sentry.io/platforms/react-native/troubleshooting/#react-native-069-and-higher

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -20,7 +20,10 @@
 import com.facebook.react.module.annotations.ReactModule;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
```
marandaneto, line 26: This does not work on Android - every version, use:
`private static final Charset UTF_8 = Charset.forName("UTF-8");` instead.

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -174,6 +178,24 @@ public void crash() {
         throw new RuntimeException("TEST - Sentry Client Crash (only works in release mode)");
     }
 
+    @ReactMethod
+    public void fetchModules(Promise promise) {
+        try {
+            InputStream stream = this.getReactApplicationContext().getResources().getAssets()
```
marandaneto, line 184: H: Use try with resources and `BufferedInputStream`.
See AndroidOptionsInitializer on Android.

sample/metro.config.js
```diff
@@ -7,8 +7,6 @@
 
 const path = require('path');
 const blacklist = require('metro-config/src/defaults/exclusionList');
-const resolve = require('metro-resolver/src/resolve');
```
marandaneto, line 10: Isnt this needed anymore?

scripts/collect-modules.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
marandaneto, line 1: M: That should be fine on macOS and Linux, can we run on CI-Windows as well?

sentry.gradle
marandaneto: H: Whats the average size of the `modules.json`?
If app size is a concern for some folks, we need to offer a way to disable this feature.
We could add an opt-out via Gradle config, See the usage of `flavorAware`.

marandaneto, line 166: Maybe `cliCleanUpTask` should also delete this file.

sentry.gradle
```diff
@@ -161,6 +188,7 @@ gradle.projectsEvaluated {
             bundleTask.finalizedBy cliTask
           }
           previousCliTask = cliTask
+          cliTask.finalizedBy modulesTask
```
marandaneto, line 191: `modulesTask` has to be executed before the packaging of the final app otherwise the modules file wont be there, I guess the packaging task should depend on `modulesTask`, I guess?

CHANGELOG.md
krystofwoldrich: I will for sure write a docs fore it and link it. I just thought I'll write it here short and detailed in docs.

sample/metro.config.js
krystofwoldrich: Not needed for some time.

scripts/collect-modules.sh
krystofwoldrich: This is used on macOS only so it's fine.

CHANGELOG.md
marandaneto: Writing here makes sense, but the position should be under the feature anyway.

sentry.gradle
markushi: What you usually do on Android is to define extra [source sets](https://developer.android.com/studio/build/build-variants#sourcesets). There should be one for asset files as well, like `android.sourceSets.main.assets.srcDirs += ['path/to/your/folder]`. But I'm not sure if you have access to `android` in this context.

krystofwoldrich: @marandaneto The one from our sample app is 1 KB.

scripts/collect-modules.sh
```diff
@@ -0,0 +1,43 @@
...
+collectModulesScript="$thisFilePath/../dist/js/tools/collectModules.js"
+
+destination="$CONFIGURATION_BUILD_DIR/$UNLOCALIZED_RESOURCES_FOLDER_PATH"
+modulesOutput="$destination/modules.json"
+if [[ -z "$SOURCE_MAP_PATH" ]]; then
+  sourceMap="$DERIVED_FILE_DIR/main.jsbundle.map"
+else
+  sourceMap="$SOURCE_MAP_PATH"
+fi
+if [[ -z "$MODULES_PATHS" ]]; then
```
marandaneto, line 37: L: Is there more than one path? otherwise `MODULES_PATH`

sample/android/gradle.properties
```diff
@@ -26,3 +26,7 @@ android.enableJetifier=true
 
 # Version of flipper SDK to use with React Native
 FLIPPER_VERSION=0.37.0
+
+sentry.collectModulesScript=../../../dist/js/tools/collectModules.js
+sentry.modulesPaths=node_modules,../..
```
marandaneto, line 31: Is this mandatory for a final App adding the SDK?

sentry.gradle
```diff
@@ -169,7 +202,21 @@ gradle.projectsEvaluated {
...
+        }
+
+        def modulesCleanUpTask = tasks.create(name: nameModulesCleanup, type: Delete) {
+            description = "clean up collected modules generated file"
+            group = 'sentry.io'
+
+            delete modulesOutput
+        }
+
+        def packageTasks = tasks.findAll { task -> task.name.startsWith("package") && !task.name.contains("Debug") && task.enabled }
```
marandaneto, line 216: Doing the search this way has side effects in some Gradle versions.
See how the Gradle plugin does https://github.com/getsentry/sentry-android-gradle-plugin/blob/fd53b091525447713d93d72f5d6f018c67aedcc8/plugin-build/src/main/kotlin/io/sentry/android/gradle/SentryPlugin.kt#L214-L216

marandaneto: Also need to check if the `TaskProvider` for this task is available on older versions of the AGP

marandaneto: This is not a blocker, but I guess this will trigger some warnings on newer AGP/Gradle versions

scripts/collect-modules.sh
krystofwoldrich: Yes, there can be more, most of the time it will be just `node_modules`, but for example our sample would not work with just one path.

sample/android/gradle.properties
krystofwoldrich: No, this is just override that is needed for our sample, but there are defaults that should be good for basically everyone

sentry.gradle
krystofwoldrich: The Gradle plugin uses `project.findTask`

We are already using this loop in the rn `sentry.gradle` plugin. For the other tasks.

sentry.gradle
```diff
@@ -153,6 +157,34 @@ gradle.projectsEvaluated {
...
+                          modulesPaths
+              ]
+
+              project.logger.info("Sentry-CollectModules arguments: ${args}")
+              commandLine(*args)
+
+              def skip = project.hasProperty('sentry.skipCollectModules')
+                  ? project.getProperty('sentry.skipCollectModules') == 'true'
+                  : false
+              enabled !skip
```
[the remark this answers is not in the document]
krystofwoldrich: I've updated the options to use the config var.

sentry.gradle
```diff
@@ -169,7 +202,21 @@ gradle.projectsEvaluated {
...
+        }
+
+        def modulesCleanUpTask = tasks.create(name: nameModulesCleanup, type: Delete) {
+            description = "clean up collected modules generated file"
+            group = 'sentry.io'
+
+            delete modulesOutput
+        }
+
+        def packageTasks = tasks.findAll { task -> task.name == 'packageRelease' && task.enabled }
```
marandaneto, line 216: H: You can't rely on `packageRelease`, you have to use `bundleTask.name` as part of the name, as shared on the Android Gradle plugin code.

sentry.gradle
```diff
@@ -169,7 +202,21 @@ gradle.projectsEvaluated {
...
+            delete "$buildDir/intermediates/assets/release/index.android.bundle.map"
+            // react native default bundle dir
+        }
+
+        def modulesCleanUpTask = tasks.create(name: nameModulesCleanup, type: Delete) {
+            description = "clean up collected modules generated file"
+            group = 'sentry.io'
+
+            delete modulesOutput
+        }
```
marandaneto, line 214: M: Why not simply reuse the `cliCleanUpTask` task as suggested?

krystofwoldrich: Because it runs before the package task and the file will be deleted before it's included in the apk.

sentry.gradle
krystofwoldrich: But the bundle task is about creating a JS bundle and I need to wait for packing.

marandaneto: That's fine but it's still a problem, you have to compose the task name with the variant name, if you don't do that, an Android app with `flavors` or custom `buildTypes` won't work, you'd be depending on the wrong task.

src/js/integrations/modulesloader.ts
```diff
@@ -0,0 +1,35 @@
...
+    let isSetup = false;
+    let modules: Record<string, string> | null;
+
+    addGlobalEventProcessor(async (event: Event) => {
+      if (!isSetup) {
+        modules = await NATIVE.fetchModules();
+        isSetup = true;
+      }
+      if (modules) {
+        event.modules = modules;
```
marandaneto, line 30: I'd append the items rather than replace it.
If one set their own `modules`, you'd overwrite them.

sentry.gradle
krystofwoldrich: Ohh, I see, thank you 🙏, I'll fix that.

marandaneto:
@krystofwoldrich is this ready for a final review os is there something missing?

krystofwoldrich:
@marandaneto It's ready for final review.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### Features
 
 - Add `maxQueueSize` option ([#2578](https://github.com/getsentry/sentry-react-native/pull/2578))
+- JS Runtime dependencies are sent in Events ([#2606](https://github.com/getsentry/sentry-react-native/pull/2606))
+  - To collect JS dependencies on iOS add `../node_modules/@sentry/react-native/scripts/collect-modules.sh` at the end of the `Bundle React Native code and images` build phase. The collection only works on Release builds. Android builds have a new step in `sentry.gradle` plugin. More in [the migration documentation](https://docs.sentry.io/platforms/react-native/migration).
```
marandaneto, line 9: We can link the migration to the specific `fragment`, Can we draft up the migration entry?

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -174,6 +181,24 @@ public void crash() {
...
+                     new BufferedInputStream(assets.open(RNSentryModule.modulesPath))) {
+            int size = stream.available();
+            byte[] buffer = new byte[size];
+            stream.read(buffer);
+            stream.close();
+            String modulesJson = new String(buffer, RNSentryModule.UTF_8);
+            promise.resolve(modulesJson);
+        } catch (FileNotFoundException e) {
+            promise.resolve(null);
+        } catch (Exception e) {
```
marandaneto, line 197:
```suggestion
        } catch (Throwable e) {
```

marandaneto: M: Also add a `logger` call.

ios/RNSentry.m
```diff
@@ -157,6 +157,16 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
...
 }
 
+RCT_EXPORT_METHOD(fetchModules:(RCTPromiseResolveBlock)resolve
+                  rejecter:(RCTPromiseRejectBlock)reject)
+{
+    NSString *filePath = [[NSBundle mainBundle] pathForResource:@"modules" ofType:@"json"];
+    NSString* modulesString = [NSString stringWithContentsOfFile:filePath
+                                                  encoding:NSUTF8StringEncoding
+                                                     error:nil];
+    resolve(modulesString);
```
marandaneto, line 167: Does this return `nil` if the file does not exist?

sentry.gradle
```diff
@@ -169,7 +215,8 @@ gradle.projectsEvaluated {
             group = 'sentry.io'
 
             delete sourcemapOutput
-            delete "$buildDir/intermediates/assets/release/index.android.bundle.map" // react native default bundle dir
+            delete "$buildDir/intermediates/assets/release/index.android.bundle.map"
```
marandaneto, line 218: H: The `release` part should use a var because this depends on the `variant` most likely, similar to how its done for other files.

src/js/tools/collectModules.ts
```diff
@@ -0,0 +1,30 @@
...
+if (!outputModulesPath) {
+  throw new Error('Second argument `modules-output-path` is missing!');
+}
+if (modulesPaths.length === 0) {
+  throw new Error('Third argument `modules-paths` is missing!');
+}
+
+const map: { sources?: string[] } = JSON.parse(readFileSync(sourceMapPath, 'utf8'));
+if (!map.sources) {
+  throw new Error('No sources found in source map!');
```
marandaneto, line 24: Since we throw `Error`, do we crash the build in case this fails?

test/tools/collectModules.test.ts
```diff
@@ -0,0 +1,45 @@
...
+        `${__dirname}/fixtures/root-module/modules`,
+      ],
+    );
+
+    expect(modules).toEqual({
+      'module-1': 'module-1-version',
+      'module-2': 'unknown'
+    });
+  });
+});
```
marandaneto, line 45: Do we need all those empty files?

ios/RNSentry.m
krystofwoldrich: Yes.

sentry.gradle
krystofwoldrich: There is a source map deleted right above it and then there is this one written explicitly, I don't really know why.

test/tools/collectModules.test.ts
krystofwoldrich: Actually not, since the files are never read.

src/js/tools/collectModules.ts
marandaneto: yes, for sure, we should not break people's build.

sentry.gradle
```diff
@@ -161,6 +193,20 @@ gradle.projectsEvaluated {
...
+          cliTask.finalizedBy modulesTask
+
+          def modulesCleanUpTask = tasks.create(name: nameModulesCleanup, type: Delete) {
+              description = "clean up collected modules generated file"
+              group = 'sentry.io'
+
+              delete modulesOutput
+          }
+
+          def packageTasks = tasks.findAll { task -> task.name.toLowerCase() == "package${variant}" && task.enabled }
```
marandaneto, line 205: M: if `variant` isn't lowercase which I think the first letter is capitalized, this would fail.

sentry.gradle
marandaneto: Not sure I understand your comment here, mind clarifying?

sentry.gradle
krystofwoldrich: That's what I thought too, but when I checked the variant was `release` lowerCase.

I'll make it safer with `equalsIgnoreCase`.

sentry.gradle
krystofwoldrich: [On this line](https://github.com/getsentry/sentry-react-native/pull/2606/files/c727f94e4b4ad137fd7d1d69822414437ee37640#diff-01c94c11105e37de1b11a82973b605f625fff5959b0386c32fd1e42a14325208R217) the correct map will be deleted.

And then there is the next line with the hard codded path.

marandaneto: nvm, you just changed the comment, so all good

marandaneto (approved):
LGTM

fabiendem:
Hi folks, the release notes https://github.com/getsentry/sentry-react-native/releases/tag/4.10.0 mentions:
> To collect JS dependencies on iOS add ../node_modules/@sentry/react-native/scripts/collect-modules.sh at the end of the Bundle React Native code and images build phase. The collection only works on Release builds. Android builds have a new step in sentry.gradle plugin. More in [the migration documentation](https://docs.sentry.io/platforms/react-native/migration#from-48x-to-49x).

But I can't see any relevant info in the migration doc, is the doc not up-to-date or am I missing something?
Cheers!

krystofwoldrich:
@fabiendem Thanks for the message, sorry for the trouble, the docs will be merged asap.

PR #2373, opened 2022-07-15
Adds Attach Screenshot Option

PaitoAnderson:
Fixes #2362

## :loudspeaker: Type of change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Adds a `attachScreenshot` option that sets the `attachScreenshot` in cocoa project and calls `setAttachScreenshot` in java project. 

## :bulb: Motivation and Context

Excited about using the new Sentry screenshot feature in React Native projects.

## :green_heart: How did you test it?
Ran the sample projects.

iOS:
image

Android:
image
(Not sure what I have to do to trigger a screenshot, didn't work for me, but did verify the setting was indeed set at init)

## :pencil: Checklist

- [x] I reviewed submitted code
- [ ] I added tests to verify changes
- [x] All tests passing
- [x] No breaking changes

Issue #2362, opened 2022-07-09
Attach screenshot


bruno-garcia:
Add option `attachScreenshot` that when opt in, attached a screenshot on error events.

marandaneto:
It's not just adding the `attachScreenshot` flag.

On iOS, native hard crashes would work, but the JS errors won't contain the screenshot, because we call `captureEnvelope` or `storeEnvelope` and those functions won't attach the screenshot before sending it over to Sentry.

On Android, We always store the envelope on the disk, and the file observer will pick it up and enrich the event with device context, that's already too late and the view might not be up to date anymore, I have to debug the Android bits once more.

I feel that https://github.com/getsentry/team-mobile/issues/11 should be addressed first with this issue in mind.
cc @bruno-garcia

bruno-garcia:
If we're triggering the screenshot from JS through the bridge it's possible the view will change, no? So possibly not exactly what was on the screen at that time? It's fine if that's the case, just good to understand and document it.

> I feel that https://github.com/getsentry/team-mobile/issues/11 should be addressed first with this issue in mind.

Not sure this one affects things as much. We can take care of screenshots on each layer: In JS if there's an error, we get a screenshot through the bridge and pass it down to the native layers in the envelope. It doesn't need to live in the JS-land either, we could do all of that in the bridge if that makes things easier. 
For the native errors we just pass the flag down so they add screenshots when they capture errors on those layers.

Would that work?

So we need to figure out:
* Might require: #1327 
* https://github.com/getsentry/sentry-java/issues/2185

marandaneto:
> If we're triggering the screenshot from JS through the bridge it's possible the view will change, no? So possibly not exactly what was on the screen at that time? It's fine if that's the case, just good to understand and document it.

If we `await` the calls, technically the view is the same because we are in the calling thread.

> The first issue is that the currentActivity is null because the Android SDK is inited after the very first Activity is created.

This is still an issue, we'd need to find a way to set the `currentActivity` via the RN SDK, maybe the RN plugin should implement its own logic for screenshots, and the EventProcessor just exposes the methods to not duplicate the same logic, similar to Slow/frozen frames.

PR #2610, opened 2022-11-09
feat: Screenshots

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
The Screenshot feature required changes in `sentry-java` and `sentry-cocoa`. The `sentry-react-native` uses the native SDKs methods to capture screenshots. Those are copied over the bridge and the envelope bytes including the screenshots are assembled on the JS/RN side and then passed back to the native layers to be sent.

In case of native crashes, native SDKs capture screenshots themselves. 

- https://github.com/getsentry/sentry-java/pull/2360
	- Waits for @romtsn to be released (Likely Wednesday 23. Nov 2022)
- https://github.com/getsentry/sentry-cocoa/pull/2384
	- Released in `7.31.1` 	

The screenshots are captured in the `client` and not in the integration due to Async/Await JS nature. In the case of multiple async integrations, the screen could be different if captured in integration.

Closes: https://github.com/getsentry/sentry-react-native/pull/2373
Closes: https://github.com/getsentry/sentry-react-native/issues/2362

Using sample app.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- [ ] Merge the new `sentry-jave` to main rebase this PR and if CI passes merge.


marandaneto:
Consider https://github.com/getsentry/sentry-react-native/issues/2362#issuecomment-1193763553
Only this is not enough.

krystofwoldrich:
> Consider [#2362 (comment)](https://github.com/getsentry/sentry-react-native/issues/2362#issuecomment-1193763553) Only this is not enough.

I know I've read it.

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -290,6 +299,58 @@ public void captureEnvelope(ReadableArray rawBytes, ReadableMap options, Promise
         promise.resolve(true);
     }
 
+    @ReactMethod
+    public void captureScreenshot(Promise promise) {
+        final Activity activity = this.getReactApplicationContext().getCurrentActivity();
+        if (activity == null
```
marandaneto, line 305: Consider this check as well https://github.com/getsentry/sentry-java/blob/38e4f116623e0dd1329f54d86dd4bbe5dd687a86/sentry-android-core/src/main/java/io/sentry/android/core/ScreenshotEventProcessor.java#L182-L186

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -290,6 +299,58 @@ public void captureEnvelope(ReadableArray rawBytes, ReadableMap options, Promise
         promise.resolve(true);
     }
 
+    @ReactMethod
+    public void captureScreenshot(Promise promise) {
+        final Activity activity = this.getReactApplicationContext().getCurrentActivity();
```
marandaneto, line 304: Maybe we could refactor `ScreenshotEventProcessor` on android that takes an `Activity` as a parma and reuse this class/method in here, so we avoid duplication and increase of bundle size.

ios/RNSentry.m
```diff
@@ -291,6 +291,33 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
     resolve(@YES);
 }
 
+RCT_EXPORT_METHOD(captureScreenshot: (RCTPromiseResolveBlock)resolve
```
marandaneto, line 294: M: Can we actually reuse this? https://github.com/getsentry/sentry-cocoa/blob/master/Sources/Sentry/SentryScreenshot.m

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -127,6 +134,9 @@ public void initNativeSdk(final ReadableMap rnOptions, Promise promise) {
                 // by default we hide.
                 options.setAttachThreads(rnOptions.getBoolean("attachThreads"));
             }
+            if (rnOptions.hasKey("attachScreenshot")) {
+                options.setAttachScreenshot(rnOptions.getBoolean("attachScreenshot"));
+            }
```
marandaneto, line 139: H: This alone won't work for Native events, See https://github.com/getsentry/sentry-react-native/issues/2362#issuecomment-1193763553
Maybe we should expose a `ScreenshotEventProcessor#setCurrentActivity(activity)` so the RN SDK can overwrite this value when initing the Android SDK.

ios/RNSentry.m
krystofwoldrich: I think we could expose `takeScreenshots` using the PrivateSDKs header.

src/js/wrapper.ts
```diff
@@ -438,6 +452,25 @@ export const NATIVE: SentryNativeWrapper = {
...
+    }
+
+    try {
+      const raw = await RNSentry.captureScreenshot();
+      return raw.map((item: NativeScreenshot) => ({
+        ...item,
+        data: new Uint8Array(item.data),
+      }));
+    } catch (e) {
+      return null;
```
marandaneto, line 470: L: Should we log it?

src/js/client.ts
```diff
@@ -204,4 +205,24 @@ export class ReactNativeClient extends BaseClient<ReactNativeClientOptions> {
       envelope[items].push(clientReportItem);
     }
   }
+
+  /**
+   * If enabled attaches a screenshot to the event hint.
+   */
+  private _attachScreenshotToEventHint(hint: EventHint): PromiseLike<EventHint> {
```
marandaneto, line 212: M: Does it make sense to be an integration or event processor instead?
If not, that's ok, but something we have to do is to at least add an integration in the `sdk.integrations` list so we can track adoption of this feature.
Only if `attachScreenshot` is enabled of course.

krystofwoldrich: As the order of the integrations is not guaranteed, and integrations can be async the screen could be captured too late. Or would that still be good enough? I know other platforms also use the event processor for the screenshots.

I'll add the Integration so we can track the adoption.

marandaneto:
@krystofwoldrich is this ready for a final review os is there something missing?
The PR description is not fully filled.

krystofwoldrich:
@marandaneto Ready for final review, it only waits for `sentry-java` to be released. I talked to @romtsn and he would like to finish something and then release it (likely this Wednesday).

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -285,16 +302,47 @@ public void captureEnvelope(ReadableArray rawBytes, ReadableMap options, Promise
...
-            logger.severe("Error while writing envelope to outbox.");
+            logger.log(SentryLevel.ERROR, "Error while writing envelope to outbox.");
         }
         promise.resolve(true);
     }
 
+    @ReactMethod
+    public void captureScreenshot(Promise promise) {
+
+        final Activity activity = this.getReactApplicationContext().getCurrentActivity();
```
marandaneto, line 313: Can you call `getCurrentActivity()` directly here?

ios/RNSentry.m
```diff
@@ -292,6 +292,35 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
...
+                  rejecter: (RCTPromiseRejectBlock)reject)
+{
+    NSArray<NSData *>* rawScreenshots = [PrivateSentrySDKOnly captureScreenshots];
+    NSMutableArray *screenshotsArray = [NSMutableArray arrayWithCapacity:[rawScreenshots count]];
+
+    int counter = 1;
+    for (NSData* raw in rawScreenshots) {
```
marandaneto, line 302: M: Can this be `nil` or 0 bytes or something? we do checks on Android but not on iOS?

marandaneto:
@krystofwoldrich Since you are blocked by the Java release, we can draft up the docs PR for RN, similar to https://docs.sentry.io/platforms/flutter/enriching-events/screenshots/

marandaneto (approved):

ios/RNSentry.m
krystofwoldrich: I don't know, but I would say, if it can, we should check it in `cocoa` and not add such a screenshot to the array. 

@brustolin Do you know, if [this](https://github.com/getsentry/sentry-cocoa/blob/f781ecba1e2ccfc37e2156f1a9300452c0c8f4c8/Sources/Sentry/SentryScreenshot.m#LL49C17-L49C17) can be `nil` sometimes or 0 size?

android/src/main/java/io/sentry/react/RNSentryModule.java
krystofwoldrich: I can, thanks.

ios/RNSentry.m
brustolin: Theoretically no. And I never saw it happening. 
Nil here is not a problem, the loop will ignore nil, but I guess you could add a check for 0 size.

krystofwoldrich: @brustolin Should I add the 0 size check in the cocoa SDK, I guess the 0 size screenshot should not be in the list.

brustolin: Good point. That would be the best solution

marandaneto:
@krystofwoldrich can we unblock this? it should not be a problem cutting a patch release on Android.

krystofwoldrich:
@marandaneto The release is out. So this is not blocked anymore.

marandaneto:
> @marandaneto The release is out. So this is not blocked anymore.

I'd way for this https://github.com/getsentry/publish/issues/1614
The release 6.9.0 isn't fully finished, not sure if it's reliable yet.

ios/RNSentry.m
krystofwoldrich: - https://github.com/getsentry/sentry-cocoa/pull/2459

marandaneto:
> > @marandaneto The release is out. So this is not blocked anymore.
> 
> I'd way wait this [getsentry/publish#1614](https://github.com/getsentry/publish/issues/1614) The release 6.9.0 isn't fully finished, not sure if it's reliable yet.

@krystofwoldrich you can unblock this -> https://github.com/getsentry/sentry-react-native/pull/2653

CHANGELOG.md
```diff
@@ -2,11 +2,12 @@
 
 ## 4.10.1
 
+### Features
```
marandaneto, line 5: @krystofwoldrich make it to `Unreleased`

krystofwoldrich: On it. :)

PR #2627, opened 2022-11-18
ref(js): Use `Scope` class rather than `Scope` type for top-level functions

lobsterkatie:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

This switches the functions defined in `sdk.ts` from using the `Scope` interface for typing to using the `Scope` class for typing, so that the types of `withScope` here and `withScope` from the main JS repo match.

Fixes https://github.com/getsentry/sentry-javascript/issues/6217.

I steered the OP of the above issue toward a different approach, so in the end he may avoid this problem regardless, but the error he was getting points out the mismatched types.

I have not tested it. Total mea culpa, I didn't want to go through and set up a whole RN SDK dev environment just for this tiny change, so I did it through the GH UI. I'm hoping that one of my esteemed colleagues might be willing to test it out and see if I've messed up the types anywhere else in the SDK. 🙏🏻

I'll leave this a draft for now, until it's tested.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

I guess _conceivably_ this could be a breaking change, but only if someone has explicitly typed their `withScope` callback, which feels pretty unlikely.

## :crystal_ball: Next steps

Figure out if any other types need to be changed elsewhere in the SDK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 - Add `maxQueueSize` option ([#2578](https://github.com/getsentry/sentry-react-native/pull/2578))
 
+### Fixes
+- Use `Scope` class rather than `Scope` type for top-level functions ([#2627](https://github.com/getsentry/sentry-react-native/pull/2627))
+
 ### Dependencies
 
 - Bump JavaScript SDK from v7.16.0 to v7.17.4 ([#2582](https://github.com/getsentry/sentry-react-native/pull/2582), [#2598](https://github.com/getsentry/sentry-react-native/pull/2598))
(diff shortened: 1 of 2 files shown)


krystofwoldrich (approved):
LGTM. I would keep it as a fix, the specific `withScope` is quite a recent change in RN, before that we were just reexporting the one from JS.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 - Add `maxQueueSize` option ([#2578](https://github.com/getsentry/sentry-react-native/pull/2578))
 
+### Fixes
+- Use `Scope` class rather than `Scope` type for top-level functions ([#2627](https://github.com/getsentry/sentry-react-native/pull/2627))
```
marandaneto, line 10:
```suggestion
### Fixes

- Use `Scope` class rather than `Scope` type for top-level functions ([#2627](https://github.com/getsentry/sentry-react-native/pull/2627))
```

Issue #2654, opened 2022-11-30
Archive failing with space in scheme-name and collect-modules.sh


rafaelmaeuer:
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [x]  iOS
- [ ]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 5.0.0-alpha.9

**`react-native` version:** 0.70.6

**Are you using Expo?**
- [x]  Yes (bare workflow)
- [ ]  No

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)
- [ ]  on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

[Link to issue]

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
	debug: false,
	dsn: Config.SENTRY_DSN,
	sampleRate: isDev ? 0.0 : 1.0,
	environment: isDev ? 'development' : 'production',
})
```

---
**I have following issue:**

With update of @sentry/react-native to `v5.0.0-alpha.9` which has latest changes from `4.10.0`, modification of `Bundle React Native code and images build phase` script is required, so I applied changes from [Migration - From 4.8.x to 4.9.x](https://docs.sentry.io/platforms/react-native/migration/#from-48x-to-49x). App-Archive is now failing when having spaces in the scheme-name: e.g. `app-name (release)`. Archive completes when using `app-name-release` as scheme-name instead.

```
Error: ENOENT: no such file or directory, open '/Users/<username>/Library/Developer/Xcode/DerivedData/<app-name-cryptic-build-id>/Build/Intermediates.noindex/ArchiveIntermediates/<app-name>'
at Object.openSync (node:fs:595:3)
    at Object.readFileSync (node:fs:463:35)
    at Object.<anonymous> (/Users/<username>/Development/<project>/<app-name>/app/node_modules/@sentry/react-native/dist/js/tools/collectModules.js:20:29)
    at Module._compile (node:internal/modules/cjs/loader:1149:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1203:10)
    at Module.load (node:internal/modules/cjs/loader:1027:32)
    at Module._load (node:internal/modules/cjs/loader:868:12)
[...]

PR #2655, opened 2022-11-30
fix(modules): Exit gracefully if source map file doesn't exist

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
If a user supplies a wrong path or the assumed path doesn't exist it would crash the build.

fixes: https://github.com/getsentry/sentry-react-native/issues/2654

## :pencil: Checklist

-  [x] I reviewed submitted code
- [x] All tests passing
- [x] No breaking changes


marandaneto:
@krystofwoldrich can we add tests to avoid regression?

krystofwoldrich:
@marandaneto Yes, I will move the validation logic to the class so it's easy to test.

src/js/tools/ModulesCollector.ts
```diff
@@ -1,6 +1,7 @@
-import { logger } from '@sentry/utils';
```
marandaneto, line 1: M: Why aren't using the `logger` anymore?

src/js/tools/ModulesCollector.ts
```diff
@@ -1,7 +1,9 @@
-/* eslint-disable no-console */
+import { logger } from '@sentry/utils';
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'fs';
 import { posix, sep } from 'path';
 
+logger.enable();
```
marandaneto, line 5: M: Is this needed? I don't see this anywhere else.

marandaneto (approved):
Left a last [comment](https://github.com/getsentry/sentry-react-native/pull/2655/commits/25b064122e55584b85d151c4ea8afd18a8961312#r1036798853) otherwise LGTM

src/js/tools/ModulesCollector.ts
krystofwoldrich: For RN SDK it's hidden in [`initAndBind`](https://github.com/getsentry/sentry-javascript/blob/8570f0d29567e5f7cebf9a18f82e019191f2ebee/packages/core/src/sdk.ts#L22)
But as this run as a standalone script, I needed to enable the logger.

Issue #2682, opened 2022-12-08
Performance monitoring (Custom Instrumentation) logs are not sent on Android


takahi5:
**OS:**

- [ ] Windows
- [x] MacOS
- [ ] Linux

**Platform:**

- [ ] iOS
- [x] Android

**SDK:**

- [x] `@sentry/react-native` (>= 1.0.0)
- [ ] `react-native-sentry` (<= 0.43.2)

**SDK version:** 4.11.0

**`react-native` version:** 0.70.6

**Are you using Expo?**

- [ ] Yes
- [x] No

**Are you using sentry.io or on-premise?**

- [x] sentry.io (SaaS)
- [ ] on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

[Link to issue]

**Configuration:**

(`@sentry/react-native`)

```js
Sentry.init({
  dsn: 'https://...@sentry.io/...'
  tracesSampleRate: 1.0, // send transaction always
});
```

---

**I have following issue:**

Performance monitoring (Custom Instrumentation) logs are not sent on Android

**Steps to reproduce:**

- Step 1
  - Create new React Native project with `npx react-native init SampleApp`
- Step 2
  - Setup sentry according to the [document](https://docs.sentry.io/platforms/react-native/)
    - `yarn add @sentry/react-native`
    - `npx @sentry/wizard -i reactNative -p ios android`
- Step 3
  - Setup Performance Monitoring according to the [document](https://docs.sentry.io/platforms/react-native/performance/)

```jsx
import * as Sentry from "@sentry/react-native";

Sentry.init({
  dsn: "https://...",
  tracesSampleRate: 1.0
});

const App: () => Node = () => {
  useEffect(() => {
    const transaction = Sentry.startTransaction({name: 'test-transaction'});
    const span = transaction.startChild({op: 'functionX'});
    // functionCallX
    span.finish();
    transaction.finish();
  }, []);
	...
}
```

- Step 4
  - Check Performance Monitoring at Sentry.io

**Actual result:**

- ❌ When I run [Custom Instrumentation](https://docs.sentry.io/platforms/react-native/performance/instrumentation/custom-instrumentation/) on Android, I cannot see any logs on sentry.io. 
  
- ✅ No problem with [Automatic Instrumentation](https://docs.sentry.io/platforms/react-native/performance/instrumentation/automatic-instrumentation/). 
- ✅ No problem with iOS. 
[...]

PR #2691, opened 2022-12-12
Add transaction op default

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
`transaction.op` is `required` on Android.

closes: https://github.com/getsentry/sentry-react-native/issues/2682

tests and sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- https://github.com/getsentry/sentry-react-native/issues/2693


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 - Add `lastEventId` method to the API ([#2675](https://github.com/getsentry/sentry-react-native/pull/2675))
 
+### Fix
+- `Sentry.startTransaction` doesn't require `op` ([#2691](https://github.com/getsentry/sentry-react-native/pull/2691))
```
marandaneto, line 10:
```suggestion
### Fix

- `Sentry.startTransaction` doesn't require `op` ([#2691](https://github.com/getsentry/sentry-react-native/pull/2691))
```

src/js/measurements.ts
```diff
@@ -48,6 +48,10 @@ const _patchStartTransaction = (
     transactionContext: TransactionContext,
     customSamplingContext?: CustomSamplingContext
   ): Transaction {
+    if (!transactionContext.op) {
+      transactionContext.op = 'default';
+    }
```
marandaneto, line 53: We'll need to do that for `startChild` as well.

src/js/measurements.ts
```diff
@@ -48,10 +48,23 @@ const _patchStartTransaction = (
     transactionContext: TransactionContext,
     customSamplingContext?: CustomSamplingContext
   ): Transaction {
+    if (!transactionContext.op) {
+      transactionContext.op = 'default';
```
marandaneto, line 52: M: Let's make a `const` for `default`, its being used in multiple laces.

marandaneto, line 52: M: I'm wondering if this should live here, the file is about measurements, is the reason that we only patch the `startTransaction` and `startChild` here? Can we move this patching else where?

marandaneto: Let's add a comment explaining why we have to patch and set the `op` as `default`.
Same for `startChild`.

marandaneto:
Let's double check if `startIdleTransaction` has to be patched as well, if its an internal and not documented API, we dont need to do that since we always have an `op`.

krystofwoldrich:
@marandaneto Checked. `startIdleTransaction` is not documented and it's only used by the SDKs.

src/js/measurements.ts
krystofwoldrich: I'll open a new issue for the refactoring.

marandaneto (approved):

PR #2710, opened 2022-12-21
Custom touch tracking property instead of accessibilityLabel

jenskuhrjorgensen:

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
This PR removes the default usage of `accessibilityLabel` for automatic touch tracking due to GDPR issues. As a replacement, the PR adds a custom `labelName` property to the `TouchEventBoundary` that can be used to set a custom property name to be used for touch tracking. 

Currently, the `sentry-react-native` uses [`accessibilityLabel` for automatic touch tracking](https://docs.sentry.io/platforms/react-native/touchevents/#tracking-specific-components). This means that if no `sentry-label` is provided it falls back to using `accessibilityLabel` as part of the uploaded breadcrumbs. However, `accessibilityLabel` often contains PII data that can help accessibility users with e.g. TalkBack or VoiceOver (this be for instance home address, phone number, social security number etc.). This means that whenever a touch event is logged to sentry.io for a component that has an `accessibilityLabel` and not a `sentry-label` the PII data will end up at Sentry. This is a GDPR issue for European companies. 

I think for a lot of users/companies it is not desirable (or even legal) to have their `accessibilityLabel` data logged by default. This PR adds a custom `labelName` property to the `TouchEventBoundary` that will then be used as fallback in case no `sentry-label` is present. This can then be set to for instance `"accessibilityLabel"` for users that would like to continue using their `accessibilityLabel` in touch tracking, but it will then make sure that it is a conscious choice to log this as it will potentially hold PII data. In my own case, we would like to use our existing `testID` property for touch tracking (currently, we have an npm patch applied to Sentry that does exactly this because logging `accessibilityLabel` is definitely not an option, an even though we could get around this by adding a `sentry-label` to all components that have an `accessibilityLabel` the default fallback is simply to dangerous for us in case we miss a `sentry-label` and an `accessibilityLabel` instead slips through), as it gives us a lot of context information about what interactions the app user performed up to a crash.

I was motivated to create this PR by our support contacts from Sentry, Will C. and Hazid M.

I replaced the `accessibilityLabel` tests with `custom-sentry-label-name` and verified that clicking the "Capture Exception" button is logging the value stored in `custom-sentry-label-name` in the Metro bundler in "Event beforeSend:" instead of the previously logged `accessibilityLabel`.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
This is a breaking change as `accessibilityLabel` will no longer be logged by default. The migration for existing users is to simply add it to your `touchEventBoundaryProps.labelName`:

```
Sentry.wrap(App, {
  touchEventBoundaryProps: {
    labelName: 'accessibilityLabel',
  },
})
```
or
```
<Sentry.TouchEventBoundary
  labelName='accessibilityLabel'
>
  <RestOfTheApp />
</Sentry.TouchEventBoundary>
```


src/js/touchevents.tsx
```diff
@@ -31,6 +31,10 @@ export type TouchEventBoundaryProps = {
    * React Node wrapped by TouchEventBoundary.
    */
   children?: React.ReactNode;
+  /**
+   * React Node wrapped by TouchEventBoundary.
```
krystofwoldrich, line 35:
```suggestion
   * Label Name used to identify the touched element.
   * Default: `sentry-label`
   * Fallback: `element.displayName`
```

krystofwoldrich:
What if we kept the `accessibilityLabel` as default? Then it would not be a breaking change but still, it would be super simple to change the behavior.

jenskuhrjorgensen:
> What if we kept the `accessibilityLabel` as default? Then it would not be a breaking change but still, it would be super simple to change the behavior.
> 
> Otherwise, I like the change, if we keep it as a breaking change we will need to point the PR to the `5.0.0` branch.

Thanks for the feedback. 

I think the "right" solution is to not use `accessibilityLabel` by default as it imposes a legal risk for some users. We already have a working patch to remove `accessibilityLabel` from the logs, so I'm fine with waiting until v5 with getting this released. I will update the target branch.

jenskuhrjorgensen:
@krystofwoldrich I created a new PR to target the v5 release
https://github.com/getsentry/sentry-react-native/pull/2712

src/js/touchevents.tsx
jenskuhrjorgensen: Good catch. I fixed the first line of the description but I don't think the default and fallback values fit in here, because setting a `sentry-label` on a specific component will actually overwrite the `labelName` set via `TouchEventBoundary`.

Issue #2085, opened 2022-02-22
Append missing context from Native SDK for iOS


marandaneto:
Similar to https://github.com/getsentry/sentry-dart/pull/761/files

Missing tags, extra, user, dist, environment, fingerprint, level, breadcrumbs...
Ideally, every scope data set in the Native SDK makes it to the Events captured by JS as well.

marandaneto:
Right RN -> iOS only merges the `contexts` and `user`, https://github.com/getsentry/sentry-react-native/blob/main/src/js/integrations/devicecontext.ts
But it should merge/fill the missing objects as well the `tags`, `extra`, `dist`, `environment`, `fingerprint`, `level`, `breadcrumbs`.

krystofwoldrich:
This might be a breaking change due to the inclusion of the event.fingerprint.

marandaneto:
> This might be a breaking change due to the inclusion of the event.fingerprint.

That's fine, its part of the v5 milestone anyway.

PR #2713, opened 2022-12-21
feat(contexts): Sync native iOS contexts to JS

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
This might be a breaking change due to the inclusion of the `event.fingerprint`.

closes: https://github.com/getsentry/sentry-react-native/issues/2085

integration tests

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
- [ ] Check breadcrumbs timestamps


src/js/breadcrumb.ts
```diff
@@ -0,0 +1,53 @@
...
+      breadcrumb.timestamp = timestamp;
+    }
+  }
+
+  return breadcrumb;
+}
+
+/**
+ * Compares two breadcrumbs and returns true if they are equal
+ * Warning: This function does not check timestamp and level, data
```
marandaneto, line 44: M: If the matter if to avoid duplication, all the fields would need to be considered.

ios/RNSentry.m
```diff
@@ -171,33 +171,27 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
...
-        NSDictionary<NSString *, id> *tempContexts = [serializedScope valueForKey:@"context"];
-
-        NSMutableDictionary<NSString *, id> *user = [NSMutableDictionary new];
-
-        NSDictionary<NSString *, id> *tempUser = [serializedScope valueForKey:@"user"];
-        if (tempUser != nil) {
-            [user addEntriesFromDictionary:[tempUser valueForKey:@"user"]];
-        } else {
-            [user setValue:PrivateSentrySDKOnly.installationID forKey:@"id"];
+        NSDictionary *serializedScope = [scope serialize];
```
marandaneto, line 178: Is there a reason to avoid the specific types such as `<NSString *, id>`?

ios/RNSentry.m
```diff
@@ -171,33 +171,27 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
                   rejecter:(RCTPromiseRejectBlock)reject)
 {
     NSLog(@"Bridge call to: deviceContexts");
-    NSMutableDictionary<NSString *, id> *contexts = [NSMutableDictionary new];
+    __block NSMutableDictionary *contexts;
```
marandaneto, line 174: What's the reason to make it a not inited `__block`?

src/js/integrations/devicecontext.ts
```diff
@@ -19,26 +21,79 @@ export class DeviceContext implements Integration {
...
 
-        if (!event.user) {
-          event.user = { ...user };
+      const nativeBreadcrumbs = Array.isArray(contexts['breadcrumbs'])
+        ? contexts['breadcrumbs'].map(breadcrumbFromObject)
+        : undefined;
+      if (nativeBreadcrumbs) {
+        event.breadcrumbs = event.breadcrumbs ?? []
+        for (const breadcrumb of nativeBreadcrumbs ?? []) {
+          const equals = (b: Breadcrumb): boolean => breadcrumbEquals(b, breadcrumb);
```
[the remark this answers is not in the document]
krystofwoldrich: Yes, it's duplication.

ios/RNSentry.m
krystofwoldrich: Yes, it's for the assigned dictionary in the configureScope to be visible outside that block so we can pass it to resolve.

ios/RNSentry.m
krystofwoldrich: Sind I didn't need the specific type it was simple to read for me. But other than that no reason. I will return it.

marandaneto: I'd rather prefer strongly typed, so the compiler helps us if we do something wrong.

src/js/integrations/devicecontext.ts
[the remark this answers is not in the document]
marandaneto: How do we avoid the dupe on Android? this has a runtime cost, wondering if we can make it better/align the solution across Android/iOS.

src/js/breadcrumb.ts
krystofwoldrich: I excluded [the timestamp because it's not passed](https://github.com/getsentry/sentry-react-native/blob/a5d7a18015e3e11c7db6e69acd41e84c8d0c2b2b/ios/RNSentry.mm#L359) to the native layer, therefore it will be always different.
[The levels are also recalculated](https://github.com/getsentry/sentry-react-native/blob/a5d7a18015e3e11c7db6e69acd41e84c8d0c2b2b/src/js/wrapper.ts#L391) so they are not matching for the same breadcrumbs.
And data are being normalized so it's also tricky to check.

I suppose we will need to change how the breadcrumbs are passed to ios.

src/js/integrations/devicecontext.ts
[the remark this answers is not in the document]
krystofwoldrich: Oh, I did not think about that.
Maybe we can just remove the JS breadcrumbs and use the iOS ones since all the JS breadcrumbs are synced down.
On android we do the same, we remove the JS breadcrumbs and use Android only.

src/js/breadcrumb.ts
krystofwoldrich: Maybe[ this will solve it](https://github.com/getsentry/sentry-react-native/pull/2713#discussion_r1061360166).

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -444,13 +444,13 @@ public void addBreadcrumb(final ReadableMap breadcrumb) {
...
                         break;
-                    case "error":
-                        breadcrumbInstance.setLevel(SentryLevel.ERROR);
-                        break;
                     case "info":
-                    default:
                         breadcrumbInstance.setLevel(SentryLevel.INFO);
                         break;
+                    case "error":
+                    default:
```
marandaneto, line 451: H: `info` is default.

src/js/wrapper.ts
```diff
@@ -566,11 +566,6 @@ export const NATIVE: SentryNativeWrapper = {
     if (level == 'log' as SeverityLevel) {
       return 'debug' as SeverityLevel;
     }
-    else if (level == 'critical' as SeverityLevel) {
-      return 'fatal' as SeverityLevel;
-    }
```
marandaneto, line 571: Why this is removed? there's no possibility of being `critical` on the JS layer anymore?

test/wrapper.test.ts
```diff
@@ -474,10 +474,6 @@ describe('Tests Native Wrapper', () => {
   });
 
   describe('_processLevel', () => {
-    test('converts deprecated levels', () => {
-      expect(NATIVE._processLevel('log' as SeverityLevel)).toBe('debug' as SeverityLevel);
```
marandaneto, line 478: H: test is removed but the wrapper still has `if` references to `log`.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
krystofwoldrich: Thanks, I've mixed up the breadcrumbs level and event level.

src/js/wrapper.ts
krystofwoldrich: Yes, `critical` was removed.

https://github.com/getsentry/sentry-javascript/blob/0f9a3dfb864e8b0c57daac7dbda333abee20d04d/packages/types/src/severity.ts#L22

krystofwoldrich:
The E2E tests are fixed here https://github.com/getsentry/sentry-react-native/pull/2687.

Otherwise, this is ready.

marandaneto (approved):

Issue #2742, opened 2023-01-12
Screenshots should be taken on the UI thread


krystofwoldrich:
This error log only appears in dev builds. Nevertheless, the screenshot-taking process should run on the UI thread.

```
2023-01-12 14:12:23.257 11693-11786 unknown:SoftAssertions  io.sentry.sample                     E  Unhandled SoftException
                                                                                                    com.facebook.react.bridge.AssertionException: Expected to run on UI thread!
                                                                                                    	at com.facebook.react.bridge.SoftAssertions.assertCondition(SoftAssertions.java:37)
                                                                                                    	at com.facebook.react.bridge.UiThreadUtil.assertOnUiThread(UiThreadUtil.java:32)
                                                                                                    	at com.facebook.react.views.view.ReactViewGroup.getChildDrawingOrder(ReactViewGroup.java:550)
                                                                                                    	at android.view.ViewGroup.getAndVerifyPreorderedIndex(ViewGroup.java:2097)
                                                                                                    	at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4287)
                                                                                                    	at com.facebook.react.views.view.ReactViewGroup.dispatchDraw(ReactViewGroup.java:810)
                                                                                                    	at android.view.View.draw(View.java:23066)
                                                                                                    	at android.view.ViewGroup.drawChild(ViewGroup.java:4529)
                                                                                                    	at com.facebook.react.views.view.ReactViewGroup.drawChild(ReactViewGroup.java:838)
                                                                                                    	at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4290)
                                                                                                    	at com.facebook.react.views.view.ReactViewGroup.dispatchDraw(ReactViewGroup.java:810)
                                                                                                    	at android.view.View.draw(View.java:23066)
                                                                                                    	at android.view.ViewGroup.drawChild(ViewGroup.java:4529)
                                                                                                    	at com.facebook.react.views.view.ReactViewGroup.drawChild(ReactViewGroup.java:838)
                                                                                                    	at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4290)
                                                                                                    	at com.facebook.react.views.view.ReactViewGroup.dispatchDraw(ReactViewGroup.java:810)
                                                                                                    	at android.view.View.draw(View.java:23198)
                                                                                                    	at android.view.View.draw(View.java:23068)
                                                                                                    	at android.view.ViewGroup.drawChild(ViewGroup.java:4529)
[...]

PR #2743, opened 2023-01-12
fix(screenshots): Run take screenshot on ui thread always

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description
The error log appears only in dev builds but suggests that the screenshots should be taken on the UI thread.

fixes: https://github.com/getsentry/sentry-react-native/issues/2742

sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Take screenshot runs on UI thread on Android
```
```suggestion
- Take screenshot runs on UI thread on Android ([#2743](https://github.com/getsentry/sentry-react-native/pull/2743))
```

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -343,15 +345,29 @@ public void captureScreenshot(Promise promise) {
             return;
         }
 
-        final byte[] raw = takeScreenshot(activity, logger, buildInfo);
-        if (raw == null) {
+        CountDownLatch doneSignal = new CountDownLatch(1);
+        final byte[][] bytesWrapper = {{}};
+        UiThreadUtil.runOnUiThread(() -> {
```
marandaneto, line 350: M: Maybe we check first if its not in the UI Thread? Using `UiThreadUtil#isOnUiThread` so we don't do the overhead unnecessarily.

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -343,15 +345,29 @@ public void captureScreenshot(Promise promise) {
...
+            bytesWrapper[0] = takeScreenshot(activity, logger, buildInfo);
+            doneSignal.countDown();
+        });
+
+        try {
+            doneSignal.await();
```
marandaneto, line 356: M: Should we set a default await time? just wondering if this can be stuck somehow.

krystofwoldrich: I don't know, but better be safe, I'll add a timeout.

android/src/main/java/io/sentry/react/RNSentryModule.java
krystofwoldrich: Make sense.

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -80,6 +83,8 @@ public class RNSentryModule extends ReactContextBaseJavaModule {
     // 16ms (slower than 60fps) to constitute slow frames.
     private static final int SLOW_FRAME_THRESHOLD = 16;
 
+    private static final int SCREENSHOT_TIMEOUT_SECONDS = 10;
```
marandaneto, line 86:
```suggestion
    private static final int SCREENSHOT_TIMEOUT_SECONDS = 2;
```

H: I believe 10 is too much, you can trigger 2 ANRs with 10s already.

android/src/main/java/io/sentry/react/RNSentryModule.java
```diff
@@ -364,6 +370,30 @@ public void captureScreenshot(Promise promise) {
...
+        }
+
+        try {
+            doneSignal.await(SCREENSHOT_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            logger.log(SentryLevel.ERROR, "Screenshot process was interrupted.");
+            return null;
+        }
+
+        return bytesWrapper[0];
```
marandaneto, line 394: M: we can lazy init the bytesWrapper and return null if there's nothing.

krystofwoldrich: I'm not sure what you mean. I can check if the first element of the wrapper length is larger than 0. But I have to create the wrapper before the runnable.

marandaneto: What I meant is that you create initing `bytesWrapper = {{}};` but actually you can init the variable wit the return of the `takeScreenshot` and return null otherwise, right now you return an empty `bytesWrapper` even if there's nothing in there, so you are locating something unnecessarily.

marandaneto: Also you just need `byte[]` and not `byte[][]`

krystofwoldrich: That is on purpose because otherwise I run into `Variable used in lambda expression should be final or effectively final`.

marandaneto: Ah gotcha, missed that, thanks, maybe add a comment to explain this?

marandaneto (approved):

android/src/main/java/io/sentry/react/RNSentryModule.java
krystofwoldrich: Sure thing, added a comment.

Issue #2210, opened 2022-04-13
Report top level integrations on RN


marandaneto:
Similar to https://github.com/getsentry/sentry-java/issues/1989 but we already send all the integrations.
It'd be useful to also send the routing libraries etc that are added manually, so we can learn the usage of it.

krystofwoldrich:
@marandaneto Routing libraries you mean the js deps of a project? -> https://github.com/getsentry/team-mobile/issues/54

Would it also make sense to report which one of our [`routingInstrumentation`](https://github.com/getsentry/sentry-react-native/blob/83be16fd96630c3e567aca2bb773d04ad5b08824/src/js/tracing/reactnativetracing.ts#L55) is being used?

marandaneto:
@krystofwoldrich nope, I mean, `ReactNavigationInstrumentation`, `ReactNavigationV4Instrumentation`, `ReactNativeTracing`, `RoutingInstrumentation `, etc... every integration/router/navigator... that has to be added manually, we'd know which integrations are being set or not, also its usage.

krystofwoldrich:
Seems like we should just change the name of the `ReactNativeTracing` integrations based on the used instrumentation since the `ReactNativeTracing` can only create `Transaction` for an app start.

PR #2798, opened 2023-02-01
feat(internal): Add routing instrumentations report in integrations

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

closes: https://github.com/getsentry/sentry-react-native/issues/2210

sample app

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/src/js/tracing/routingInstrumentation.ts
+++ b/src/js/tracing/routingInstrumentation.ts
@@ -11,6 +11,10 @@ export type OnConfirmRoute = (context: TransactionContext) => void;
 
 export interface RoutingInstrumentationInstance {
   /**
+   * Name of the routing instrumentation
+   */
+  readonly name: string;
+  /**
    * Registers a listener that's called on every route change with a `TransactionContext`.
    *
    * Do not overwrite this unless you know what you are doing.
@@ -39,6 +43,8 @@ export interface RoutingInstrumentationInstance {
  */
 export class RoutingInstrumentation implements RoutingInstrumentationInstance {
   public static instrumentationName: string = 'base-routing-instrumentation';
+
+  public readonly name: string = RoutingInstrumentation.instrumentationName;
 
   protected _getCurrentHub?: () => Hub;
   protected _beforeNavigate?: BeforeNavigate;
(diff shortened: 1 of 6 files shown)


src/js/tracing/routingInstrumentation.ts
```diff
@@ -40,6 +44,8 @@ export interface RoutingInstrumentationInstance {
 export class RoutingInstrumentation implements RoutingInstrumentationInstance {
   public static instrumentationName: string = 'base-routing-instrumentation';
 
+  public readonly name: string = RoutingInstrumentation.instrumentationName;
```
marandaneto, line 47: L: Should we kill `instrumentationName` then and use `name` only?

marandaneto:
Can we also do that for other features that require a setup, such as https://docs.sentry.io/platforms/react-native/touchevents/ and https://docs.sentry.io/platforms/react-native/performance/instrumentation/automatic-instrumentation/#react-profiler

src/js/tracing/routingInstrumentation.ts
krystofwoldrich: We are actually using the static value outside of the class.
https://github.com/getsentry/sentry-react-native/blob/77a228f62596771bde0e7ddc3b01257a863cab7a/src/js/tracing/reactnavigationv4.ts#L370
So I would like to keep it, as we have id and name in integrations.

src/js/touchevents.tsx
```diff
@@ -70,6 +72,15 @@ class TouchEventBoundary extends React.Component<TouchEventBoundaryProps> {
...
   };
 
+  public readonly name: string = 'TouchEventBoundary';
+
+  /**
+   * Registers the TouchEventBoundary as a Sentry Integration.
+   */
+  public componentDidMount(): void {
+    getCurrentHub().getClient()?.addIntegration?.(createIntegration(this.name));
+  }
```
marandaneto, line 82: L: I believe this could be called multiple times, we just need to be sure that we don't add duplicates to the list.

marandaneto (approved):
Left a last comment otherwise LGTM

src/js/touchevents.tsx
krystofwoldrich: I'm aware of that, I didn't put any check here, as the function checks if the integration was installed.

PR #2808, opened 2023-02-06
Update reactnativeerrorhandlers.ts

andrew-schenk:
Adds a try/catch guard to keep react-native-web from crashing due to a missing global object `ErrorUtils`

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Add in a try/catch around ErrorUtils functionality

This is to keep react-native-web from crashing. This can occur in a react-native project that is using the web version of storybook. Components and stories rendered in that environment will crash due to using `ErrorUtils` global object which does not exist in react-native-web. For iOS and Android consumers, the functionality would be unchanged.

In a storybook.js project.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/src/js/integrations/reactnativeerrorhandlers.ts
+++ b/src/js/integrations/reactnativeerrorhandlers.ts
@@ -161,69 +161,72 @@ export class ReactNativeErrorHandlers implements Integration {
     if (this._options.onerror) {
       let handlingFatal = false;
 
-      const defaultHandler =
-        ErrorUtils.getGlobalHandler && ErrorUtils.getGlobalHandler();
+      try {
+        const defaultHandler =
+          ErrorUtils.getGlobalHandler && ErrorUtils.getGlobalHandler();
 
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      ErrorUtils.setGlobalHandler(async (error: any, isFatal?: boolean) => {
-        // We want to handle fatals, but only in production mode.
-        const shouldHandleFatal = isFatal && !__DEV__;
-        if (shouldHandleFatal) {
-          if (handlingFatal) {
-            logger.log(
-              'Encountered multiple fatals in a row. The latest:',
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        ErrorUtils.setGlobalHandler(async (error: any, isFatal?: boolean) => {
+          // We want to handle fatals, but only in production mode.
+          const shouldHandleFatal = isFatal && !__DEV__;
+          if (shouldHandleFatal) {
+            if (handlingFatal) {
+              logger.log(
+                'Encountered multiple fatals in a row. The latest:',
+                error
+              );
+              return;
+            }
+            handlingFatal = true;
+          }
+
+          const currentHub = getCurrentHub();
+          const client = currentHub.getClient<ReactNativeClient>();
+          const scope = currentHub.getScope();
+
+          if (!client) {
+            logger.error(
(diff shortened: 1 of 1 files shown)


src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -161,69 +161,72 @@ export class ReactNativeErrorHandlers implements Integration {
...
-          });
-        } else {
-          // If in dev, we call the default handler anyway and hope the error will be sent
-          // Just for a better dev experience
-          defaultHandler(error, isFatal);
-        }
-      });
+          }
+        });
+      } catch () {
```
andrew-schenk, line 228:
```suggestion
      } catch(e) {
```

krystofwoldrich (changes requested):
Thank you for the PR. Could we rather than wrapping the code in try-catch, add a guard at the begging of the function checking if `ErrorUtils` exist?

```js
if (!ErrorUtils) {
    logger.warn('ErrorUtils not found. Can be caused by different environment for example react-native-web.');
	return;
}
```

Are JS errors captured when the global handler is not set? Because we do not add the web global error handlers by default in the RN SDK.

The web handlers: https://github.com/getsentry/sentry-javascript/blob/2aa4e94b036675245290596884959e06dcced044/packages/browser/src/integrations/globalhandlers.ts#L24

andrew-schenk:
> Thank you for the PR. We would like to make `react-native-web` compatible if only such a small change is necessary, but I'm not sure if it's enough.
> 
> Could we rather than wrapping the code in try-catch, add a guard at the begging of the function checking if `ErrorUtils` exist? Linting will fail if `console.log` is used, you can replace it with `logger`.
> 
> ```js
> if (!ErrorUtils) {
>     logger.warn('ErrorUtils not found. Can be caused by different environment for example react-native-web.');
> 	return;
> }
> ```
> 
> CI Build failed: https://github.com/getsentry/sentry-react-native/actions/runs/4107836949/jobs/7101832808
> 
> Are JS errors captured when the global handler is not set? Because we do not add the web global error handlers by default in the RN SDK.
> 
> The web handlers: https://github.com/getsentry/sentry-javascript/blob/2aa4e94b036675245290596884959e06dcced044/packages/browser/src/integrations/globalhandlers.ts#L24

```js
if (!ErrorUtils) {
```
will still throw an undefined error in react-native-web. Switching it up to this works.
```js
if(!global.ErrorUtils) {
```

In terms of making this library react-native-web compatible. I would say more work would need to be done. This change only makes Sentry fail silently which is helpful for my use case of putting RN components into a RNW storybook.

krystofwoldrich:
@andrew-schenk I've made some small changes, so ts and lint don't complain. Like that it would work for you?

src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -161,11 +161,17 @@ export class ReactNativeErrorHandlers implements Integration {
     if (this._options.onerror) {
       let handlingFatal = false;
 
+      const errorUtils = RN_GLOBAL_OBJ.ErrorUtils;
```
andrew-schenk, line 164: missing the import for `RN_GLOBAL_OBJ`

krystofwoldrich: It was already imported in the file.

krystofwoldrich (approved):
LGTM 🚀

Issue #2813, opened 2023-02-08
Set the sentryClientName on Native SDKs during SDK init


krystofwoldrich:
### Description

- Meta issue: https://github.com/getsentry/team-mobile/issues/48

RN should set `sentry.javascript.react-native` name on Native SDKs.

krystofwoldrich:
We also need to pass a new name to the native SDK. This needs an exposed method on Android and iOS to set the name.

PR #2814, opened 2023-02-08
chore(native): Add `react-native` postfix to native SDKs

krystofwoldrich:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Set's RN SDK name and version to the Native SDKs.

- https://github.com/getsentry/team-mobile/issues/48
- closes: https://github.com/getsentry/sentry-react-native/issues/2813
## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- ~~blocked by: https://github.com/getsentry/sentry-java/issues/2525~~
#skip-changelog


android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -102,6 +103,17 @@ Activity getCurrentActivity() {
 
     public void initNativeSdk(final ReadableMap rnOptions, Promise promise) {
         SentryAndroid.init(this.getReactApplicationContext(), options -> {
+            if (rnOptions.hasKey("_metadata")) {
```
marandaneto, line 106: Shouldn't this come from `event.sdk`? why `_metadata`?

krystofwoldrich: We are setting this during SDK init and we cant access `event`. 

`_metadata` it's the name of this property we inherit from javascript.

https://github.com/getsentry/sentry-javascript/blob/2aa4e94b036675245290596884959e06dcced044/packages/types/src/options.ts#L184

marandaneto: Gotcha, I am just trying to understand how this ends in the `event.sdk` on RN? otherwise this info isn't even set.

krystofwoldrich: Got it, from `options._metadata.sdk` it's added to `event.sdk` in the `baseclient` implementation that RN client extends.

https://github.com/getsentry/sentry-javascript/blob/2aa4e94b036675245290596884959e06dcced044/packages/core/src/envelope.ts#L24

marandaneto: so in this case the `event.sdk` is already set and when you read the `sdk` in the native layers, you could read directly from `event.sdk` and not from `event. _metadata`, right?
I'd say we could even delete `event. _metadata` from the event before writing it to the disk, it helps with serialization/deserialization/loading the file to memory performance.

krystofwoldrich: `event._metadata` doesn't exist, there is `event.sdk` from the JS, yes.
`client_options._metadata` exists and that is what we pass down to initialize the native SDKs with the same client name and version.

marandaneto: Indeed, I was somehow comparing the `event` and the `options`, of course, we use the `options` to init the SDK, my bad.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -102,6 +103,17 @@ Activity getCurrentActivity() {
...
+                @Nullable final ReadableMap metadata = rnOptions.getMap("_metadata");
+                @Nullable final ReadableMap sdkMap = metadata != null ? metadata.getMap("sdk") : null;
+                @Nullable final String sdkMapName = sdkMap != null ? sdkMap.getString("name") : null;
+                @Nullable final String sdkMapVersion = sdkMap != null ? sdkMap.getString("version") : null;
+                if (sdkMapName != null && sdkMapVersion != null) {
+                    @NonNull final SdkVersion sdkVersion = new SdkVersion(sdkMapName, sdkMapVersion);
+                    options.setSentryClientName(sdkMapName + "/" + sdkMapVersion);
+                    options.setSdkVersion(sdkVersion);
```
marandaneto, line 114: We need to be sure that this is working, I might have not tested well.
See https://github.com/getsentry/team-mobile/issues/48#issuecomment-1424684160

marandaneto:
We need to check this https://github.com/getsentry/develop/pull/832
Maybe we need to be more explicit about the SDK naming (for the native SDKs).
Maybe we are looking into `sentry.java.android.react-native` and `sentry.cocoa.react-native.` if I got it right?

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
krystofwoldrich: I've only checked that the set SDK name appears in the `event.sdk` in Sentry.

krystofwoldrich:
> We need to check this [getsentry/develop#832](https://github.com/getsentry/develop/pull/832) Maybe we need to be more explicit about the SDK naming (for the native SDKs). Maybe we are looking into `sentry.java.android.react-native` and `sentry.cocoa.react-native.` if I got it right?

Yes. I understand the same.

`sentry.java.android.react-native` with a version of the `sentry.java.android`
`sentry.cocoa.react-native` with a version of the `sentry.cocoa`

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Breaking changes
+
+- Events captured by Native SDKs are reporting as RN SDK ([#2814](https://github.com/getsentry/sentry-react-native/pull/2814))
```
marandaneto, line 7: L: I don't think we need this, it's just the internal SDK name.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
marandaneto: M: There's a simpler approach here https://github.com/getsentry/sentry-dart/pull/1320/files
Since we just need to append `.flutter` at the end of the name.

marandaneto:
@krystofwoldrich I believe we can easily fix this since it's possible to do on iOS and Android, right? you can check Flutter.

krystofwoldrich:
@marandaneto Yes.

marandaneto:
@krystofwoldrich dont forget to ping @kahest and the data team about updating the SDK family sheet.

marandaneto (approved):

PR #2826, opened 2023-02-13
feat(sdk): Add `in_foreground` to event app context

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

- https://github.com/getsentry/team-mobile/issues/36

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


src/js/integrations/devicecontext.ts
```diff
@@ -47,7 +48,14 @@ export class DeviceContext implements Integration {
         event.user = nativeUser;
       }
 
-      const nativeContext = native.context;
+      let nativeContext = native.context;
+      if (AppState.currentState !== 'unknown') {
+        nativeContext = nativeContext || {};
+        nativeContext.app = {
+          ...nativeContext.app,
+          in_foreground: AppState.currentState === 'active',
```
bitsandfoxes, line 56: Doesn't it need to be set to `"false"` at some point too? Or are you passing the `nativeContext` to the native SDKs and they set it for you?

krystofwoldrich: The `in_foreground` is merged into what I get from native SDKs. As I think the value that RN reports is important for RN devs.

If the `AppState.currentState === 'active'` => `true`, else `false`.

Thanks for the comment I've added a test to make it clear how it works.

bitsandfoxes (approved):
Awesome!

src/js/integrations/devicecontext.ts
bitsandfoxes: Ahh, I see now, gotcha. Thanks!

marandaneto, line 56: Looking at https://reactnative.dev/docs/appstate#app-states
There are edge cases for Android and iOS, wondering if we should do something about it.

krystofwoldrich: Based on this:
https://developer.android.com/guide/components/activities/activity-lifecycle#coordinating-activities
https://github.com/getsentry/sentry-java/blob/fe30606689aa4c20db711bb91bfd369169ee18f5/sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java#L105

The `sentry-java` impl behaves the same as `active` in RN.

Issue #2744, opened 2023-01-12
Issue calling Sentry.captureMessage() on iOS


TheTreek:
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [x]  iOS
- [ ]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 4.12.0

**`react-native` version:** 0.64.4

**Are you using Expo?**
- [ ]  Yes
- [x]  No

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)
- [ ]  on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

[Link to issue]

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  dsn: 'https://...@sentry.io/...'
  tracesSampleRate: 0.25,
  captureFailedRequestsEnabled: true,
  debug: true,
  enableNdkScopeSync: true,
});
```

---
**I have following issue:**

When calling `Sentry.captureMessage("Custom message", 'error')` on iOS this error is logged:
```
 ERROR  Sentry Logger [error]: Error while sending event: [Error: Exception in HostFunction: Malformed calls from JS: field sizes are different.

[[42,81],[7,0],[[146,100,1673548149834,false]],379]]
```
I was able to determine this is happening @sentry/react-native calls `RNSentry.captureEnvelope()` (node_modules/@sentry/react-native/dist/js/wrapper.js:72). Looks like a react-native bridging issue.

**Steps to reproduce:**
- Call Sentry.captureMessage("Custom message here", 'error') on iOS
- See that it fails to capture the message.

**Actual result:**
```
ERROR  Sentry Logger [error]: Error while sending event: [Error: Exception in HostFunction: Malformed calls from JS: field sizes are different.

[[42,81],[7,0],[[142,100,1673548954018,false]],354]]
```

**Expected result:**

[Expected result]

krystofwoldrich:
Thanks for the message.

Here is another issue with the same error and more details about it on Android.
- https://github.com/getsentry/sentry-react-native/issues/2751

krystofwoldrich:
@TheTreek How did you track down the `RNSentry.captureEnvelope()`? Did you use [this](https://stackoverflow.com/questions/55084493/malformed-calls-from-js-field-sizes-are-different)?

[...]

PR #2852, opened 2023-02-23
fix(sdk): Transfer large envelopes (tens of MB) over bridge without crash

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Arrays larger than `196603` will crash RN Bridge, params of the called function will disappear and `Malformed JS Call` error will be thrown. Since we support attachments 200KB can be easily reached. 

The implementation was tested using generated large data.

```js
const dataSize = 20e6; // 20MB
const simulatedData = new Uint8Array(dataSize);
for (let i = 0; i < dataSize; i++) {
  simulatedData[i] = i % 256;
}
scope.addAttachment({ data: simulatedData, filename: 'big.bin', contentType: 'application/octet-stream' });
```

Android simulator was sending 20MB successfully larger envelops started crashing the app (50MB and more)

iOS simulator was sending envelopes up to 60MB without problems.

Screenshot 2023-02-23 at 11 23 29

closes: https://github.com/getsentry/sentry-react-native/issues/2744

sample app, unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- [ ] Add the patch to v4.


marandaneto:
@krystofwoldrich if the size is a problem, [base64 is 1/3 bigger](https://en.wikipedia.org/wiki/Base64) than the raw array of bytes by definition.
Also, transferring a String would be a problem for binary attachments, you need to be sure that the encoding is not going to break, for example, images, etc.
Is there any other solution here?

marandaneto:
@krystofwoldrich if the size is a problem, [base64 is 1/3 bigger](https://en.wikipedia.org/wiki/Base64) than the raw array of bytes by definition.
Also, transferring a String would be a problem for binary attachments, you need to be sure that the encoding is not going to break, for example, images, etc.
Is there any other solution here?

krystofwoldrich:
@marandaneto 
The problem is [array length](https://github.com/getsentry/sentry-react-native/pull/2852#issue-1596594013) not the size in bytes itself.

I've checked how FS libraries are handling this, as they also need to transfer large binary data and they do it the same.
- https://github.com/itinance/react-native-fs/blob/master/FS.common.js#L396

marandaneto:
> @marandaneto The problem is [array length](https://github.com/getsentry/sentry-react-native/pull/2852#issue-1596594013) not the size in bytes itself.
> 
> I've checked how FS libraries are handling this, as they also need to transfer large binary data and they do it the same.
> 
> * [itinance/react-native-fs@`master`/FS.common.js#L396](https://github.com/itinance/react-native-fs/blob/master/FS.common.js?rgh-link-date=2023-03-02T09%3A43%3A38Z#L396)

The thing is that the bigger, the more the overhead of serialization/deserialization to base64.
The description and title are about bigger in size and not in length, I am a bit confused now, can you elaborate?

krystofwoldrich:
@marandaneto 

`196603` is the max array length that we can send over the bridge. Any larger array (no matter the array content) will disappear from the parameters causing the malformed JS call.

In our case, it's a bytes array so this length represents about 200 KB.

krystofwoldrich:
https://github.com/facebook/react-native/issues/36354

src/js/wrapper.ts
```diff
@@ -150,13 +152,19 @@ export const NATIVE: SentryNativeWrapper = {
...
+      newBytes.set(envelopeBytes);
+      newBytes.set(bytesItemHeader, envelopeBytes.length);
+      newBytes.set(EOL, envelopeBytes.length + bytesItemHeader.length);
+      newBytes.set(bytesPayload, envelopeBytes.length + bytesItemHeader.length + EOL.length);
+      newBytes.set(EOL, envelopeBytes.length + bytesItemHeader.length + EOL.length + bytesPayload.length);
+      envelopeBytes = newBytes;
     }
 
-    await RNSentry.captureEnvelope(envelopeBytes, { store: hardCrashed });
+    await RNSentry.captureEnvelope(fromByteArray(envelopeBytes), { store: hardCrashed });
```
marandaneto, line 167: M: Instead of passing a base64 string, why don't we pass the stringified envelope/JSON? that's how it used to be before moving to byte array.

krystofwoldrich: The main reason to move to the byte array was attachments, as those come as a byte array (or string).

Second, by assembling the byte array in JS we avoid possible mistakes in the byte length. Originally the string we send to the native layer for calculation of the byte length.

We could change to some hybrid solution of creating JSON envelopes without the lengths and only encoding attachments. The `Uint8Array` either must be copied to `int array` or encoded into a string. For large attachments 5-20 MB this takes roughly the same time.

marandaneto: If they actually take roughly the same time, we'd spare vendoring the base64 class at least (bundle size).
Your call here, just trying to ponder the trade-offs.

marandaneto:
If this has to be fixed ASAP, go for it, otherwise, I'd consider trying to get feedback on https://github.com/facebook/react-native/issues/36354 just to be sure we are doing the right thing.

src/js/wrapper.ts
krystofwoldrich: True, I've not thought about the bundle size.

krystofwoldrich:
It affects only Hermes dev builds, so I will wait for the https://github.com/facebook/react-native/issues/36354 resolution.
This fix would increase the bundle size.

krystofwoldrich:
Closing this as the bug only happens in dev env and by fixing https://github.com/facebook/react-native/issues/36354 in Hermes, there should not be any changes required on our end.

krystofwoldrich:
I've reopened this because we got a message from a customer, that the error is happening also in their Release build. 

So let's make this change, even though it means vendoring one extra library.

I've compared here different [approaches](https://github.com/krystofwoldrich/repro-large-array/blob/main/App.tsx#L32) but encoding to base64 comes out the best, second fastest in JS, and smallest size increase compared to other string encoded options also it's one method to process the data on the native side.

marandaneto:
> I've reopened this because we got a message from a customer, that the error is happening also in their Release build.
> 
> So let's make this change, even though it means vendoring one extra library.
> 
> I've compared here different [approaches](https://github.com/krystofwoldrich/repro-large-array/blob/main/App.tsx?rgh-link-date=2023-05-15T09%3A46%3A44Z#L32) but encoding to base64 comes out the best, second fastest in JS, and smallest size increase compared to other string encoded options also it's one method to process the data on the native side.

Can we ask for a minimal reproducible example before merging this? this has been tested a few times, and the outcome was only during `dev`.
Can we try to reach out again to RN/Hermes folks? I'd only do this change if we can either reproduce ourselves or get told by RN/Hermes this is indeed a problem, otherwise, we are just adding overhead to the SDK but without the full picture of it.

marandaneto:
> > I've reopened this because we got a message from a customer, that the error is happening also in their Release build.
> > So let's make this change, even though it means vendoring one extra library.
> > I've compared here different [approaches](https://github.com/krystofwoldrich/repro-large-array/blob/main/App.tsx?rgh-link-date=2023-05-15T10%3A10%3A50Z#L32) but encoding to base64 comes out the best, second fastest in JS, and smallest size increase compared to other string encoded options also it's one method to process the data on the native side.
> 
> Can we ask for a minimal reproducible example before merging this? this has been tested a few times, and the outcome was only during `dev`. Can we try to reach out again to RN/Hermes folks? I'd only do this change if we can either reproduce ourselves or get told by RN/Hermes this is indeed a problem, otherwise, we are just adding overhead to the SDK but without the full picture of it.

Any follow-up on that?

krystofwoldrich:
@marandaneto We never managed to reproduce it in release builds. 

The issue seems to be in RN itself not in Hermes, based on [this comment](https://github.com/facebook/hermes/issues/994#issuecomment-1548278990), I have in plan to revisit this try to see where the error is coming from to move this forward.

krystofwoldrich:
With the recent addition of native profiling, the SDK crashes on the byte array size limit more often than before.

Therefore we should address this.

ios/RNSentry.mm
```diff
@@ -396,16 +396,12 @@ - (NSDictionary*) fetchNativeStackFramesBy: (NSArray<NSNumber*>*)instructionsAdd
               });
 }
 
-RCT_EXPORT_METHOD(captureEnvelope:(NSArray * _Nonnull)bytes
+RCT_EXPORT_METHOD(captureEnvelope:(NSString * _Nonnull)bytes
```
lucas-zimerman, line 399: show we call `bytes` to `rawBytes to align with the android parameter name?

src/js/vendor/base64-js/fromByteArray.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+function encodeChunk(uint8: Uint8Array | number[], start: number, end: number): string {
+  let tmp;
+  const output = [];
+  for (let i = start; i < end; i += 3) {
+    tmp = ((uint8[i] << 16) & 0xff0000) + ((uint8[i + 1] << 8) & 0xff00) + (uint8[i + 2] & 0xff);
+    output.push(tripletToBase64(tmp));
+  }
+  // console.log('encodeChunk', output.length, uint8.length, start, end);
```
lucas-zimerman, line 45:
```suggestion
```

src/js/vendor/base64-js/fromByteArray.ts
```diff
@@ -0,0 +1,74 @@
...
+  return output.join('');
+}
+
+/**
+ * Converts a Uint8Array or Array of bytes into a string representation of base64.
+ */
+export function fromByteArray(uint8: Uint8Array | number[]): string {
```
lucas-zimerman, line 52: This is a nitpick, but what do you think of calling this function `ToBase64` ?
`FromByteArray` works great in this case, but since the output is a string, it is not very clear at first glance that the output will be a base64 string.

lucas-zimerman (approved):
LGTM!

yogendrajs:
@krystofwoldrich & team: Thanks for merging this, when will this be released? Do you know if there is a tag that I can use while on?

kahest:
@yogendrajs this was released with version [5.15.0](https://github.com/getsentry/sentry-react-native/releases/tag/5.15.0)

Issue #2879, opened 2023-03-09
The `abs_path` of the stack frame is `index.android.bundle` which is not a valid URL


birdofpreyru:
**OS:**
- [ ]  Windows
- [ ]  MacOS
- [x]  Linux

**Platform:**
- [ ]  iOS
- [x]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 5.1.0

**`react-native` version:** 0.71.4

**Are you using Expo?**
- [ ]  Yes
- [x]  No

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)
- [ ]  on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

[Link to issue](https://dr-pogodin-studio.sentry.io/issues/3988722801)

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  dsn: 'A_VALID_DSN_OF_SENTRY_PROJECT',
  release: 'A_VALID_RELEASE',
  dist: 'A_VALID_DIST',
});
```

---
**I have following issue:**

My React Native (v0.70.4) app project was correctly set up and working with Sentry (v4.8.0), in particular the source map upload and resolution used to work fine. Yesterday I updated my app, among other stuff I have upgraded `@sentry/react-native` v4.8.0 &rArr; v.5.1.0; and `react-native` v.0.70.4 &rArr; v0.71.4, without touching anything else related to the Sentry setup. After this the source map  resolution in new error events fails with the message:

> **Invalid Absolute Path URL**
> The `abs_path` of the stack frame is `index.android.bundle` which is not a valid URL.

I checked that my build artifacts are still correctly uploaded, and look no different than before. The `abs_path` values in the event JSON have changed though:
- For the new events, received after the upgrade, I see the values like `"abs_path":"index.android.bundle"`
- In events received before the upgrade these values were like `"abs_path":"app:///index.android.bundle"`.

Thus, in my current understanding, it is caused either by a bug or a feature in the latest releases of `@sentry/react-native`. I wasn't able to find any relevant info in the Sentry migration guide for v5, or in the trouble-shooring, general setup instructions, existing issues in the repo; thus asking for help to further troubleshoot and fix it.

andrewtremblay:
I'm seeing this as well in my react native android build.
`@sentry/react-native`: `^5.1.0`
[...]

PR #2891, opened 2023-03-13
fix(frames): remove non absolute non url frame.abs_path coming from default stack parser

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

In JS SDK `frame.abs_path` was added and that caused the `from.abs_path` to be relative. This lead to source maps failure in the product.

related changes in JS getsentry/sentry-javascript#7167

fixes: https://github.com/getsentry/sentry-react-native/issues/2879

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- [x] Add test so this won't happen again

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Remove non URL `frame.abs_path` which was causing source maps to fail ([#2876](https://github.com/getsentry/sentry-react-native/pull/2876))
+
 ### Dependencies
 
 - Bump Cocoa SDK from v8.2.0 to v8.3.0 ([#2876](https://github.com/getsentry/sentry-react-native/pull/2876))
--- a/src/js/sdk.tsx
+++ b/src/js/sdk.tsx
@@ -126,6 +126,7 @@ export function init(passedOptions: ReactNativeOptions): void {
                 ? `${appPrefix}${frame.filename}`
                 : `${appPrefix}/${frame.filename}`;
           }
+          delete frame.abs_path;
         }
         return frame;
       },
--- a/src/js/sdk.tsx
+++ b/src/js/sdk.tsx
@@ -1,13 +1,12 @@
 import type { Scope } from '@sentry/core';
 import { getIntegrationsToSetup, Hub, initAndBind, makeMain, setExtra } from '@sentry/core';
-import { RewriteFrames } from '@sentry/integrations';
 import {
   defaultIntegrations as reactDefaultIntegrations,
   defaultStackParser,
   getCurrentHub,
   makeFetchTransport,
 } from '@sentry/react';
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Remove non URL `frame.abs_path` which was causing source maps to fail ([#2876](https://github.com/getsentry/sentry-react-native/pull/2876))
```
```suggestion
- Remove non URL `frame.abs_path` which was causing source maps to fail ([#2891](https://github.com/getsentry/sentry-react-native/pull/2891))
```

src/js/sdk.tsx
```diff
@@ -126,6 +126,7 @@ export function init(passedOptions: ReactNativeOptions): void {
                 ? `${appPrefix}${frame.filename}`
                 : `${appPrefix}/${frame.filename}`;
           }
+          delete frame.abs_path;
```
marandaneto, line 129: M: `abs_path` is part of the protocol, so symbolication should work anyway, for fixing this issue is fine as of now, but I wonder if we should not fix ingestion when both `filename` and `abs_path` is provided? can you double check with the stack trace team?

marandaneto (approved):

src/js/sdk.tsx
krystofwoldrich: [Relay does exactly that](https://github.com/getsentry/relay/blob/1e470ec87aee8fb9214844a1cf5eaa5257ea3d69/relay-general/src/store/normalize/stacktrace.rs#L27-L29) it copies it from filename, so until we need the two fields in the SDK I think we can save a bit the envelope size and send only the filename.

krystofwoldrich: The reason why it didn't work is becuae the default frame parse now copies the name before the changes needed for RN. 

I've talked to the web team and I still don't exactly understand what they use it for, but we can adjust that later.

krystofwoldrich: So it's used for identifying new Debug Meta Image Type Source Maps, I will have to talk more to the web team on how it works.

https://develop.sentry.dev/sdk/event-payloads/debugmeta/#source-map-images

krystofwoldrich:
@marandaneto I've added tests for the rewrite frames. Could you review it too?

src/js/sdk.tsx
marandaneto: alright, let's create a repo maintenance issue for that, so we don't forget it.

marandaneto (approved):
I liked the approach of splitting up the `RewriteFrames` and making it testable + adding unit tests for multiple cases, thank you!

src/js/sdk.tsx
krystofwoldrich: https://github.com/getsentry/sentry-react-native/issues/2892

Issue #2884, opened 2023-03-10
Extend Device Context with CPU info


kahest:
Our [device context docs](https://develop.sentry.dev/sdk/event-payloads/contexts/#device-context) describe a few new CPU related fields which are not provided by the mobile SDKs yet.

- `processor_count`
- `cpu_description`
- `processor_frequency`

Need to write the native parts to get the data (or reproduce from https://github.com/getsentry/sentry-java/issues/2561 and https://github.com/getsentry/sentry-cocoa/issues/2740).

krystofwoldrich:
Types are already present in JS SDK so RN can use them right away.
- https://github.com/getsentry/sentry-javascript/blob/2aa4e94b036675245290596884959e06dcced044/packages/types/src/context.ts#L49

krystofwoldrich:
@philipphofmann @brustolin Is there a `sentry-cocoa` release coming soon so we can unblock this?

- https://github.com/getsentry/sentry-cocoa/pull/2855

PR #2984, opened 2023-04-19
feature(ios): Add processor count to device info

krystofwoldrich:

- [x] New feature

## :scroll: Description

Add CPU count from iOS. Android works already since the version it was added to `sentry-java` no changes in RN are needed.

- closes: https://github.com/getsentry/sentry-react-native/issues/2884

sample app

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


ios/RNSentry.mm
```diff
@@ -207,6 +207,19 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
         }
     }];
 
+    NSDictionary<NSString *, id> *extraContext = [PrivateSentrySDKOnly getExtraContext];
```
marandaneto, line 210: Is that a new method? we don't have that on Flutter, wondering if we should.

ios/RNSentry.mm
```diff
@@ -207,6 +207,19 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
...
+
+    NSMutableDictionary<NSString *, NSDictionary<NSString *, id> *> *deviceContext = [contexts[@"context"][@"device"] mutableCopy];
+    [deviceContext addEntriesFromDictionary:extraContext[@"device"]];
```
marandaneto, line 213: M: What if `extraContext` is null? Or what if `extraContext` does not have `device`, does ObjC handle that automatically or would that throw?

ios/RNSentry.mm
krystofwoldrich: Yes, it's a new method, [added here](https://github.com/getsentry/sentry-cocoa/pull/2855/files#diff-127399cc36e15a55aa8f32df346b6407d4e9faab0cf65b94188cbd8ba2c9bbed).

I believe flutter should have it too.

ios/RNSentry.mm
krystofwoldrich: Currently, these two keys are always there but I agree it's not safe.

marandaneto: Check https://github.com/getsentry/sentry-dart/pull/1402
I always check if there's a value and it's the type I expect.
https://en.wikipedia.org/wiki/Defensive_programming

marandaneto (approved):

PR #2992, opened 2023-04-20
chore(internal): Add User Interaction Tracing as Integration

krystofwoldrich:

- [x] Enhancement

## :scroll: Description

Adds dummy integration so we can track adoption.

integration tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
#skip-changelog

### DIFF
--- a/test/client.test.ts
+++ b/test/client.test.ts
@@ -564,7 +564,7 @@ describe('Tests ReactNativeClient', () => {
     });
   });
 
-  describe('register enabled user interactions tracing as integrations', () => {
+  describe('user interactions tracing as integrations', () => {
     test('register user interactions tracing', () => {
       const client = new ReactNativeClient(
         mockedOptions({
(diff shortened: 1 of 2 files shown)


test/client.test.ts
```diff
@@ -563,6 +563,24 @@ describe('Tests ReactNativeClient', () => {
       expect(client.getIntegrationById('MockRoutingInstrumentation')).toBeTruthy();
     });
   });
+
+  describe('register enabled user interactions tracing as integrations', () => {
```
marandaneto, line 567: M: Can we add a test that we don't add an integration if it's not installed?

marandaneto (approved):
Left a [comment](https://github.com/getsentry/sentry-react-native/pull/2992#discussion_r1173397902) otherwise LGTM.

test/client.test.ts
krystofwoldrich: Thanks. Will add that before merging.

Issue #2973, opened 2023-04-17
NSInvalidArgumentException -[NSTaggedPointerString sentry_sanitize]: unrecognized selector sent to instance 


OskarAtJoint:
**OS:**
- [ ]  Windows
- [ ]  MacOS
- [ ]  Linux

**Platform:**
- [X]  iOS
- [ ]  Android

**SDK:**
- [X]  `@sentry/react-native` (>= 1.0.0) 5.3.0
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 0.0.0

**`react-native` version:** 0.0.0

**Are you using Expo?**
- [ ]  Yes
- [X]  No

**Are you using sentry.io or on-premise?**
- [ ]  sentry.io (SaaS)
- [ ]  on-premise
---
**I have following issue:**

Sentry native crash if data is not a "record"
eg:
Sentry.addBreadcrumb({ data: 123 })
@sentry/react-native 5.3.0

OK its not valid type but it shouldnt crash!
it works in eg 4.2.3 (no crash)

**Steps to reproduce:**
Do Sentry.addBreadcrumb({ data: 123 })

**Actual result:**

2023-04-17 13:50:06.847799+0200 app[36380:3420582] -[__NSCFNumber sentry_sanitize]: unrecognized selector sent to instance 0xb6db535c86b9f994
2023-04-17 13:50:06.869393+0200 app[36380:3420582] [native] Exception '-[__NSCFNumber sentry_sanitize]: unrecognized selector sent to instance 0xb6db535c86b9f994' was thrown while invoking addBreadcrumb on target RNSentry with params (
        {
        data = 123;
        level = info;
        timestamp = "1681732206.23";
    }
)

**Expected result:**

No crash!

krystofwoldrich:
Hi,
thank you for the message, 

yes, I agree, it's a wrong type but we should not crash on it.

PR #2997, opened 2023-04-21
fix(breadcrumbs): Always convert breadcrumb data to a valid object

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

- closes: https://github.com/getsentry/sentry-react-native/issues/2973

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


src/js/breadcrumb.ts
```diff
@@ -34,7 +34,7 @@ export function breadcrumbFromObject(candidate: BreadcrumbCandidate): Breadcrumb
   if (typeof candidate.timestamp === 'string') {
     const timestamp = Date.parse(candidate.timestamp);
     if (!isNaN(timestamp)) {
-      breadcrumb.timestamp = timestamp;
+      breadcrumb.timestamp = timestamp / 1000; // breadcrumb timestamp is in seconds
```
marandaneto, line 37: M: Shouldn't this be done during `Date.parse`?

src/js/scope.ts
```diff
@@ -59,12 +60,22 @@ export class ReactNativeScope extends Scope {
...
     };
-    NATIVE.addBreadcrumb(mergedBreadcrumb);
-    return super.addBreadcrumb(mergedBreadcrumb, maxBreadcrumbs);
+
+    const countBefore = this._breadcrumbs.length;
+    super.addBreadcrumb(mergedBreadcrumb, maxBreadcrumbs);
+    const countAfter = this._breadcrumbs.length;
+
+    const breadcrumbAdded = countAfter > countBefore;
+    if (breadcrumbAdded) {
```
marandaneto, line 74: I don't follow why we need the `breadcrumbAdded` condition and `countBefore`/`countAfter`, mind elaborating?

src/js/utils/normalize.ts
```diff
@@ -0,0 +1,15 @@
+import { normalize } from '@sentry/utils';
+
+/**
+ * Converts any input into a valid record with string keys.
+ */
+export function convertToNormalizedObject(wat: unknown): Record<string, unknown> {
+  const normalized: unknown = normalize(wat);
+  if (typeof normalized !== 'object') {
+    return {
+      unknown: normalized,
```
marandaneto, line 10: L: if `unknown` is the new key in case it's not a Key/Value obj, we can use `value` instead similar to Java for `Contexts`.

marandaneto: Or `unknown` is a type here?

src/js/utils/normalize.ts
```diff
@@ -0,0 +1,15 @@
+import { normalize } from '@sentry/utils';
+
+/**
+ * Converts any input into a valid record with string keys.
+ */
+export function convertToNormalizedObject(wat: unknown): Record<string, unknown> {
```
marandaneto, line 6:
```suggestion
export function convertToNormalizedObject(data: unknown): Record<string, unknown> {
```

M: Always prefer more meaningful names.

src/js/wrapper.ts
```diff
@@ -381,7 +381,6 @@ export const NATIVE: SentryNativeWrapper = {
       ...breadcrumb,
       // Process and convert deprecated levels
       level: breadcrumb.level ? this._processLevel(breadcrumb.level) : undefined,
-      data: breadcrumb.data ? normalize(breadcrumb.data) : undefined,
```
marandaneto, line 384: Did we move this somewhere?

src/js/utils/normalize.ts
marandaneto, line 6: L: Did you check if the JS SDK always has this? sounds like possible.

src/js/breadcrumb.ts
marandaneto: M: Should this be a different PR or get more elaborated in the changelog?

src/js/utils/normalize.ts
krystofwoldrich: `unknown` is a key here. Sure let's use `value` instead, I would love to keep this consistent.

src/js/utils/normalize.ts
krystofwoldrich: It's used in JS a lot to name variables that you don't know what will be inside. 
https://github.com/getsentry/sentry-javascript/blob/2aa4e94b036675245290596884959e06dcced044/packages/utils/src/is.ts#L16

src/js/wrapper.ts
krystofwoldrich: Yes, it's now here.
https://github.com/getsentry/sentry-react-native/blob/699a4f21602f14c6edc6ab9ae42d0e430e9e27c5/src/js/scope.ts#L66

src/js/utils/normalize.ts
marandaneto: That is fine, but it always comes from `breadcrumb.data`, so lets call `data`?

krystofwoldrich: I've checked, as there are a lot of these kinds of util methods, but nothing handy for this.

src/js/scope.ts
krystofwoldrich: Thanks for the catch, it's not needed. The oldest breadcrumb gets removed if the max is reached.

marandaneto (approved):

Issue #2668, opened 2022-12-05
Profiling for RN


krystofwoldrich:
This issue tracks all related tasks to Sentry React Native Profiling support.

```[tasklist]
### In Development
- [ ] https://github.com/getsentry/sentry-react-native/pull/3349
- [ ] https://github.com/getsentry/sentry-react-native/pull/3397
- [ ] https://github.com/getsentry/relay/pull/2706
- [ ] https://github.com/getsentry/sentry/issues/59989
- [ ] https://github.com/getsentry/sentry/pull/58771
- [ ] https://github.com/getsentry/relay/pull/2685
- [ ] https://github.com/getsentry/sentry-react-native/issues/3641
``` 

```[tasklist] 
### Alpha since https://github.com/getsentry/sentry-react-native/releases/tag/5.8.0
- [x] Mark as alpha
- [x] Release SDK with code snippet in the changelog
- [ ] https://github.com/getsentry/sentry-docs/pull/7228
```

```[tasklist]
### In Progress
- [ ] https://github.com/getsentry/sentry-react-native/pull/3057
- [ ] https://github.com/getsentry/sentry-javascript/pull/8270
- [ ] https://github.com/getsentry/sentry-react-native/pull/3141
```

```[tasklist]
### Planned
- [ ] https://github.com/getsentry/sentry-react-native/issues/3337
- [x] Investigate HBC Bundle Profile symbolication with and without debug info
```

[Private Notion page link.](https://www.notion.so/sentry/RN-Profiling-5c5e43b91e8145ae9393fdcbaedd3771)

---

This issue serves to gather interest and comments regarding Profiling for RN. 

More about the Profiling product [here](https://docs.sentry.io/product/profiling/).

Android and iOS support profiling therefore RN native could offer an option to turn it on in the native layers.

Thumbs up if you are interested.

marandaneto:
Right now that would be possible if you [init](https://docs.sentry.io/platforms/react-native/manual-setup/native-init/) the SDK manually.

Enable Performance and Profiling directly on the Native SDKs, for example, Android.
https://docs.sentry.io/platforms/android/performance/
https://docs.sentry.io/platforms/android/profiling/

The same steps for iOS, would work for Android and iOS native code only, not in the JS bits nor C bits for Android.

krystofwoldrich:
Linking opened Profiling discussion.

- https://github.com/getsentry/sentry-react-native/discussions/3009

[...]

PR #3057, opened 2023-05-09
feat(profiling): Add Hermes JS Profiling (experimental)

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

Related to: https://github.com/getsentry/sentry-react-native/issues/2668
Doesn't close the issue as this is only the first step. Native profiling will follow.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -615,6 +625,46 @@ public void disableNativeFramesTracking() {
         }
     }
 
+    public void startProfiling(Promise promise) {
+        final WritableMap result = new WritableNativeMap();
+        try {
+            RNSentryModuleImpl.javaScriptExecutorFactory.startSamplingProfiler();
+        } catch (UnsupportedOperationException e) {
```
marandaneto, line 632:
```suggestion
        } catch (Throwable e) {
```

Just for the sake of safety.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -615,6 +625,46 @@ public void disableNativeFramesTracking() {
...
+            String error = javaScriptExecutorFactory.toString() + "does not support Sampling Profiler";
+            logger.log(SentryLevel.ERROR, error);
+            result.putString("error", error);
```
marandaneto, line 635: L: If we log the error in the Native layer, we probably don't need to send the error back?

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -615,6 +625,46 @@ public void disableNativeFramesTracking() {
...
+        }
+        promise.resolve(result);
+    }
+
+    public void stopProfiling(Promise promise) {
+        final WritableMap result = new WritableNativeMap();
+        try {
```
marandaneto, line 642: M: Better to use try with resources to release the buffers automatically.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -615,6 +625,46 @@ public void disableNativeFramesTracking() {
...
+            final File output = File.createTempFile(
+        "sampling-profiler-trace", ".cpuprofile", reactApplicationContext.getCacheDir());
+            RNSentryModuleImpl.javaScriptExecutorFactory.stopSamplingProfiler(output.getPath());
+
+            StringBuilder text = new StringBuilder();
```
marandaneto, line 647: L: Would make sense to use `StringBuffer` as well?

android/src/main/java/io/sentry/react/RNSentryPackage.java
```diff
@@ -17,7 +18,7 @@ public class RNSentryPackage extends TurboReactPackage {
   @Override
   public NativeModule getModule(String name, ReactApplicationContext reactContext) {
     if (name.equals(RNSentryModuleImpl.NAME)) {
-      return new RNSentryModule(reactContext);
+      return new io.sentry.react.RNSentryModule(reactContext);
```
marandaneto, line 21: L: Why do we need to add the package import here? do we have a clash?

ios/RNSentry.mm
```diff
@@ -498,6 +503,46 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
...
 
+static NSString* const enabledProfilingMessage = @"Use SENTRY_PROFILING_ENABLED env to build with enabled profiling";
+
+RCT_EXPORT_SYNCHRONOUS_TYPED_METHOD(NSDictionary *, startProfiling)
+{
+#ifdef SENTRY_PROFILING_ENABLED
+    try {
+        facebook::hermes::HermesRuntime::enableSamplingProfiler();
+        return @{ @"started": @YES };
+    } catch (const std::exception& ex) {
```
marandaneto, line 514: I'm not entirely sure how `std::exception` works on iOS but it's not used in Cocoa, the only reference is:
https://github.com/getsentry/sentry-cocoa/blob/b9b0f0a4b268f1ed4b30104486d690931dab63e1/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_CPPException.cpp#L134
I know that exceptions don't work that well in iOS, maybe better to check with @brustolin or @philipphofmann here if this is necessary or if there's any other way to guard this function.

sample-new-architecture/android/app/src/main/java/com/samplenewarchitecture/MainApplication.java
```diff
@@ -1,21 +1,32 @@
...
 import com.facebook.react.ReactPackage;
+import com.facebook.react.bridge.JavaScriptExecutor;
+import com.facebook.react.bridge.JavaScriptExecutorFactory;
+import com.facebook.react.common.JavascriptException;
 import com.facebook.react.defaults.DefaultNewArchitectureEntryPoint;
 import com.facebook.react.defaults.DefaultReactNativeHost;
 import com.facebook.soloader.SoLoader;
 import java.util.List;
 
+import javax.annotation.Nullable;
```
marandaneto, line 18: M: On Android, we use the JetBrains annotations, let's stick to this one.

sample-new-architecture/android/app/src/main/java/com/samplenewarchitecture/MainApplication.java
```diff
@@ -1,21 +1,32 @@
...
+
+import io.sentry.react.RNSentryModuleImpl;
 import io.sentry.react.RNSentryPackage;
 
 public class MainApplication extends Application implements ReactApplication {
 
+  private final JavaScriptExecutorFactory mJavaScriptExecutorFactory = new HermesExecutorFactory();
```
marandaneto, line 25:
```suggestion
  private final JavaScriptExecutorFactory javaScriptExecutorFactory = new HermesExecutorFactory();
```

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -68,6 +72,8 @@ public class RNSentryModuleImpl {
 
     public static final String NAME = "RNSentry";
 
+    private static JavaScriptExecutorFactory javaScriptExecutorFactory = null;
```
marandaneto, line 75: L: is there a way to not make it static?

sample-new-architecture/ios/sampleNewArchitecture.xcodeproj/project.pbxproj
```diff
@@ -523,6 +523,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = "org.reactjs.native.example.$(PRODUCT_NAME:rfc1034identifier)";
 				PRODUCT_NAME = sampleNewArchitecture;
 				RCT_NEW_ARCH_ENABLED = 1;
+				SENTRY_PROFILING_ENABLED = 1;
```
marandaneto, line 526: Will that be required in People's apps?

src/js/profiling/cache.ts
```diff
@@ -0,0 +1,75 @@
+import type { ThreadCpuProfile } from './types';
+
+/**
+ * Creates a cache that evicts keys in fifo order
+ * @param size {Number}
+ */
+// sentry-javascript/packages/browser/src/profiling/cache.ts
+// https://github.com/getsentry/sentry-javascript/blob/bf8c0551e2a969083199f014d07ac2a617e323fc/packages/browser/src/profiling/cache.ts#L7
```
marandaneto, line 8: Can we reuse the JS class instead?

marandaneto:
@krystofwoldrich could some of the TS types live in/share with the JS SDK? it seels like some of them are duplicate/can be reused.

sample-new-architecture/ios/sampleNewArchitecture.xcodeproj/project.pbxproj
krystofwoldrich: This was just from testing. I've removed it.
People will need to set `SENTRY_PROFILING_ENABLED=1` when installing pods. This will add compiler flag and the profiling code will be included in the build.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
krystofwoldrich: I need to update this code, this was from the first try when I thought we will need the `javaScriptExecutorFactory` to use it's start and stop profiling methods.

sample-new-architecture/ios/sampleNewArchitecture.xcodeproj/project.pbxproj
marandaneto: I see, wondering if there's a way to avoid this, let's pair up.

ios/RNSentry.mm
brustolin: Objc try/catch not that good, but sometimes is not possible to avoided it.
But this is C++ code, and I dont think this try/catch will be a problem.

marandaneto: Gotcha, just in case, a bit of basic guidelines https://learn.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-170#basic-guidelines

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
krystofwoldrich: Make sense to log only one, let's do it in the JS layer for users' convenience, so it's easier to see the log.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
krystofwoldrich: Do you mean for the file name?

ios/RNSentry.mm
krystofwoldrich: This code can throw an exception if Hermes was compiled without a profiler.

https://github.com/krystofwoldrich/hermes/blob/4bcd62206c672c5b9674cab1399931ae8c275829/API/hermes/hermes.cpp#L1070-L1076

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
marandaneto: Nope, I meant instead `StringBuilder` sine you are using Buffer everywhere.

marandaneto: If you dont need thread safety, `StringBuilder` is fine and faster.

android/src/main/java/io/sentry/react/RNSentryPackage.java
krystofwoldrich: Thanks, just a typo, removed.

krystofwoldrich:
Linking related JS PR
- https://github.com/getsentry/sentry-javascript/pull/8270

src/js/profiling/cache.ts
krystofwoldrich: Yes, that's why I left the TODO comment. [PR here](https://github.com/getsentry/sentry-javascript/pull/8270).

sample-new-architecture/src/App.tsx
```diff
@@ -85,6 +85,9 @@ Sentry.init({
   // otherwise they will not work.
   // release: 'myapp@1.2.3+1',
   // dist: `1`,
+  _experiments: {
+    profileSampleRate: 1,
```
JonasBa, line 89: Just to be consistent with tracing and browser/node profiling - we pluralize these to `tracesSampleRate` and `profilesSampleRate`

src/js/profiling/convertHermesProfile.ts
```diff
@@ -0,0 +1,94 @@
...
+    return {
+      stack_id: hermesSample.sf,
+      thread_id: hermesSample.tid,
+      elapsed_since_start_ns: elapsed_since_start_ns.toFixed(0),
+    };
+  });
+
+  const hermesFrameChildToParentMap = new Map<Hermes.StackFrameId, Hermes.StackFrameId>();
+  const framesMap = new Map<string, ThreadCpuFrame>();
+  for (const key in hermesProfile.stackFrames) {
```
JonasBa, line 41: you can early return here if Map.has(key) and skip parsing duplicate frames

src/js/profiling/convertHermesProfile.ts
```diff
@@ -0,0 +1,94 @@
...
+import { parseHermesStackFrameName } from './hermes';
+import type { RawThreadCpuProfile } from './types';
+
+const UNKNOWN_FRAME_FUNCTION_NAME = '<unknown>';
+const UNKNOWN_STACK_ID = -1;
+const JS_THREAD_NAME = 'JavaScriptThread';
+const JS_THREAD_PRIORITY = 1;
+
+/**
+ * Converts a Hermes profile to a Sentry profile.
```
JonasBa, line 14: Can you describe roughly how the conversion is done? There is a lot happening in this function that might not be immediately obvious to the reader

src/js/profiling/integration.ts
```diff
@@ -0,0 +1,190 @@
...
+      typeof this._currentProfileTimeout === 'number' && clearTimeout(this._currentProfileTimeout);
+      this._finishCurrentProfile();
+
+      const shouldStartProfiling = this._shouldStartProfiling(transaction);
+      if (!shouldStartProfiling) {
+        return;
+      }
+
+      this._startNewProfile(transaction);
+      setTimeout(
```
JonasBa, line 64: I dont think you are ever setting this._currentProfileTimeout?

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
krystofwoldrich: I'll keep `StringBuilder` as we don't need thread safety in this case.

src/js/profiling/integration.ts
krystofwoldrich: Thank you, that's a good catch. I will also add a test for it.

sample-new-architecture/src/App.tsx
krystofwoldrich: Thanks, I've overlooked to Java and there it is in singular.

src/js/profiling/convertHermesProfile.ts
krystofwoldrich: `hermesStacks` is a set, it contains leafs of the hermes stack tree (root is main func call).

As the leafs are unique, looping over them and walking up the tree to the roor will generate unique stack only.

Or maybe I've overlooked something?

src/js/profiling/convertHermesProfile.ts
```diff
@@ -0,0 +1,94 @@
...
+  for (const jsThreadId of jsThreads) {
+    thread_metadata[jsThreadId] = {
+      name: JS_THREAD_NAME,
+      priority: JS_THREAD_PRIORITY,
+    };
+  }
+
+  return {
+    samples,
+    frames: Array.from(framesMap.entries())
```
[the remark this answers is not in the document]
krystofwoldrich: I've added `hermesStackFrameIdToSentryFrameIdMap`. Since looping over the keys doesn't have to be ordered.

krystofwoldrich:
Thank you for all the comments, I've fixed the code, could you review it again?

Also, check the jsDocs for the mapping algorithm if it's understandable.

krystofwoldrich:
Sanity check that the conversion to Sentry Profile format works correctly.
Profile uploaded to Sentry on the left. Raw Hermes Profile converted to the Chrome Dev Tools format (no source maps, no bundle) on the left.

Screenshot 2023-06-20 at 16 58 12

src/js/profiling/hermes.ts
```diff
@@ -0,0 +1,91 @@
...
+  fileName?: string;
+}
+
+const FILE_NAME_REGEX = /([a-zA-Z0-9_-]+\.[a-zA-Z0-9_-]+)(?::(\d+:\d+))?\)/;
+/**
+ * Hermes Profile Stack Frame Name contains function name and file path.
+ *
+ * `foo(/path/to/file.js:1:2)` -> `{ function: 'foo', fileName: 'file.js' }`
+ */
+export function parseHermesStackFrameName(name: string): HermesStackFrameNameParsed {
```
JonasBa, line 57: I have a hunch this will be perf sensitive. Have you done any benchmarking on this?

src/js/profiling/convertHermesProfile.ts
```diff
@@ -0,0 +1,160 @@
...
+  const { frames, hermesStackFrameIdToSentryFrameIdMap } = mapFrames(hermesProfile.stackFrames);
+
+  const { stacks, hermesStackToSentryStackMap } = mapStacks(
+    hermesStacks,
+    hermesProfile.stackFrames,
+    hermesStackFrameIdToSentryFrameIdMap,
+);
+
+  for (const sample of samples) {
+    sample.stack_id = hermesStackToSentryStackMap.get(sample.stack_id) ?? UNKNOWN_STACK_ID;
```
JonasBa, line 41: Why do you need nullish coalescing here? Is it for safety? If you are not sure of the exact case of when it might happen then I would rather elect to throw (especially during beta phase) so that we can find cases where this happens and add better handling/comments for why we need the fallback

src/js/profiling/hermes.ts
```diff
@@ -0,0 +1,91 @@
...
+export interface HermesStackFrameNameParsed {
+  function?: string;
+  fileName?: string;
+}
+
+const FILE_NAME_REGEX = /([a-zA-Z0-9_-]+\.[a-zA-Z0-9_-]+)(?::(\d+:\d+))?\)/;
+/**
+ * Hermes Profile Stack Frame Name contains function name and file path.
+ *
+ * `foo(/path/to/file.js:1:2)` -> `{ function: 'foo', fileName: 'file.js' }`
```
JonasBa, line 55: does line and column info correspond to call site or function definition?

src/js/profiling/hermes.ts
krystofwoldrich: I didn't do benchmarking on this, I will take a look on it.

src/js/profiling/hermes.ts
krystofwoldrich: call site - but we only parse the func name and the path

Can you share a bit more context, could it be a trouble somewhere?

JonasBa: No, just something worth knowing tbqh. Some profilers can output line/col of where a fn is defined - this can have an impact on the interpretation of the charts/graphs as well as reduce payload size.

src/js/profiling/hermes.ts
JonasBa: Yeah, I wouldn't worry too much about it, def a topic for a different PR. Just something to be mindful about as we might be parsing a lot of frames here

src/js/profiling/convertHermesProfile.ts
krystofwoldrich: I've changed it here.
https://github.com/getsentry/sentry-react-native/blob/bfe91bbf99ddaeb333e3a50121057fd3388ca7c4/src/js/profiling/convertHermesProfile.ts#L43

krystofwoldrich: I know throwing an error would be better to get to know about the issue, but it will also crash the app, so `logger.error` feels like a good middle ground.

src/js/profiling/hermes.ts
krystofwoldrich: The original code with regex takes ~10ms on my simulator. The new code without the regex takes ~4ms. Thanks for pointing this out.

src/js/profiling/convertHermesProfile.ts
marandaneto: https://develop.sentry.dev/sdk/philosophy/#degrade-gracefully
I'd not throw either.

krystofwoldrich:
@marandaneto @JonasBa I've addressed the comments. The PR is ready for a final review.

JonasBa (approved):

krystofwoldrich:
@JonasBa Thanks. The CI just timed out, re-run should fix it.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -79,6 +82,7 @@ public class RNSentryModuleImpl {
     private final PackageInfo packageInfo;
     private FrameMetricsAggregator frameMetricsAggregator = null;
     private boolean androidXAvailable;
+    private boolean debugEnabled = false;
```
marandaneto, line 85: You can just check that directly from `HubAdapter().options.isDebug` or something similar.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -616,6 +621,48 @@ public void disableNativeFramesTracking() {
...
+
+            final File output = File.createTempFile(
+                "sampling-profiler-trace", ".cpuprofile", reactApplicationContext.getCacheDir());
+
+            if (this.debugEnabled) {
+                logger.log(SentryLevel.INFO, "Profile saved to: " + output.getAbsolutePath());
+            }
+
+            try (final BufferedReader br = new BufferedReader(new FileReader(output));) {
+                HermesSamplingProfiler.dumpSampledTraceToFile(output.getPath());
```
marandaneto, line 649: Is there a way to avoid the IO and read directly from HermesSamplingProfiler?

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -616,6 +621,48 @@ public void disableNativeFramesTracking() {
...
+        return result;
+    }
+
+    public WritableMap stopProfiling() {
+        final WritableMap result = new WritableNativeMap();
+        try {
+            HermesSamplingProfiler.disable();
+
+            final File output = File.createTempFile(
+                "sampling-profiler-trace", ".cpuprofile", reactApplicationContext.getCacheDir());
```
marandaneto, line 642: If we always use the same file, are we overwriting and/or deleting the old files? otherwise, it might be better to use a generated name to avoid overwriting/using the wrong one.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -616,6 +621,48 @@ public void disableNativeFramesTracking() {
...
+                String line;
+                while ((line = br.readLine()) != null) {
+                    text.append(line);
+                    text.append('\n');
+                }
+
+                result.putString("profile", text.toString());
+            }
+        } catch (Throwable e) {
+            result.putString("error", e.toString());
```
marandaneto, line 661: Should we just log in to the native plugins instead of returning the stringified error avoiding the back-and-forth of messages in the native bridges?

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -616,6 +621,48 @@ public void disableNativeFramesTracking() {
...
+                HermesSamplingProfiler.dumpSampledTraceToFile(output.getPath());
+
+                final StringBuilder text = new StringBuilder();
+                String line;
+                while ((line = br.readLine()) != null) {
+                    text.append(line);
+                    text.append('\n');
+                }
+
+                result.putString("profile", text.toString());
```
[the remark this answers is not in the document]
krystofwoldrich: It will be slower because we will have to convert the `byte array` to `RN WritableArray` and then in JS first from `int array` to `byte array` and then to `string` and then parse the JSON.

Right now we just pass the string and in JS we only convert JSON which is super fast (JSON conversion is a native C++ function of the JS VM).

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
krystofwoldrich: We are overwriting the file, but since there is only one sampling profiler and always when we stop we read the data and send them to JS, it should not be a problem.

marandaneto: Can we delete the file in the finally block at least?

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
krystofwoldrich: At the moment the interface has only dumpToFileExported. 
https://github.com/facebook/react-native/blob/4d40cebe4017509098d889fdbb75761e895a21e3/packages/react-native/ReactAndroid/src/main/java/com/facebook/hermes/instrumentation/HermesSamplingProfiler.java#L13-L32

We could make our own wrapper around the C++ Hermes static library, but it would require more patching of the apps to ensure that our cpp code compiles. At the moment we don't have any Cpp dependencies as we use the Java wrapper that is part of RN.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
krystofwoldrich: From developer experience, at the moment people will see the error right away in the Metro DevServer logs, but yes, we keep the native bridge a bit more busy.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
marandaneto: Gotcha, lets keep it as it is now, can we open a feature request in the RN repo instead? I believe this would he beneficial for future profilers, it's leas overhead to transform the profile into something i stead dumping to a file.

krystofwoldrich: Yes, definitely, a feature request on the RN repo makes sense.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -616,6 +619,57 @@ public void disableNativeFramesTracking() {
...
+            }
+        } catch (Throwable e) {
+            result.putString("error", e.toString());
+        } finally {
+            if (output != null && !isDebug) {
```
marandaneto, line 663: M: The file should be deleted also if in debug mode.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -616,6 +619,57 @@ public void disableNativeFramesTracking() {
...
+                final boolean wasProfileSuccessfullyDeleted = output.delete();
```
marandaneto, line 664: H: delete can throw, wrap in a try catch as well for safety.

marandaneto (approved):
Left 2 more comments otherwise LGTM.

krystofwoldrich:
- Blocked by https://github.com/getsentry/team-mobile/issues/11

Sentry backend stopped accepting Profile and Transaction in separate envelopes on 16th June. Because of the current implementation of sending envelopes on Android, we can send Transaction and Profile in one envelope. Android opens the envelope and repacks the envelope items. This will be solved by the new way of enriching events as then Android will blindly send the envelope without unpacking it.

PR #3093, opened 2023-05-30
Fix issue with Sentry React Native SDK's enableNative variable

mateioprea:
## :loudspeaker: Type of change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This commit addresses a problem with the Sentry React Native SDK where the this.enableNative variable was not correctly reinitialized to true after reinitializing the SDK. The issue arose when calling the `closeNativeSdk()` method, which set `this.enableNative` to `false`. Without reinitializing it to true, subsequent reinitializations of the SDK would prevent stopping the SDK afterwards.

To fix this, I added a line in the initNativeSdk function to explicitly set `this.enableNative = true` after reinitializing the native SDK.

This change resolves the problem and ensures that the enableNative variable is properly managed during SDK reinitialization, allowing for the correct functioning of the Sentry React Native SDK.

## :bulb: Motivation and Context

The motivation behind this change is to ensure that the enableNative variable is correctly managed during SDK reinitialization, allowing for the proper functioning of the Sentry React Native SDK.

This change fixes the issue and ensures that the enableNative variable is reset to true after reinitializing the native SDK, thereby enabling the native SDK functionality as intended.

## :green_heart: How did you test it?

I conducted testing of this change on my local application to ensure its effectiveness. The following steps were taken during the testing process:

1. Set up a local development environment with the Sentry React Native SDK integrated into my application.
2. Created a scenario where the closeNativeSdk() method is called to disable the native SDK and subsequently attempted to reinitialize the SDK.
3. Verified that prior to the change, the enableNative variable was not reinitialized to true after reinitializing the SDK, leading to the native SDK functionality not being enabled.
4. Applied the proposed change to the codebase, which includes resetting the enableNative variable to true in the initNativeSdk function.
5. Repeated the scenario where the closeNativeSdk() method is called and then attempted to reinitialize the SDK.
6. Verified that after applying the change, the enableNative variable was correctly reinitialized to true, enabling the native SDK functionality upon reinitialization.
7. Conducted additional testing by calling the closeNativeSdk() method again after reinitializing the SDK.
8. Ensured that the native SDK was successfully stopped and disabled as expected, preventing any further native SDK functionality.
9. Ensured that the rest of the application's functionality remained unaffected by the change.
10. By following these steps and observing the expected behavior, I confirmed that the change effectively resolves the issue with the enableNative variable and allows for subsequent reinitializations of the SDK while still being able to stop and disable the native SDK when desired.

[...]

PR #3099, opened 2023-06-04
Disable enableNative if Native SDK was not initialized

### LINKED ISSUE #3027: SDK with `enableNative: true` falls back to JS if native is not available
### Description

At the moment RN SDK throws an error when `enableNative: true` with Expo Go, to make DEX better we should not enforceNative but only warn about the situation and use JS as a fallback.

This is especially an improvement for Expo users and `react-native-web` users.

~~We add `forceNative` flag to keep the original behavior for users who want it.~~

lucas-zimerman:
sets the flag `EnableNative` if `makeNativeTransportFactory` failed to be created.

Platform.OS could be 'Android' on Expo so we can not rely on that value in order to set EnableNative to true/false

Validated issues:
https://sentry-sdks.sentry.io/issues/4227272790/events/e80ef7a5799c43c69e9d69c53df24b16/?project=5428561

https://sentry-sdks.sentry.io/issues/4227268747/events/47a07470db4641ea90e5c7f124e0c623/?project=5428561

https://sentry-sdks.sentry.io/issues/4227271973/events/d94ead040c684d459b172e1458c8504e/?project=5428561

Close #3027


marandaneto:
Is this somehow a dupe of https://github.com/getsentry/sentry-react-native/pull/3093 ?

lucas-zimerman:
> Is this somehow a dupe of #3093 ?

Seems to be a different case.
on the wrapper, this.enableNative has a fixed flow, it starts as true by default, and it changes to false when closing the SDK. But once reinitialised, this flow isn't reset so `true` will stay with the previous value set that was false.

PR https://github.com/getsentry/sentry-react-native/pull/3093 fixes this flow where this PR fixes the initial value of enableNative to false if the NativeTransport isn't available.

src/js/sdk.tsx
```diff
@@ -65,17 +65,25 @@ export function init(passedOptions: ReactNativeOptions): void {
...
     ?? DEFAULT_OPTIONS.maxQueueSize;
+
+  let enableNative = passedOptions.enableNative !== undefined
+    ? passedOptions.enableNative
+    : DEFAULT_OPTIONS.enableNative;
+  // If custom transport factory fails the SDK won't initialize
+  let chosenTransport: ((transportOptions: ReactNativeTransportOptions) => Transport) | null
+    = passedOptions.transport
+    || makeNativeTransportFactory({
+      enableNative: enableNative
```
krystofwoldrich, line 76:
```suggestion
      enableNative,
```

src/js/sdk.tsx
```diff
@@ -65,17 +65,25 @@ export function init(passedOptions: ReactNativeOptions): void {
...
+    });
+  if (chosenTransport == null) {
```
krystofwoldrich, line 78: M: Could we make the decision directly based on `NATIVE.isNativeAvailable` so even if people supply their own transport the native is disabled if not available?

krystofwoldrich:
Let's add one or two tests into `sdk.test.ts`. Something to ensure that we are disabling native and maybe one for the case when people supply their own transport but native is not available.

```js
    it('falls back to fetch if native flag enabled but native not available', () => {
      (NATIVE.isNativeAvailable as jest.Mock).mockImplementation(() => false);
      init({ enableNative: true });
      expect(usedOptions()?.enableNative).toEqual(false);
      expect(NATIVE.isNativeAvailable).toBeCalled();
    });
```

Besides that, it looks good, thanks for the events examples.

PS.: I've noticed none of the stack traces is symbolicated, but I see symbolicate calls in the breadcrumbs, strange.

src/js/sdk.tsx
lucas-zimerman: I implemented that, since this only affects the value of EnableNative, I rolled back the original logic state since we no longer need the altered condition to check if enableNative should be set to true or false.

lucas-zimerman:
issue in regard to the stack trace was created here: https://github.com/getsentry/sentry-react-native/issues/3104

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Disable enableNative if Native SDK was not initialized ([#3099](https://github.com/getsentry/sentry-react-native/pull/3099))
```
```suggestion
- Disable `enableNative` if Native SDK is not available ([#3099](https://github.com/getsentry/sentry-react-native/pull/3099))
```

src/js/sdk.tsx
```diff
@@ -65,15 +65,18 @@ export function init(passedOptions: ReactNativeOptions): void {
     // eslint-disable-next-line deprecation/deprecation
     ?? passedOptions.transportOptions?.bufferSize
     ?? DEFAULT_OPTIONS.maxQueueSize;
+
+  const enableNative = passedOptions.enableNative !== undefined
+    ? passedOptions.enableNative
+    : NATIVE.isNativeAvailable();
```
krystofwoldrich, line 71: This looks good, now that it's not dependent on the transport, but the `const enableNative` should be false also in the case when `passedOptions.enableNative === true` and `NATIVE.isNativeAvailable() === false`

test/sdk.test.ts
```diff
@@ -372,7 +417,7 @@ describe('Tests the SDK functionality', () => {
         integrations: [
           <Integration>{
             name: 'HttpClient',
-            setupOnce: () => {},
+            setupOnce: () => { },
```
lucas-zimerman, line 420:
```suggestion
            setupOnce: () => {},
```

src/js/sdk.tsx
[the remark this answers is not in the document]
krystofwoldrich: Awesome, thanks.

krystofwoldrich (approved):
Looks good. Thank you! 🚀🚀

PR #3109, opened 2023-06-07
Fix Lint on utils\normalize.ts

lucas-zimerman:
Quick fix/ignore to lint warnings/errors  on the `normalize.ts` file

#skip-changelog


lucas-zimerman:
https://github.com/getsentry/sentry-react-native/actions/runs/5201833318/jobs/9382558431
image

Somehow it doesn't trigger an fail

krystofwoldrich (approved):
Thank you. Good catch.

Issue #3081, opened 2023-05-19
Add Expo RewriteFrames to `sentry-react-native` to better support Expo Bare workflow


krystofwoldrich:
### Description

We have recently added [a function](https://github.com/getsentry/sentry-react-native/pull/3050) to report if an app is running with Expo or not.

We can use it to automatically fix the bundle name for users using `sentry-react-native` directly in their Expo apps.

https://github.com/expo/sentry-expo/blob/689f78092da814557e2e268e24bc0bac8366f969/src/sentry.ts#L62-L70

This will eliminate these kinds of issues: https://github.com/getsentry/sentry-react-native/issues/3078

Issue #3078, opened 2023-05-18
JS Error is reported as Java Exception


fantasyado:
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [x]  iOS
- [ ]  Android

**SDK:**
- [x]  `@sentry/react-native` (5.5.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 0.0.0

**`react-native` version:** 0.70.5

**Are you using Expo?**
- [x]  Yes -- bare flow
- [ ]  No

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)
- [ ]  on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

[Link to issue](https://aspiraconnect.sentry.io/issues/4191461462/?alert_rule_id=1038348&alert_timestamp=1684404390608&alert_type=email&environment=production&project=5174057&referrer=alert_email)

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  dsn: 'https://...@sentry.io/...'
  // other options
  enabled: !__DEV__
});
```

---
**I have following issue:**

[Description]

**Steps to reproduce:**
- run xcode build script 
   ``` 
    xcodebuild -workspace $schemeName.xcworkspace   -scheme $schemeName -configuration Release -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 14 Pro Max,OS=16.1'  
  ```
- install the ipa to simulator and trigger a error

**Actual result:**

[Actual result]
The source map seems not used to display the error source code location.
<img width="699" alt="image" src="https://github.com/getsentry/sentry-react-native/assets/10321851/f19b1895-5baa-4d3e-80ad-86d2e2e03402">
and when I check the sourceMaps tab in sentry.io it shows no  bundle connected to this release

<img width="836" alt="image" src="https://github.com/getsentry/sentry-react-native/assets/10321851/0190793a-a789-4f73-b0e8-457da33f21f9">

**Expected result:**

[Expected result]
The source map should be used to parse the error stacktrace.

Please help on this, Thanks !

fantasyado:
update:   after I set release and dist in Sentry.init and doing 
```
export SENTRY_ORG=my-org
export SENTRY_PROJECT=my-project 
```
in the build script , the source map uploading seems normal on sentry like this 
[...]

PR #3115, opened 2023-06-13
feat(expo): Add support for Expo hash named bundles in RewriteFrames

krystofwoldrich:

- [x] Enhancement
- [x] Refactoring

## :scroll: Description

This code was part of `sentry-expo` package by moving it into `sentry-react-native` users using this package directly will have working source maps out of the box without changing the rewriteframes integration themself. 

Technically this could be a breaking change as before we kept the hash bundle name, but since then source maps are not working it will only add more for new users and will not affect current non-expo users and expo users with their own rewriteframes.    

closes: https://github.com/getsentry/sentry-react-native/issues/3081

unit tests, expo app

https://github.com/getsentry/sentry-react-native/issues/3078#issuecomment-1554208985

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/integrations/rewriteframes.ts
+++ b/src/js/integrations/rewriteframes.ts
@@ -1,28 +1,42 @@
 import { RewriteFrames } from '@sentry/integrations';
 import type { StackFrame } from '@sentry/types';
+import { Platform } from 'react-native';
+
+import { isExpo } from '../utils/environment';
 
 /**
  * Creates React Native default rewrite frames integration
  * which appends app:// to the beginning of the filename
- * and removes file://, 'address at' prefixes and CodePush postfix.
+ * and removes file://, 'address at' prefixes, CodePush postfix,
+ * and Expo bundle postfix.
  */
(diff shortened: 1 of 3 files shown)


src/js/integrations/rewriteframes.ts
```diff
@@ -1,28 +1,42 @@
...
-            frame.filename.indexOf('/') === 0 ? `${appPrefix}${frame.filename}` : `${appPrefix}/${frame.filename}`;
-        }
-        delete frame.abs_path;
+      if (frame.filename === '[native code]' || frame.filename === 'native') {
+        return frame;
       }
+
+      // Expo adds hash to the end of bundle names
+      if (isExpo()) {
+        frame.filename = Platform.OS === 'android' ? 'app:///index.android.bundle' : 'app:///main.jsbundle';
```
marandaneto, line 32: M: Since it's technically possible to run on Web/Desktop as well, better to check if `Platform.OS` is from the Apple ecosyste, and fallback to something else.

marandaneto (approved):

PR #3133, opened 2023-06-19
feat(android): 🌟 add support for React Native 0.73

peterlazar1993:
## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :bulb: Motivation and Context
Starting from React Native v0.73 , all libraries will need to be updated with namespace due to the upgrade to AGP 8
https://github.com/react-native-community/discussions-and-proposals/issues/671

OS | Implemented
-- | --
iOS | ❌
Android | ✅

## :green_heart: How did you test it?
I tested the changes on RN 0.71.11 and everything works!

## :pencil: Checklist

- [x] I reviewed submitted code
- [ ] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] All tests passing
- [x] No breaking changes

PR #3135, opened 2023-06-19
RN-0.73 change Android configs for RN 0.73 compatibility

felipemillhouse:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

React Native 0.73 will depend on Android Gradle Plugin (AGP) 8.x. This will require all the libraries to specify a namespace in their build.gradle file.

Details
I'd like to share some of the upcoming changes that will happen in 0.73, which is still a bit far, but we'd rather start earlier rather than later.

React Native 0.73 will depend on Android Gradle Plugin (AGP) 8.x, which brings a lot of improvements for Android apps but also a series of notable changes.

Most importantly:

App/Library Developers will now have to install and use [Java 17](https://developer.android.com/build/releases/gradle-plugin#jdk-17-agp)
Library Developers will have to specify a namespace in [their build.gradle file](https://developer.android.com/build/releases/gradle-plugin#namespace-dsl).
Specifically, the last change is a breaking change and will make libraries that are not specifying a namespace incompatible with React Native 0.73 (your project won't build).

Support for namespace was added in AGP 7.3.x, which ships with React Native 0.71. Libraries that published a new version with a namespace declared for 0.71 or 0.72 don't need further update. So we invite library authors to do those changes as soon as possible so by the time 0.73 is out, most of the apps are adapted.

What you need to change
Library authors will have to update their `android/build.gradle` file as follows:

```git
android {
+   namespace = "io.sentry.react"
    ...
}
```
and remove the package definition from their AndroidManifest.xml:

```git
<?xml version="1.0" encoding="utf-8"?>
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-          package="io.sentry.react">
+          >
...
</manifest>
```
Further reading
[Official Google documentation on namespaces is here](https://developer.android.com/build/configure-app-module#set-namespace).

React Native 0.73 compatibility

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -12,6 +12,7 @@ if (isNewArchitectureEnabled()) {
 }
 
 android {
+    namespace = "io.sentry.react";
     compileSdkVersion safeExtGet('compileSdkVersion', 31)
 
     defaultConfig {
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -1,5 +1,4 @@
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="io.sentry.react">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <application>


krystofwoldrich:
Thank you for the PR and for sharing the information.

We already have a PR with these changes. See https://github.com/getsentry/sentry-react-native/pull/3133.

I'll close this as a duplicate.

PR #3140, opened 2023-06-22
Filter beforeSendTransaction from the Native SDK

### LINKED ISSUE #3069: Missing filter for `beforeSendTransaction` on Native Layer
```
    const { beforeSend, beforeBreadcrumb, integrations, ...filteredOptions } = options;
    /* eslint-enable @typescript-eslint/unbound-method,@typescript-eslint/no-unused-vars */
    const nativeIsReady = await RNSentry.initNativeSdk(filteredOptions);
```
The current filter on the wrapper only applies for `beforeSend` but doesn't filter the `beforeSendTransaction`. We should filter in order to avoid issues with the Native SDK on iOS

# Proposed Solution

a function that receives the React Native options, and return a new object that only contains the valid objects, this avoids users adding custom parameters to the options and the same being sent to the Native Layer, but also prevents that new parameters doesn't get added automatically to the Native layer.

lucas-zimerman:
Includes `beforeSendTransaction` to the list of filtered items from the Native SDK.
Close #3069

TODO: tests for beforeSendTransaction and also the other cases that were ignored.

### DIFF
--- a/test/wrapper.test.ts
+++ b/test/wrapper.test.ts
@@ -145,36 +145,6 @@ describe('Tests Native Wrapper', () => {
       expect(NATIVE.enableNative).toBe(true);
     });
 
-    test('filter beforeSend when initializing Native SDK', async () => {
-      await NATIVE.initNativeSdk({
-        dsn: 'test',
-        enableNative: true,
-        autoInitializeNativeSdk: true,
-        beforeSend: jest.fn(),
-      });
-
-      expect(RNSentry.initNativeSdk).toBeCalled();
-      // @ts-ignore mock value
-      const initParameter = RNSentry.initNativeSdk.mock.calls[0][0];
-      expect(initParameter).not.toHaveProperty('beforeSend');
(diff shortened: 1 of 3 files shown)


test/wrapper.test.ts
```diff
@@ -130,6 +130,66 @@ describe('Tests Native Wrapper', () => {
...
+        autoInitializeNativeSdk: true,
+        beforeSend: jest.fn(),
+      });
+
+      expect(RNSentry.initNativeSdk).toBeCalled();
+      // @ts-ignore mock value
+      const initParameter = RNSentry.initNativeSdk.mock.calls[0][0];
+      expect(initParameter).not.toHaveProperty('beforeSend');
+      expect(NATIVE.enableNative).toBe(true);
+    });
```
krystofwoldrich, line 176: This test is here twice.

```suggestion
```

krystofwoldrich (approved):
Thank you. Looks good. 🚀

PR #3159, opened 2023-07-03
Make 'collect modules' more monorepo friendly on Android

liamjones:
- [x] Bugfix (?)
- [x] Enhancement

## :scroll: Description
Change Android's 'collect modules' gradle task to use a `$projectDir` rather than `$reactRoot` for locating where it should put the `modules.json` output file.

This _should_ more consistently put it in the correct place automatically for a monorepo. `$reactRoot` can be the app directory (like with `sample-new-architecture`) but, depending on monorepo setup, it can also be the root of the monorepo (as it is for us). This change allows it to work for both cases (and shouldn't affect non-monorepo projects).

Additionally, while I was testing this, it seemed like the `modulesPaths` for the sample app was set incorrectly (it was going above the top directory of the repo) so I've altered it.

In our monorepo setup the `modules.json` was being put in the wrong place, and thus didn't become part of the produced apk.

Manually tested in our monorepo directly by running `./gradlew assemsbleRelease` from the `android` directory of one of our apps.

Manually tested via `sample-new-architecture` in the `@sentry/react-native` repo to check I got the same APK output before and after my changes (bar the changes present inside `modules.json` due to fixing the `modulesPaths`).

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/sample-new-architecture/android/app/build.gradle
+++ b/sample-new-architecture/android/app/build.gradle
@@ -62,7 +62,7 @@ project.ext.sentryCli = [
     collectModulesScript: "../../../dist/js/tools/collectModules.js",
     modulesPaths: [
       "node_modules",
-      "../..",
+      "../node_modules",
     ],
     skipCollectModules: false,
 ]
--- a/sentry.gradle
+++ b/sentry.gradle
@@ -44,7 +44,7 @@ gradle.projectsEvaluated {
         if (reactRoot == null) {
             reactRoot = props.get("root").get() // RN 0.71 and above
         }
-        def modulesOutput = "$reactRoot/android/app/src/main/assets/modules.json"
+        def modulesOutput = "$projectDir/src/main/assets/modules.json"
         def modulesTask = null
 
         (shouldCleanUp, bundleOutput, sourcemapOutput) = forceSourceMapOutputFromBundleTask(bundleTask)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 - Use `android.namespace` for AGP 8 and RN 0.73 ([#3133](https://github.com/getsentry/sentry-react-native/pull/3133))
+- Make 'collect modules' more monorepo friendly on Android ([#3159](https://github.com/getsentry/sentry-react-native/pull/3159))
 
 ### Dependencies
 


liamjones:
Hmm... I'm also wondering if, on the iOS side, this is the most sensible default for the `collect-modules.sh`?: 

https://github.com/getsentry/sentry-react-native/blob/5.6.0/scripts/collect-modules.sh#L42

E.g. it's currently using the app's directory location (`$PWD`) as the basis of its lookup. Again, this results in it choosing the wrong directory in our monorepo. It picks the app's node_modules (which is empty) rather than the repo's node_modules.

Yes, I could manually override via `MODULES_PATHS` (an aside; should be singular?) it it already knows that sentry-react-native (and thus node_modules) exists elsewhere: 

https://github.com/getsentry/sentry-react-native/blob/5.6.0/scripts/collect-modules.sh#L31

It could potentially be changed to be relative to the script (e.g. `$thisFilePath`). Alternatively, and possibly more robust,  RN uses node itself to resolve these locations, like this:

https://github.com/facebook/react-native/blob/259a31689334968d607978c8d335c6da444cc253/packages/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/utils/PathUtils.kt#L84

Thoughts?

sample-new-architecture/android/app/build.gradle
```diff
@@ -62,7 +62,7 @@ project.ext.sentryCli = [
     collectModulesScript: "../../../dist/js/tools/collectModules.js",
     modulesPaths: [
       "node_modules",
-      "../..",
+      "../node_modules",
```
krystofwoldrich, line 65: I agree `../node_modules` makes sense to add, but `../..` should stay to pick up the `@sentry/react-native` package.

sentry.gradle
```diff
@@ -44,7 +44,7 @@ gradle.projectsEvaluated {
         if (reactRoot == null) {
             reactRoot = props.get("root").get() // RN 0.71 and above
         }
-        def modulesOutput = "$reactRoot/android/app/src/main/assets/modules.json"
+        def modulesOutput = "$projectDir/src/main/assets/modules.json"
```
krystofwoldrich, line 47: This will work for the modules, but I think the issue is with the `$reactRoot` not being set to the root of your react native application project.

Because we use `$reactRoot` at other places like https://github.com/getsentry/sentry-react-native/blob/b093f67712a98bc80b03d5f85b58b84fb16f1adb/sentry.gradle#L103 and I assume those would not work for you too?

krystofwoldrich:
Thank you for bringing this up, we definitely want to make the `sentry.gradle` mono repo friendly. I think it would be best to fix all the occurrences of the `$reactRoot`.

What do you think?

sentry.gradle
liamjones: I wonder if this is just a misconfiguration on our monorepo then?

The [`app/build.gradle` says](https://github.com/facebook/react-native/blob/main/packages/react-native/template/android/app/build.gradle#L11):

```gradle
//   The root of your project, i.e. where "package.json" lives. Default is '..'
```

Now, I read that as location 1 below but it sounds like you're saying it should be 2?

```
.
└── monorepo/ (1)
    ├── node_modules/
    │   ├── @sentry/
    │   │   └── react-native
    │   ├── react-native
    │   └── something-shared
    └── packages/
        └── an-app/ (2)
            └── node_modules/
                └── something-unique
```

Maybe I need to clarify with the RN maintainers... 🤔

krystofwoldrich: @liamjones Have you found out some new information? What tool are you using to manage/create the monorepo?

liamjones: With my altered setup I now have `react.root` and `react.entryFile` set to the default in `app/build.gradle`, as you were expecting, and just an override for `react.reactNativeDir`.

liamjones:
@krystofwoldrich now I've fixed my monorepo setup and thought about it further, I'm going to withdraw this PR.

I _could_ still make the change mentioned in https://github.com/getsentry/sentry-react-native/pull/3159#issuecomment-1618437790 but now I think the current behaviour is actually preferable. Currently it 'selects' the direct RN project's node_modules as the source to collect from, this makes sense for a default. Yes, in our case that's empty because everything has been hoisted to the root of the monorepo but our monorepo is very targeted towards producing multiple RN apps. If someone else's monorepo was more split between RN apps and the web then the RN app's dependencies wouldn't necessarily have been hoisted.

I'll raise a separate PR for the sample app config change specifically.

krystofwoldrich:
@liamjones Thank you for all the details and explanation of your setup, even though this PR is closed the information is valuable and we will use it in the future to make our SDK easier to set up in monorepo scenarios.

Issue #3157, opened 2023-07-03
Android - `modulesCleanUpTask` doesn't work


liamjones:
**OS:**
- [x]  MacOS

**Platform:**
- [x]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)

**SDK version:** 5.6.0

**`react-native` version:** 0.71.11

**Are you using Expo?**
- [x]  Yes (bare workflow)

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)

**Configuration:**

N/A

---
**I have following issue:**

The created `modules.json` for Android builds doesn't get cleaned up after the build is finished

**Steps to reproduce:**
- Have project git repo in clean state
- Run `assembleRelease`
- Look at `git status`

**Actual result:**

New file created `android/app/src/main/assets/modules.json`

**Expected result:**

Clean git status

liamjones:
This seems to be working with the sample app against the current head of the main branch (https://github.com/getsentry/sentry-react-native/commit/59b869787a2213f654afe344111300a2e09ce69a) unless the build fails.

Let me check there isn't something weird going on in our Android project which caused this...

liamjones:
This is also in a monorepo but the create and delete steps both use the same path so, if it can create the file, I don't think the monorepo should have any effect on it being unable to delete the file.

PR #3161, opened 2023-07-04
fix(gradle): Use application variant instead of varian output to hook to correct package task for modules cleanup

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

The variant output name includes architecture but the package task has only the application variant in the name.

demo-x86_64-release vs. demoRelease

closes: https://github.com/getsentry/sentry-react-native/issues/3157

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


marandaneto:
Have you tested this using normal apk generation and bundle generation, with splitting per architecture enabled or not?
We need to ensure this does not break any other way to generate your apks or bundles.

krystofwoldrich:
@marandaneto Yes, I've tested bundle and apks with arch splitting and without.

For the splitting, I used this setting.

```groovy
    splits {
        abi {
            reset()
            enable true
            universalApk true  // If true, also generate a universal APK
            include "armeabi-v7a", "x86", "arm64-v8a", "x86_64"
        }
    }
```

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 
 - Warn users about multiple versions of `promise` package which can cause unexpected behavior like undefined `Promise.allSettled` ([#3162](https://github.com/getsentry/sentry-react-native/pull/3162))
 - Event is enriched with all the Android context on the JS layer and you can filter/modify all the data in the `beforeSend` callback similarly to iOS. ([#3170](https://github.com/getsentry/sentry-react-native/pull/3170))
+- Use application variant instead of varian output to hook to correct package task for modules cleanup ([#3161](https://github.com/getsentry/sentry-react-native/pull/3161))
```
romtsn, line 30:
```suggestion
- Use application variant instead of variant output to hook to correct package task for modules cleanup ([#3161](https://github.com/getsentry/sentry-react-native/pull/3161))
```

liamjones:
@krystofwoldrich Thanks for the fix! I think the changelog is still wrong in `main` post merge? This wasn't released in 5.8.1, right?

PR #2900, opened 2023-03-16
POC: Support artifact bundles (DebugID)

krystofwoldrich:
## :loudspeaker: Type of change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

### How to test

```bash
yarn
yarn build
cd sample-new-architecture
yarn
cd sample-new-architecture/android
./gradlew assembleRelease
adb install app/build/outputs/apk/release/app-release.apk
```

## :bulb: Motivation and Context

- https://github.com/getsentry/sentry-react-native/issues/2892

## :pencil: Checklist

- [ ] I reviewed submitted code
- [ ] I added tests to verify changes
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] All tests passing
- [ ] No breaking changes

Issue #2947, opened 2023-03-31
Add support for Source Maps Debug ID in Hermes bundle


krystofwoldrich:
### Description

Based on the initial testing, the current way of injecting Debug ID into a bundle file works only for plain text JS files.

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/sentry-cli/pull/1667
- [ ] https://github.com/getsentry/sentry-react-native/pull/3164
- [ ] https://github.com/getsentry/sentry-react-native/issues/2892
- [ ] https://github.com/getsentry/sentry-react-native/pull/2900
``` 

Debug ID must be propagated to Hermes Combined Source Map, but CLI at the moment doesn't automatically upload combined source maps.

```[tasklist]
### Related tasks
- [ ] https://github.com/getsentry/sentry-cli/issues/1569
```

marandaneto:
@krystofwoldrich you can link it to https://github.com/getsentry/team-mobile/issues/133

krystofwoldrich:
We might be able to use `getPolyfills` to add the debug id holder object to the JS runtime 

https://facebook.github.io/metro/docs/configuration#getpolyfills

shanamatthews:
Hey @krystofwoldrich - I know we have docs on [uploading source maps with Hermes](https://docs.sentry.io/platforms/react-native/manual-setup/hermes/), are there any updates needed based on these improvements?

krystofwoldrich:
- Follow up https://github.com/getsentry/sentry-react-native/issues/3326

PR #3164, opened 2023-07-05
Add Debug Ids support for JSC and Hermes bundles

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Adds Sentry Metro Plugin that injects Debug IDs to bundles.

Adds additional tooling that enables Debug IDs in Hermes source maps.

This PR adds Sentry Metro Serializer, which wraps the default or user-supplied serializer, and adds a Sentry Debug ID module which contains the Debug ID code snipped. The serializer also creates a Debug ID by hashing the content of the intermediate bundle, that is the bundle object before written to disk. 

- https://github.com/getsentry/sentry-react-native/issues/2947

- closes: https://github.com/getsentry/sentry-react-native/pull/2900

sample app, ci

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- [x] https://github.com/getsentry/sentry-cli/pull/1754
- [ ] Update docs
- [ ] Update wizard
- [x] Add metro plugin tests
- [x] Update e2e rn tests to use debug ids and hermesc source map
- [ ] Add migration docs


krystofwoldrich:
iOS e2e tests are not happy as they need https://github.com/getsentry/sentry-cli/pull/1754 release to work. 

Besides that this RP is ready for a review.

src/js/tools/sentryMetroSerializer.ts
```diff
@@ -0,0 +1,256 @@
...
+ * Creates the default Metro plain bundle serializer.
+ * This is used when the user does not provide a custom serializer.
+ */
+const createDefaultMetroSerializer = (serializerConfig: Partial<SerializerConfigT>): MetroSerializer => {
+  return (entryPoint, preModules, graph, options) => {
+    const createModuleId = options.createModuleId;
+    for (const module of graph.dependencies.values()) {
+      options.createModuleId(module.path);
+    }
+    const modules = [...graph.dependencies.values()].sort((a, b) => createModuleId(a.path) - createModuleId(b.path));
```
AbhiPrasad, line 127: l: Is there a way we can avoid the double iteration through the dependencies (once in for loop, once to construct final array?)

src/js/tools/sentryMetroSerializer.ts
```diff
@@ -0,0 +1,256 @@
...
+
+    let bundle = baseJSBundle(entryPoint, preModules, graph, options);
+    if (options.sentryBundleCallback) {
+      bundle = options.sentryBundleCallback(bundle);
+    }
+    const { code } = bundleToString(bundle);
+    if (serializerConfig.processModuleFilter === undefined) {
+      // processModuleFilter is undefined when processing build request from the dev server
+      return code;
```
AbhiPrasad, line 136: should we be running `sentryBundleCallback` in dev server?

We should also move `const modules` below this early return.

src/js/tools/sentryMetroSerializer.ts
AbhiPrasad: Considering we combine `preModules` and `modules` below, do we need to sort all of them together?

src/js/tools/sentryMetroSerializer.ts
```diff
@@ -0,0 +1,256 @@
...
+function determineDebugIdFromBundleSource(code: string): string | undefined {
+  const match = code.match(
+    /sentry-dbid-([0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12})/,
+  );
+
+  if (match) {
+    return match[1];
+  } else {
+    return undefined;
+  }
```
AbhiPrasad, line 165: l: Can this be a ternary return?

src/js/tools/sentryMetroSerializer.ts
```diff
@@ -0,0 +1,256 @@
...
+    // Add debug id module to the preModules
+    let modifiedPreModules: readonly Module<MixedOutput>[];
+    const containsDebugIdModule = preModules.some(module => module.path === DEBUG_ID_MODULE_PATH);
+    if (!containsDebugIdModule) {
+      const debugIdModule = createDebugIdModule(DEBUG_ID_PLACE_HOLDER);
+      const tmpPreModules = [...preModules];
+      if (tmpPreModules.length > 0 && tmpPreModules[0] !== undefined && tmpPreModules[0].path === PRELUDE_MODULE_PATH) {
+        // prelude module must be first as it measures the bundle startup time
+        tmpPreModules.unshift(preModules[0]);
+        tmpPreModules[1] = debugIdModule;
```
AbhiPrasad, line 189: l: can we just do a swap instead of unshift?

src/js/tools/sentryMetroSerializer.ts
```diff
@@ -0,0 +1,256 @@
...
+    const serializer = customSerializer || createDefaultMetroSerializer(this);
+    options.sentryBundleCallback = sentryBundleCallback;
+
+    // Add debug id module to the preModules
+    let modifiedPreModules: readonly Module<MixedOutput>[];
+    const containsDebugIdModule = preModules.some(module => module.path === DEBUG_ID_MODULE_PATH);
+    if (!containsDebugIdModule) {
+      const debugIdModule = createDebugIdModule(DEBUG_ID_PLACE_HOLDER);
+      const tmpPreModules = [...preModules];
+      if (tmpPreModules.length > 0 && tmpPreModules[0] !== undefined && tmpPreModules[0].path === PRELUDE_MODULE_PATH) {
```
AbhiPrasad, line 186: m: I'd like it if both the logic to generate `modifiedPreModules` and `bundleCode`/`bundleMapString` was extracted into helper functions so there is a clear input/output.

src/js/tools/sentryMetroSerializer.ts
```diff
@@ -0,0 +1,256 @@
...
+    const bundleCodeWithDebugId =
+      indexOfSourceMapComment === -1
+        ? // If source map comment is missing lets just add the debug id comment
+          `${bundleCode}\n${debugIdComment}`
+        : // If source map comment is present lets add the debug id comment before it
+          `${bundleCode.substring(0, indexOfSourceMapComment) + debugIdComment}\n${bundleCode.substring(
+            indexOfSourceMapComment,
+          )}`;
+
+    if (graph.transformOptions.hot === true) {
```
AbhiPrasad, line 242:
```suggestion
    if (graph.transformOptions.hot) {
```

src/js/tools/sentryMetroSerializer.ts
```diff
@@ -0,0 +1,256 @@
...
+      throw new Error(
+        'Debug ID was not found in the bundle. Call `options.sentryBundleCallback` if you are using a custom serializer.',
+      );
+    }
+    if (graph.transformOptions.hot !== true) {
+      // eslint-disable-next-line no-console
+      console.log('info ' + `Bundle Debug ID: ${debugId}`);
+    }
+
+    currentDebugIdModule?.output[0].data.code.replace(DEBUG_ID_PLACE_HOLDER, debugId);
```
AbhiPrasad, line 230: l: if `currentDebugIdModule` is not defined, should we just early return above?

src/js/tools/sentryMetroSerializer.ts
```diff
@@ -0,0 +1,256 @@
...
+};
+
+const DEBUG_ID_PLACE_HOLDER = '__debug_id_place_holder__';
+const DEBUG_ID_MODULE_PATH = '__debugid__';
+const PRELUDE_MODULE_PATH = '__prelude__';
+const SOURCE_MAP_COMMENT = '//# sourceMappingURL=';
+const DEBUG_ID_COMMENT = '//# debugId=';
+
+const createDebugIdSnippet = (debugId: string): string => {
+  return `var _sentryDebugIds={},_sentryDebugIdIdentifier="";void 0===_sentryDebugIds&&(_sentryDebugIds={});try{var stack=(new Error).stack;stack&&(_sentryDebugIds[stack]="${debugId}",_sentryDebugIdIdentifier="sentry-dbid-${debugId}")}catch(e){}`;
```
AbhiPrasad, line 45: How do we ensure this stays in sync with `scripts/sentry-debugid-injection-snippet.js`? Should we write a test?

AbhiPrasad (commented):
1st pass, excited for this!

src/js/tools/sentryMetroSerializer.ts
krystofwoldrich: I don't think so. The `debugIdModule` doesn't exist in the array till this moment.

Input: ['prelude', 'other1', 'other2']
Wanted output: ['prelude', 'debugId', 'other1', 'other2']

typings/metro.d.ts
```diff
@@ -0,0 +1,44 @@
+declare module 'metro/src/DeltaBundler/Serializers/baseJSBundle' {
```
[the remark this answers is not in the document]
krystofwoldrich: Good point, these are translated from Flow from the Metro package.

Added license and moved to vendor dir.

src/js/tools/sentryMetroSerializer.ts
krystofwoldrich: We have to sort the modules, otherwise, the source map would be broken, but I checked `baseJSBundle` and we can avoid the module id creation as it's be done in `baseJSBundle`.

src/js/tools/sentryMetroSerializer.ts
krystofwoldrich: I agree, if hot we can completely skip all debug id related modifications.

src/js/tools/sentryMetroSerializer.ts
krystofwoldrich: There is already an early return when the Debug ID was not found in the final bundle code. 

I've adjusted the code so there is always Debug ID module either existing one or created by us.

src/js/tools/sentryMetroSerializer.ts
krystofwoldrich: Added a test for this.

src/js/tools/sentryMetroSerializer.ts
```diff
@@ -0,0 +1,256 @@
...
+import * as bundleToString from 'metro/src/lib/bundleToString';
+import CountingSet from 'metro/src/lib/CountingSet';
+import * as countLines from 'metro/src/lib/countLines';
+
+type SourceMap = Record<string, unknown>;
+
+type ExpectedSerializedConfigThisContext = Partial<SerializerConfigT>;
+
+export type VirtualJSOutput = {
+  type: string;
```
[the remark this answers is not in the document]
krystofwoldrich: Thanks, actually looking at this again, there is only one possible type here `js/script/virtual`

krystofwoldrich:
@AbhiPrasad Thanks for the comments, I've shaken things up, should be better now.

scripts/has-sourcemap-debugid.js
```diff
@@ -0,0 +1,31 @@
...
+if (!fs.existsSync(sourceMapPath)) {
+  console.log(`${sourceMapPath} does not exist.`);
+  process.exit(1);
+}
+
+let sourceMap;
+try {
+  sourceMap = JSON.parse(fs.readFileSync(sourceMapPath, 'utf8'));
+} catch (e) {
+  console.log(`${sourceMapPath} is not valid JSON`, e);
```
AbhiPrasad, line 20:
```suggestion
  console.log(`Sourcemap at ${sourceMapPath} was unable to be read.`, e);
```

scripts/has-sourcemap-debugid.js
```diff
@@ -0,0 +1,31 @@
+const process = require('process');
+const fs = require('fs');
+
+const sourceMapPath = process.argv[2];
+
+if (!sourceMapPath) {
+  console.log('Add source map path as first argument of the script.');
```
AbhiPrasad, line 7: should we `console.warn` or `console.error` these? Ditto for the console logs that are below this.

src/js/tools/sentryMetroSerializer.ts
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * Creates a Metro serializer that adds Debug ID module to the plain bundle.
+ * The Debug ID module is a virtual module that provides a debug ID in runtime.
+ *
+ * RAM Bundles do not support custom serializers.
+ */
+export const createSentryMetroSerializer = (customSerializer?: MetroSerializer): MetroSerializer => {
+  return async function (entryPoint, preModules, graph, options) {
+    const serializer = customSerializer || createDefaultMetroSerializer();
```
AbhiPrasad, line 24: l: let's move this call above the `return async function` so we don't construct default serializer every time?

src/js/tools/sentryMetroSerializer.ts
```diff
@@ -0,0 +1,167 @@
...
+    const awaitedResult = await serializerResult;
+    if (typeof awaitedResult === 'string') {
+      code = awaitedResult;
+    } else {
+      code = awaitedResult.code;
+      map = awaitedResult.map;
+    }
+  }
+
+  return { code, map };
```
AbhiPrasad, line 126: let's avoid mutating code/map in favour of multiple return statements.

```suggestion
  if (typeof serializerResult === 'string') {
    return { code: serializerResult, map: '{}' };
   }
  
  if ('map' in serializerResult) {
    return { code: serializerResult.code, map: serializerResult.map };
  }
  
  const awaitedResult = await serializerResult;
  if (typeof awaitedResult === 'string') {
    return { code: awaitedResult, map: '{}' };
  }
  
  return { code: awaitedResult.code, map: awaitedResult.map };
```

AbhiPrasad (approved):
Really good work @krystofwoldrich

brustolin (approved):
The part related to cocoa LGTM.

scripts/has-sourcemap-debugid.js
krystofwoldrich: I agree that's more correct, but I had issues reading std err output in https://github.com/getsentry/sentry-react-native/pull/3164/files#diff-01c94c11105e37de1b11a82973b605f625fff5959b0386c32fd1e42a14325208R125-R126

So that's why I kept it `console.log`. More important is that we have the log.

src/js/tools/sentryMetroSerializer.ts
krystofwoldrich, line 24:
```suggestion
  const serializer = customSerializer || createDefaultMetroSerializer();
  return async function (entryPoint, preModules, graph, options) {
```

sentry.gradle
```diff
@@ -94,71 +96,107 @@ gradle.projectsEvaluated {
...
+
+                  // Add release and dist for backward compatibility if no Debug ID detected in output soruce map
+                  def process = ["node", hasSourceMapDebugIdScript, sourcemapOutput].execute(null, new File("$reactRoot"))
+                  project.logger.lifecycle("Check generated source map for Debug ID: ${process.text}")
+                  def notIncludeRelease = "$bundleCommand" == "bundle" && process.exitValue() == 0
+                  def not = notIncludeRelease ? 'not ' : ''
+                  project.logger.lifecycle("Sentry Source Maps upload will ${not}include the release name and dist.")
+                  extraArgs.addAll(notIncludeRelease ? [] : [
+                      "--release", releaseName,
+                      "--dist", versionCode
```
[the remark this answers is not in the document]
krystofwoldrich: Answered in https://github.com/getsentry/sentry-react-native/discussions/3474

PR #3200, opened 2023-07-19
Fix bug preventing native SDK from being called after reinitialization

mateioprea:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This commit addresses a bug in the Sentry React Native integration that occurs when calling `Sentry.init({ enableNative: true })`, followed by `Sentry.close()`, and then re-calling `Sentry.init({ enableNative: true })`. The bug prevents the native SDK from being invoked after the reinitialization.

The issue resides in the `closeNativeSdk()` method, where the `enableNative` flag is incorrectly used to determine whether the native SDK should be closed. This flag is not being updated correctly after reinitialization, causing the SDK to remain disabled.

With this change, the `nativeIsReady` flag is used to determine whether the native SDK should be closed, ensuring that subsequent reinitialization correctly enables the native SDK functionality.

This fix ensures that the native SDK will be called as expected after reinitializing with enableNative: true, resolving the issue where the SDK was not invoked due to the bug in the `closeNativeSdk()` method.

In my App.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


krystofwoldrich:
Hi,
thank you for the details and for informing us about the regression,

I believe the fix should be in the `isNativeAvailable` and not in the `closeMethod`. 

The `closeMethod` correctly flips the `enableNative` flag. But `isNativeAvailable` should not read from it.

mateioprea:
Fair enough! I will change it there

mateioprea:
@krystofwoldrich done!

krystofwoldrich:
@mateioprea Looks good, I'm thinking about some tests to prevent this in the future, but nothing comes to mind.

Can you update the changelog?

mateioprea:
@krystofwoldrich done. I had my previous PR opened and I made a mistake in the CHANGELOG file, but i fixed it. 

I cannot think of any test cases for now as well. 

Thank you!

mateioprea:
@krystofwoldrich any news here? :D

krystofwoldrich (approved):
Thank you for the detailed information and the fix. Looks good. 🚀

Tested with the sample app.

Issue #2118, opened 2022-03-07
Linked error are not symbolicated in Debug builds


Yupeng-li:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) 

Which SDK and version?
@sentry/react-native 3.2.13

### Steps to Reproduce

I need to report inner error when throwing exceptions, and [LinkedErrors](https://getsentry.github.io/sentry-javascript/classes/node.linkederrors.html) is the solution I found. To link two error, I assigned the inner error to the `cause` property of the final error.  

```js
//....
finalError.cause = innerError
Sentry.captureException(finalError)
```

When I tested, both the higher level error and the inner error are logged to sentry, however the stack trace was minified. But if I don't link them, the stack trace worked fine, so I think it's a bug. Here is a mini repo to reproduce the issue. https://github.com/Yupeng-li/sentry-test

The repo is a react-native app with two simple buttons: one throws an error and another throws an error with `cause` property. 

<img src="https://user-images.githubusercontent.com/31921080/157045959-d8963770-3fdd-416f-be68-e848d8ab9b82.png" width=20% height=20%>

Please modify the DSN before running it.

### Expected Result
Linked error should have original stack trace.  

This is what I got when I reported the error directly. (Throw an error)
![image](https://user-images.githubusercontent.com/31921080/157046821-591205de-66b8-401c-aefd-ebcc90bafb9d.png)

### Actual Result
This is what I got when I reported the error with `cause`.  (Throw an error with cause)
![linked errors](https://user-images.githubusercontent.com/31921080/157045491-60d53d30-ee51-46f5-9650-632fd8b90ad7.jpg)

marandaneto:
This integration is within the [node](https://github.com/getsentry/sentry-javascript/tree/master/packages/node) package, wondering if that makes any difference.
@AbhiPrasad or @kamilogorek do you know what's going on? thanks

AbhiPrasad:
This is strange, it seems that `RewriteFrames` doesn't update the stacktrace here?

https://github.com/getsentry/sentry-react-native/blob/16cfa52fcae796c51d1880b2ad6d9fc73dbede59/src/js/sdk.tsx#L72

Not sure if it's a general SDK bug, or just based on how react native sdk is configuring `RewriteFrames`, we need to test with a stock browser SDK app that uses `LinkedErrors` + `RewriteFrames`.

marandaneto:
Thanks @AbhiPrasad 
[...]

PR #3201, opened 2023-07-19
Add Native Linked Errors (RN Mixed stack traces)

krystofwoldrich:

- [x] New feature

## :scroll: Description

In https://github.com/facebook/react-native/pull/36925 we have added native errors to the JS errors coming from native Trubo modules calls. This PR implements `NativeLinkedErrors` integration that is able to parse these errors so they can be sent to Sentry.

- Related JS change: https://github.com/getsentry/sentry-javascript/pull/8766

JVM:
Screenshot 2023-07-19 at 16 58 02

iOS:
Screenshot 2023-08-09 at 11 29 47

sample app, integration tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


src/js/integrations/nativelinkederrors.ts
```diff
@@ -227,81 +229,3 @@ export class NativeLinkedErrors implements Integration {
     return NATIVE.fetchNativeStackFramesBy(instructionsAddr);
   }
 }
-
-// TODO: Needs to be exported from @sentry/browser
```
marandaneto: Since this is exported now, did we need to adapt the code or is it the very same code? otherwise, the adaptations have to remain here.

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Features
+
+- Add support for React Native mixed stacktraces ([#3201](https://github.com/getsentry/sentry-react-native/pull/3201))
+
+  In the current `react-native@nightly` JS errors from native modules can contain native JVM or Objective-C exception stack trace.
```
marandaneto, line 9: Does the `nightly` version have a preview number or something? or this is going to be outdated very soon.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -704,6 +704,15 @@ public void fetchNativeSdkInfo(Promise promise) {
         }
     }
 
+    public void fetchNativePackageName(Promise promise) {
+        promise.resolve(packageInfo.packageName);
+    }
+
+    public void fetchNativeStackFramesBy(Promise promise) {
+        logger.log(SentryLevel.ERROR, "This method works only on iOS.");
```
marandaneto, line 712: Can we actually not call this method at all? so the log isn't needed and we spare 1L, a smaller bundle size.

ios/RNSentry.mm
```diff
@@ -195,6 +199,86 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
...
 }
 
+RCT_EXPORT_METHOD(fetchNativePackageName:(RCTPromiseResolveBlock)resolve
+                  rejecter:(RCTPromiseRejectBlock)reject)
+{
+    NSString *packageName = [[NSBundle mainBundle] executablePath];
+    resolve(packageName);
+}
+
+- (NSDictionary*) fetchNativeStackFramesBy: (NSArray<NSNumber*>*)instructionsAddr
```
marandaneto, line 209: Can this be simplified by https://github.com/getsentry/sentry-dart/blob/35005743b3b663c45136b546c2fd11ab2e36fd39/flutter/ios/Classes/SentryFlutterPluginApple.swift#L242 ?

src/js/integrations/nativelinkederrors.ts
```diff
@@ -0,0 +1,231 @@
...
+
+interface LinkedErrorsOptions {
+  key: string;
+  limit: number;
+}
+
+/**
+ * Processes JS and RN native linked errors.
+ */
+export class NativeLinkedErrors implements Integration {
```
marandaneto, line 30: Does the integration handle well if its using an older version of RN that don't contain the linked errors at all?

src/js/sdk.tsx
```diff
@@ -38,6 +39,7 @@ import { NATIVE } from './wrapper';
 const IGNORED_DEFAULT_INTEGRATIONS = [
   'GlobalHandlers', // We will use the react-native internal handlers
   'TryCatch', // We don't need this
+  'LinkedErrors', // We replace this with `NativeLinkedError`
```
marandaneto, line 42: What's about https://github.com/getsentry/sentry-react-native/issues/2118
Does this PR help with that?

src/js/sdk.tsx
```diff
@@ -106,6 +108,7 @@ export function init(passedOptions: ReactNativeOptions): void {
       ),
     ]);
 
+    defaultIntegrations.push(new NativeLinkedErrors());
```
marandaneto, line 111: What's about https://github.com/getsentry/sentry-react-native/issues/2118
Does this PR help with that?

src/js/integrations/nativelinkederrors.ts
krystofwoldrich: It was copied without a change. So export from JS will work.

CHANGELOG.md
krystofwoldrich: Yes, they have a datatime timestamp, but since I assume the feature is going to stay and move to release, I didn't want to fix it on some old nightly build.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
krystofwoldrich: It's not called but we have to implement at least an empty method. Make sense, I'll remove the code to make the bundle smaller.

ios/RNSentry.mm
krystofwoldrich: It would only simplified the end of the function: https://github.com/getsentry/sentry-react-native/pull/3201/files#diff-ed0b1c4e6c4d33d23050f30ecbe94615487ed47eb7232cedd651092566b9cfa6R259-R265

And also the method `getDebugImages` gets all loaded images which is very inefficient. It retrieves all loaded images but based on the stack addresses only a handful is needed. (We are talking ones/tens vs 100s)

src/js/integrations/nativelinkederrors.ts
krystofwoldrich: Yes, prev RN versions would not throw an error in JS at all or it would be a plain JS error.

Tested here: https://github.com/getsentry/sentry-react-native/pull/3201/files/d3ea823bbc2048300e907d76105991ea02e6c520#diff-efd6c71dcc055b46a67752fd77362be1c8bc3499caf7616416841302da5e8145R19-R74

src/js/sdk.tsx
krystofwoldrich: Thanks for mentioning this error, sadly no, in debug only the top-level js error would be symbolicated.

CHANGELOG.md
marandaneto: if someone is using an older `nightly`, it won't work, maybe put between `()` the timestamp?

ios/RNSentry.mm
```diff
@@ -195,6 +199,86 @@ - (void)setEventEnvironmentTag:(SentryEvent *)event
...
+                               symbolicate: (SymbolicateCallbackType) symbolicate
+{
+  BOOL shouldSymbolicateLocally = [SentrySDK.options debug];
```
marandaneto, line 212: H: Does the Cocoa SDK do that too based on the `debug` flag? I believe that this should be based on the build type and not on the sentry options debug flag.

CHANGELOG.md
krystofwoldrich: Okay, I see, I will add it.

ios/RNSentry.mm
krystofwoldrich: Cocoa does it also based on the `debug` flag from clients' options.
https://github.com/getsentry/sentry-cocoa/pull/3079

marandaneto: Ok, lets keep as it is then

marandaneto (approved):

PR #3211, opened 2023-07-24
WIP: Ignore FrameStalling when app goes to background.

lucas-zimerman:
This is an initial PR to align the decision of how to handle stops the _iterateloop from the StallFrame logic and reactivate it if it was previously running  when the device is again active.


src/js/tracing/stalltracking.ts
```diff
@@ -61,6 +63,22 @@ export class StallTrackingInstrumentation {
...
+
+    this.backgroundEventListener = this.backgroundEventListener.bind(this);
+    AppState.addEventListener('change', this.backgroundEventListener);
+  }
+
+  private backgroundEventListener(state: AppStateStatus) {
+    if (state === "active" as AppStateStatus) {
+      this._isBackground = false;
+      if (this._timeout != null) {
+        this._lastIntervalMs = Math.floor(timestampInSeconds() * 1000);
```
krystofwoldrich, line 75: Why do we do the `Math.floor` here?

krystofwoldrich:
This looks great, I think the current solution is nice and simple. 

@lucas-zimerman  Can we add some tests? There will be a bit of mocking but it would be good to cover the edge cases and lock the logic.

lucas-zimerman:
oh prettier auto applied some formatting to the changelog file, I'll fix it.

lucas-zimerman:
test isolated from StallTracker because of some flaky when the stalltracker is tested before the client.test or wrapper.test

.github/workflows/buildandtest.yml
```diff
@@ -25,7 +25,9 @@ jobs:
         if: ${{ steps.cache.outputs['cache-hit'] != 'true' }}
         run: yarn install
       - name: Test
-        run: yarn test
+        run: yarn test:no-stall
+      - name: Test Stall Tracking
+        run: yarn test /test/tracing/stalltracking.test.ts
```
krystofwoldrich, line 30: I'll take a look at the test order before merging this. I would like to avoid merging this specific workaround to main.

src/js/tracing/stalltracking.ts
```diff
@@ -61,6 +65,11 @@ export class StallTrackingInstrumentation {
 
   public constructor(options: StallTrackingOptions = { minimumStallThreshold: 50 }) {
     this._minimumStallThreshold = options.minimumStallThreshold;
+
+    this._backgroundEventListener = this._backgroundEventListener.bind(this);
+    // Avoids throwing any error if using React Native on a environment that doesn't implement AppState.
+    // eslint-disable-next-line @typescript-eslint/unbound-method
+    AppState?.addEventListener('change', this._backgroundEventListener);
```
krystofwoldrich, line 72: We should also check `AppState.isAvailable`.

test/tracing/stalltracking.background.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    // Check if _isBackground is set to true
+    expect(stallTracking['_isBackground']).toBe(true);
+  });
+  it('Stall tracking should not call _iteration when state is active but _timeout is null', () => {
+    const stallTracking = new StallTrackingInstrumentation();
+    stallTracking['_timeout'] = null;
+    // Mock _iteration
+    stallTracking['_iteration'] = () => {
+      throw new Error('_iteration should not be called by _backgroundEventListener');
+    }; // Create a fake timeout to simulate a running interval
```
krystofwoldrich, line 33: Let's also use `jest.fn` here and then check that `not.called()`.

src/js/tracing/stalltracking.ts
```diff
@@ -215,6 +224,21 @@ export class StallTrackingInstrumentation {
...
   }
 
+  /**
+   * Switch that enables the iteraction once app moves from background to foreground.
+   */
+  private _backgroundEventListener(state: AppStateStatus): void {
+    if (state === ('active' as AppStateStatus)) {
+      this._isBackground = false;
+      if (this._timeout != null) {
+        this._lastIntervalMs = Math.floor(timestampInSeconds() * 1000);
```
krystofwoldrich, line 234: Why is the Math.floor here? We don't do that at the other places in the file.

krystofwoldrich:
I really like splitting the test into multiple files, it makes it much easier to read. I have a few small comments otherwise it looks good 🚀.

.github/workflows/buildandtest.yml
krystofwoldrich, line 30:
```suggestion
        run: yarn test
```

package.json
```diff
@@ -25,6 +25,7 @@
     "downlevel": "downlevel-dts dist ts3.8/dist --to=3.8",
     "clean": "rimraf dist coverage",
     "test": "jest",
+    "test:no-stall": "jest --testPathIgnorePatterns=test/tracing/stalltracking.test.ts  --testPathIgnorePatterns=test/e2e/",
```
krystofwoldrich, line 28:
```suggestion
```

.github/workflows/buildandtest.yml
krystofwoldrich, line 30:
```suggestion
        run: yarn test
```

.prettierignore
```diff
@@ -1,2 +1,3 @@
 /test/react-native/versions
+changelog.md
```
krystofwoldrich, line 2: L: Just so it doesn't fail on case sensitive fs.
```suggestion
CHANGELOG.md
```

src/js/tracing/stalltracking.ts
```diff
@@ -215,6 +226,21 @@ export class StallTrackingInstrumentation {
...
+   */
+  private _backgroundEventListener(state: AppStateStatus): void {
+    if (state === ('active' as AppStateStatus)) {
+      this._isBackground = false;
+      if (this._timeout != null) {
+        this._lastIntervalMs = timestampInSeconds() * 1000;
+        this._iteration();
+      }
+    } else {
+      this._isBackground = true;
```
krystofwoldrich, line 240: H: Should we clear the timeout here to avoid a possible race condition where we set `isBackground` but the app gets paused before the `_iteration` gets executed?

```suggestion
      this._isBackground = true;
      this._timeout !== null && clearTimeout(this._timeout);
```

lucas-zimerman, line 240: Yes, but I would prefer using the below code since clearTimeout is a void function
```suggestion
      this._isBackground = true;
      clearTimeout(this._timeout);
```

src/js/tracing/stalltracking.ts
```diff
@@ -215,6 +226,22 @@ export class StallTrackingInstrumentation {
...
+      clearTimeout(this._timeout);
```
lucas-zimerman, line 241:
```suggestion
      this._timeout !== null && clearTimeout(this._timeout);
```

src/js/tracing/stalltracking.ts
lucas-zimerman: I have appleid your suggestion on another commit so I'll be resolving this one.

krystofwoldrich (approved):
🚀

PR #3170, opened 2023-07-10
Add SentrySerialized to ReactNativeWritable convertor, load device context from sentry-java scope

krystofwoldrich:
## :loudspeaker: Type of change

- [x] Bugfix
- [x] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

Uses the new serialization interface from `sentry-java` to sync scope from Android SDK to the React Native SDK.

## :bulb: Motivation and Context

- fixes: https://github.com/getsentry/sentry-react-native/issues/2146
- closes: https://github.com/getsentry/team-mobile/issues/11
- blocked by: https://github.com/getsentry/sentry-java/pull/2814

## :pencil: Checklist

- [x] I reviewed submitted code
- [ ] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- [ ] Add tests for the JavaConvertor

PR #3213, opened 2023-07-25
Fix node version for e2e tests

krystofwoldrich:
We have to fix the major version of Node.JS for RN E2E tests, as RN 0.65.3 won't build with the newer node versions due to unsupported OSSL. 

https://stackoverflow.com/questions/69394632/webpack-build-failing-with-err-ossl-evp-unsupported

And at the same time, we have to fix the major version for the newer RN releases as they support only Node.JS 18 and above.

https://github.com/facebook/react-native/blob/main/packages/react-native/package.json#L24

#skip-changelog

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -179,8 +179,14 @@ jobs:
           echo "SENTRY_DIST=$SENTRY_DIST"
 
       - uses: actions/setup-node@v3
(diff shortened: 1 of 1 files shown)


krystofwoldrich:
Included in https://github.com/getsentry/sentry-react-native/pull/3170

PR #3220, opened 2023-07-31
Background spans

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This Integrations will add background span to auto-created transactions. This helps users to discover why their transactions are longer than expected. Background spans are not added to the end of the transaction where they would artificially extend its duration.

This is an experimental feature at the moment.
 
React Native has issues with long transactions and most of them are caused by the app in the background. 

- [x] Sample app. 
- [x] Unit tests 

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


src/js/integrations/backgroundspans.ts
```diff
@@ -0,0 +1,110 @@
...
+        this._currentBackgroundStartTimestamp = undefined;
+      } else if (state === <AppStateStatus>'background' && !this._currentBackgroundStartTimestamp) {
+        this._currentBackgroundStartTimestamp = timestampInSeconds();
+      }
+    });
+  }
+
+  /**
+   * Creates a background span and adds it to the transaction
+   * Patches the transaction's finish method to remove trailing background span
```
AbhiPrasad, line 54: if it's an idle transaction, can we use https://github.com/getsentry/sentry-javascript/blob/develop/packages/core/src/tracing/idletransaction.ts#L202?

src/js/integrations/backgroundspans.ts
```diff
@@ -0,0 +1,110 @@
...
+      return;
+    }
+
+    tx.startChild({
+      startTimestamp,
+      description: 'App in background',
+      op: BACKGROUND_SPAN_OP,
+    }).finish(endTimestamp);
+
+    if ((tx as unknown as Record<string, boolean>).__backgroundSpan === true) {
```
AbhiPrasad, line 67:
```suggestion
    if ((tx as unknown as Record<string, boolean>).__backgroundSpan) {
```

src/js/integrations/backgroundspans.ts
krystofwoldrich: Yes, we can check if it is IdleTx and then used the callback.

src/js/integrations/backgroundspans.ts
```diff
@@ -0,0 +1,120 @@
...
+      op: BACKGROUND_SPAN_OP,
+    }).finish(endTimestamp);
+
+    if ((tx as unknown as Record<string, boolean>).__backgroundSpan) {
+      return;
+    }
+    (tx as unknown as Record<string, boolean>).__backgroundSpan = true;
+
+    if (tx instanceof IdleTransaction) {
+      tx.registerBeforeFinishCallback((finishingIdleTx: IdleTransaction) => {
```
AbhiPrasad, line 73: Looking at this again, could we use the `finishTransaction` lifecycle hook instead of monkeypatching?

https://github.com/getsentry/sentry-javascript/blob/fac2be8cb7b46962220f2260d418c030eb4bf63e/packages/core/src/baseclient.ts#L410

krystofwoldrich:
After reviewing this approach we will not proceed with the background spans but we will cancel the idleTransaction when the app goes to background.

PR #3221, opened 2023-07-31
chore(sample): Change environment to `dev` to ensure 100% sampling

krystofwoldrich:
We want to receive all events from the sample for development and debugging purposes.

#skip-changelog 

> Environments that contain "debug", "dev", "local", "qa", and "test" are not sampled on the server.


lucas-zimerman (approved):
LGTM!

PR #3227, opened 2023-08-04
chore(wrap): Warn users if root component mounts before `Sentry.init`

krystofwoldrich:

- [x] Enhancement

## :scroll: Description

Switching the order causes issues with the App Start tracking. 

sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/tracing/reactnativeprofiler.tsx
+++ b/src/js/tracing/reactnativeprofiler.tsx
@@ -20,7 +20,7 @@ export class ReactNativeProfiler extends Profiler {
     if (!client) {
       // We can't use logger here because this will be logged before the `Sentry.init`.
       // eslint-disable-next-line no-console
-      console.warn('App Start Span could not be finished. `Sentry.wrap` was called before `Sentry.init`.');
+      __DEV__ && console.warn('App Start Span could not be finished. `Sentry.wrap` was called before `Sentry.init`.');
     } else {
       const tracingIntegration = getCurrentHub().getIntegration(
         ReactNativeTracing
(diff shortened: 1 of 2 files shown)


src/js/tracing/reactnativeprofiler.tsx
```diff
@@ -14,17 +14,23 @@ export class ReactNativeProfiler extends Profiler {
    */
   public componentDidMount(): void {
     super.componentDidMount();
-    getCurrentHub().getClient()?.addIntegration?.(createIntegration(this.name));
+    const client = getCurrentHub().getClient();
+    client && client.addIntegration && client.addIntegration(createIntegration(this.name));
```
lucas-zimerman, line 18: code looks good!

What about altering slightly the reading of the code, removing the else condition and storing the hub so we dont need to get the current hub twice:

```typescript
    const hub = getCurrentHub();
    const client = hub.getClient();

    if (!client) {
      // We can't use logger here because this will be logged before the `Sentry.init`.
      // eslint-disable-next-line no-console
      __DEV__ && console.warn('App Start Span could not be finished. `Sentry.wrap` was called before `Sentry.init`.');
      return;
    }

    client.addIntegration && client.addIntegration(createIntegration(this.name));
    const tracingIntegration = hub.getIntegration(
      ReactNativeTracing
    );

    tracingIntegration
      && this._mountSpan
      && typeof this._mountSpan.endTimestamp !== 'undefined'
      // The first root component mount is the app start finish.
      && tracingIntegration.onAppStartFinish(this._mountSpan.endTimestamp);
```

krystofwoldrich: Yes, looks good.

lucas-zimerman (approved):
LGTM!

Issue #1042, opened 2020-08-25
Ability to disable autoUpload source maps


trangcongthanh:
**OS:**
- [x]  Windows
- [ ]  MacOS
- [ ]  Linux

**Platform:**
- [ ]  iOS
- [x]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.7.1)
- [ ]  `react-native-sentry` (<= 0.43.2)

**`react-native` version:** 0.63.2 with hermes

**Are you using Expo?**
- [ ]  Yes
- [x]  No

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)
- [ ]  on-premise

---
**I have following issue:**

Look like source maps got uploaded everytime I run `react-native run-android` event with debug mode. It take alot of time to start to dev my app. If it upload my local source maps, will it affect to our production? And I don't want to use sentry for local dev.

Looking around documents but not found any where mention How to disable autoUpload source maps.

iBotPeaches:
Could this be an added feature? Now that Release builds automatically try to upload assets, team members don't even have creds to upload it - thus it fails. I mean Release builds are uncommon locally, but they have to happen for certain things to test.

I thought there would be some simple environment variable or even property file, but the only way I've found out to patch this is overriding our gradle file with something like:

```
gradle.projectsEvaluated {
    def tasksToSkip = project.tasks.matching {
        it.name.contains("SentryUpload")
    }
    tasksToSkip.all { enabled = false }
}
```

jennmueng:
I agree with this, we will investigate adding this in the near future.

marandaneto:
you can tho https://docs.sentry.io/platforms/react-native/sourcemaps/#2-disable-the-automatic-source-maps-upload-script
a boolean would be nicer, so I'd rather keep it open

marandaneto:
So we should offer this flag to be flipped via the Sentry configuration on your Gradle file, we should also document that on our [debug symbols](https://docs.sentry.io/platforms/react-native/upload-debug/) docs.

sreuter:
It would be really great to just have an environment setting to achieve this, e.g. disable it for local development to prevent the console from being spammed.

<img width="628" alt="Screen Shot 2022-05-23 at 12 24 53 pm" src="https://user-images.githubusercontent.com/550246/169731795-9f53f880-418c-4530-93c8-84d0918757fe.png">

marandaneto:
[...]

PR #3296, opened 2023-09-20
feat(sdk): Disable sourcemap based on ENV

Tobbb:
DISABLE_UPLOAD_SOURCEMAP_SENTRY, the name can be discussed

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

disable task in sentry.gradle. This part is only for android, the iOS parts will be made in the sentry-wizard. ill link it here later

when building test-builds there is no need to generate source-map for every build. it might be good to save some CI/CD time

https://github.com/getsentry/sentry-react-native/issues/1042

manual build and verified

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

fix sentry-wizard

### DIFF
--- a/sentry.gradle
+++ b/sentry.gradle
@@ -159,7 +159,7 @@ gradle.projectsEvaluated {
 
               enabled true
           }
-
+          cliTask.onlyIf { !System.getenv("DISABLE_UPLOAD_SOURCEMAP_SENTRY") }
           modulesTask = tasks.create(nameModulesTask, Exec) {
               description = "collect javascript modules from bundle source map"
               group = 'sentry.io'


krystofwoldrich:
@Tobbb Thank you for the PR, this is a good start and will work as you mentioned for the React Native Sentry Source Map upload on Android.

I believe we should also consider projects using [Sentry Android Gradle Plugin](https://docs.sentry.io/platforms/android/configuration/gradle/) and as you mentioned iOS. 

Regarding iOS, `sentry-wizard` is an option, but it will require existing projects to manually make the change. 
We could also include `scripts/sentry-xcode-bundle.sh` or similar so these changes will be propagated with the package upgrade in the future.

Tobbb:
> @Tobbb Thank you for the PR, this is a good start and will work as you mentioned for the React Native Sentry Source Map upload on Android.
> 
> I believe we should also consider projects using [Sentry Android Gradle Plugin](https://docs.sentry.io/platforms/android/configuration/gradle/) and as you mentioned iOS.
> 
> Regarding iOS, `sentry-wizard` is an option, but it will require existing projects to manually make the change. We could also include `scripts/sentry-xcode-bundle.sh` or similar so these changes will be propagated with the package upgrade in the future.

Yes it's a good point, It feels kinda awkward having the logic only being applied in scope of the wizard. perhaps applying the patch to Xcode build phases would make more sense and only use wizard to generate credential files. but it might be out of the scope for this

PR #3298, opened 2023-09-21
Add support for AGP8+

pvegh:
by enabling buildConfig since it's using custom buildConfig field
So that consumer projects don't have to globally enable this themselves leading to performance impact

- [ ] Bugfix
- [ ] New feature
- [ x ] Enhancement
- [ ] Refactoring

## :scroll: Description
Since AGP8 this flag is disabled by default. Consumer projects have to enable this flag in them globally if the library doesn't do this correctly itself, but uses custom buildConfig fields.

No errors and better build performance than the workaround in consuming projects would be.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -19,6 +19,13 @@ android {
         namespace = "io.sentry.react"
     }
 
+    def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
+    if (agpVersion.tokenize('.')[0].toInteger() >= 8) {
+        buildFeatures {
+            buildConfig = true
+        }
+    }
+
     defaultConfig {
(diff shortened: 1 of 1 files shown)


krystofwoldrich:
Thank you for the enhancement. Let's check if CI passes.

pvegh:
I'm not sure my changes would cause the failed checks @krystofwoldrich

pvegh:
One of your paying customers is eagerly waiting for this, I even created the patch to make things faster.
We'd like to see a quick release with this. Thanks!

krystofwoldrich (approved):
Thank you for preparing the SDK for AGP 8. 🚀

Issue #3317, opened 2023-10-02
`attachScreenshot: true` still takes screenshots even when `enabled: false`


liamjones:
**Platform:**
- [x]  iOS
- [x]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)

**SDK version:** 5.9.2

**`react-native` version:** 0.71.13

**Are you using Expo?**
- [x]  No

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)

**If you are using sentry.io, please post a link to your issue so we can take a look:**

N/A

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  dsn: 'https://a8e9796651e6b88bb70e505002fc359d@o169785.ingest.sentry.io/4505951781191680',
  enabled: false,
  attachScreenshot: true,
});
```

---
**I have following issue:**

[Description]

**Steps to reproduce:**
- Configure Sentry with `attachScreenshot: true` and `enabled: false`.
- Connect debugger and set breakpoints in 'take screenshot' functions of Android/iOS native modules
- Call `Sentry.captureException()`

**Actual result:**

Breakpoints hit as native modules try to capture a screenshot

**Expected result:**

Screenshots not taken as no reporting is going to happen

krystofwoldrich:
Hi, thank you for the information,
I agree, I would also expect the screenshot not to be taken when the SDK is disabled.

krystofwoldrich:
This is caused by the Screenshot integration being hooked to captureException and not using the event processor.

https://github.com/getsentry/sentry-react-native/blob/5af0e04ecdb7e90a485aeb261ad96a743bc0898f/src/js/client.ts#L55

PR #3328, opened 2023-10-08
Fix: Should not take screenshot when SDK is disabled

ziyad-elabid-nw:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This PR suggests a solution for not taking screenshot when the SDK is diabled.

related to this issue : #3317 

I wrote a unit test for the case when enabled is false

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@
 ### Fixes
 
 - Change log output to show what paths are considered when collecting modules ([#3316](https://github.com/getsentry/sentry-react-native/pull/3316))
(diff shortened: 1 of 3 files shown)


krystofwoldrich:
@ziyad-elabid-nw Thank you for opening the PR, sadly not including the integration doesn't fix the issue, as the Screenshots are taken regardless. 

I've updated the integration itself [here](https://github.com/getsentry/sentry-react-native/pull/3333), so the unexpected behavior is removed. When the SDK is disabled integrations are added as usual but the `setup` method is not called.

PR #3331, opened 2023-10-09
fix(tracing): Use deprecated tracing origins option if set

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

Until the next major version, where we can remove the deprecated options, there are 3 ways to set trace propagation targets. The oldest `tracingOrigins` option was not working as the default value of the newer option always overwrote it.

sample app, unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/test/tracing/reactnativetracing.test.ts
+++ b/test/tracing/reactnativetracing.test.ts
@@ -209,6 +209,56 @@ describe('ReactNativeTracing', () => {
         }),
       );
     });
+
+    it('client tracePropagationTargets takes priority over integration options', () => {
+      const instrumentOutgoingRequests = jest.spyOn(SentryBrowser, 'instrumentOutgoingRequests');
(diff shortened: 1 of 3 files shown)


test/tracing/reactnativetracing.test.ts
```diff
@@ -117,6 +119,98 @@ describe('ReactNativeTracing', () => {
     jest.clearAllMocks();
   });
 
+  describe('trace propagation targets', () => {
```
Lms24, line 122: l/recommendation: Afaik, in the JS SDK we have tests that show which option is taken if multiple of these properties are specified. Admittedly it's an edge case, so feel free to disregard (not a blocker).

Lms24 (approved):
LGTM, one optional suggestion.

Issue #3311, opened 2023-09-29
Error `IntrinsicAttributes` when passing App component to `Sentry.wrap`


dprevost-LMI:
**OS:**
- [x]  MacOS

**SDK:**
- [x]  `@sentry/react-native` (5.7.1)

**Are you using Expo?**
- [x]  No

---
[Description]
I upgraded from 4.2.4 to 5.7.1 and started to have the below error when doing `Sentry.wrap` on my App component.
```
	"message": "Argument of type 'typeof App' is not assignable to parameter of type 'ComponentType<IntrinsicAttributes>'.\n  Type 'typeof App' is not assignable to type 'ComponentClass<IntrinsicAttributes, any>'.\n    Types of parameters 'props' and 'props' are incompatible.\n      Type 'IntrinsicAttributes' is not assignable to type '(AppStateProps & AppOwnProps) | Readonly<AppStateProps & AppOwnProps>'.
```

It seems to come from [this commit](https://github.com/getsentry/sentry-react-native/commit/3ffcddd6248046202afca78f6b9af8e3f591202c#diff-210c7a3eefc2c034ae4f9720256eaf1b7adbceb49eb6d547500931f9a51928f6) in file `src/js/sdk.tsx`

Where `wrap` now expects the component to extend `JSX.IntrinsicAttributes`
```
-export function wrap<P>(
+export function wrap<P extends JSX.IntrinsicAttributes>(
```
Weirdly, removing my props object from the App component makes the error disappear.

I found three ways to solve this for now, but it seems more like patches:
1. Remove my props from the components (which I can't)
2. Have one of my props interfaces extend JSX.IntrinsicAttributes
3. Remove `extends JSX.IntrinsicAttributes` from sdk.ts using a patch

**Expected result:**
I would expect to be able to pass my App component extending React.PureComponent with Props object to the `Sentry.wrap` without having errors

Can you explain why `extend JSX.IntrinsicAttributes` is now required. Do you know how to work around/fix my issue the proper way?

Code snippet
```TS

interface AppStateProps /*extends React.JSX.IntrinsicAttributes *//* This fix the issue*/ {
  profileId?: string;
}

interface AppOwnProps {
  routingInstrumentation?: ReactNavigationInstrumentation;
  appBootTransaction?: Transaction;
  onTimeToReady?(): void | Promise<void>;
}

class App extends React.PureComponent<AppStateProps & AppOwnProps> {
...
}

[...]

PR #3332, opened 2023-10-09
fix(sdk): `Sentry.wrap` doesn't enforce any keys on the wrapped component props

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

closes: https://github.com/getsentry/sentry-react-native/issues/3311
adds test to prevent this regression in the future

unit test, sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/sdk.tsx
+++ b/src/js/sdk.tsx
@@ -149,9 +149,7 @@ export function init(passedOptions: ReactNativeOptions): void {
 /**
  * Inits the Sentry React Native SDK with automatic instrumentation and wrapped features.
  */
-// Deprecated in https://github.com/DefinitelyTyped/DefinitelyTyped/commit/f1b25591890978a92c610ce575ea2ba2bbde6a89
-// eslint-disable-next-line deprecation/deprecation
-export function wrap<P extends JSX.IntrinsicAttributes>(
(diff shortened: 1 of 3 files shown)


src/js/sdk.tsx
```diff
@@ -149,9 +149,7 @@ export function init(passedOptions: ReactNativeOptions): void {
 /**
  * Inits the Sentry React Native SDK with automatic instrumentation and wrapped features.
  */
-// Deprecated in https://github.com/DefinitelyTyped/DefinitelyTyped/commit/f1b25591890978a92c610ce575ea2ba2bbde6a89
-// eslint-disable-next-line deprecation/deprecation
-export function wrap<P extends JSX.IntrinsicAttributes>(
+export function wrap<P extends Record<string, unknown>>(
```
lforst, line 152: Hm, we probably want to infer the return type. Let me get back to this. In an hour or so.

krystofwoldrich: I thought about that too, but so far no one complained about that. But the `JSX.IntrinsicAttributes` forced users to specify a `key` attribute in the props.

lforst (approved):

PR #3339, opened 2023-10-13
Feat: Report as app context what's the current screen during the error

### LINKED ISSUE #2994: Report as app context what's the current screen during the error
- Backlink https://github.com/getsentry/team-sdks/issues/18

This will be depending on active navigation integrations as that is the only way to know the active screen in RN.

lucas-zimerman:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This feature adds the current screen name into the app's context.

The tag is being set here
https://github.com/getsentry/sentry-react-native/blob/21465bc0471021968029445e6f85b284019c9283/src/js/tracing/reactnativetracing.ts#L438-L460

and later is being checked by the device context to be inserted into native events.

We could later remove this tag but it will become a break change so for the time being we can keep both the tag and the context.

I think the current implementation is OK, the only bother that I have with it is checking the tags for this information, ideally there should have another parameter to retrieve this information. I was thinking of using 
```typescript
      scope.setContext('app', { 'view_names':  [context.name ] });
```
but that removes  the app context instead of updating it.

NOTE: Tests will be added once the main logic form this code is approved.

https://github.com/getsentry/sentry-react-native/issues/2994

https://sentry-sdks.sentry.io/issues/4058613036/events/ae3a1c915f2e4fc993938748da6199d2/?project=5428561

## :pencil: Checklist

- [x] All tests passing
- [X] No breaking changes

## :crystal_ball: Next steps

Close #2994


src/js/integrations/devicecontext.ts
```diff
@@ -54,6 +54,10 @@ export class DeviceContext implements Integration {
         };
       }
       if (nativeContexts) {
+        const routingName = event.tags?.['routing.route.name'];
+        if (routingName) {
+          nativeContexts.app['view_names'] = [routingName];
+        }
```
krystofwoldrich, line 60: Since we would like to remove the tag in the next major release of the SDK. Let's add the `app.view_names` in https://github.com/getsentry/sentry-react-native/blob/a27397ef27b60c808768965a4dff6efaa6a8b81e/src/js/tracing/reactnativetracing.ts#L458

Here let's make sure the values from `context.app` from RN Event are not overwritten by the native.

src/js/integrations/devicecontext.ts
```diff
@@ -49,12 +49,16 @@ export class DeviceContext implements Integration {
       if (AppState.currentState !== 'unknown') {
         nativeContexts = nativeContexts || {};
         nativeContexts.app = {
+          ...event.contexts?.app,
```
lucas-zimerman, line 52:
```suggestion
```

src/js/tracing/reactnativetracing.ts
```diff
@@ -454,7 +454,10 @@ export class ReactNativeTracing implements Integration {
           },
         });
       }
-
+      scope.setContext('app', { view_names: [context.name] });
```
krystofwoldrich, line 457: We should check if `app` exists in the scope, so we don't overwrite other keys set by a user, only the `view_names`.

src/js/tracing/reactnativetracing.ts
```diff
@@ -350,6 +360,17 @@ export class ReactNativeTracing implements Integration {
     return this._inflightInteractionTransaction;
   }
 
+  /**
+   *  Sets the current view name into the app context.
+   *  @param event Le event.
+   */
+  private _getCurrentViewEventProcessor(event: Event): Event {
+    if (event.contexts && this._currentViewName) {
+      event.contexts.app = { ...{ view_names: [this._currentViewName] }, ...event.contexts.app };
```
krystofwoldrich, line 369: We can avoid the creation of the object wrapping `view_names`.

```suggestion
      event.contexts.app = { view_names: [this._currentViewName], ...event.contexts.app };
```

src/js/tracing/reactnativetracing.ts
```diff
@@ -255,6 +262,9 @@ export class ReactNativeTracing implements Integration {
       logger.log('[ReactNativeTracing] Not instrumenting route changes as routingInstrumentation has not been set.');
     }
 
+    this._getCurrentViewEventProcessor = this._getCurrentViewEventProcessor.bind(this);
+    addGlobalEventProcessor(async (event: Event) => this._getCurrentViewEventProcessor(event));
```
krystofwoldrich, line 266: We can right away pass the function with binded this as a global event processor. And avoid creating the arrow function. 

```suggestion
    addGlobalEventProcessor(this._getCurrentViewEventProcessor.bind(this));
```

krystofwoldrich:
Thanks this looks good. I just have two performance suggestions to avoid creating unnecessary objects.

krystofwoldrich (approved):
Thank you! Looks good.

Issue #2801, opened 2023-02-02
Detect `event.environment` automatically if not specified in the options


krystofwoldrich:
### Description

We can detect by `__DEV__` is RN was build for `production` or for `development`/`debug` we can set the `event.environment` automatically if not defined in the options.

- context  https://github.com/getsentry/team-mobile/issues/73#issuecomment-1323428698
	- rn sets default `production` not auto-detection at the moment

PR #3362, opened 2023-10-24
feat(sdk): Automatically detect environment if not set by User

krystofwoldrich:

- [x] New feature

## :scroll: Description

Detection is based on the JS environment and the global `__DEV__` property user can override the value.

This change might affect users not setting the environment option at the moment. As their dev builds will now have `development` environment.

- https://github.com/getsentry/sentry-react-native/issues/2801

sample app, tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/utils/environment.ts
+++ b/src/js/utils/environment.ts
@@ -35,3 +35,8 @@ export function getHermesVersion(): string | undefined {
     RN_GLOBAL_OBJ.HermesInternal.getRuntimeProperties()['OSS Release Version']
   );
 }
+
+/** Returns default environment based on __DEV__ */
+export function getDefaultEnvironment(): 'development' | 'production' {
+  return __DEV__ ? 'development' : 'production';
+}
--- a/src/js/utils/environment.ts
+++ b/src/js/utils/environment.ts
@@ -38,5 +38,5 @@ export function getHermesVersion(): string | undefined {
 
 /** Returns default environment based on __DEV__ */
 export function getDefaultEnvironment(): 'development' | 'production' {
-  return __DEV__ ? 'development' : 'production';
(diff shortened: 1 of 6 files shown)


src/js/utils/environment.ts
```diff
@@ -35,3 +35,8 @@ export function getHermesVersion(): string | undefined {
     RN_GLOBAL_OBJ.HermesInternal.getRuntimeProperties()['OSS Release Version']
   );
 }
+
+/** Returns default environment based on __DEV__ */
+export function getDefaultEnvironment(): 'development' | 'production' {
+  return __DEV__ ? 'development' : 'production';
```
lforst, line 41: Do we know 100% whether `__DEV__` is always defined? If not I would suggest doing:

```suggestion
  return typeof __DEV__ !== undefined && __DEV__ ? 'development' : 'production';
```

That way it doesn't throw if it isn't defined.

lforst (approved):

src/js/utils/environment.ts
lucas-zimerman: doesn't typedef `__DEV__` returns 'undefined' instead of undefined? Additionally if `__DEV__` is not false, what's the expected return? fallback to the original empty string? if so maybe we could do the following
```suggestion
  return typeof __DEV__ !== 'undefined' ? (__DEV__ ? 'development' : 'production') :  '';
```

lucas-zimerman: I did a quick test and ` typeof __DEV__ !== undefined ` return true for undefined and false for 'undefined' so the code would keep breaking because it would pass to the second condition at any case.

Lms24 (approved):

Issue #3263, opened 2023-08-31
Migrate Expo Configuration Plugin


krystofwoldrich:
- This is part of https://github.com/getsentry/sentry-react-native/issues/3262

The goal is to move Expo Config Plugin from `sentry-expo` to `sentry-react-native`.

krystofwoldrich:
Released in https://github.com/getsentry/sentry-react-native/releases/tag/5.16.0

```js
const { withSentry } = require('@sentry/react-native/expo');

const config = {...};

module.exports = withSentry(config, {
  url: 'https://sentry.io/',
  project: 'project-slug', // Or use SENTRY_PROJECT env
  organization: 'org-slug', // Or use SENTRY_ORG env
});
```

PR #3391, opened 2023-11-14
add Expo config plugin

kbrandwijk:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This PR adds the Expo config plugin that used to be part of the `sentry-expo` package.
The plugin implementation is identical, except for moving the configuration properties over from the (deprecated) `postPublish` configuration in app.json to config plugin properties that is the default for config plugins.

Closes https://github.com/getsentry/sentry-react-native/issues/3263
For more details, see the linked issue and the parent issue.

Tested by building and testing the config plugin locally, and adding it to a local Expo test project to see if the config plugin was applied successfully.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## :crystal_ball: Next steps
:shipit:


krystofwoldrich:
@kbrandwijk Thank you, this is a good start. There are a few minor things we'll need to adjust.

I've changed the target to `expo` branch so we can get all changes there before merging to `main` and releasing.

I would also like to add a simple Expo app to test a build with the plugin in CI. But we can do that in another PR.

package.json
```diff
@@ -81,6 +88,8 @@
     "eslint": "^7.6.0",
     "eslint-plugin-react": "^7.20.6",
     "eslint-plugin-react-native": "^3.8.1",
+    "expo": "^47.0.0",
+    "expo-module-scripts": "^3.1.0",
```
krystofwoldrich, line 92: Could we only add `@expo/config-plugins`? 

I believe that is the only dependency used in the plugin.

package.json
```diff
@@ -18,12 +18,16 @@
...
+    "build": "yarn build:sdk && yarn downlevel && yarn build:tools && yarn build:plugin",
     "build:sdk": "tsc -p tsconfig.build.json",
     "build:sdk:watch": "tsc -p tsconfig.build.json -w --preserveWatchOutput",
     "build:tools": "tsc -p tsconfig.build.tools.json",
+    "build:plugin": "expo-module build plugin",
     "downlevel": "downlevel-dts dist ts3.8/dist --to=3.8",
     "clean": "rimraf dist coverage",
+    "clean:plugin": "expo-module clean plugin",
+    "prepare": "expo-module prepare",
+    "prepublishOnly": "expo-module prepublishOnly",
```
krystofwoldrich, line 30: These commands should not be necessary. At the moment the `yarn install` fails on the `prepare`. 

I think we can leave them out.

package.json
```diff
@@ -18,12 +18,16 @@
   },
   "main": "dist/js/index.js",
   "scripts": {
-    "build": "yarn build:sdk && yarn downlevel && yarn build:tools",
+    "build": "yarn build:sdk && yarn downlevel && yarn build:tools && yarn build:plugin",
     "build:sdk": "tsc -p tsconfig.build.json",
     "build:sdk:watch": "tsc -p tsconfig.build.json -w --preserveWatchOutput",
     "build:tools": "tsc -p tsconfig.build.tools.json",
+    "build:plugin": "expo-module build plugin",
```
krystofwoldrich, line 25:
```suggestion
    "build:plugin": "EXPO_NONINTERACTIVE=true expo-module build plugin",
```

Let's keep it without watchdog by default like the other commands.

package.json
```diff
@@ -33,10 +37,12 @@
     "lint": "yarn lint:eslint && yarn lint:prettier",
     "lint:eslint": "eslint --config .eslintrc.js .",
     "lint:prettier": "prettier --check \"{src,test,scripts}/**/**.ts\"",
+    "lint:plugin": "expo-module lint plugin",
```
krystofwoldrich, line 40: I think this might not be needed. We can add the plugin to `{projectRoot}tsconfig.json`

```
    "plugin/**/*.ts",
```

and then if needed any exception or lint changes can go to `.eslintrc.js`.

krystofwoldrich:
We can move the test to `/test/expo-plugin/*` and change the naming from `-test.ts` to `.test.ts`. 

We keep all tests in `/test`.

package.json
kbrandwijk: We recommend against adding `@expo/config-plugins` directly, because it's very easy to create version conflicts with it.

krystofwoldrich:
To keep the PR small and easy to understand I created some follow-up tasks.

- https://github.com/getsentry/sentry-react-native/issues/3393

package.json
kbrandwijk: Removed

kbrandwijk:
> @kbrandwijk So with this PR the configuration of the plugin would look like this:
> 
> `app.config.js`
> 
> ```js
> const { withSentry } = require('@sentry/react-native/plugin/build/withSentry');
> 
> const config = {
> 	name: 'my app',
> };
> 
> module.exports = withSentry(config, {
> 	url: 'https://example.sentry.url/',
> 	authToken: 'example-token',
> 	project: 'project-slug',
> 	organization: 'org-slug',
> });
> ```
> 
> OR in JSON
> 
> ```json
> {
> 	"name": "my app",
> 	"plugins": [
> 		[
> 			"@sentry/react-native/plugin/build/withSentry",
> 			{
> 				"url": "https://example.sentry.url/",
> 				"authToken": "example-token",
> 				"project": "project-slug",
> 				"organization": "org-slug"
> 			}
> 		]
> 	]
> }
> ```

Correct

package.json
krystofwoldrich: Okay, I see let's keep `expo`.

kbrandwijk:
Thank you for the PR feedback. After applying the lint suggestion, I addressed most of the issues, but was left with a few nasty ones that probably require a bit more work to fix (around the xcode types for example). I disabled those for now, I suggest we circle back to these in a follow up ticket.

krystofwoldrich:
@kbrandwijk Sounds good. Thank you.

kbrandwijk:
> > @kbrandwijk So with this PR the configuration of the plugin would look like this:
> > `app.config.js`
> > ```js
> > const { withSentry } = require('@sentry/react-native/plugin/build/withSentry');
> > 
> > const config = {
> > 	name: 'my app',
> > };
> > 
> > module.exports = withSentry(config, {
> > 	url: 'https://example.sentry.url/',
> > 	authToken: 'example-token',
> > 	project: 'project-slug',
> > 	organization: 'org-slug',
> > });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > OR in JSON
> > ```json
> > {
> > 	"name": "my app",
> > 	"plugins": [
> > 		[
> > 			"@sentry/react-native/plugin/build/withSentry",
> > 			{
> > 				"url": "https://example.sentry.url/",
> > 				"authToken": "example-token",
> > 				"project": "project-slug",
> > 				"organization": "org-slug"
> > 			}
> > 		]
> > 	]
> > }
> > ```
> 
> Correct

Actually, for end user purposes, it would just be:
```json
{
	"name": "my app",
	"plugins": [
		[
			"@sentry/react-native",
			{
				"url": "https://example.sentry.url/",
				"authToken": "example-token",
				"project": "project-slug",
				"organization": "org-slug"
			}
		]
	]
}
```

plugin/src/withSentry.ts
```diff
@@ -0,0 +1,70 @@
...
+
+const withSentry: ConfigPlugin<PluginProps | void> = (config, props) => {
+  const sentryProperties = getSentryProperties(props);
+  let cfg = config;
+  if (sentryProperties !== null) {
+    try {
+      cfg = withSentryAndroid(cfg, sentryProperties);
+    } catch (e) {
+      WarningAggregator.addWarningAndroid(
+        'sentry-expo',
```
brentvatne, line 25: it looks like we left some sentry-expo references in here, should update to this lib name

plugin/src/withSentryAndroid.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  // Use the same location that sentry-wizard uses
+  // See: https://github.com/getsentry/sentry-wizard/blob/e9b4522f27a852069c862bd458bdf9b07cab6e33/lib/Steps/Integrations/ReactNative.ts#L232
+  const pattern = /^android {/m;
+
+  if (!buildGradle.match(pattern)) {
+    WarningAggregator.addWarningAndroid(
+      'sentry-expo',
+      'Could not find react.gradle script in android/app/build.gradle. Please open a bug report at https://github.com/expo/sentry-expo.',
```
brentvatne, line 44: we should replace this url with https://github.com/getsentry/sentry-react-native

objectiveSee:
Any update on when this release will be stable? We are setting up Sentry for our app that uses Expo 49 and I am having issues with source maps when using [Expo's sentry library](https://github.com/expo/sentry-expo). If this alpha is almost ready then we might just wait to use this instead.

krystofwoldrich:
@objectiveSee We are planning two releases before the stable, which should come in the upcoming weeks.

PR #3349, opened 2023-10-17
feat(profiling): Add Native iOS profiles to Hermes JS profiles  

krystofwoldrich:
## :loudspeaker: Type of change

- [x] New feature

## :scroll: Description

This PR adds native iOS profiles to currently existing RN Hermes profiles.

Hermes and Native profiles are merged on a device and sent as one profiling event in the envelope with associated Transaction.

## :bulb: Motivation and Context

- [x] Add Support for Mixed Profiles to Sentry
	- https://github.com/getsentry/sentry/blob/9c487382d459203ebe4109f6b677507f03ca0891/src/sentry/profiles/task.py#L276
	- https://github.com/getsentry/sentry/blob/9c487382d459203ebe4109f6b677507f03ca0891/src/sentry/profiles/task.py#L347
- [x] https://github.com/getsentry/relay/pull/2629
- [x] https://github.com/getsentry/sentry/pull/47983

## :green_heart: How did you test it?
sample app, unit tests

## :pencil: Checklist

- [x] I reviewed submitted code
- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- [ ] https://github.com/getsentry/sentry/pull/58771

PR #3397, opened 2023-11-15
feat(profiling): Add RN/Android mixed profiles

krystofwoldrich:

- [x] New feature

## :scroll: Description

This PR enabled the RN SDK to send Android Profiles with an attached SentryProfile. In this case a Hermes Profile.

The Java Module implementation depends on `sentry-java` changes:
- https://github.com/getsentry/sentry-java/pull/3051
- https://github.com/getsentry/sentry-java/pull/3052
- https://github.com/getsentry/sentry-java/pull/3016
- https://github.com/getsentry/sentry-java/pull/3054

- https://github.com/getsentry/sentry-react-native/pull/3349

RN sample app

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- [x] Add JVM Debug Meta Images

### Release: 
- [x] https://github.com/getsentry/sentry-java/pull/3051
- [x] https://github.com/getsentry/sentry-java/pull/3052
- [x] https://github.com/getsentry/sentry-java/pull/3016
- [x] https://github.com/getsentry/sentry-java/pull/3054

### Blocked by
- [x] https://github.com/getsentry/sentry/issues/61142


krystofwoldrich:
This PR is ready to be reviewed.

Merge only after Sentry UI can display the JS profile attached to the Android profile.

Add note for Self hosted users to the changelog

viglia:
@krystofwoldrich on the profiling (`vroom`) side you should be unblocked now. 
I've merged this morning the last PR.

viglia (commented):
@krystofwoldrich at a high level it looks good.

@JonasBa can give a more detailed look at the `js/ts` profiler specifically.

Let me pull in @stefanosiano as well for the `android` profiler.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -633,27 +694,26 @@ public WritableMap stopProfiling() {
...
+            result.putString("profile", readStringFromFile(output));
 
-                final StringBuilder text = new StringBuilder();
-                String line;
-                while ((line = br.readLine()) != null) {
-                    text.append(line);
-                    text.append('\n');
-                }
+            WritableMap androidProfile = new WritableNativeMap();
+            byte[] androidProfileBytes = FileUtils.readBytesFromFile(end.traceFile.getPath(), maxTraceFileSize);
```
stefanosiano, line 710: is this happening in the main thread (i'm not really into react)?
In the Android layer we do it when building the envelope to send

krystofwoldrich: Currently, the RN Native Modules are not running on the main thread. But the RN docs recommends avoiding that assumption and spawning a worker thread.

But since this is a sync function also the JS code using this is sync, we would anyway have to block the thread to wait for the worker. Unless we make the JS code async too. Does that sound plausible? @stefanosiano

krystofwoldrich: Example profile: https://sentry-sdks.sentry.io/profiling/profile/sentry-react-native/0cd899c326af461b9d20a73408eaee81/flamegraph/?colorCoding=by+system+vs+application+frame&fov=0%2C0%2C5139650048%2C41&query=&sorting=call+order&tid=23160&view=top+down

stefanosiano: We can keep it as is for now. We can reevaluate if native modules will run on the main thread, to avoid other complexities (profile is still required to be sent in the same envelope of the transaction).
Also, continuous profiling (or a custom Android profiler) will likely change this, so let’s keep it as is

stefanosiano (approved):

Issue #3406, opened 2023-11-20
Cannot upload: You must either specify a release or have debug ids injected into your sources


daniel-johns-99:
**OS:**
- [ ]  Windows
- [X]  MacOS
- [ ]  Linux

**Platform:**
- [X]  iOS
- [ ]  Android

**SDK:**
- [X]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 5.13.0

**`react-native` version:** 0.67.2

**Are you using Expo?**
- [ ]  Yes
- [X]  No

**Are you using sentry.io or on-premise?**
- [X]  sentry.io (SaaS)
- [ ]  on-premise

**Configuration:**

(`@sentry/react-native`)
```
set -e
export SENTRY_PROPERTIES=sentry.properties
export EXTRA_PACKAGER_ARGS="--sourcemap-output $DERIVED_FILE_DIR/main.jsbundle.map"
export NODE_BINARY=node
WITH_ENVIRONMENT="../node_modules/react-native/scripts/xcode/with-environment.sh"
REACT_NATIVE_XCODE="../node_modules/react-native/scripts/react-native-xcode.sh"
SENTRY_XCODE="../node_modules/@sentry/react-native/scripts/sentry-xcode.sh"
BUNDLE_REACT_NATIVE="/bin/sh $SENTRY_XCODE $REACT_NATIVE_XCODE"
 
/bin/sh -c "$BUNDLE_REACT_NATIVE"
```

---
**I have the following issue:**

**Steps to reproduce:**
- Run a release build on iOS
- Note that the build fails with the following errors messages
```
Using React Native Packager bundle and source map.
Processing react-native sourcemaps for Sentry upload.
> Analyzing 2 sources
> Rewriting sources
> Adding source map references
error: Cannot upload: You must either specify a release or have debug ids injected into your sources
```

**Expected result:**

Build passes

I can't see a way to specify a release when running the sentry-xcode.sh script and I can't see any where in the React Native documentation which mentions how to have debug ids injection into the source.
Apologies for the lack of information in this bug Report but I tried passing SENTRY_LOG_LEVEL=debug and it doesnt provide any additioanl data which helps.

I'll note that based on https://github.com/getsentry/sentry-cli/blob/9e967e32ce5a2c092514f73414e9ad9aac20048a/src/utils/xcode.rs#L35 the following values are present in the build

```
[...]

PR #3409, opened 2023-11-21
fix(metro): Conditionally use Set or CountingSet in Sentry Metro plugin

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

This PR checks if `CountingSet` exists, if not NodeJS Set is used as a fallback.

CountingSet was added in Metro 0.72.0 before NodeJS Set was used.

This PR also fixed the tests so that the parent dir node modules are not used during the build of the test Apps. As due to that the e2e tests were OK although they should not.

- fixes: https://github.com/getsentry/sentry-react-native/issues/3406

e2e tests, locally run 0.65.3 build

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/tools/utils.ts
+++ b/src/js/tools/utils.ts
@@ -77,7 +77,9 @@ export function determineDebugIdFromBundleSource(code: string): string | undefin
 }
 
 /**
- * CountingSet was added in Metro 0.71.0 before that NodeJS Set was used.
+ * CountingSet was added in Metro 0.72.0 before that NodeJS Set was used.
+ *
+ * https://github.com/facebook/metro/blob/fc29a1177f883144674cf85a813b58567f69d545/packages/metro/src/lib/CountingSet.js
  */
 function resolveSetCreator(): () => CountingSet<string> {
   try {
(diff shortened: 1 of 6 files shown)


src/js/tools/utils.ts
```diff
@@ -74,3 +75,22 @@ export function determineDebugIdFromBundleSource(code: string): string | undefin
...
 }
+
+/**
+ * CountingSet was added in Metro 0.71.0 before that NodeJS Set was used.
+ */
+function resolveSetCreator(): () => CountingSet<string> {
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const { default: MetroSet } = require('metro/src/lib/CountingSet');
+    return () => new MetroSet();
```
AbhiPrasad, line 86: l: should we debug log?

AbhiPrasad (approved):
Where are the docs for `CountingSet`?

src/js/tools/utils.ts
krystofwoldrich: I was thinking about it, but the RN CLI doesn't forward the output.

krystofwoldrich:
CountingSet added here:
https://github.com/facebook/metro/blob/fc29a1177f883144674cf85a813b58567f69d545/packages/metro/src/lib/CountingSet.js
https://github.com/facebook/metro/blob/v0.72.0/packages/metro/src/lib/CountingSet.js

PR #3465, opened 2023-12-13
feat(expo): Dynamically resolve default integrations based on platform

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

Currently, the same default integrations are added for all types of builds. This causes multiple warnings and error logs during event processing. Although events are processed correctly this might confuse the SDK users.

### Before:

Screenshot 2023-12-13 at 15 10 38

### After:

Screenshot 2023-12-13 at 15 13 56

Note: The SDK Info warning will be removed in the next PR with Expo Modules implementation. 

sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


src/js/integrations/default.ts
```diff
@@ -0,0 +1,87 @@
...
+  integrations.push(new SdkInfo());
+  integrations.push(new ReactNativeInfo());
+
+  if (__DEV__ && notWeb()) {
+    integrations.push(new DebugSymbolicator());
+  }
+
+  integrations.push(createReactNativeRewriteFrames());
+
+  if (options.enableNative) {
```
lucas-zimerman, line 65: users can set this parameter to true, should we also check for `notweb()` here?

src/js/integrations/default.ts
```diff
@@ -0,0 +1,87 @@
...
+    integrations.push(new DeviceContext());
+    integrations.push(new ModulesLoader());
+    if (options.attachScreenshot) {
```
lucas-zimerman, line 68: should we warn users if debug is enabled that attachScreenshot will not be enabled because it's not supported on the web environment?

lucas-zimerman (approved):
The comments that I made also applies to the old behaviour so it's not blocking this PR, with that I am approving the PR :D

src/js/integrations/default.ts
krystofwoldrich: I see the point, I think it makes sense, but for now, I would leave the warning only for cases when we expected something to work and it doesn't. So in happy scenarios, there are no warnings.

src/js/integrations/default.ts
krystofwoldrich: I thought about that too, probably `isNativeAvailable` would be even better, since there can be the case when we are not on the web but native is not available.

But since the `enableNative` was in the past enforcing I wanted to keep that.

Issue #3490, opened 2023-12-28
Upload debug symbols fails on iOS in monorepo.


joarkosberg:
The script `sentry-xcode-debug-files.sh` does not enable overriding the `WITH_ENVIRONMENT` variable. So in a monorepo, it does not find the with environment script. 
All other environment variables I have needed to change are configurable. So would be nice if this one was too.

Looked into creating PR, with looking it up as a environment variable, but unsure if that is the preferable way.

krystofwoldrich:
Thank you, we will review the PR.

PR #3496, opened 2024-01-05
Dont use static path to react native in xcode debug files script.

joarkosberg:
- [x] Bugfix

## :scroll: Description
Issue: https://github.com/getsentry/sentry-react-native/issues/3490 
The script `sentry-xcode-debug-files.sh` refers to `WITH_ENVIRONMENT` with a static path. (unsure if static is the correct word here)
So in a monorepo for example, the path is not correct.
All other environment variables I have needed to change are configurable or dynamic. 

So update this one also to be dynamic, with a lookup like is done some other places in this respository and with help from issue comment!

* Tested in own monorepo. 

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes (I think so)

### DIFF
--- a/scripts/sentry-xcode-debug-files.sh
+++ b/scripts/sentry-xcode-debug-files.sh
@@ -6,7 +6,12 @@
 set -x -e
 
 # load envs if loader file exists (since rn 0.68)
-WITH_ENVIRONMENT="../node_modules/react-native/scripts/xcode/with-environment.sh"
+nodePath="node"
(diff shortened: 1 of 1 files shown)


krystofwoldrich:
Thank you for the PR, I've pulled your commit and tested the code, I've run into one issue, the node resolution I've originally suggested doesn't work for envs without default system node, so I've opted for simple overwrite of the `WITH_ENVIRONMENT`.

- You can check the new PR here https://github.com/getsentry/sentry-react-native/pull/3525

PR #3501, opened 2024-01-08
feat(expo): Update `getSentryExpoConfig`

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

This PR updates `getSentryExpoConfig`:
- matches method signature with `getDefaultConfig` so it can be a drop-in replacement
- adds this method to sentry expo sample application
- remove the check of `graph.transformOptions.hot` which was used to check if the function is running in development or production build but for web builds this option is always true
	-  now the function only relies on the presence of Debug ID, which is currently only created for production builds
- add extra empty lines around the warning otherwise it might be overwritten by the interactive progress bar

sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


lucas-zimerman (approved):
LGTM!

PR #3509, opened 2024-01-09
dependabot: configure to use groups for npm

mdtro:
#skip-changelog

Configure Dependabot to:

- Only open security update PRs.
- Group dependencies where necessary to keep versions aligned and reduce Dependabot spam.


krystofwoldrich (approved):
Looks good thank you!

PR #3515, opened 2024-01-10
fix(expo): Use `@sentry/react-native/expo` as plugin name in expo upload

krystofwoldrich:
Although the previous option also works,
we promote in all docs and previous messages/changelogs the `/expo` export, so we should use it here too.


lucas-zimerman (approved):
LGTM!

PR #3518, opened 2024-01-11
fix: errors from InternalBytecode.js are no longer marked as in_app

### LINKED ISSUE #3348: Do not mark `InternalBytecode.js` as `in_app` frame
### Description

React Native applications with Hermes include `InternalBytecode.js` 

https://github.com/krystofwoldrich/hermes/blob/4bcd62206c672c5b9674cab1399931ae8c275829/lib/InternalBytecode/InternalBytecode.cpp#L14

Screenshot 2023-10-17 at 12 40 10

At the moment these frames are marked as `in_app: true`, but they should not be.

lucas-zimerman:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

By default InternalBytecode is marked as inApp but in reality it's an generated file outside of the user code so it should be flagged as  false.

closes #3348

I haven't found in time an error that could be reproduced so instead, I retrieved from the internet the raw Stack Trace with users that had errors with the `InternalBytecode`, made a test with it and then I fixed the `rewriteStackFrame` in order to pass the test. 

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Errors from InternalBytecode.js are no longer marked as in_app ([#3518](https://github.com/getsentry/sentry-react-native/pull/3518))
+
 ## 5.15.2
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


krystofwoldrich (approved):
Looks good, thank you! 🚀

PR #3523, opened 2024-01-15
fix(xcode): node can't be overwritten in `sentry-xcode-debug-files.sh`

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

Before this PR the `$NODE_BINARY` could not be overwritten from the `with_enviroment` script. But only by explicitly setting the `$NODE_BINARY` env before the build. In new RN versions setting it with in the env script is common.

sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


bitsandfoxes (approved):
:shipit:

Issue #3475, opened 2023-12-15
Android Java Source Context is not uploaded


krystofwoldrich:
### Description

For Android Project [this](https://github.com/getsentry/sentry-react-native/compare/main...kw-include-java-sources) should be all that is needed to upload Java source context.

But for the RN Android projects the source code is not uploaded.

krystofwoldrich:
- assembleRelease variant is missing `assmbleProvider` in some cases, this branch adds fallback https://github.com/getsentry/sentry-android-gradle-plugin/compare/main...rz/chore/assemble-task-fallback

- combination of proguard and source context is tracked in this issue https://github.com/getsentry/sentry-java/issues/3147

PR #3532, opened 2024-01-18
misc(sample): Add SAGP JVM source context

krystofwoldrich:
This PR enables JVM source context in the react native sample application.

This PR adds how to develop with SAGP guide, which was needed to debug the source context.

 This PR adds a simple native crash module to test the source context. This is needed because the frames of the native crash in the SDK are excluded as it's not an application code. 
 
 #skip-changelog 
 
- closes https://github.com/getsentry/sentry-react-native/issues/3475


romtsn (approved):
LGTM!

PR #2933, opened 2023-03-27
feat(tracing): Add `enableTracing` option

krystofwoldrich:
## :loudspeaker: Type of change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Add `enableTracing` option

## :bulb: Motivation and Context

- closes: https://github.com/getsentry/sentry-react-native/issues/2923

## :green_heart: How did you test it?
sample app

## :pencil: Checklist

- [x] I reviewed submitted code
- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [x] All tests passing
- [x] No breaking changes

PR #3533, opened 2024-01-18
fix(tracing): Performance Tracing should be disabled by default

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

- introduced by https://github.com/getsentry/sentry-react-native/pull/2933

unit tests, sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/integrations/default.ts
+++ b/src/js/integrations/default.ts
@@ -1,4 +1,3 @@
-import { hasTracingEnabled } from '@sentry/core';
 import { HttpClient } from '@sentry/integrations';
 import { Integrations as BrowserReactIntegrations } from '@sentry/react';
 import type { Integration } from '@sentry/types';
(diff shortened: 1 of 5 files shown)


src/js/integrations/default.ts
```diff
@@ -77,7 +76,11 @@ export function getDefaultIntegrations(options: ReactNativeClientOptions): Integ
     }
   }
 
-  if (hasTracingEnabled(options) && options.enableAutoPerformanceTracing) {
+  const hasTracingEnabled =
```
kahest, line 79: `l`: maybe worth a comment why `hasTracingEnabled(options)` from core utils doesn't work

src/js/sdk.tsx
```diff
@@ -76,8 +76,11 @@ export function init(passedOptions: ReactNativeOptions): void {
     stackParser: stackParserFromStackParserOptions(passedOptions.stackParser || defaultStackParser),
     beforeBreadcrumb: safeFactory(passedOptions.beforeBreadcrumb, { loggerMessage: 'The beforeBreadcrumb threw an error' }),
     initialScope: safeFactory(passedOptions.initialScope, { loggerMessage: 'The initialScope threw an error' }),
-    tracesSampler: safeTracesSampler(passedOptions.tracesSampler),
   };
+  if ('tracesSampler' in options) {
```
kahest, line 80: why is this necessary? also, the impl of `safeTracesSampler` looks strange to my untrained (in TS/JS) eyes, it's basically `if (tracesSampler) { ... } else { return tracesSampler; }`
probably totally fine, just trying to understand & learn :)

kahest: ok yeah that's what I assumed - and we want to make sure we don't set a falsy value and therefore guard it with the added `if`?

krystofwoldrich: Yes, because if we set undefined key is created on the options object which can cause issues with the key existence checks.

kahest (approved):
LGTM! 🚀

Issue #3419, opened 2023-11-23
Optimize CI


krystofwoldrich:
### Description

Summary of CI improvements we can do to our current CI.

```[tasklist]
### Improvements
- [ ] Cache the plain E2E app builds before adding Sentry
- [ ] Create Job dependencies, only build the SDK once
- [ ] Improve e2e measurements of app start time
- [ ] Do not ran ci for changelog changes
- [x] Use Linux runners for Android e2e app builds
```

kahest:
about Linux runners, maybe useful for Android runs: https://github.com/getsentry/sentry-java/pull/3071/files#diff-fc6c05b53998713db6ac269f92471cf26730b2cf242c2460ff908b0799d9b303

PR #3545, opened 2024-01-23
Ref: Skip CI for commits without code changes

lucas-zimerman:
#3419

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Goal is to make CI faster when a PR only alters files such as changelog, readme,...
This was done with a reusable job that allows different group of jobs to know if they need to run or not.

It frees resources to Sentry (mostly the Mac Virtual machines) when small changes are made and also allow tests to pass faster when code-less changes are made.

The last PRs have a comment `TEST:...` that checks specific scenarios, for example, only changing the changelog to not trigger the build to run, altering some source file to check if the build and tests will run, breaking the Branch check (worst case) to validate that even if the job fails, the remain jobs will continue working.

## :pencil: Checklist

- [ ] I added tests to verify changes (no permanent tests but tested by simulating commits to test specific cases)
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


lucas-zimerman:
Current Commit test:
- ignore .yml files temporarily (changes added in order to skip the builds for this PR)
- .prettierignore altered and ignored

src/js/sdk.tsx
```diff
@@ -1,3 +1,4 @@
+// TEST
```
krystofwoldrich, line 1: We should remove this.
```suggestion
```

.github/workflows/skip-ci.yml
```diff
@@ -0,0 +1,36 @@
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check diff from Pull Request
+        id: check_diff
+        run: |
+          skipList=(".github/CODEOWNERS" ".prettierignore")
+          # Ignores changelog.md, readme.md,...
+          fileChangesArray=($(git diff --name-only  origin/${{ github.base_ref }}..origin/${{ github.head_ref }} | grep -v '\.md$'))
```
krystofwoldrich, line 26: We only have 5 md files, could we just add them to the skip list to make this more explicit? 

Screenshot 2024-01-31 at 12 19 00

krystofwoldrich:
Possible I've missed some benefits of the job, but could we use https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore?

lucas-zimerman:
> Possible I've missed some benefits of the job, but could we use https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore?

Yeah we can, the idea was to reuse the list across the jobs, instead of having to place it on each workflow/trigger

we can place the following snippet on each file, but it seems like the build and test jobs are stucked forever, despite falling on Tte path-ignore list ( could be a github bug)
```
    paths-ignore:
      - '.prettierignore'
      - 'CHANGELOG.md'
      - 'README.md'
      - '.github/CODEOWNERS'
      - '.github/pull_request_template.md'
      - '.github/ISSUE_TEMPLATE/BUG_REPORT.md'
      - '.github/ISSUE_TEMPLATE/FEATURE_REQUEST.md'
```

But the main difference is with paths ignore we'll have to place this snippet on each workflow and the end result will look like this
image
Where with the reusable workflow, we'll have to add an additional job to run it but the end result will not hide the jobs but mark as skipped.
image


For the time being, I applied the requested change, too bad github doesn't have more reusable options for this use case.

.github/workflows/buildandtest.yml
```diff
@@ -5,8 +5,23 @@ on:
...
   pull_request:
-
+    paths-ignore:
+      - '.prettierignore'
+      - 'CHANGELOG.md'
+      - 'README.md'
+      - '.github/CODEOWNERS'
+      - '.github/pull_request_template.md'
+      - '.github/ISSUE_TEMPLATE/BUG_REPORT.md'
+      - '.github/ISSUE_TEMPLATE/FEATURE_REQUEST.md'
```
krystofwoldrich, line 24:
```suggestion
      - '.github/ISSUE_TEMPLATE/FEATURE_REQUEST.md'

```

krystofwoldrich:
@lucas-zimerman Looks good, thank you for all the tests and screenshots. 

I will check with @kahest the required jobs settings and then we can merge it.

krystofwoldrich (approved):
I approve this to see if GH allows the merge or if it will be blocked by the required jobs.

krystofwoldrich:
This can be merged after we remove the required Test and Build jobs.

krystofwoldrich:
I reverted the PR back to skip-ci action as it's compatible with the required jobs. The ignore-paths are not working for the repository as they don't allow us to require certain jobs like Test and Build.

krystofwoldrich (approved):
Looks good! 🚀

PR #3550, opened 2024-01-25
feat(integrations): Add Spotlight integration

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Attempts to match Browser Integration as close as possible
- https://github.com/getsentry/spotlight/blob/996754bc915572165cabcf66337afd74e64f7eb8/packages/overlay/src/integrations/sentry/sentry-integration.ts#L15

This PR adds Spotlight integration for React Native and Expo applications.

The integration auto-detects the dev server hostname and expects spotlight/sidecar to be available on the same hostname.

Screenshot 2024-01-25 at 00 01 32
Screenshot 2024-01-25 at 00 01 20
Screenshot 2024-01-25 at 00 00 33

sample apps

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- [x] Add tests to verify the original envelope is not changed

### DIFF
--- a/src/js/integrations/spotlight.ts
+++ b/src/js/integrations/spotlight.ts
@@ -1,0 +1,98 @@
+import type { Client, Envelope, EventProcessor, Integration } from '@sentry/types';
+import { logger, serializeEnvelope } from '@sentry/utils';
+
+import { makeUtf8TextEncoder } from '../transports/TextEncoder';
+import { ReactNativeLibraries } from '../utils/rnlibraries';
+
+type SpotlightReactNativeIntegrationOptions = {
+  /**
+   * The URL of the Sidecar instance to connect and forward events to.
+   * If not set, Spotlight will try to connect to the Sidecar running on localhost:8969.
+   *
+   * @default "http://localhost:8969/stream"
+   */
+  sidecarUrl?: string;
+};
+
+/**
+ *
+ */
+export function Spotlight({
+  sidecarUrl = getDefaultSidecarUrl(),
+}: SpotlightReactNativeIntegrationOptions = {}): Integration {
+  logger.info('[Spotlight] Using Sidecar URL', sidecarUrl);
+
+  return {
+    name: 'Spotlight',
+
+    setupOnce(_: (callback: EventProcessor) => void, getCurrentHub) {
+      const client = getCurrentHub().getClient();
+      if (client) {
+        setup(client, sidecarUrl);
+      } else {
+        logger.warn('[Spotlight] Could not initialize Sidecar integration due to missing Client');
+      }
+    },
+  };
+};
+
+function setup(client: Client, sidecarUrl: string): void {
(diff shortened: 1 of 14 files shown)


src/js/integrations/spotlight.ts
```diff
@@ -0,0 +1,96 @@
...
+  if (!client.on) {
+    return;
+  }
+
+  client.on('beforeEnvelope', (originalEnvelope: Envelope) => {
+    // TODO: This is a workaround for spotlight/sidecar not supporting images
+    const spotlightEnvelope: Envelope = [...originalEnvelope];
+    const envelopeItems = [...originalEnvelope[1]].filter(
+      item => typeof item[0].content_type !== 'string' || !item[0].content_type.startsWith('image'),
+    );
```
Lms24, line 53: Hmm this is a good point. I think we just support text data at the moment.

I think we should handle this in the sidecar somehow because we don't want SDKs to have too much of that logic. Ideally, Spotlight (similarly to Relay lol) accepts all kinds of payloads and just deals with removing incompatible stuff on its end. 

Additionally, WDYT, would it be valuable to render the images in the Spotlight UI?

src/js/integrations/spotlight.ts
```diff
@@ -0,0 +1,96 @@
...
+        err,
+      );
+    });
+  });
+}
+
+function getDefaultSidecarUrl(): string {
+  try {
+    const { url } = ReactNativeLibraries.Devtools?.getDevServer();
+    return `http://${getHostnameFromString(url)}:8969/stream`;
```
Lms24, line 76: neat!

Lms24 (approved):
Nice! I assume the usage instructions are gonna be similar to web frontend in the sense that we tell users to add the integration in dev mode? Can RN tree-shake the integration out in prod builds? 

Other than that, LGTM!

src/js/integrations/spotlight.ts
Lms24: Btw, no change required from my end. The workaround is fine for now, I just think we should handle this in the spotlight sidecar in the future

src/js/integrations/spotlight.ts
krystofwoldrich: Thank you.

src/js/integrations/spotlight.ts
krystofwoldrich: I think rendering the screenshots would be especially useful for crashes/uncaught errors. So the developer doesn't have to try to crash the app again and carefully watch what was on the screen at that moment (or create a screenrecording).

krystofwoldrich:
Yes, we tell users to enable it in dev builds, the same as on JS. Sadly RN only removes unimported files, so the integration code stays. This is something to revisit when RN supports tree shaking.

PR #3553, opened 2024-01-25
fix(debugsymbolicator): Don't trace debug symbolicator source context Metro Dev Server requests

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description
Example Erro produced by the DebugSymbolicator integration when trying to download the Source Context for Development Builds Symbolication. https://sentry-sdks.sentry.io/issues/4058613036/events/5612d59c349740c79bbd3c01a317ad8e/

These requests should not be traced by Sentry, as they are part of the SDK implementation and present only in dev builds.

Screenshot 2024-01-25 at 16 51 12

sample apps

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


src/js/utils/xhr.ts
```diff
@@ -0,0 +1,51 @@
...
+export const XHR_READYSTATE_DONE = 4;
+
+let originalXhrOpen: XMLHttpRequest['open'] | null = null;
+let originalXhrSend: XMLHttpRequest['send'] | null = null;
+
+/**
+ * Saves original XHR methods so that they can be used later
+ */
+export function preserveXMLHttpRequest(customGlobal: { XMLHttpRequest?: typeof XMLHttpRequest } = RN_GLOBAL_OBJ): void {
+  if (!customGlobal.XMLHttpRequest || !customGlobal.XMLHttpRequest.prototype) {
```
mydea, line 20: Another way to achieve this could also be like I did here for Spotlight: https://github.com/getsentry/spotlight/pull/335

that's fetch, but may also work here - basically we leverage the `__sentry_original__` prop we keep on monkeypatched APIs:

```ts
export function getNativeFetchImplementation(): FetchImpl {
  if (fetchIsWrapped(window.fetch)) {
    return window.fetch.__sentry_original__;
  }

  return window.fetch;
}

function fetchIsWrapped(fetchImpl: FetchImpl): fetchImpl is WrappedFetchImpl {
  return '__sentry_original__' in fetchImpl;
}
```

krystofwoldrich: Thank you, I'll use that.

mydea (approved):
LGTM!

PR #3559, opened 2024-01-27
fix(spotlight): Stop tracing Sidecar network request in JS

krystofwoldrich:

- [x] Bugfix
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

The current approach makes the network request from the Spotlight Integration to the Sidecart visible in breadcrumbs, performance, and errors if one occurs. This PR changes to the StealthXhr implementation which uses the `__sentry_original__` methods, so the calls are not traced by Sentry.

sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


src/js/integrations/spotlight.ts
```diff
@@ -56,19 +57,26 @@ function sendEnvelopesToSidecar(client: Client, sidecarUrl: string): void {
...
-        'Content-Type': 'application/x-sentry-envelope',
-      },
-      mode: 'cors',
-    }).catch(err => {
-      logger.error(
-        "[Spotlight] Sentry SDK can't connect to Spotlight is it running? See https://spotlightjs.com to download it.",
-        err,
-      );
-    });
+    const xhr = createStealthXhr();
```
lucas-zimerman, line 60: XMLHttpRequest is always available with Sidecar? because createStealthXhr can throw an exception.

krystofwoldrich: `XMLHttpRequest` is always available in RN as far as I know.

krystofwoldrich: I've changed the throw to null return.

mydea (approved):

Issue #3569, opened 2024-01-30
Make sentry-expo-upload-sourcemaps read ENV files


mrljsh:
The ﻿sentry-expo-upload-sourcemaps script currently requires users to manually set environment variables when invoking the command-line interface. This process could be streamlined by adding support for automatically reading environment variables from ﻿.env files.

I would make PR for this but I am unsure if you would use dotenv or some other external tool for reading .env files.

krystofwoldrich:
@mrljsh Thank you for creating a separate issue,
please feel free to create a PR!

Thank you for considering the inclusion of other libraries, since we will likely only need the parse function of `dotenv`, we can vendor it. You can copy it from the link and allow maintainers to edit the PR we will add the appropriate license and annotations.

https://github.com/motdotla/dotenv/blob/cfd735d7d00f01e3772a35b541b1b141fd5a5c5c/lib/main.js#L9-L48

Issue #3546, opened 2024-01-23
Third party frames are not symbolicated


pax-k:
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [x]  iOS
- [x]  Android

**SDK:**
- [x]  `"@sentry/react-native": "5.16.0"`

**SDK version:** 5.16.0

**`react-native` version:** 0.73.2

**Are you using Expo?**
- [x]  Yes
- [ ]  No

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)
- [ ]  on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

[[Link to issue]](https://hyphen-ul.sentry.io/issues/4907607577/?project=4504792017666048&query=is%3Aunresolved&referrer=issue-stream&stream_index=3)

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  enabled: ENV.APP_VARIANT !== 'development',
  dsn: ENV.SENTRY_DSN,
  debug: false,
  attachStacktrace: true,
  enableAutoSessionTracking: true,
  attachViewHierarchy: true,
  environment: ENV.APP_VARIANT,
  tracesSampleRate: 1.0,
  enableNdkScopeSync: true,
  _experiments: {
    profilesSampleRate: 1.0,
  },
  enableAppHangTracking: true,
  appHangTimeoutInterval: 5,
  enableAutoPerformanceTracing: true,
  enableTracing: true,
  enableCaptureFailedRequests: true,
  enableWatchdogTerminationTracking: true,
  sessionTrackingIntervalMillis: 10000,
  integrations: [new CaptureConsole({ levels: ['warn', 'error'] }), new HttpClient()],
  beforeBreadcrumb: (breadcrumb) => {
    if (
      ENV.APP_VARIANT === 'production' &&
      breadcrumb.category === 'console' &&
      breadcrumb.level === 'debug'
    ) {
      return null;
    }
    return breadcrumb;
  },
});
```

(`app.config.js`)

```
export default {
  expo: {
    name: getVariantConfigs()[APP_VARIANT].name,
    slug: 'xxx',
    owner: 'zzz',
    version: '0.1.7',
    orientation: 'portrait',
    icon: './assets/icon.png',
    userInterfaceStyle: 'light',
    scheme: 'xxx',
    jsEngine: 'hermes',
    sdkVersion: '50.0.0',
    splash: {
[...]

PR #3571, opened 2024-01-30
feat(expo): read .env file in upload script

quinlanj:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Read from `.env` file if it exists in the user's project directory, based on https://github.com/getsentry/sentry-react-native/issues/3546#issuecomment-1911987720

Fixes https://github.com/getsentry/sentry-react-native/issues/3569

Most of the expo commands are of [this form](https://github.com/expo/expo/blob/97005b5a2528c2349e2c27d6d678f85e6f8eb754/packages/%40expo/cli/src/prebuild/prebuildAsync.ts#L65-L66): 
1. Set `NODE_ENV` to development if unset (unless if the command is `expo export`, where the metro bundler will output differently depending on NODE_ENV so we default it to production)
2. Read from `.env` file - `@expo/env` is a dependency of the `expo/cli` so we make a best effort to load it at runtime. 

test project

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/scripts/expo-upload-sourcemaps.js
+++ b/scripts/expo-upload-sourcemaps.js
@@ -106,7 +106,7 @@ const projectRoot = '.'; // Assume script is run from the project root
 try {
   require('@expo/env').load(projectRoot);
 } catch (error) {
-  console.warn('⚠️ Failed to load environment variables from @expo/env.');
+  console.warn('⚠️ Failed to load environment variables using @expo/env.');
   console.warn(error);
 }
 
--- a/scripts/expo-upload-sourcemaps.js
+++ b/scripts/expo-upload-sourcemaps.js
@@ -101,7 +101,6 @@ function groupAssets(assetPaths) {
   return groups;
 }
 
-process.env.NODE_ENV = process.env.NODE_ENV || 'development';
 const projectRoot = '.'; // Assume script is run from the project root
 try {
   require('@expo/env').load(projectRoot);
--- a/scripts/expo-upload-sourcemaps.js
+++ b/scripts/expo-upload-sourcemaps.js
@@ -101,6 +101,7 @@ function groupAssets(assetPaths) {
   return groups;
 }
 
+process.env.NODE_ENV = process.env.NODE_ENV || 'development'; // Ensures precedence .env.development > .env (the same as @expo/cli)
 const projectRoot = '.'; // Assume script is run from the project root
 try {
   require('@expo/env').load(projectRoot);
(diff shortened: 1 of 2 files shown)


scripts/expo-upload-sourcemaps.js
```diff
@@ -101,6 +101,15 @@ function groupAssets(assetPaths) {
   return groups;
 }
 
+process.env.NODE_ENV = process.env.NODE_ENV || 'development';
+const projectRoot = '.'; // Assume script is run from the project root
+try {
+  require('@expo/env').load(projectRoot);
+} catch (error) {
+  console.warn('⚠️ Failed to load environment variables from @expo/env.');
```
krystofwoldrich, line 109:
```suggestion
  console.warn('⚠️ Failed to load environment variables using @expo/env.');
```

To me `from` sounds a bit confusing.

scripts/expo-upload-sourcemaps.js
```diff
@@ -101,6 +101,15 @@ function groupAssets(assetPaths) {
   return groups;
 }
 
+process.env.NODE_ENV = process.env.NODE_ENV || 'development';
```
krystofwoldrich, line 104: Does it have any side effects? Because if it doesn't I think we can leave it out.

scripts/expo-upload-sourcemaps.js
```diff
@@ -101,6 +101,15 @@ function groupAssets(assetPaths) {
...
+const projectRoot = '.'; // Assume script is run from the project root
```
krystofwoldrich, line 105: I think that is safe to assume. ✅

scripts/expo-upload-sourcemaps.js
quinlanj: So `@expo/env` is a shim over the dotenv package, where falling back to 'development' would affect the precedence of `.env` files that are applied: https://github.com/bkeepers/dotenv?tab=readme-ov-file#customizing-rails

For most of the expo commands, we fallback to `development` unless there is a good reason to do something else (notable exceptions are defaulting to `production` for metro bundling purposes) . I can remove it though, I understand that this may add too much complexity for a simple script

krystofwoldrich: Thank you for pointing that out, let's keep the fallback and add a small comment about why.

krystofwoldrich (approved):
Thank you, looks good. 🚀

PR #3588, opened 2024-02-12
feat(tracing): Add automatic tracing of time to initial display for react-navigation

krystofwoldrich:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This PR adds automatic TTID spans and measurements for React Navigation version 5 and newer. When TTID enables new `navigation.processing` span is added. This represents the duration from navigation dispatch to the new screen resolved in JS. This was the original minimum transaction duration.

### Technical details

On iOS React Navigation uses ViewControllers to create RN Screens. We swizzle the RNSScreen view controller `viewDidAppear` to start listening for the next new frame. We can't start listening to the new frame when the JS layer is done with the navigation, because the next frame after that will be from the old screen/animation to the new screen.

A current limitation is that this works only for native view controller animations. Animations initiated from JS happen after the view `didAppear`. 

On Android React Navigation creates Fragments to create RN Screens. We hook into `onFragmentViewCreated` where we check if the fragment is `RNSScreenFragment`. If that is we hook onto the event dispatcher attached to that fragment and wait for `appeareEvent` and then for the next rendered frame.

We can't use the JS `viewApear` hook as this is processed by the JS Engine asynchronously from the main application thread and might happen after (depending on the JS loop queue) the first frame of the screen is rendered.

### Known limitations

- Needs native code so it doesn't work in Expo Go
- On iOS JS initiated animations, happen after the view controller first frame
- There doesn't seem to be a way to hook into the View event dispatcher used by the RNScreens on iOS, so we had to use swizzling.

### How to enable TTID

```js
import * as Sentry from "@sentry/react-native";

const routingInstrumentation = new Sentry.ReactNavigationInstrumentation({
});

Sentry.init({
  integrations: [
    new Sentry.ReactNativeTracing({routingInstrumentation}),
  ],
});
```

### Examples in Sentry

Example 1: https://sentry-sdks.sentry.io/discover/sentry-react-native:7432443eb31a4166be70d4b8d63b040b
Screenshot 2024-02-16 at 10 40 30

Example 2: https://sentry-sdks.sentry.io/discover/sentry-react-native:4429cff3025b4ec8aa2b128f1d564459
Screenshot 2024-02-16 at 10 43 10

sample app, unit and integration tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- [x] https://github.com/getsentry/sentry-cocoa/pull/3683
- [x] https://github.com/getsentry/sentry-cocoa/pull/3684
- [x] Bump `sentry-cocoa`, above changes will be included in `8.21.0`


ios/RNSentryRNSScreen.m
```diff
@@ -0,0 +1,26 @@
+#import <Sentry/SentryFramesTracker.h>
+#import <Sentry/SentryDependencyContainer.h>
+#import <Sentry/SentrySwizzle.h>
+
+#import "RNSentryRNSScreen.h"
+#import "RNSentryDependencyContainer.h"
+
+@implementation RNSentryRNSScreen
+
++ (void)swizzleViewDidAppear {
```
philipphofmann, line 10: `h`: Using swizzling to start the frames tracker seems weird to me. Why did you choose that approach?

krystofwoldrich: The frame tracker is already running ([initialized here](https://github.com/getsentry/sentry-react-native/blob/dfadf15d2fb88910293eec14f78bf8c9670b90f6/ios/RNSentryDependencyContainer.m#L30)), 
but the listener is added in the swizzled `viewDidAppear` of the RNSScreen ([added listener](https://github.com/getsentry/sentry-react-native/blob/dfadf15d2fb88910293eec14f78bf8c9670b90f6/ios/RNSentryFramesTrackerListener.m#L27)).

philipphofmann: Why do you need the listener? We can also discuss this on a call and document the outcome here. I'm unsure if it's required.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -123,21 +126,53 @@ public class RNSentryModuleImpl {
...
+      return this.reactApplicationContext.getCurrentActivity();
+    }
+
+    private @NotNull Runnable createEmitNewFrameEvent() {
+        final @NotNull SentryDateProvider dateProvider = new SentryAndroidDateProvider();
+
+        return () -> {
+          final SentryDate endDate = dateProvider.now();
+          WritableMap event = Arguments.createMap();
+          event.putDouble("newFrameTimestampInSeconds", endDate.nanoTimestamp() / 1e9);
```
markushi, line 154: Not sure if it's fine, but `nanoTimestamp()` will only give you a millisecond precision, [see the issue linked here](https://github.com/getsentry/sentry-java/issues/3210).
I see the duration is calculated via `this._latestTransaction?.startTimestamp`, but couldn't find the setter for `.startTimestamp`.

markushi (approved):
Looking good! I mainly focused on the Android parts, left one comment up for discussion

ios/RNSentryDependencyContainer.m
```diff
@@ -0,0 +1,35 @@
...
+
++ (instancetype)sharedInstance
+{
+  return instance;
+}
+
+- (instancetype)init
+{
+    return self;
+}
```
brustolin, line 25: l: A good practice for singletons in Objc would be like the following. 
Your code is good, so no need to change it if you dont want to, just a objc tip.

```suggestion
@implementation RNSentryDependencyContainer {
     NSObject *sentryDependencyContainerLock;
 }

+ (instancetype)sharedInstance
{
    static RNSentryDependencyContainer *instance = nil;
    static dispatch_once_t onceToken;
    dispatch_once(&onceToken, ^{ instance = [[self alloc] init]; });
    return instance;
}

- (instancetype)init
{
    if (self = [super init]) {
        sentryDependencyContainerLock = [[NSObject alloc] init];
    }
    return self;
}
```

ios/RNSentryRNSScreen.m
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  SEL selector = NSSelectorFromString(@"viewDidAppear:");
+  SentrySwizzleInstanceMethod(rnsscreenclass, selector, SentrySWReturnType(void),
+      SentrySWArguments(BOOL animated), SentrySWReplacement({
+          [[[RNSentryDependencyContainer sharedInstance] framesTrackerListener] startListening];
+          SentrySWCallOriginal(animated);
+      }),
+      SentrySwizzleModeOncePerClass, (void *)selector);
+}
```
brustolin, line 24: h: Now that Im reading the full code, I dont think you should start listening when the viewDidAppear is called,
you can start listening at the beginning of the app. When viewDidAppear is called you just report a for the ongoing UI transaction.

krystofwoldrich: So instead of adding a listener in `viewDidAppear` I would set some flag and then I would wait for the next frame?

Won't that cause https://github.com/getsentry/sentry-react-native/blob/89818402b1846d4170a03414a4b495d63433b8fc/ios/RNSentryFramesTrackerListener.m#L16 to be executed for every frame in the app?

Or did I misunderstand?

brustolin: I see. you remove the listener from frame track every time you emit and event. I haven't notice it before.
Well it's almost like using a flag, yeah, I dont think you need to change this.

brustolin (approved):
It looks good from the cocoa point of view! 😉

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
krystofwoldrich: Thank you for the note, the RN measurements have the nanoseconds precision ([RN impl of performance now](https://github.com/facebook/react-native/blob/d09c02f9e2d468e4d0bde51890e312ae7003a3e6/packages/react-native/ReactCommon/cxxreact/JSExecutor.cpp#L28)), this is used for the `startTimestamp`.

But it doesn't seem to be an issue, looking at the test transactions.

src/js/tracing/reactnavigation.ts
```diff
@@ -188,8 +222,51 @@ export class ReactNavigationInstrumentation extends InternalRoutingInstrumentati
...
+
+          !routeHasBeenSeen &&
+            this._newScreenFrameEventEmitter?.once(
+              NewFrameEventName,
+              ({ newFrameTimestampInSeconds }: NewFrameEvent) => {
+                if (!latestTtidSpan) {
+                  return;
+                }
+
+                if (spanToJSON(latestTtidSpan).parent_span_id !== getActiveSpan()?.spanContext().spanId) {
```
lucas-zimerman, line 243: why are you using `spanToJSON(latestTtidSpan).parent_span_id` instead of `latestTtidSpan.parentSpanId`?

krystofwoldrich: Because `latestTtidSpan.parentSpanId` will be removed in JS SDK v8.

PR #3600, opened 2024-02-14
fix(ci): Only check for diff in to skip ci in pull requests

krystofwoldrich:
- follow up to https://github.com/getsentry/sentry-react-native/pull/3545

Skip CI job should check the diff only in pull requests.

#skip-changelog


lucas-zimerman (approved):
LGTM!

PR #3630, opened 2024-02-26
fix(expo): Ensure authToken is not written to application package

krystofwoldrich:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Processed Expo plugin configuration is saved in plain text in the application package.

This PR removes the `authToken` property to ensure the value won't be saved in the app package.

expo sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- [ ] Update docs to emphasize use of `SENTRY_AUTH_TOKEN` env
- [ ] Inform users about the possibly exposed token

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixes
 
 - Option `enabled: false` ensures no events are sent ([#3606](https://github.com/getsentry/sentry-react-native/pull/3606))
+- Remove Expo Plugin `authToken` option from application bundle ([#3630](https://github.com/getsentry/sentry-react-native/pull/3630))
 
 ### Dependencies
 
--- a/plugin/src/withSentry.ts
+++ b/plugin/src/withSentry.ts
@@ -14,6 +14,12 @@ interface PluginProps {
 
 const withSentryPlugin: ConfigPlugin<PluginProps | void> = (config, props) => {
   const sentryProperties = getSentryProperties(props);
+
+  if (props && props.authToken) {
+    // If not removed, the plugin config with the authToken will be written to the application package
+    delete props.authToken;
+  }
+
   let cfg = config;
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,11 @@
 ### Fixes
 
 - Option `enabled: false` ensures no events are sent ([#3606](https://github.com/getsentry/sentry-react-native/pull/3606))
+- Remove Expo Plugin `authToken` option from application bundle ([#3630](https://github.com/getsentry/sentry-react-native/pull/3630))
+  - Expo Configuration is saved in plain text and is not secure.
+    Please, rotate your token if you published an app with
+    the auth token in the plugin config. Use `SENTRY_AUTH_TOKEN` env.
+    Read more in the [docs](https://docs.sentry.io/platforms/react-native/manual-setup/expo/).
```
kahest, line 16:
```suggestion
- Don't add Expo Plugin option `authToken` to application bundle ([#3630](https://github.com/getsentry/sentry-react-native/pull/3630))
  - Expo plugin configurations are generelly stored in plain text, and are also automatically added to built app bundles, and are therefore considered insecure.
  - You should not set the auth token in the plugin config except for local testing. Instead, use the `SENTRY_AUTH_TOKEN` env variable, as pointed out in our [docs](https://docs.sentry.io/platforms/react-native/manual-setup/expo/).
  - In addition to showing a warning, we are now actively removing an `authToken` from the plugin config if it was set.
  - If you had set the auth token in the plugin config previously, **and** built and published an app with that config, you should [rotate your token](https://docs.sentry.io/product/accounts/auth-tokens/).
```

kahest (approved):

PR #3654, opened 2024-03-05
feat(performance): Add Time to Full Display and manual API for TTID

krystofwoldrich:
- [x] New feature

## :scroll: Description

This PR adds manual API for measuring TTID and TTFD.

TTID and TTFD use React Native Native Components, `android.view.View` on Android and `UIView` on iOS. This enables the SDK to reliably hook into the rendering loop and wait for the next frame after the native component is created.

Why the need for native components?

We tested exposing the hook for the next frame to JS runtime, which turned out unreliable. When waiting for the component to mount and receiving the mount event back JS and based on that starting listening for the next frame, depending on the JS loop, the listen for the next frame might be significantly (10s, 100s ms) delayed. This only works in ideal scenarios, when JS runtime starts the listen immediately after receiving the mount event. 

What does the API look like?

When combined with auto instrumentation only the components are needed. The Time To Display is only measured when `record=true`, any consequent re-renders are ignored until a new `activeSpan`.

The Time To Display spans can be started manually.

```js
const initialDisplaySpan = startTimeToInitialDisplaySpan();
const fullDisplaySpan = startTimeToFullDisplaySpan();

function MyComponent() {
	return <View>
		<TimeToInitialDisplay record={true} />
		<TimeToFullDisplay record={false} />
	</View>
}
```

sample app, unit and integration tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- [ ] Follow up PR with native implementation for new architecture
- [ ] Add the documentation


src/js/tracing/timetodisplay.tsx
```diff
@@ -0,0 +1,238 @@
...
+    return;
+  }
+
+  span.end(frameTimestampSeconds);
+
+  span.setStatus('ok');
+  logger.debug(`[TimeToDisplay] ${spanToJSON(span).description} span updated with end timestamp.`);
+
+  setSpanDurationAsMeasurement('time_to_full_display', span);
+}
```
brustolin, line 238: h: We need to check whether TTID span is finished already, otherwise we need to wait for it and then automatically close TTFD with the same end.

brustolin (commented):
Cocoa code looks good.
There is one thing to fix in the TypeScript behaviour.

markushi (approved):
Looking good, left a few minor comments

src/js/tracing/timetodisplay.tsx
krystofwoldrich: Thank you, I fixed it here https://github.com/getsentry/sentry-react-native/pull/3654/commits/c06e638609e43b8bf1b0263fbea25dfc26a251ad

android/src/main/java/io/sentry/react/RNSentryOnDrawReporterManager.java
```diff
@@ -0,0 +1,141 @@
...
+                return;
+            }
+
+            @Nullable Activity activity = reactContext.getCurrentActivity();
+            if (activity == null || emitter == null || buildInfo == null) {
+                return;
+            }
+
+            FirstDrawDoneListener
+                    .registerForNextDraw(activity, emitter, buildInfo);
```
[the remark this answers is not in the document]
krystofwoldrich: I checked with our sample app, the code is executed on the main thread.

brustolin (approved):
LGTM!

PR #3658, opened 2024-03-05
fix(ci): `diff_check` should not fail when all files are filtered

krystofwoldrich:
This PR fixes `skip-ci` jobs which failed when only `*.md` files were changed.

`grep -v '\.md$'` returns code `1` when all lines are filtered out.

#skip-changelog


lucas-zimerman (approved):
lgtm!

PR #3664, opened 2024-03-06
feat(expo): Added ability to provide sentryUrl for source maps upload

Acetyld:
## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
Added sentryUrl for people that are self hosting

## :bulb: Motivation and Context
Its also possible to define URL inside app.json and this was missing from the command.

## :green_heart: How did you test it?
I tested it locally.

## :pencil: Checklist

- [X] I reviewed submitted code
- [ ] I added tests to verify changes
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] All tests passing
- [ ] No breaking changes

PR #3679, opened 2024-03-14
feat(expo): Added ability to provide sentryUrl for source maps upload

krystofwoldrich:
- Pulled https://github.com/getsentry/sentry-react-native/pull/3664

Fixed changelog

#skip-changelog


kahest (approved):

yukukotani:
@krystofwoldrich Is it intended that `sentryUrl` is required? IMO, `sentryUrl` should be optional with the default value `https://sentry.io/`.

PR #3688, opened 2024-03-19
misc(expo-sample): Add expo-router auto instrumentation

krystofwoldrich:
Follow our docs and move `Sentry.init` to `app/_layout`

https://docs.sentry.io/platforms/react-native/manual-setup/expo/

This is important especially when the default route changes and async routes are enabled. Sentry would not load until the route would be required.

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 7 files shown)


lucas-zimerman (approved):
LGTM!

PR #3689, opened 2024-03-19
misc(sample): Do not execute Sentry in NodeJS during npx expo export

krystofwoldrich:
During Expo export, the JS code is executed, which spams the export output.

This PR adds env to the sample app to avoid this.

#skip-changelog


lucas-zimerman (approved):
LGTM!

PR #3691, opened 2024-03-19
misc(sample): Disable Expo Web static routes, export command hangs

krystofwoldrich:
Disable static output in the Expo sample app, this option causes the export command to hang.

I was  unable to determine why, but it happens when running the command multiple times, which is common during the SDK development.

- related issue https://github.com/expo/eas-cli/issues/1907


krystofwoldrich:
Seems like a local machine issue, closing the PR.

PR #3699, opened 2024-03-20
fix(timetodisplay): Change TimeToDisplay unsupported message from error to warning

krystofwoldrich:

- [x] Enhancement
- [x] Refactoring

## :scroll: Description

The error log will trigger a stack trace log in React which is an unwanted side-effect. The log then looks like a crash/uncaught error. Plus the stack trace is not symbolicated so it's not useful.

This PR changes the log to a warning and uses Timeout to avoid the React stack printing.

Before:
Screenshot 2024-03-20 at 19 16 30

After:
Screenshot 2024-03-21 at 16 37 44
Screenshot 2024-03-21 at 16 37 39

sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/tracing/timetodisplay.tsx
+++ b/src/js/tracing/timetodisplay.tsx
@@ -61,7 +61,9 @@ function TimeToDisplay(props: {
 
   if (__DEV__ && !nativeComponentMissingLogged && !nativeComponentExists) {
     nativeComponentMissingLogged = true;
-    logger.log('TimeToInitialDisplay and TimeToFullDisplay are not supported on the web, Expo Go and New Architecture. Run native build or report an issue at https://github.com/getsentry/sentry-react-native');
+    setTimeout(() => {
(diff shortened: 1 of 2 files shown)


src/js/tracing/timetodisplay.tsx
```diff
@@ -61,7 +61,9 @@ function TimeToDisplay(props: {
 
   if (__DEV__ && !nativeComponentMissingLogged && !nativeComponentExists) {
     nativeComponentMissingLogged = true;
-    logger.error('RNSentryOnDrawReporter is not available on the web, Expo Go and New Architecture. Run native build or report an issue at https://github.com/getsentry/sentry-react-native');
+    setTimeout(() => {
```
lucas-zimerman, line 64: Just a NIT so future contributors understand why we are using setTimeout here.

```suggestion
    // Using setTimeout with a delay of 0 milliseconds to defer execution and avoid printing the React stack trace.
    setTimeout(() => {
```

lucas-zimerman (approved):
LGTM! (added a nit comment in the code for future reference)

PR #3721, opened 2024-03-27
feat(integrations): Update integration to new JS interface

krystofwoldrich:
Updates the React Native integration to use the new factory functions returning plain JS object style.

- needs https://github.com/getsentry/sentry-react-native/pull/3720


src/js/integrations/debugsymbolicator.ts
```diff
@@ -20,255 +19,163 @@ export type ReactNativeError = Error & {
...
-          getFramesToPop(hint.originalException as Error),
-        );
-        symbolicatedFrames && this._replaceExceptionFramesInEvent(event, symbolicatedFrames);
-      } else if (hint.syntheticException && isErrorLike(hint.syntheticException)) {
-        // syntheticException is Error object
-        const symbolicatedFrames = await this._symbolicate(
-          hint.syntheticException.stack,
-          getFramesToPop(hint.syntheticException),
-        );
+export const debugSymbolicatorIntegration: IntegrationFn = () => {
```
AbhiPrasad, line 22: we should backport the integrations to a v5 release so that users can migrate to the functional integrations more incrementally.

You can look at https://github.com/getsentry/sentry-javascript/blob/fdcd1adfe65becd9c605412cdd474eb676714bc4/packages/integrations/src/dedupe.ts#L9-L45 as an example of this in `v7`.

AbhiPrasad (changes requested):

src/js/integrations/debugsymbolicator.ts
krystofwoldrich: Yes, I agree, I'll prepare it.

krystofwoldrich:
- closed in favour of backwards compatible changes from https://github.com/getsentry/sentry-react-native/pull/3814

Issue #3724, opened 2024-03-28
Enabling Sentry.ReactNativeTracing results in Cannot read properties of undefined (reading 'remove') when sending error.


the-simian:
**OS:**
- [x]  Windows
- [x]  Linux

**Platform:**
- [ ]  iOS
- [x]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)

**SDK version:** 5.19.1

**`react-native` version:** 0.73.6

**Are you using Expo?**
- [x]  Yes

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)

**If you are using sentry.io, please post a link to your issue so we can take a look:**
[The issue is logged but breaks the dev server.](https://lifeguides.sentry.io/issues/5085465590/?environment=development&project=4506758892093440&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=0)

[Link to issue]

**Configuration:**

(`@sentry/react-native`)
```ts
import * as Sentry from '@sentry/react-native';
import { Slot, useNavigationContainerRef } from 'expo-router';

const routingInstrumentation = new Sentry.ReactNavigationInstrumentation();

Sentry.init({
  dsn: 'My-DSN',
  debug: true,
  tracesSampleRate: 1.0,
  integrations: [
    new Sentry.ReactNativeTracing({
        enableNativeFramesTracking: Constants.appOwnership !== 'expo',
      enableUserInteractionTracing: true,
      routingInstrumentation,
    }),
  ],
});

export function RootLayout() {
  useEffect(() => {
    if (navRef) {
      routingInstrumentation.registerNavigationContainer(navRef);
    }
  }, [navRef]);

   return <Slot />
}



export default Sentry.wrap(RootLayout);
```

---
**I have the following issue:**

When I willfully throw a test error via Sentry, with the above configuration, the dev server is halted with an error emitting from sentry itself. Here is the trace:

```sh
TypeError: Cannot read properties of undefined (reading 'remove')
    at callback (/home/the-simian/<app>/node_modules/@sentry/react-native/dist/js/tracing/transaction.js:28:22)
    at IdleTransaction.end (/home/the-simian/<app>/node_modules/@sentry/core/cjs/tracing/idletransaction.js:160:9)
    at IdleTransaction.finish (/home/the-simian/<app>/node_modules/@sentry/core/cjs/tracing/span.js:475:17)
[...]

PR #3727, opened 2024-04-02
fix(perf): Cancel spans in background doesn't crash in environments without AppState

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

Some RN environments do not have App State available, we have to check for it before use. Cancel background spans func was missing this check. 

- fixes: https://github.com/getsentry/sentry-react-native/issues/3724

sample app, tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


lucas-zimerman (approved):
Nice one, LGTM!

PR #3738, opened 2024-04-04
fix(ci): Ensure `diff-check` action works with PRs from forks

krystofwoldrich:
This PR makes the `diff-check` more generic, so the check works also for PR from forks.

#skip-changelog


lucas-zimerman (approved):

lucas-zimerman (approved):
LGTM!

PR #3740, opened 2024-04-04
fix(ci): Only install `xcbeautify` if missing

krystofwoldrich:
The new version of `xcbeautify`, 2.x.x, is not compatible with the `macos12` runners, but the runners include an older version, 1.6.0.

- https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md
- https://github.com/cpisciotta/xcbeautify/releases/tag/2.0.1

This PR adds a check to only install the tool if missing.

#skip-changelog


lucas-zimerman (approved):
LGTM!

PR #3759, opened 2024-04-17
fix(sample): Remove duplicate ruby version constrain in RN sample

krystofwoldrich:
#skip-changelog 

The duplicate constrain from `.ruby-version` and `Gemfile` makes ruby version managers fail to select correct Ruby version.


lucas-zimerman (approved):
LGTM!

PR #3767, opened 2024-04-18
fix(sample): Fix bare sample Pods to work with dynamic frameworks

krystofwoldrich:
Fixes dynamic frameworks builds and add test to CI for it.

#skip-changelog 

To save CI time I only included legacy architecture and dynamic frameworks. The new arch and dynamic doesn't seem to give any new errors/

### DIFF
--- a/samples/react-native/ios/Podfile
+++ b/samples/react-native/ios/Podfile
@@ -23,6 +23,8 @@ if linkage != nil
   use_frameworks! :linkage => linkage.to_sym
   pre_install do |installer|
     installer.pod_targets.each do |pod|
+      # RNScreens and RNReanimated don't work when dynamically linked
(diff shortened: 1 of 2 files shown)


samples/react-native/ios/Podfile
```diff
@@ -21,6 +21,15 @@ linkage = ENV['USE_FRAMEWORKS']
...
   use_frameworks! :linkage => linkage.to_sym
+  pre_install do |installer|
+    installer.pod_targets.each do |pod|
+      if pod.name.eql?('RNReanimated') || pod.name.eql?('RNScreens')
+        def pod.build_type;
+          Pod::BuildType.static_library
+        end
+      end
+    end
+  end
```
brustolin, line 32: m: I believe it would be nice to have a comment here to explain why we do this.

brustolin (approved):

PR #3777, opened 2024-04-25
fix(sample): RN Sample doesn't show multiple promise packages warning

krystofwoldrich:
This PR fixes the bare RN sample multiple promise package versions. 

It overwrites the default resolve function to ensure promise from the sample node_modules is used.

#skip-changelog


lucas-zimerman (approved):
LGTM!

PR #3771, opened 2024-04-21
feat(ReactNativeClientOptions): Remove "tunnel" option

AliReza99:
Remove the "tunnel" option from the ReactNativeClientOptions type due to lack of support in React Native.

This pull request removes the "tunnel" option from the ReactNativeClientOptions type. While passing this option currently does not create issues, its presence can cause confusion, leading developers to believe that the feature works when it does not. Removing it ensures clarity and avoids potential misunderstandings in React Native projects.

PR #3787, opened 2024-04-29
fix(options): Remove "tunnel" from SDK options

krystofwoldrich:
Removes confusing unsupported `tunnel` option

- original https://github.com/getsentry/sentry-react-native/pull/3771


lucas-zimerman (approved):
LGTM!

PR #3852, opened 2024-06-03
(1) ref(tracing): extract app start to a standalone integration

krystofwoldrich:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring

### Before this PR
- AppStart Tracking is embedded in `ReactNativeTracing` implementation. 
- When automatic routing instrumentation is enabled appstart is attached to the first transaction.
- When no routing instrumentation app start is standalone idle transaction.

```js
import Sentry from '@sentry/react-native';

Sentry.init({
  tracesSampleRate: 1.0,
  integrations: [
      enableAppStartTracking: true, // default true
    }),
  ],
});
```

### After this PR
- AppStart Tracking is standalone integration.
- When automatic routing instrumentation is enabled appstart is attached to the first transaction. No change, but this behaviour can be overwritten by user now.
- When no routing instrumentation app start is standalone transaction. Linked by a traceId to the first transaction after start.
- AppStart has `standalone` option which let user change the behaviour. By default we check presence of routing instrumentation like before.
- Moving the tracking to a standalone integration will let us measure usage of this feature (enabled integrations telemetry).

```js
import Sentry from '@sentry/react-native';

Sentry.init({
  tracesSampleRate: 1.0,
  enableAppStartTracking: true, // default true
  integrations: [
    Sentry.appStartIntegration({
      standalone: false, // default false
    }),
  ],
});
```

### Examples After

Screenshot 2024-08-06 at 10 46 40
Screenshot 2024-08-06 at 10 44 24

tests, sample app

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


src/js/tracing/reactnativeprofiler.tsx
```diff
@@ -47,12 +44,7 @@ export class ReactNativeProfiler extends Profiler {
...
     client.addIntegration && client.addIntegration(createIntegration(this.name));
-
-    const endTimestamp = this._mountSpan && typeof spanToJSON(this._mountSpan).timestamp
-    const tracingIntegration = client.getIntegrationByName && client.getIntegrationByName<ReactNativeTracing>('ReactNativeTracing');
-    tracingIntegration
-      && typeof endTimestamp === 'number'
-      // The first root component mount is the app start finish.
-      && tracingIntegration.onAppStartFinish(endTimestamp);
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    captureAppStart();
```
lucas-zimerman, line 48: Q:Shouldn't we wait for it?

lucas-zimerman (approved):
Non-blocking question, LGTM!

src/js/tracing/reactnativeprofiler.tsx
krystofwoldrich: We only trigger the capture of app start here, recording the timestamp and we don't await it as the fetch from native layer can happen async later

PR #3861, opened 2024-06-05
fix(logs): Add missing App Start logs when the span is dropped

krystofwoldrich:

- [x] Bugfix
- [x] New feature
- [x] Enhancement

## :scroll: Description

Adds warning logs when `debug: true` to places in the SDK code where it's decided that App Start span won't be added.

This will makes it easier for our customer to see why their transactions are missing app start spans as well for us when debugging the SDK.

sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fix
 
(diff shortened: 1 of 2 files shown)


buenaflor (approved):
👍

PR #3864, opened 2024-06-05
ci(ios): allow larger app size diff (up to 1 MiB) 

vaind:
bumps the expected size after recent sentry-cocoa update

#skip-changelog


krystofwoldrich:
@vaind Thank you this is a good catch. 

A lot of the times these tests fail on the start up time of the app without Sentry is higher than with Sentry. And so it's easy to overlook that something else was broken.

krystofwoldrich (approved):

PR #3867, opened 2024-06-06
fix(ttd): Make all options of startTTID optional

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

This function should not require any options.

sample app


lucas-zimerman (approved):
LGTM!

PR #3868, opened 2024-06-06
fix(ttd): Add span ids to time to display debug logs

krystofwoldrich:

- [x] Enhancement

## :scroll: Description

Having the Span IDs in the debug logs makes it possible to relate data from Sentry to the logs.

### DIFF
--- a/src/js/tracing/timetodisplay.tsx
+++ b/src/js/tracing/timetodisplay.tsx
@@ -258,15 +258,16 @@ function updateFullDisplaySpan(frameTimestampSeconds: number, passedInitialDispl
     return;
   }
 
-  if (spanToJSON(span).timestamp) {
-    logger.warn(`[TimeToDisplay] ${spanToJSON(span).description} (${span.spanContext().spanId}) span already ended.`);
+  const spanJSON = spanToJSON(span);
+  if (spanJSON.timestamp) {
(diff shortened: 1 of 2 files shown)


src/js/tracing/timetodisplay.tsx
```diff
@@ -247,25 +248,25 @@ function updateFullDisplaySpan(frameTimestampSeconds: number, passedInitialDispl
...
   const span = startTimeToFullDisplaySpan();
   if (!span) {
-    logger.warn(`[TimeToDisplay] No span found or created, possibly performance is disabled.`);
+    logger.warn(`[TimeToDisplay] No TimeToFullDisplay span found or created, possibly performance is disabled.`);
     return;
   }
 
   if (spanToJSON(span).timestamp) {
-    logger.warn(`[TimeToDisplay] ${spanToJSON(span).description} span already ended.`);
+    logger.warn(`[TimeToDisplay] ${spanToJSON(span).description} (${span.spanContext().spanId}) span already ended.`);
```
lucas-zimerman, line 262: Suggestion: It would be nice to store the result of `spanToJSON` so that we we don't need to call it twice on  on  lines 261, 262 and 263

lucas-zimerman (approved):
I left a minor suggestion but nothing blocking the PR, LGTM!

src/js/tracing/timetodisplay.tsx
krystofwoldrich: Yeah, I agree, I will change that.

PR #3869, opened 2024-06-06
fix(ttfd): Use TTID end timestamp when TTFD should be updated with an earlier timestamp

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

This PR fixes a race condition when TTFD update would be called after TTID, but the provided timestamp was set to earlier point in time than TTID end timestamp.

sample app, unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/tracing/timetodisplay.tsx
+++ b/src/js/tracing/timetodisplay.tsx
@@ -262,7 +262,12 @@ function updateFullDisplaySpan(frameTimestampSeconds: number, passedInitialDispl
     return;
   }
 
-  span.end(frameTimestampSeconds);
+  if (initialDisplayEndTimestamp > frameTimestampSeconds) {
+    logger.warn(`[TimeToDisplay] Using initial display end. Full display end frame timestamp is before initial display end.`);
+    span.end(initialDisplayEndTimestamp);
+  } else {
+    span.end(frameTimestampSeconds);
+  }
 
   span.setStatus('ok');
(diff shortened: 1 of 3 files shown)


src/js/tracing/timetodisplay.tsx
```diff
@@ -262,7 +262,12 @@ function updateFullDisplaySpan(frameTimestampSeconds: number, passedInitialDispl
...
-  span.end(frameTimestampSeconds);
+  if (initialDisplayEndTimestamp > frameTimestampSeconds) {
+    logger.warn(`[TimeToDisplay] Using initial display end. Full display end frame timestamp is before initial display end.`);
+    span.end(initialDisplayEndTimestamp);
+  } else {
+    span.end(frameTimestampSeconds);
+  }
 
   span.setStatus('ok');
   logger.debug(`[TimeToDisplay] ${spanToJSON(span).description} span updated with end timestamp.`);
```
lucas-zimerman, line 273: Not related to this PR:
I noticed on line 273 and also 261 that we are calling `spanToJSON(span).description` twice, ideally we could store this value on a constant if description doesn't change.

Also, When users have the logger disabled, we could avoid calling spanToJSON,  allowing a small performance slight performance improvement by avoiding a call to a json serializer when it's not needed.

lucas-zimerman (approved):
LGTM!

PR #3875, opened 2024-06-07
ci: switch e2e, sample app and testflight to large runners

krystofwoldrich:

- [x] Enhancement

## :scroll: Description

Even with the `macos-14` which are m1 runners the time to build the iOS RN project takes from 10 to 17 (outliers 30-40m) minutes (not counting the waiting time for the runner).

This PR aims to test the large runners and make a decision if the time benefit is worth switching.

With the `xlarge` the build times from the run below range from 4 to 7 minutes (not counting the waiting time for the runner).

#skip-changelog


krystofwoldrich:
When not counting the outliers and including waiting times for the runners. The diff is ~3 – 5 minutes. Smaller than I expected closing.

PR #3881, opened 2024-06-11
fix(performance): app start frames can start with zeroed values

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

The native implementations dropped zero value measurements, which could cause app start frames not to be measured as frames on transaction start would be `null`.

This PR change native impls to return zero values and adds zero value check to processing final frames.

sample app, unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/tracing/nativeframes.ts
+++ b/src/js/tracing/nativeframes.ts
@@ -192,6 +192,17 @@ export class NativeFramesInstrumentation {
       },
     };
 
+    if (
+      measurements.frames_frozen.value <= 0 &&
+      measurements.frames_slow.value <= 0 &&
+      measurements.frames_total.value <= 0
+    ) {
+      logger.warn(
+        `[NativeFrames] Detected zero slow or frozen frames. Not adding measurements to traceId (${traceId}).`,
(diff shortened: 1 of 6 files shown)


src/js/tracing/nativeframes.ts
```diff
@@ -192,6 +192,17 @@ export class NativeFramesInstrumentation {
       },
     };
 
+    if (
+      measurements.frames_frozen.value <= 0 &&
+      measurements.frames_slow.value <= 0 &&
+      measurements.frames_total.value <= 0
+    ) {
```
lucas-zimerman, line 199: So we want to add the measurement when all the values are below zero?

krystofwoldrich: Yes, that's correct and that's as well the current behaviour of the other SDKs.

https://github.com/getsentry/sentry-cocoa/blob/bee0cc30381f47a3f82730c423a8cd14ebb6c483/Sources/Sentry/SentryFramesTracker.m#L359

PR #3884, opened 2024-06-12
fix(performance): UIKit can't end after JS execution starts

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

UIKit init end when the native `AppDelegate.application` method is called, this is before the JS engine starts.

This PR adds check, which only records the UIKit init start if the end was recorder after JS start. 

This logic will let manual init setups to record the UIKit init end and auto init setup to record only the start.

Before
Screenshot 2024-06-12 at 11 45 38

After
Screenshot 2024-06-12 at 11 27 19

sample app, unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/tracing/reactnativetracing.ts
+++ b/src/js/tracing/reactnativetracing.ts
@@ -558,7 +558,7 @@ export class ReactNativeTracing implements Integration {
         return this._createUIKitSpan(appStartSpan, span);
       }
 
-      return appStartSpan.startChild({
+      appStartSpan.startChild({
         op: appStartSpan.op,
         description: span.description,
         startTimestamp: span.start_timestamp_ms / 1000,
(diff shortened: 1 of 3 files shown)


src/js/tracing/reactnativetracing.ts
```diff
@@ -557,6 +561,33 @@ export class ReactNativeTracing implements Integration {
...
+  /**
+   * UIKit init is measured by the native layers till the native SDK start
+   * RN initializes the native SDK later, the end timestamp would be wrong
+   */
+  private _createUIKitSpan(parentSpan: Span, nativeUIKitSpan: NativeAppStartResponse['spans'][number]): Span {
+    const bundleStart = getBundleStartTimestampMs();
+
+    // If UIKit init ends after the bundle start the native SDK was auto initialize
+    // and so the end timestamp is incorrect
+    // The timestamps can't equal as after UIKit RN initializes
```
philipphofmann, line 573: `l`: 
```suggestion
    // If UIKit init ends after the bundle start, the native SDK was auto-initialized
    // and so the end timestamp is incorrect.
    // The timestamps can't equal, as RN initializes after UIKit.
```

src/js/tracing/reactnativetracing.ts
```diff
@@ -557,6 +561,33 @@ export class ReactNativeTracing implements Integration {
...
+    if (bundleStart && bundleStart < nativeUIKitSpan.end_timestamp_ms) {
+      return parentSpan.startChild({
+        op: parentSpan.op,
+        description: 'UIKit init start',
+        startTimestamp: nativeUIKitSpan.start_timestamp_ms / 1000,
+        endTimestamp: nativeUIKitSpan.start_timestamp_ms / 1000,
```
philipphofmann, line 579: `h`: This creates a span with a duration of 0ms, which has no value. Wouldn't it be better to set the endTimestamp to the startTimestamp of the span `JS Bundle Execution Before React Root`?

krystofwoldrich: Such a span would not be named `UIKit init` anymore (UIKit, RN Modules, JS Engine). It could be named `Native (layers) init`. But it becomes unclear what the span represents.

I would still keep the 0ms UIKit init start so it's clear when the `Native (layers) init` starts. So users don't have to look up docs due to opaque naming.

krystofwoldrich: I've added `Native Runtime Init` span which cover the time from UIKit Init start to bundle execution start.

philipphofmann: To me, a UIKit Init span of 0ms is confusing cause UIKit for sure takes longer to initialize. If we don't know how long it took to initialize, I would rather drop the span. I think having only the native runtime init span should be enough.

krystofwoldrich: `UIKit Init span of 0ms` Yes, I partially agree, that's why I named it `UIKit Init Start` to imply that's when the init started. Sadly we currently don't have any other way to mark single point in time. 

I think it's useful for context. It's very common question what `App Start includes`, when it starts, when it ends. And I believe `UIKit Init Start` will helps us avoid the question when `Native Runtime Init` starts, the end it clear as the `JS Exec` span starts.

krystofwoldrich: Zero valued spans are used also in other SDK to create marks.
 
Screenshot 2024-06-13 at 16 11 11

philipphofmann: What about calling it then `UIKit Init to JS Exec Start` or something? Basically reflecting what it is in the name instead of having a 0 ms span.

philipphofmann (approved):
I don't want the 0ms discussion to block this PR. Decide what is best on your own.

PR #3894, opened 2024-06-17
fix(ci): Remove JS V8 dropped packages from the update CI

krystofwoldrich:
JS V8 removed some packages used in RN V5.

The use of these packages is removed in RN V6.

We have to update the CI on main, otherwise the updater job keeps failing.

### DIFF
--- a/.github/workflows/update-deps.yml
+++ b/.github/workflows/update-deps.yml
@@ -7,6 +7,7 @@ on:
   # And on on every PR merge so we get the updated dependencies ASAP, and to make sure the changelog doesn't conflict.
   push:
     branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/update-deps.yml
```diff
@@ -7,6 +7,7 @@ on:
   # And on on every PR merge so we get the updated dependencies ASAP, and to make sure the changelog doesn't conflict.
   push:
     branches:
+      - '**'
```
krystofwoldrich, line 10: TMP to test, will remove before merge.

lucas-zimerman (approved):
LGTM!

PR #3899, opened 2024-06-18
feat: touch breadcrumbs info improvements

vaind:
- cleanup & fix touchevent ignore behaviour
- feat: support annotated components in touch events
- deduplicate subsequent items in the component tree/path

Example event: https://sentry-sdks.sentry.io/issues/5502756331/events/13026ba7f65a4eab8ec5dfd835c11888/
image

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2687,7 +2687,7 @@ We are looking into ways making this more stable and plan to re-enable it again 
 
 ## v0.23.2
 
-- Fixed #228 again ¯\\*(ツ)*/¯
+- Fixed #228 again ¯\\_(ツ)_/¯
 
 ## v0.23.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Improve touch event component info if annotated with [`@sentry/babel-plugin-component-annotate`](https://www.npmjs.com/package/@sentry/babel-plugin-component-annotate) ([#3899](https://github.com/getsentry/sentry-react-native/pull/3899))
+
 ## 5.24.1
 
 ### Fixes
@@ -642,7 +648,7 @@ This release is compatible with `expo@50.0.0-preview.6` and newer.
   });
   ```
 
-  Read more at https://github.com/getsentry/sentry-javascript/blob/develop/CHANGELOG.md#7690
+  Read more at <https://github.com/getsentry/sentry-javascript/blob/develop/CHANGELOG.md#7690>
 
 - Report current screen in `contexts.app.view_names` ([#3339](https://github.com/getsentry/sentry-react-native/pull/3339))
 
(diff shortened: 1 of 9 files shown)


src/js/touchevents.tsx
```diff
@@ -113,18 +122,17 @@ class TouchEventBoundary extends React.Component<TouchEventBoundaryProps> {
...
-  ): void {
+  private _logTouchEvent(touchPath: TouchedComponentInfo[], label?: string): void {
     const level = 'info' as SeverityLevel;
+
+    const root = touchPath[0];
+    const detail = label ? label : `${root.name}${root.file ? ` (${root.file})` : ''}`;
+
     const crumb = {
       category: this.props.breadcrumbCategory,
-      data: { componentTree: componentTreeNames },
```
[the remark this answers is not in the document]
lucas-zimerman: is there a sample picture on how it was before/after this change?

CHANGELOG.md
```diff
@@ -2680,7 +2687,7 @@ We are looking into ways making this more stable and plan to re-enable it again
 
 ## v0.23.2
 
-- Fixed #228 again ¯\\_(ツ)_/¯
+- Fixed #228 again ¯\\*(ツ)*/¯
```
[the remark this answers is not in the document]
vaind: interestingly, it renders the same

test/touchevents.test.tsx
```diff
@@ -160,15 +160,15 @@ describe('TouchEventBoundary._onTouchStart', () => {
     expect(addBreadcrumb).toBeCalledWith({
       category: defaultProps.breadcrumbCategory,
       data: {
-        componentTree: ['Styled(View2)', 'Styled(View)'],
+        path: [{ name: 'Styled(View)' }],
```
krystofwoldrich, line 163: `'Styled(View2)'` is not in the path because it has a label which is ignored, was this intentional?

Before when label was ignored the code took the next value, but now when ignored label is read it will drop the value from the path.

src/js/touchevents.tsx
```diff
@@ -166,80 +174,91 @@ class TouchEventBoundary extends React.Component<TouchEventBoundaryProps> {
...
+      return false;
+    }
+    if (value.label && this._isNameIgnored(value.label)) {
+      return false;
+    }
+
+    // Deduplicate same subsequent items.
+    if (touchPath.length > 0 && JSON.stringify(touchPath[touchPath.length - 1]) === JSON.stringify(value)) {
+      return false;
+    }
```
krystofwoldrich, line 257: Was this happening a lot in your testing @vaind ?

Let's add a test for it.

test/touchevents.test.tsx
vaind: I thought the original behavior was a bug - since you specified you want to ignore something, I'd expect the element to be ignored, not just the label (and element taken in path). I can make it work the original way, of course, if you think it's necessary.

src/js/touchevents.tsx
vaind: it did. And it was tested as part of another test case but I've added a custom one for it.

test/touchevents.test.tsx
krystofwoldrich: No, I also think it was a bug, the previous behaviour seems very unexpected to me as well!

I just wanted to make sure we are making conscious decision to change this.

src/js/touchevents.tsx
krystofwoldrich: Aaa, thanks, I've missed that. Thanks for splitting it.

krystofwoldrich (approved):
Looks good, thank you!

PR #3902, opened 2024-06-19
feat(e2e): Enable Replay for E2E tests

krystofwoldrich:

- [x] Enhancement

## :scroll: Description

This PR enables replays for RN E2E tests.

This should ensure we have the latest state of RN Replay available in Sentry.

e2e tests


krystofwoldrich:
Just enabling replay for e2e test is not enough, the test end before segment is uploaded.

Note: Currently we don't have a way to wait for Replay flush.

PR #3905, opened 2024-06-19
ci: release iOS metrics startup test restrictions

vaind:
The startup time doesn't seem to be impacted by the SDK (is that OK or do we measure it wrong?), so it may be lower then without the SDK

#skip-changelog


krystofwoldrich (approved):
LGTM 🚀

krystofwoldrich:
I think this make sense to avoid failing CI for no reason.

But we should check how the app start is measured, maybe we are not correctly waiting for the first React Component render?

PR #3911, opened 2024-06-21
fix(scope): Hooks into Global and Isolations Scopes for Sync with the native scope

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

This PR adds necessary scope changes to enables hybrid SDKs (React Native) to synchronize their scopes to the native SDKs which are using the Hub and Scopes architecture.

The following variants were considered.

### Change static RN SDK APIs to operate on Global Scope

This will cause all integration from JS Core to stop working as they do not use the RN SDK APIs.

### Add Default Scope which all SDK APIs will use

This adds new scope to the mix and increases complexity.

### Set Global and Isolation scope as one reference.

Due to default falls back any interaction with the static APIs before Sentry.init will break the sync. The isolation scope will be created on first call before RN setup.

### Use scope change listener

Not suitable for sync as it returns whole scope, comparing diff is too expensive. 

### **Hook into scope changes**

This work with API calls before init, the date before init won't be sync the same as in the previous implementation. We patch our own object, in the future we can use listeners for individual changes.

sample app, unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

#skip-changelog


src/js/utils/worldwide.ts
```diff
@@ -26,5 +27,14 @@ export interface ReactNativeInternalGlobal extends InternalGlobal {
   nativePerformanceNow?: () => number;
 }
 
+/**
+ * Minimum required Sentry Carrier from JS Core SDK
+ * https://github.com/getsentry/sentry-javascript/blob/874df8e5b28e2f0f530b0e651067cdd2aa3e5cf7/packages/core/src/carrier.ts#L18
+ */
+export type SentryCarrier = {
```
lucas-zimerman, line 34: Q: Is this being used?

src/js/sdk.tsx
```diff
@@ -44,6 +45,8 @@ export function init(passedOptions: ReactNativeOptions): void {
   }
 
   useEncodePolyfill();
+  enableSyncToNative(getGlobalScope());
+  enableSyncToNative(getIsolationScope());
```
lucas-zimerman, line 49: M: We only need to sync when native is enabled, when disabled we could use the original Scope

```suggestion
  if (enableNative) {
     enableSyncToNative(getGlobalScope());
     enableSyncToNative(getIsolationScope());
  }
```

krystofwoldrich: Yes, that's a good point.

test/scopeSync.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    });
+
+    describe('addBreadcrumb', () => {
+      it('it only syncs once per scope', () => {
+        enableSyncToNative(scope);
+        enableSyncToNative(scope);
+
+        scope.addBreadcrumb({ message: 'test' });
+
+        expect(NATIVE.addBreadcrumb).toBeCalledTimes(1);
```
lucas-zimerman, line 32: Can we test the original call from the Scope were also called just to make sure we didn't miss calling them?

lucas-zimerman (approved):
I asked some minor changes, but other than that it looks good for merge, good job on the solution chosen here!

lucas-zimerman (commented):
Actually, there is one thing that is bothering me, with your implementation, doesn't it have a chance to break if users uses minification on their code? in that case, the name setTag may change but not the string literal setTag.

krystofwoldrich:
I've check the minified output of the Metro bundler and it correctly recognizes the dynamic access to the properties and does not minify the scope.

```
__d((function(g,r,i,a,m,e,d){Object.defineProperty(e,"__esModule",{value:!0}),e.enableSyncToNative=function(n){if(t.has(n))return;t.set(n,!0),n.setUser=(0,r(d[0]).wrapFunction)(n.setUser,(function(t){return function(c){return r(d[1]).NATIVE.setUser(c),t.call(n,c)}})),n.setTag=(0,r(d[0])
```

We dynamically access scope properties in the core as well, so even if changed in  RN the scope would not be minified.

https://github.com/getsentry/sentry-javascript/blob/874df8e5b28e2f0f530b0e651067cdd2aa3e5cf7/packages/core/src/utils/spanOnScope.ts#L28

mydea (approved):
this makes sense to me 👍

PR #3846, opened 2024-05-31
feat(replay): Add breadcrumbs mapping from RN to RRWeb format

vaind:
## :scroll: Description

image

* [x] navigation 
* [x] tap - closes #3818 
   Note: identifiers can be improved after #2863 

## Dependencies

Blocked by 
- [x] https://github.com/getsentry/sentry-java/pull/3448
- [x] https://github.com/getsentry/sentry-cocoa/pull/4027

PR #3912, opened 2024-06-24
feat(replay): Add network breadcrumbs

krystofwoldrich:

- [x] New feature

## :scroll: Description

This PR adds network breadcrumbs to RN replays.

RN sample app, Android works, ~iOS waiting for https://github.com/getsentry/sentry-react-native/pull/3846~, iOS working

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

Screenshot 2024-06-24 at 16 36 58

Screenshot 2024-06-25 at 10 59 14


src/js/replay/networkUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+  if (RN_GLOBAL_OBJ.TextEncoder) {
+    return new RN_GLOBAL_OBJ.TextEncoder().encode(input);
+  }
+  return utf8ToBytes(input);
+}
+
+function _serializeFormData(formData: FormData): string {
+  // This is a bit simplified, but gives us a decent estimate
+  // This converts e.g. { name: 'Anne Smith', age: 13 } to 'name=Anne+Smith&age=13'
+  // @ts-expect-error passing FormData to URLSearchParams actually works
```
vaind, line 62:
```suggestion
  // @ts-expect-error passing FormData to URLSearchParams won't correctly serialize `File` entries, which is fine for this use-case. See https://github.com/microsoft/TypeScript/issues/30584
```

vaind (approved):
LGTM but there are some conflicts with what I've done today in the base PR.

test/replay/networkUtils.test.ts
```diff
@@ -0,0 +1,59 @@
...
+      expect(getBodySize('abcd')).toBe(4);
+      // Emojis are correctly counted as mutliple characters
+      expect(getBodySize('With emoji: 😈')).toBe(16);
+    });
+
+    it('works with URLSearchParams', () => {
+      const params = new URLSearchParams();
+      params.append('name', 'Jane');
+      params.append('age', '42');
+      params.append('emoji', '😈');
```
lucas-zimerman, line 33: Q: Should we also test with non english characters?

src/js/replay/networkUtils.ts
```diff
@@ -0,0 +1,64 @@
+import { RN_GLOBAL_OBJ } from '../utils/worldwide';
+import { utf8ToBytes } from '../vendor';
+
+/** Convert a Content-Length header to number/undefined.  */
```
lucas-zimerman, line 4: NIT
```suggestion
/** Convert a Content-Length header to number/undefined. */
```

lucas-zimerman (approved):
LGTM!

src/js/replay/networkUtils.ts
krystofwoldrich: Sounds like something we should add also to the JS repo where I copied this from.

`won't correctly serialize `File` entries, which is fine for this use-case. See https://github.com/microsoft/TypeScript/issues/30584`

test/replay/networkUtils.test.ts
krystofwoldrich: Probably not necessary, the main goal is to check that the function doesn't just could string length.

Issue #3913, opened 2024-06-25
Automatically add `@sentry/babel-plugin-component-annotate` using `withSentry`


krystofwoldrich:
### Description

`@sentry/babel-plugin-component-annotate` plugin exists as a standalone package and as such can be added to `babel.config.js`. 

- This introduces another config file Sentry should be added.
- As well as a new place `@sentry/wizard` should update.
- Gives RN SDK no control over the file which will be annotated.

Web bundler plugins include the babel plugin in their config (https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/5e53faa75d4b263cb604f8c2c5ee53cae31a6941/packages/bundler-plugin-core/src/index.ts#L621) and filter the files on which the plugin is executed. RN SDK can do the same using `babelTransformerPath` option of Metro. We can create a middleware which only adds the plugin for selected files (exclude node_modules for example).

- No new Sentry config file
- No changes to Wizard (only enabled flag)
- RN has control over the execution

```js
withSentry(metroConfig, { reactComponentAnnotation: { enabled: true } })
```
 
- part of https://github.com/getsentry/sentry-react-native/issues/2703

PR #3916, opened 2024-06-26
feat: Add Sentry Babel Transformer

krystofwoldrich:

- [x] New feature

## :scroll: Description

This PR adds `@sentry/babel-plugin-component-annotate` to `@sentry/react-native/metro`.

- Similar to the rest of our JS Bundlers Plugins https://github.com/getsentry/sentry-javascript-bundler-plugins/pull/469

Example use RN 0.72 and newer:

```js
// For Expo
const { getSentryExpoConfig } = require("@sentry/react-native/metro");
const config = getSentryExpoConfig(__dirname, { annotateReactComponents: true });

// For RN
const m = mergeConfig(getDefaultConfig(__dirname), config);

module.exports = withSentryConfig(m, {
  annotateReactComponents: true, // default false
});
```

For RN 0.65 to 0.71 (these version did not have default config in metro.config.js but applied it later, that changed in 0.72):
```
const config  = {
  transformer: {
	babelTransformerPath: require.resolve('metro-babel-transformer'),
	// this must be adjust to appropriate babel transformer for given RN version
  },
};
```

- closes: https://github.com/getsentry/sentry-react-native/issues/3913

sample apps, e2e tests, unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- [ ] Add docs


CHANGELOG.md
```diff
@@ -7,6 +7,18 @@
 - Improve touch event component info if annotated with [`@sentry/babel-plugin-component-annotate`](https://www.npmjs.com/package/@sentry/babel-plugin-component-annotate) ([#3899](https://github.com/getsentry/sentry-react-native/pull/3899))
```
vaind, line 7: maybe remove this from the changelog?

CHANGELOG.md
```diff
@@ -7,6 +7,18 @@
...
 - Add replay breadcrumbs for touch & navigation events ([#3846](https://github.com/getsentry/sentry-react-native/pull/3846))
 - Add network data to Session Replays ([#3912](https://github.com/getsentry/sentry-react-native/pull/3912))
+- Add `annotateReactComponents` option to `@sentry/react-native/metro` ([#3916](https://github.com/getsentry/sentry-react-native/pull/3916))
```
vaind, line 10: How about just `annotateComponents`?

vaind (commented):
This goes well beyond my understanding of the ecosystem but the I approve of the idea, because it's exactly what I wanted to do at first, I just didn't know how.

CHANGELOG.md
krystofwoldrich: Thanks good point, I'll update the changelog.

src/js/tools/sentryBabelTransformer.ts
```diff
@@ -0,0 +1,43 @@
+import componentAnnotatePlugin from '@sentry/babel-plugin-component-annotate';
+
+import { enableLogger } from './enableLogger';
+import { loadDefaultBabelTransformer } from './sentryBabelTransformerUtils';
+import type { BabelTransformer, BabelTransformerArgs } from './vendor/metro/metroBabelTransformer';
+
+enableLogger();
```
lforst, line 7: Do we for sure want this call here?

src/js/tools/metroconfig.ts
```diff
@@ -64,6 +96,38 @@ function loadExpoMetroConfigModule(): {
   }
 }
 
+/**
+ * Adds Sentry Babel transformer to the Metro config.
+ */
+export function withSentryBabelTransformer(config: MetroConfig): MetroConfig {
+  const defaultBabelTransformerPath = config.transformer && config.transformer.babelTransformerPath;
+  logger.debug('Default Babel transformer path from `config.transformer`:', defaultBabelTransformerPath);
```
lforst, line 104: The Sentry SDK logger is mostly intended to be used when the SDK is initialized. For build tooling we usually don't rely on it. I would do a good ol' console.log instead.

src/js/tools/metroconfig.ts
```diff
@@ -64,6 +96,38 @@ function loadExpoMetroConfigModule(): {
...
+    process.on('exit', () => {
+      cleanDefaultBabelTransformerPath();
+    });
+  }
+
+  return {
+    ...config,
+    transformer: {
+      ...config.transformer,
+      babelTransformerPath: require.resolve('./sentryBabelTransformer'),
```
lforst, line 126: I don't know about the state of ESM/CJS in react native but `require.resolve` will not work in ESM.

src/js/tools/sentryBabelTransformerUtils.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/**
+ * Reads default Babel transformer path from the project root.
+ */
+export function readDefaultBabelTransformerPath(): string | undefined {
+  try {
+    if (fs.existsSync(getDefaultBabelTransformerPath())) {
+      return fs.readFileSync(getDefaultBabelTransformerPath()).toString();
+    }
```
lforst, line 29: I think this pattern of fs.exists then fs.readFile should be avoided. It technically is a race condition. readFile throws anyhow if the file doesn't exist.

lforst (approved):
Take with a grain of salt because I am no expert but nothing in this PR looks completely off to me.

src/js/tools/sentryBabelTransformer.ts
krystofwoldrich: It only enables it when `env.SENTRY_LOG_LEVEL === 'debug'`. The babelTransformer is created when the file loads, so I call it here not to miss any logs (if enabled).

src/js/tools/metroconfig.ts
krystofwoldrich: The `@sentry/react-native/metro` is CJS, same as the `metro` package.

lucas-zimerman (approved):
Looks good to me from my part, LGTM!

0Calories:
Thanks for expanding support to react native! This looks great

PR #3920, opened 2024-06-28
Fix: Symbolicate `error.cause ` on debug builds

lucas-zimerman:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

we were limited to only parse the stack-trace of the original error, with this change, we will parse the error and the cause of it.

Close: #2118

The changes were made using in regard to what Sentry JavaScript currently support (an error with  a cause error, anything else is discarded by the SDK so I limited the scope to only parse the original error and the cause of the original error)

Unit test, and also with an app:

Before: https://sentry-sdks.sentry.io/issues/5546274920/

After: https://sentry-sdks.sentry.io/issues/5548578775/

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps


src/js/integrations/debugsymbolicator.ts
```diff
@@ -28,6 +30,11 @@ export type ReactNativeError = Error & {
   componentStack?: string;
 };
 
+type ErrorLike = {
+  stack: string
+};
```
krystofwoldrich, line 35: Let's move this type to `utils/error`

src/js/integrations/debugsymbolicator.ts
```diff
@@ -51,13 +58,18 @@ export const DebugSymbolicator = convertIntegrationFnToClass(
...
-    const symbolicatedFrames = await symbolicate(
-      hint.originalException.stack,
-      getFramesToPop(hint.originalException as Error),
-    );
-    symbolicatedFrames && replaceExceptionFramesInEvent(event, symbolicatedFrames);
+    const errorGroup = getExceptionGroup(hint.originalException)
+    for (const [index, error] of errorGroup.entries()) {
+      const symbolicatedFrames = await symbolicate(
+        error.stack,
+        getFramesToPop(error as Error),
```
krystofwoldrich, line 67: We can relax the condition and `ExtendedError` can extend `ErrorLike` if it doesn't bereak anything else.

src/js/integrations/debugsymbolicator.ts
```diff
@@ -200,3 +212,16 @@ async function addSourceContext(frame: SentryStackFrame): Promise<void> {
...
 }
+
+/**
+ * Return a list containing the original exception and also the cause if found.
+ *
+ * @param originalException The original exception.
+ */
+function getExceptionGroup(originalException: ErrorLike): ErrorLike[] {
+  const errorGroup: ErrorLike[] = [originalException];
+  const cause = (originalException as { cause?: unknown }).cause;
```
krystofwoldrich, line 223: We have to walk the tree of causes. It can have more than one layer.

```
erorr3
error2.cause = error3
error.cause = error2
```

test/integrations/debugsymbolicator.test.ts
```diff
@@ -334,5 +358,152 @@ describe('Debug Symbolicator Integration', () => {
         },
       });
     });
+
+    it('should symbolicate multiple error with cause ', async () => {
```
krystofwoldrich, line 362: Let's extend the tests with one more cause in the chain.

krystofwoldrich:
The lint job is failing, please run `yarn fix`.

test/integrations/debugsymbolicator.test.ts
lucas-zimerman: I Added more tests to check for multiple causes and also if there is a different amount of processed errors compared to the items  found on the hint.

src/js/integrations/debugsymbolicator.ts
lucas-zimerman: You gave me the idea of just extending the `ExtendedError` with an optional cause error.

src/js/integrations/debugsymbolicator.ts
lucas-zimerman: I removed it since it is no longer required with the new changes.

krystofwoldrich (approved):
Thank you for the additional tests, makes it much easier to review. 

🚀 Looks good.

PR #3927, opened 2024-07-02
feat(native): Set current screen on native scope

krystofwoldrich:

- [x] New feature

## :scroll: Description

This PR adds current screen on native scopes. This enables native replay integration to track the RN screens and use them them to fill the `urls.field`

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/android/src/main/java/io/sentry/react/RNSentryBreadcrumb.java
+++ b/android/src/main/java/io/sentry/react/RNSentryBreadcrumb.java
@@ -1,0 +1,84 @@
+package io.sentry.react;
+
+import com.facebook.react.bridge.ReadableMap;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Map;
+
+import io.sentry.Breadcrumb;
+import io.sentry.SentryLevel;
+
+public class RNSentryBreadcrumb {
+
+    @Nullable
+    public static String getCurrentScreenFrom(ReadableMap from) {
+        final @Nullable String maybeCategory = from.hasKey("category") ? from.getString("category") : null;
+        if (maybeCategory == null) {
+            return null;
+        }
+
+        final @Nullable ReadableMap maybeData = from.hasKey("data") ? from.getMap("data") : null;
+        if (maybeData == null) {
+            return null;
+        }
+
+        final @Nullable String maybeCurrentScreen = from.hasKey("to") ? from.getString("to") : null;
+
+        return maybeCurrentScreen;
+    }
+
+    @NotNull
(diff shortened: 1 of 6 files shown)


android/src/main/java/io/sentry/react/RNSentryBreadcrumb.java
```diff
@@ -0,0 +1,87 @@
...
+import io.sentry.SentryLevel;
+
+public class RNSentryBreadcrumb {
+
+    @Nullable
+    public static String getCurrentScreenFrom(ReadableMap from) {
+        final @Nullable String maybeCategory = from.hasKey("category") ? from.getString("category") : null;
+        if (maybeCategory == null || !maybeCategory.equals("navigation")) {
+            return null;
+        }
```
lucas-zimerman, line 20: does  getString throws if not found? Otherwise we could do
```suggestion
        final @Nullable String maybeCategory = from.getString("category");
        if (!"navigation".equals(maybeCategory)) {
            return null;
        }
```

android/src/main/java/io/sentry/react/RNSentryBreadcrumb.java
```diff
@@ -0,0 +1,87 @@
...
+
+        final @Nullable ReadableMap maybeData = from.hasKey("data") ? from.getMap("data") : null;
+        if (maybeData == null) {
+            return null;
+        }
```
lucas-zimerman, line 25: suggestion: does the same thing at the end but we don't need to have a parameter maybe when we are sure we have the data after checking if the key exist.
```suggestion
        if (from.hasKey("data")) {
            return null;
        }
        final @Nullable ReadableMap data = from.getMap("data");
```

android/src/main/java/io/sentry/react/RNSentryBreadcrumb.java
```diff
@@ -0,0 +1,87 @@
...
+
+        try {
+            // getString might throw if cast to string fails (data.to is not enforced by TS to be a string)
+            return maybeData.hasKey("to") ? maybeData.getString("to") : null;
```
lucas-zimerman, line 29: If the suggestions above are accepted we can change this too
```suggestion
            return data.hasKey("to") ? data.getString("to") : null;
```

lucas-zimerman (approved):
Left some suggestions that will not change the code logic so I am approving it, LGTM!

krystofwoldrich:
@lucas-zimerman Thank you for the review, at first I thought the same, we can simplify the null check, but sadly no.

The maps can contain key with null values.

```java
        WritableMap test = new WritableNativeMap();
        test.putString("keyWithNullValue", null);
```

PR #3935, opened 2024-07-09
Update CHANGELOG.md

krystofwoldrich:
- We don't want to point towards the Babel plugin yet, as it will be included in https://github.com/getsentry/sentry-react-native/pull/3916

#skip-changelog


buenaflor (approved):
🚀

PR #3949, opened 2024-07-15
Change React Native Version import to the one from Platform

svbutko:

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
The import for selection of React Native version, both for type and value happens via JS constant, while there's an documented way to get it from `Platform`: https://reactnative.dev/docs/platform#constants

It's also available in 0.65 which makes it work for all versions:
https://reactnative-archive-august-2023.netlify.app/docs/0.65/platform#constants

While working with other repos than `react-native` like `react-native-tvos` or `react-native-web` with certain TS configs, there are issues like this one:

```
node_modules/@sentry/react-native/dist/js/utils/rnlibrariesinterface.d.ts:1:37 - error TS7016: Could not find a declaration file for module 'react-native/Libraries/Core/ReactNativeVersion'. '/Users/**/zattoo/aurora/node_modules/react-native/Libraries/Core/ReactNativeVersion.js' implicitly has an 'any' type.
  If the 'react-native-tvos' package actually exposes this module, try adding a new declaration (.d.ts) file containing `declare module 'react-native/Libraries/Core/ReactNativeVersion';`

1 import type { version as RNV } from 'react-native/Libraries/Core/ReactNativeVersion';
```

This PR solves this import issue

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


krystofwoldrich:
Hi @svbutko,
thank you for the fix. Excuse the long waiting time, I was OoO.

It definitely make sense to use the public documented export instead of the internal lib. 

Could you check why `yarn test` fails after the change. I've already added changelog, but we need the tests to succeed to merge the change.

svbutko:
Hi @krystofwoldrich 
No problem at all!

Checked the tests, and added a fix for potential missing constants (e.g. testing or going beyond supported platforms)

krystofwoldrich (approved):
Thank you @svbutko!

It looks good now. 🚀

PR #3952, opened 2024-07-15
bump Android SDK 7.12.0

bruno-garcia:
https://github.com/getsentry/sentry/issues/70065


romtsn (approved):
Approving, but we'll have to re-trigger CI when the artifact gets to maven

PR #3968, opened 2024-07-25
fix: use full path to metro

paulschreiber:
## :scroll: Description
Sample code cleanup. Use a full path (instead of a relative path) for
```js
const { withSentryConfig } = require('@sentry/react-native/metro');
```
Matches what is in your [docs](https://docs.sentry.io/platforms/react-native/manual-setup/metro/).

#skip-changelog


krystofwoldrich:
Hi @paulschreiber,
thank you for the PR, sadly this change is not possible at the moment, as the sample application doesn't have `@sentry/react-native` as a dependency. 

The reason is we always want to used the latest state of the file from the repository. 

We are going to change to a monorepo structure in the future, then the proposed change will be possible.

Until then, I'll close this PR.

Issue #3835, opened 2024-05-22
extremely long auto transactions for react-navigation


brandonpearcy:
I saw sent here after a few rounds of conversation with Sentry customer support.

**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [x]  iOS
- [x]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 5.22.0

**`react-native` version:** 0.72.7

**Are you using Expo?**
- [ ]  Yes
- [x]  No

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)
- [ ]  on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

Here are some sample transactions that are extremely long:

https://heylo.sentry.io/performance/trace/1e7f705bab9749adb96338cb45f29323/?fov=0%2C73193135.7121582&node=txn-f037f7af2d2a42a59d9a4c3f8cc77bf2&pageEnd=2024-05-22T20%3A38%3A53.000&pageStart=2024-05-22T20%3A36%3A10.000&timestamp=1716410316

https://heylo.sentry.io/performance/trace/8f8e52c16a6545a3a2f6f571584c6ee4/?fov=0%2C238348109.14990234&node=txn-e1756085d2824a5c8b988214b492ee24&statsPeriod=24h&timestamp=1716394519

**Configuration:**

```
    Sentry.init({
      appHangTimeoutInterval: 5, // seconds
      dsn: heyloConfig.SENTRY_DSN,
      enableAutoPerformanceTracing: true,
      enableAutoSessionTracking: true,
      environment: Config.BUILD_VARIANT_STRING,
      integrations: [
        new Sentry.ReactNativeTracing({
          enableAppStartTracking: true,
          enableNativeFramesTracking: true,
          enableStallTracking: true,
          enableUserInteractionTracing: true,
          routingInstrumentation,
          // Uncomment to sample subsequent navigations to a route already seen
          // ignoreEmptyBackNavigationTransactions: true,
        }),
      ],
      // Set tracesSampleRate to 1.0 to capture 100% of transactions for performance monitoring.
      // We recommend adjusting this value in production.
      tracesSampleRate: 0.01,
    })
```

---
**I have the following issue:**

[...]

PR #3974, opened 2024-07-29
fix(tracing): Drop app start data older than one minute

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

This PR adds filter to app start data which drops all app starts which are older than one minute in production builds. 

We can do this as a minute or more time difference between transaction start and app start is a clear signal the data are not related and result would be incorrect and long (minutes in duration) transaction.

We skip the filter in development builds, as bundling of the Javascript bundle can take more than a minute and so we want to capture the data. 

- fix: https://github.com/getsentry/sentry-react-native/issues/3835

integration tests, sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


lucas-zimerman (approved):
LGTM!

PR #3993, opened 2024-08-05
test: e2e replay

vaind:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Tests that a replay is part of the error event when session replay options are enabled.

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes


krystofwoldrich:
### [Session replays uploaded from this PR CI](https://sentry-sdks.sentry.io/replays/?project=5428561&query=release%3Ae2e-refs-pull-3993-merge&statsPeriod=90d)

test/react-native/rn.patch.app.js
```diff
@@ -26,6 +26,17 @@ Sentry.init({
   release: '${SENTRY_RELEASE}',
   dist: '${SENTRY_DIST}',
   dsn: 'https://1df17bd4e543fdb31351dee1768bb679@o447951.ingest.sentry.io/5428561',
+  debug: true,
+  _experiments: {
+    replaysSessionSampleRate: 1.0,
+    replaysOnErrorSampleRate: 1.0,
```
krystofwoldrich, line 32: Currently all the tests are executed in one application session.

Let's set only `replaysOnErrorSampleRate` as that will ensure 
- the buffer to session Replay conversion is tested 
- as well the buffered replay id is assigned to the error event

test/e2e/test/e2e.test.ts
```diff
@@ -131,6 +131,30 @@ describe('End to end tests for common events', () => {
...
+    const eventId = await waitForEventId();
+    const sentryEvent = await fetchEvent(eventId);
+    expect(sentryEvent.eventID).toMatch(eventId);
+
+    expect(sentryEvent.contexts).toBeDefined();
+    const replay = sentryEvent.contexts!['replay'] as any;
+    expect(replay).toBeDefined();
+    expect(replay.replay_id.length).toBe(32);
+
+    const replayInfo = await fetchReplay(replay.replay_id);
```
krystofwoldrich, line 147: I thought possibly the Android replays are uploaded due to a short time limit, but if I understand the code correctly, this fetch will wait up to 10 minutes, checking the endpoint every 1s. 

Is that right? @vaind

vaind: yeah, nothing to do with timeouts. Replay capture doesn't work at all on android emulator in GHA at the moment

vaind: Also no screenshots are captured by appium, if you look into an artifact for any of the Android jobs:

image


From what I can tell it's an old/recurring issue of the emulator on GHA, see https://github.com/actions/runner-images/issues/6814

test/e2e/test/e2e.test.ts
```diff
@@ -131,6 +132,33 @@ describe('End to end tests for common events', () => {
...
+      await element.click();
+
+      const eventId = await waitForEventId();
+      const sentryEvent = await fetchEvent(eventId);
+      expect(sentryEvent.eventID).toMatch(eventId);
+
+      expect(sentryEvent.contexts).toBeDefined();
+      const replay = sentryEvent.contexts!['replay'] as any;
+      expect(replay).toBeDefined();
+      expect(replay.replay_id.length).toBe(32);
```
lucas-zimerman, line 148: Suggestion: We could use a more specific test to make it clear that we are dealing with an UUID instead of just checking for a fixed number.

```suggestion
const uuidRegex = /^[0-9a-f]{8}[0-9a-f]{4}[4][0-9a-f]{3}[89ab][0-9a-f]{3}[0-9a-f]{12}$/i;
expect(uuidRegex.test(replay.replay_id)).toBe(true);
```

test/e2e/test/e2e.test.ts
```diff
@@ -131,6 +132,33 @@ describe('End to end tests for common events', () => {
...
+
+      const replayInfo = await fetchReplay(replay.replay_id);
+      expect(replayInfo).toBeDefined();
+      expect(replayInfo.data.duration).toBeGreaterThan(0);
+      expect(replayInfo.data.count_segments).toBeGreaterThan(0);
+
+      const video = await fetchReplaySegmentVideo(replay.replay_id, 0);
+      expect(video).toBeDefined();
+      expect(video.size).toBeGreaterThan(1000);
+      expect(await video.slice(4, 12).text()).toMatch('ftypmp42');
```
lucas-zimerman, line 158: Q: Why validating 1000 here?

test/e2e/test/e2e.test.ts
vaind: Thanks for the suggestion but I think it's an overkill here. The replay ID is read from the sentry server API so should already be a valid UUID. Additionally, it's used to fetch from other APIs by this ID so those calls would faile. It's not unlike all the other tests where we use an error ID to fetch the event from the API - there are no checks prior to API requests either.

test/e2e/test/e2e.test.ts
vaind: arbitrary, it just should be reasonably large. This number worked fine so far.

test/e2e/test/e2e.test.ts
lucas-zimerman: Oh I see, sound good to me

vaind:
replaced by #4277

PR #3996, opened 2024-08-06
(2) ref(tracing): Extract NativeFrames as standalone integration

krystofwoldrich:
## :loudspeaker: Type of change

- [x] Refactoring

## :scroll: Description

This PR extracts Native Frames Tracking to a standalone integration. This is possible via client event `spanStart` and `spanEnd` which is available in JS v8. Before `ReactNativeTracing` integration ensured the Stall Tracking is started and stopped on transaction start and stop.

Moving the tracking to a standalone integration will let us measure usage of this feature (enabled integrations telemetry).

This PR introduces user facing breaking change.

### Before

```js
import Sentry from '@sentry/react-native';
Sentry.init({
  tracesSampleRate: 1.0,
  integrations: [
    new Sentry.ReactNativeTracing({
      enableNativeFramesTracking: true, // default true
    }),
  ],
});
```

### After
```js
import Sentry from '@sentry/react-native';
Sentry.init({
  tracesSampleRate: 1.0,
  enableNativeFramesTracking: true, // default true
});
```

## :green_heart: How did you test it?
sample app, tests

## :pencil: Checklist

- [x] I reviewed submitted code
- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [x] All tests passing
- [ ] No breaking changes

## Predecessor to this PR
- https://github.com/getsentry/sentry-react-native/pull/3852

PR #3997, opened 2024-08-06
(3) ref(tracing): Extract Stall Tracking to a standalone integration

krystofwoldrich:

- [x] Refactoring

## :scroll: Description

This PR extracts Stall Tracking to a standalone integration. This is possible via client event `spanStart` and `spanEnd` which is available in JS v8. Before `ReactNativeTracing` integration ensured the Stall Tracking is started and stopped on transaction start and stop.

Moving the tracking to a standalone integration will let us measure usage of this feature (enabled integrations telemetry).

This PR introduces user facing breaking change.

### Before

```js
import Sentry from '@sentry/react-native';
Sentry.init({
  tracesSampleRate: 1.0,
  integrations: [
      enableStallTracking: true, // default true
    }),
  ],
});
```

### After
```js
import Sentry from '@sentry/react-native';
Sentry.init({
  tracesSampleRate: 1.0,
  enableStallTracking: true, // default true
});
```

sample app, tests

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## Predecessor to this PR
- https://github.com/getsentry/sentry-react-native/pull/3996


src/js/options.ts
```diff
@@ -206,6 +206,11 @@ export interface BaseReactNativeOptions {
    */
   enableNativeFramesTracking?: boolean;
 
+  /**
+   * Track when and how long the JS event loop stalls for. Adds stalls as measurements to all transactions.
+   */
```
lucas-zimerman, line 211: NIT: It's nice to have the default value on the comments so it doesn't generate doubts with users.

```suggestion
   *
   * @default true
   */
```

test/sdk.test.ts
```diff
@@ -456,6 +456,29 @@ describe('Tests the SDK functionality', () => {
...
 
+    it('no stall tracking integration by default', () => {
+      init({});
+
+      expectNotIntegration('StallTracking');
+    });
+
+    it('when tracing enabled stall tracking integration added by default', () => {
+      init({
+        tracesSampleRate: 0.5,
```
lucas-zimerman, line 467: Q: If we need tracesSampleRate in order to enable stall tracking, shouldn't we also disable it if traceSamplerate is 0?

test/sdk.test.ts
```diff
@@ -456,6 +456,29 @@ describe('Tests the SDK functionality', () => {
       expectNotIntegration('NativeFrames');
     });
 
+    it('no stall tracking integration by default', () => {
```
lucas-zimerman, line 459: NIT: By default it adds the integration, unless the tracing sample rate is not set.

```suggestion
    it('when tracing not set stall tracking integration not added by default', () => {
```

lucas-zimerman (approved):
Some small nits but non blocking, other than that, LGTM!

test/sdk.test.ts
krystofwoldrich: Thanks, yeah I see it's a bit unclear.

test/sdk.test.ts
krystofwoldrich: I thought about it as well, but I think changing the other options based on other options just adds code which effectively does nothing.

The enable options are effectively just shortcuts for adding the integrations.

But if there is a case which would break we should change it.

PR #3998, opened 2024-08-06
(5) ref(tracing): Remove ReactNativeTracing deprecated options and use new integrations style

krystofwoldrich:

- [x] Refactoring

- Removed deprecated ReactNativeTracing option `idleTimeout` use `idleTimeoutMs` instead
- Removed deprecated ReactNativeTracing option `maxTransactionDuration` use `finalTimeoutMs` instead

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


lucas-zimerman (commented):
Looks good to me, but tests are failing with this change, ideally the tests should be passing before we merge it.

src/js/tracing/integrations/appStart.ts
```diff
@@ -123,11 +123,10 @@ export const appStartIntegration = ({
     }
   };
 
-  const afterAllSetup = (client: Client): void => {
+  const afterAllSetup = (_client: Client): void => {
     if (standaloneUserOption === undefined) {
       // If not user defined, set based on the routing instrumentation presence
-      standalone = !client.getIntegrationByName<ReactNativeTracing>(ReactNativeTracing.id)?.options
-        .routingInstrumentation;
+      // FIXME: standalone = getReactNativeTracingIntegration(client)?.options.routingInstrumentation;
```
lucas-zimerman, line 129: Q: Are you gonna fix it on a follow up PR?

lucas-zimerman (approved):
Small question but looks to me, LGTM!

src/js/tracing/integrations/appStart.ts
lucas-zimerman, line 129: If we are not going to fix it right away, we should at least add a `todo` so it doesn't get lost.

```suggestion
      // TODO: FIXME: standalone = getReactNativeTracingIntegration(client)?.options.routingInstrumentation;
```

krystofwoldrich: Thank you for the catch, this will likely change in the future, but for now I left it as is, I removed the comment.

krystofwoldrich:
THank you for the review, could you give it one more look.

I have fixed the comment and also the tests + changelog.

src/js/tracing/integrations/appStart.ts
```diff
@@ -22,7 +22,8 @@ import {
   APP_START_WARM as APP_START_WARM_OP,
   UI_LOAD as UI_LOAD_OP,
 } from '../ops';
-import { ReactNativeTracing } from '../reactnativetracing';
+import { getReactNativeTracingIntegration } from '../reactnativetracing';
+// import { getReactNativeTracingIntegration } from '../reactnativetracing';
```
lucas-zimerman, line 26:
```suggestion

```

lucas-zimerman (approved):
I left a comment to remove some redundant comment, but other than that it looks good to me!

PR #4000, opened 2024-08-08
(6) fix(tracing): ReactNativeTracing and initial navigation spans have to be created after integrations setup

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

This PR removes the dependency of initializing tracing integrations like `NativeFramesTracking` and `StallTracking` before `ReactNativeTracing` integration.

This dependency was introduced in https://github.com/getsentry/sentry-react-native/pull/3996 and https://github.com/getsentry/sentry-react-native/pull/3997 when extracting the mentioned into standalone integration.

The cause of the dependency was that ReactNativeTracing start initiali navigation span during the integrations setup, but should do it after all integrations are setup.

integration tests, sample app 

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

#skip-changelog


lucas-zimerman (approved):
LGTM!

src/js/tracing/reactnativetracing.ts
```diff
@@ -111,6 +111,17 @@ export const reactNativeTracingIntegration = (
...
+
+    instrumentOutgoingRequests({
+      traceFetch: finalOptions.traceFetch,
+      traceXHR: finalOptions.traceXHR,
+      shouldCreateSpanForRequest: finalOptions.shouldCreateSpanForRequest,
+      tracePropagationTargets: client.getOptions().tracePropagationTargets || DEFAULT_TRACE_PROPAGATION_TARGETS,
+    });
+  };
+
+  const afterAllSetup = (): void => {
```
lucas-zimerman, line 124: ● ReactNativeTracing › Tracing Instrumentation › With routing instrumentation › Cancels route transaction when app goes to background 

Seems to be failing on this specific PR, could you double-check if it broke with this changes?

lucas-zimerman, line 124: ● ReactNativeTracing › Tracing Instrumentation › With routing instrumentation › Cancels route transaction when app goes to background 

Seems to be failing on this specific PR, could you double-check if it broke with this changes?

lucas-zimerman (changes requested):

src/js/tracing/reactnativetracing.ts
krystofwoldrich: I forgot to update the test to call the newly added `afterAllSetup`. Thank you.

krystofwoldrich:
I've fixed the JS tests.

The e2e fails due to https://github.com/getsentry/sentry-react-native/pull/4004 unrelated to this PR, it will be fixed when I merge main to v6.

lucas-zimerman (approved):
LGTM!

PR #4006, opened 2024-08-09
Feat: Allow users to exclude Replay package from Web

lucas-zimerman:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


src/js/tools/metroconfig.ts
```diff
@@ -146,6 +157,26 @@ function withSentryDebugId(config: MetroConfig): MetroConfig {
...
 }
 
+function excludeSentryWebReplay(config: MetroConfig): MetroConfig {
+  const originalResolver = config.resolver?.resolveRequest;
+
+  return {
+    ...config,
+    resolver: {
+      ...config.resolver,
+      resolveRequest: (context, moduleName, platform) => {
```
krystofwoldrich, line 167: Would it make sense to include web replay on `platform=web`?

So it work for web out of the box and when users want to optimalize the bundle size and are not using Replay they can use the option `includeWebReplay`.

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@
...
+  const { getSentryExpoConfig } = require("@sentry/react-native/metro");
+  const config = getSentryExpoConfig(__dirname, { includeWebReplay: true });
+
+  // For RN
+  const { getDefaultConfig } = require('@react-native/metro-config');
+  const { withSentryConfig } = require('@sentry/react-native/metro');
+  module.exports = withSentryConfig(getDefaultConfig(__dirname), { includeWebReplay: true });
+
+```
+
```
krystofwoldrich, line 20: Nit. Just moved some white spaces around.

````suggestion
- Exclude Sentry Web Replay by default, reducing the code in 130KB. ([#4006](https://github.com/getsentry/sentry-react-native/pull/4006))
  - You can keep Sentry Web Replay by setting `includeWebReplay` to `true` in your metro config as shown in the snippet:

  ```js
  // For Expo
  const { getSentryExpoConfig } = require("@sentry/react-native/metro");
  const config = getSentryExpoConfig(__dirname, { includeWebReplay: true });

  // For RN
  const { getDefaultConfig } = require('@react-native/metro-config');
  const { withSentryConfig } = require('@sentry/react-native/metro');
  module.exports = withSentryConfig(getDefaultConfig(__dirname), { includeWebReplay: true });
  ```

````

File: src/js/tools/metroconfig.ts [hunk 3]
```diff
@@ -18,6 +18,11 @@ export interface SentryMetroConfigOptions {
    * @default false
    */
   annotateReactComponents?: boolean;
+  /**
```
krystofwoldrich, line 21:
```suggestion

  /**
```

krystofwoldrich:
Awesome size reduction, thank you!

I left some small suggestions/nits to changelog. And one comment about behaviour for rn-web.

src/js/tools/metroconfig.ts
```diff
@@ -146,6 +158,29 @@ function withSentryDebugId(config: MetroConfig): MetroConfig {
   };
 }
 
+/**
+ *
+ */
+export function excludeSentryWebReplay(config: MetroConfig, includeWebReplay: boolean | undefined): MetroConfig {
```
krystofwoldrich, line 164: Let's rename `excludeSentryWebReplay` to `includeSentryWebReplay` to match the flag name `includeWebReplay`.

To make it easier to read

`excludeSentryWebReplay(config, true)` -> web replay is included (but looks like exclude true)

vs

`includeSentryWebReplay(config, true)` -> web replay is included

src/js/tools/metroconfig.ts
```diff
@@ -146,6 +158,29 @@ function withSentryDebugId(config: MetroConfig): MetroConfig {
   };
 }
 
+/**
+ *
```
krystofwoldrich, line 162:
```suggestion
 * Includes `@sentry/replay` packages based on the `includeWebReplay` flag and current bundle `platform`.
```

test/tools/metroconfig.test.ts
```diff
@@ -108,9 +112,116 @@ describe('metroconfig', () => {
...
+
+    test('removes Web Replay when platform is web and includeWebReplay is false', () => {
+      const modifiedConfig = excludeSentryWebReplay(config, false);
+      const result = resolveRequest(modifiedConfig, contextMock, '@sentry/replay', 'web');
+
+      expect(result).toEqual({ type: 'empty' });
+      expect(originalResolverMock).not.toHaveBeenCalled();
+    });
+
+    /* Test expected to fail but not required since excludeSentryWebReplay is not run when includeWebReplay is true.
```
krystofwoldrich, line 151: This test should not fail, if a user forces to include the replay package it should be included.

test/tools/metroconfig.test.ts
```diff
@@ -108,9 +112,116 @@ describe('metroconfig', () => {
...
+    test('removes Web Replay when platform is android and includeWebReplay is undefined', () => {
+      const modifiedConfig = excludeSentryWebReplay(config, undefined);
+      const result = resolveRequest(modifiedConfig, contextMock, '@sentry/replay', 'android');
+
+      expect(result).toEqual({ type: 'empty' });
+      expect(originalResolverMock).not.toHaveBeenCalled();
+    });
+
+    /* Test expected to fail but not required since excludeSentryWebReplay is not run when includeWebReplay is true.
+    test('keep Web Replay when platform is ios and includeWebReplay is true', () => {
```
krystofwoldrich, line 177: The same as the test above.

test/tools/metroconfig.test.ts
```diff
@@ -108,9 +112,116 @@ describe('metroconfig', () => {
...
+
+    test('removes Web Replay when platform is ios and includeWebReplay is undefined', () => {
+      const modifiedConfig = excludeSentryWebReplay(config, undefined);
+      const result = resolveRequest(modifiedConfig, contextMock, '@sentry/replay', 'ios');
+
+      expect(result).toEqual({ type: 'empty' });
+      expect(originalResolverMock).not.toHaveBeenCalled();
+    });
+
+    test('calls originalResolver when moduleName is not @sentry/replay', () => {
```
krystofwoldrich, line 201: Can you add tests for indludeReplay=false and module not @sentry/replay?

At the moment an empty module would be returned always.

src/js/tools/metroconfig.ts
```diff
@@ -146,6 +158,29 @@ function withSentryDebugId(config: MetroConfig): MetroConfig {
...
+ */
+export function excludeSentryWebReplay(config: MetroConfig, includeWebReplay: boolean | undefined): MetroConfig {
+  const originalResolver = config.resolver?.resolveRequest;
+
+  return {
+    ...config,
+    resolver: {
+      ...config.resolver,
+      resolveRequest: (context, moduleName, platform) => {
+        if (includeWebReplay === false || (platform !== 'web' && moduleName.includes('@sentry/replay'))) {
```
krystofwoldrich, line 172: After the flag change this will simplify to

```suggestion
        if (excludeWebReplay && moduleName.includes('@sentry/replay')) {
```

src/js/tools/metroconfig.ts
```diff
@@ -18,6 +18,11 @@ export interface SentryMetroConfigOptions {
    * @default false
    */
   annotateReactComponents?: boolean;
+  /**
+   * Adds the Sentry replay package for web.
+   * @default false (Mobile) undefined (Web)
```
krystofwoldrich, line 23:
```suggestion
   * @default false (Mobile) true (Web)
```

krystofwoldrich:
There is one lint issue.

But also none of the 0.65 tests builds. We have to fix that before moving forward with this.

lucas-zimerman:
Visual Sample from metro 0.66 (old metro where the pattern changed)

Including Web replay
image

Excluding Web replay
image

src/js/tools/metroconfig.ts
```diff
@@ -158,26 +158,83 @@ function withSentryDebugId(config: MetroConfig): MetroConfig {
   };
 }
 
+type ResolverThreeParams = (
```
krystofwoldrich, line 161: Can we use `CustomResolver` type form `metro-resolver`?

src/js/tools/metroconfig.ts
```diff
@@ -158,26 +158,83 @@ function withSentryDebugId(config: MetroConfig): MetroConfig {
...
+  context: CustomResolutionContext,
+  moduleName: string,
+  platform: string | null,
+) => Resolution;
+type ResolverFourParams = (
```
krystofwoldrich, line 166: Can we call this `CustomResolverBeforeMetro_0_68`

src/js/tools/metroconfig.ts
```diff
@@ -158,26 +158,83 @@ function withSentryDebugId(config: MetroConfig): MetroConfig {
...
+) => Resolution;
+
 /**
  * Includes `@sentry/replay` packages based on the `includeWebReplay` flag and current bundle `platform`.
  */
 export function withSentryResolver(config: MetroConfig, includeWebReplay: boolean | undefined): MetroConfig {
-  const originalResolver = config.resolver?.resolveRequest;
+  const originalResolver = config.resolver?.resolveRequest as ResolverThreeParams | ResolverFourParams;
+
+  const hasSentryReplay = (platform: string | null, moduleName: string): boolean => {
```
krystofwoldrich, line 179: `hasSentryReplay` -> `shouldIncludeSentryReplay`

src/js/tools/metroconfig.ts
```diff
@@ -158,26 +158,83 @@ function withSentryDebugId(config: MetroConfig): MetroConfig {
...
+    return (
+      (includeWebReplay === false || (includeWebReplay === undefined && platform !== 'web')) &&
+      moduleName.includes('@sentry/replay')
+    );
+  };
+
+  let resolver: ResolverThreeParams | ResolverFourParams;
+
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
```
krystofwoldrich, line 188: Can we avoid this by checking `context.resolveRequest === sentryResolverRequest` inside of the `sentryResolverRequest`?

Than we could have one sentryResolverRequest implementation.

src/js/tools/metroconfig.ts
```diff
@@ -158,26 +158,83 @@ function withSentryDebugId(config: MetroConfig): MetroConfig {
...
+      }
+      if (originalResolver) {
+        return originalResolver(context, moduleName, platform);
+      }
+      return context.resolveRequest(context, moduleName, platform);
+    };
+  } else {
+    // On older Metro, the given context from resolver is not the defaultResolver but the called function itself.
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-var-requires, import/no-extraneous-dependencies
+    const defaultResolver = require('metro-resolver').resolve;
```
krystofwoldrich, line 206: Let's only require this when `originalResolver` is `undefined`

In case there is an issue with the require so users have a way out.

Let's also do the require in `try catch` and on error let's log out from what path we tried the resolver and where can user open issue. And how to opt out -> setting `includeWebReplay=true`

similar to https://github.com/getsentry/sentry-react-native/blob/13a38606494c437f28c974fc2544a583d5a8f4f1/packages/core/src/js/tools/vendor/metro/utils.ts#L99

and on success let's once (the first time) log out that we are using the fallback and the path.

src/js/tools/metroconfig.ts
```diff
@@ -158,26 +158,83 @@ function withSentryDebugId(config: MetroConfig): MetroConfig {
...
+type ResolverThreeParams = (
+  context: CustomResolutionContext,
+  moduleName: string,
+  platform: string | null,
+) => Resolution;
+type ResolverFourParams = (
+  context: CustomResolutionContext,
+  moduleName: string,
+  platform: string | null,
+  realModuleName?: string,
```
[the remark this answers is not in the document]
lucas-zimerman: the last parameter is the module name

src/js/tools/metroconfig.ts
```diff
@@ -158,26 +158,83 @@ function withSentryDebugId(config: MetroConfig): MetroConfig {
...
 /**
  * Includes `@sentry/replay` packages based on the `includeWebReplay` flag and current bundle `platform`.
  */
 export function withSentryResolver(config: MetroConfig, includeWebReplay: boolean | undefined): MetroConfig {
-  const originalResolver = config.resolver?.resolveRequest;
+  const originalResolver = config.resolver?.resolveRequest as ResolverThreeParams | ResolverFourParams;
+
+  const hasSentryReplay = (platform: string | null, moduleName: string): boolean => {
+    return (
+      (includeWebReplay === false || (includeWebReplay === undefined && platform !== 'web')) &&
```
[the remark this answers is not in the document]
lucas-zimerman: I think it's set to null when someone tries to run in some unsupported platform (not ios,windows, ios, android, web).
But I like the idea of including replay when platform is null

src/js/tools/metroconfig.ts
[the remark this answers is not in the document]
lucas-zimerman: I see why the confusion now, I was following at one point this example https://github.com/facebook/metro/issues/542#issuecomment-1059821963 and later on the code implemented by metro, on the code snippet on that link the module name are inverted

src/js/tools/metroconfig.ts
lucas-zimerman: atm users can define the resolveRequest on MetroConfig to not enable it or even set includeWebReplay as true so that withSentryResolver is not executed.

src/js/tools/metroconfig.ts
lucas-zimerman: we now have one SentryResolverRequest implementation

krystofwoldrich: > Can we avoid this by checking context.resolveRequest === sentryResolverRequest inside of the sentryResolverRequest?

I ment can we avoid `require(metro)` to check the version? Requiring the metro package doesn't have to be accurate due to how hoisting of node modules works. 

We had issues with this in the past when requiring `defaultSerializer` from metro.

If we instead check `context.resolveRequest === sentryResolverRequest` we know that we are running in the old metro version. This should be safer as we are not relying on the hoister module to be the correct one.

src/js/tools/metroconfig.ts
```diff
@@ -18,6 +18,11 @@ export interface SentryMetroConfigOptions {
    * @default false
    */
   annotateReactComponents?: boolean;
+  /**
+   * Adds the Sentry replay package for web.
+   * @default false (Mobile) true (Web)
```
krystofwoldrich, line 23: Since the resolver got more complex due to older Metro versions, let's change this to opt in on all platforms, to avoid breaking builds on accident.

We can update the default later, when we get feedback on the feature.

```suggestion
   * @default true
```

src/js/tools/metroconfig.ts
```diff
@@ -46,6 +51,9 @@ export function withSentryConfig(
   if (annotateReactComponents) {
     newConfig = withSentryBabelTransformer(newConfig);
   }
+  if (includeWebReplay !== true) {
```
krystofwoldrich, line 54:
```suggestion
  if (options.includeWebReplay === false) {
```

src/js/tools/metroconfig.ts
```diff
@@ -146,6 +158,66 @@ function withSentryDebugId(config: MetroConfig): MetroConfig {
...
+      return oldMetroModuleName
+        ? originalResolver(context, moduleName, platform, oldMetroModuleName)
+        : originalResolver(context, moduleName, platform);
+    }
+
+    // Prior 0.68, resolve context.resolveRequest is sentryResolver itself, where on later version it is the default resolver.
+    if (context.resolveRequest === sentryResolverRequest) {
+      // eslint-disable-next-line no-console
+      console.error(
+        `Error: [@sentry/react-native/metro] Cannot desolve the defaultResolver on Metro older than 0.68.
```
krystofwoldrich, line 198:
```suggestion
        `Error: [@sentry/react-native/metro] Can not resolve the defaultResolver on Metro older than 0.68.
```

krystofwoldrich (approved):
Thank you, this was more complex than expected, due to the older Metro versions.

Looks great now!

src/js/tools/metroconfig.ts
```diff
@@ -46,6 +51,9 @@ export function withSentryConfig(
   if (annotateReactComponents) {
     newConfig = withSentryBabelTransformer(newConfig);
   }
+  if (options.includeWebReplay === false) {
```
krystofwoldrich, line 54:
```suggestion
  if (includeWebReplay === false) {
```

Issue #2797, opened 2023-01-30
`Sentry.crashedLastRun` in React Native


krystofwoldrich:
### Description

The native SDK can give users information about `crashedLastRun` but RN SDK doesn't have such a feature. It might be useful for RN devs too.

Some user patched this feature in their app like this, but it's not working in the last RN SDK version.
```cpp
RCT_EXPORT_METHOD(didCrashLastLaunch:(RCTPromiseResolveBlock)resolve
                  rejecter:(RCTPromiseRejectBlock)reject)
{
    if ([SentrySDK crashedLastRun]) {
        resolve(@YES);
    } else {
        resolve(@NO);
    }
}
```

marandaneto:
I don't think we have this feature on the Android SDK tho, so let's confirm and raise an issue on the Android SDK as well if needed.

krystofwoldrich:
I think this PR added it 
- https://github.com/getsentry/sentry-java/pull/1739

MikaReesu:
Hello, I was trying to use the isCrashedLastRun on Android with Kotlin but it didn't work. 

Is it still not implemented? 

Thank you for your answers.

krystofwoldrich:
@MikaReesu Hi, thank you for the message,
this is not implemented in RN yet.

Only in the native SDKs (cocoa, java) at the moment.

MikaReesu:
I see, thank you.

LonelyCpp:
This would be quite helpful!

krystofwoldrich:
@LonelyCpp Thanks for letting us know you're interested.

PR #4014, opened 2024-08-09
Implement crashedLastRun

janicduplessis:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This exposes `crashedLastRun` from the iOS sdk and `isCrashedLastRun` from the android sdk to react native.

It is useful to know if the app crashed in the previous launch.

Fixes https://github.com/getsentry/sentry-react-native/issues/2797

Tested in an app on iOS and Android by making it crash then checking that it returns true on next open, or false if didn't crash before.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


krystofwoldrich:
@janicduplessis Thank you for the PR, we'll review this shortly.

src/js/sdk.tsx
```diff
@@ -246,3 +246,14 @@ export function configureScope(callback: (scope: Scope) => void): ReturnType<Hub
   };
   getCurrentHub().configureScope(safeCallback);
 }
+
+/**
+ * Returns if the app crashed in the last run.
+ */
+export async function crashedLastRun(): Promise<boolean> {
```
krystofwoldrich, line 253: Overall everything looks good. I have only one comment.

All the fallbacks result to `false` which might be false positive, as we are just unable to read the value. To avoid that I think returning `boolean | undefined` is better, as users of the API can decide what to do if the value was not retrieved.  

```suggestion
export async function crashedLastRun(): Promise<boolean | undefined> {
```

lucas-zimerman: Also agree with it, returning false when we are not certain could give a false impression that the app didn't crash

janicduplessis:
Updated to return null when the sdk does. Also added it to the sample app.

krystofwoldrich:
Awesome, thank you, I'll review the changes soon.

src/js/NativeRNSentry.ts
```diff
@@ -46,6 +46,7 @@ export interface Spec extends TurboModule {
   initNativeReactNavigationNewFrameTracking(): Promise<void>;
   captureReplay(isHardCrash: boolean): Promise<string | undefined | null>;
   getCurrentReplayId(): string | undefined | null;
+  crashedLastRun(): Promise<boolean | null>;
```
krystofwoldrich, line 49: iOS nil is converted to undefined and Java null to null, it should not cause any issues, but just so the typing is correct.

```suggestion
  crashedLastRun(): Promise<boolean | undefined | null>;
```

src/js/wrapper.ts
```diff
@@ -642,6 +644,17 @@ export const NATIVE: SentryNativeWrapper = {
...
 
+  async crashedLastRun(): Promise<boolean | null> {
+    if (!this.enableNative) {
+      return null;
+    }
+    if (!this._isModuleLoaded(RNSentry)) {
+      return null;
+    }
+
+    return RNSentry.crashedLastRun();
```
krystofwoldrich, line 655: Related to the native type comment, ensuring only null is passed back if the bool is not evaluated.

```suggestion
    const result = RNSentry.crashedLastRun();
    return typeof result === 'boolean' ? result : null;
```

krystofwoldrich:
I've added simple test for the JS interface, but besides that it looks good.

Thank you!

krystofwoldrich (approved):

PR #4029, opened 2024-08-16
ref: Change native frames from `traceId` to `spanId` 

krystofwoldrich:

- [x] Enhancement
- [x] Refactoring

## :scroll: Description

This PR changes the native frames from attaching to `traceId` which can be the same for multiple spans to `spanId`. 

- This fixes potencial situations when multiple root spans share the same trace id. This is possible since JS v8, before Transaction (now Root Spans) would always have unique trace id. 
- This is needed for attaching the native frames as span level measurements. 

sample app, unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 5.29.0
 
 ### Features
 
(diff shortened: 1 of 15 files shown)


krystofwoldrich:
This merge was a mistake from VSCode GH plugin. I'll open a new PR.

PR #4030, opened 2024-08-16
ref: Change native frames from `traceId` to `spanId` 

krystofwoldrich:

- [x] Enhancement
- [x] Refactoring

## :scroll: Description

This PR changes the native frames from attaching to `traceId` which can be the same for multiple spans to `spanId`. 

- This fixes potencial situations when multiple root spans share the same trace id. This is possible since JS v8, before Transaction (now Root Spans) would always have unique trace id. 
- This is needed for attaching the native frames as span level measurements. 

sample app, unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/src/js/integrations/default.ts
+++ b/src/js/integrations/default.ts
@@ -12,6 +12,8 @@ import {
   browserGlobalHandlersIntegration,
   browserLinkedErrorsIntegration,
   browserReplayIntegration,
+  createNativeFramesIntegrations,
+  createReactNativeRewriteFrames,
   debugSymbolicatorIntegration,
   dedupeIntegration,
   deviceContextIntegration,
@@ -24,7 +26,6 @@ import {
   inboundFiltersIntegration,
   mobileReplayIntegration,
   modulesLoaderIntegration,
-  nativeFramesIntegration,
   nativeLinkedErrorsIntegration,
   nativeReleaseIntegration,
   reactNativeErrorHandlersIntegration,
@@ -36,7 +37,6 @@ import {
   userInteractionIntegration,
   viewHierarchyIntegration,
 } from './exports';
-import { createReactNativeRewriteFrames } from './rewriteframes';
 
 /**
  * Returns the default ReactNative integrations based on the current environment.
@@ -104,8 +104,9 @@ export function getDefaultIntegrations(options: ReactNativeClientOptions): Integ
   if (hasTracingEnabled && options.enableAppStartTracking) {
     integrations.push(appStartIntegration());
   }
-  if (hasTracingEnabled && options.enableNativeFramesTracking) {
-    integrations.push(nativeFramesIntegration());
+  const nativeFrames = createNativeFramesIntegrations(hasTracingEnabled && options.enableNativeFramesTracking);
+  if (nativeFrames) {
+    integrations.push(nativeFrames);
   }
   if (hasTracingEnabled && options.enableStallTracking) {
     integrations.push(stallTrackingIntegration());
--- a/src/js/integrations/default.ts
+++ b/src/js/integrations/default.ts
@@ -104,9 +104,11 @@ export function getDefaultIntegrations(options: ReactNativeClientOptions): Integ
   if (hasTracingEnabled && options.enableAppStartTracking) {
     integrations.push(appStartIntegration());
   }
-  const nativeFrames = createNativeFramesIntegrations(hasTracingEnabled && options.enableNativeFramesTracking);
-  if (nativeFrames) {
-    integrations.push(nativeFrames);
+  const nativeFramesIntegrationInstance = createNativeFramesIntegrations(
+    hasTracingEnabled && options.enableNativeFramesTracking,
+  );
+  if (nativeFramesIntegrationInstance) {
+    integrations.push(nativeFramesIntegrationInstance);
   }
   if (hasTracingEnabled && options.enableStallTracking) {
     integrations.push(stallTrackingIntegration());
--- a/src/js/tracing/integrations/nativeFrames.ts
+++ b/src/js/tracing/integrations/nativeFrames.ts
@@ -12,16 +12,16 @@ import { NATIVE } from '../../wrapper';
 const FETCH_FRAMES_TIMEOUT_MS = 2_000;
 
 /**
- * Timeout from the final native frames fetch to processing the associated transaction.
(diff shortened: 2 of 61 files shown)


src/js/integrations/default.ts
```diff
@@ -104,8 +104,9 @@ export function getDefaultIntegrations(options: ReactNativeClientOptions): Integ
   if (hasTracingEnabled && options.enableAppStartTracking) {
     integrations.push(appStartIntegration());
   }
-  if (hasTracingEnabled && options.enableNativeFramesTracking) {
-    integrations.push(nativeFramesIntegration());
+  const nativeFrames = createNativeFramesIntegrations(hasTracingEnabled && options.enableNativeFramesTracking);
+  if (nativeFrames) {
+    integrations.push(nativeFrames);
```
lucas-zimerman, line 109: I will suggest changing the name of nativeFrames to nativeFramesIntegration to keep it consistent with the names being pushed below
```suggestion
  const nativeFramesIntegratin = createNativeFramesIntegrations(hasTracingEnabled && options.enableNativeFramesTracking);
  if (nativeFrames) {
    integrations.push(nativeFramesIntegration);
```

src/js/utils/AsyncExpiringMap.ts
```diff
@@ -0,0 +1,163 @@
...
+    };
+    this._map.set(key, entry);
+
+    promise.then(
+      value => {
+        entry.value = value;
+        entry.expiresAt = Date.now() + this._ttl;
+        entry.promise = null;
+      },
+      () => {
```
lucas-zimerman, line 51: Q: Should we log when an error occurs ?

src/js/tracing/integrations/nativeFrames.ts
```diff
@@ -1,239 +1,201 @@
...
+ * Timeout from the start of a span to fetching the associated native frames.
+ */
+const FETCH_FRAMES_TIMEOUT_MS = 2_000;
+
+/**
+ * Timeout from the final native frames fetch to processing the associated transaction.
+ * If the transaction is not processed by this time, the native frames will be dropped
+ * and not added to the event.
+ */
+const END_FRAMES_TIMEOUT_MS = 2_000;
```
lucas-zimerman, line 19: I can not really see the difference of `END_FRAMES_TIMEOUT_MS` and `START_FRAMES_TIMEOUT_MS` by the comments since they are exactly the same. Does the Start frame apply to the start of the native frames fetch?

src/js/tracing/integrations/nativeFrames.ts
```diff
@@ -1,239 +1,201 @@
...
-  ): Promise<FramesMeasurements | null> => {
-    if (_finishFrames.has(traceId)) {
-      logger.debug(`[NativeFrames] Native end frames already fetched for trace id (${traceId}).`);
-      return Promise.resolve(_prepareMeasurements(traceId, finalEndTimestamp, startFrames));
+  const processEvent = async (event: Event): Promise<Event> => {
+    if (
+      event.type !== 'transaction' ||
+      !event.transaction ||
+      !event.contexts ||
+      !event.contexts.trace ||
```
lucas-zimerman, line 152: We could check if contexts exits by checking trace itself.

```suggestion
      !event.contexts?.trace ||
```

test/tracing/integrations/nativeframes.test.ts
```diff
@@ -216,6 +229,17 @@ describe('NativeFramesInstrumentation', () => {
   });
 
   it('does not set measurements on a transaction event for which finishFrames times out.', async () => {
+    asyncProcessorBeforeNativeFrames = async (event: Event) => {
+      await Promise.resolve();
+      await Promise.resolve();
+      await Promise.resolve();
+
+      global.Date.now = jest.fn(() => mockDate.getTime() + 2100); // hardcoded final frames timeout 2000ms
```
lucas-zimerman, line 237:
```suggestion
      global.Date.now = jest.fn(() => mockDate.getTime() + 2100); // hardcoded final frames timeout 2100ms
```

test/tracing/integrations/nativeframes.test.ts
```diff
@@ -216,6 +229,17 @@ describe('NativeFramesInstrumentation', () => {
...
+      await jest.runOnlyPendingTimersAsync();
+      await jest.advanceTimersByTimeAsync(2100); // hardcoded final frames timeout 2000ms
```
lucas-zimerman, line 239:
```suggestion
      await jest.advanceTimersByTimeAsync(2100); // hardcoded final frames timeout 2100ms
```

src/js/utils/AsyncExpiringMap.ts
```diff
@@ -0,0 +1,163 @@
...
+      return remainingTime > 0 ? remainingTime : 0;
+    }
+    return undefined;
+  }
+
+  /**
+   * Remove expired entries.
+   */
+  public cleanup(): void {
+    const now = Date.now();
```
lucas-zimerman, line 130: Q: Shouldn't we add a lock between cleanup and set, or add some race condition test between them?

lucas-zimerman (approved):
Non-blocking questions and also some minor suggestions. Let me know if you need more reviews.

src/js/utils/AsyncExpiringMap.ts
krystofwoldrich: No, it's responsibility of the user of the Map to handle the promise rejection.

test/tracing/integrations/nativeframes.test.ts
```diff
@@ -216,6 +229,17 @@ describe('NativeFramesInstrumentation', () => {
   });
 
   it('does not set measurements on a transaction event for which finishFrames times out.', async () => {
+    asyncProcessorBeforeNativeFrames = async (event: Event) => {
+      await Promise.resolve();
+      await Promise.resolve();
+      await Promise.resolve();
```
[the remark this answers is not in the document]
krystofwoldrich: There are multiple background promises that I need to ensure are resolved -> so the expiring map get cleaned with the fake timers enabled. It's a bit sketch but it works.

test/tracing/integrations/nativeframes.test.ts
krystofwoldrich: The timeout is 2000, that's why we move by 2100 to ensure it was triggered.

Issue #4035, opened 2024-08-19
'ReactCommon/RCTTurboModule.h' file not found RN 0.75 New Architecture


serhii-yalla:
**OS:**
- [ ]  Windows
- [ ]  MacOS
- [X]  Linux

**Platform:**
- [X]  iOS
- [ ]  Android

**SDK:**
- [X]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 5.30.0

**`react-native` version:** 0.75.1

**Are you using Expo?**
- [ ]  Yes
- [X]  No

**Are you using sentry.io or on-premise?**
- [X]  sentry.io (SaaS)
- [ ]  on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

[Link to issue]

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  dsn: 'https://...@sentry.io/...'
  // other options
});
```

or

(`react-native-sentry`)
```
Sentry.config(
  'https://...@sentry.io/...'
  // other options
 ).install();
```

---
**I have the following issue:**

Can't build app in XCode with 0.75.1 react native and new arch enabled.
```
'ReactCommon/RCTTurboModule.h' file not found
In file included from /Users/pprnd/Documents/yalla_cashier/node_modules/@sentry/react-native/ios/RNSentry.mm:35:
```

**Steps to reproduce:**
- Step 1 Set up app with 0.75.1 react native and add @sentry/react-native 5.30.0
- Step 2 Install deps, pods and try to build app

**Actual result:**

![Image](https://github.com/user-attachments/assets/a8aeb59f-4f1f-4cfd-a103-cf88a02b1cc1)

**Expected result:**

Successful app build

krystofwoldrich:
Thank you @serhii-yalla for the report, we will investigate this.

serhii-yalla:
@krystofwoldrich When can I expect new release?

thespacemanatee:
@krystofwoldrich we're also eagerly waiting for the release!

krystofwoldrich:
This fix will be released later today.

frankcalise:
I am experiencing this issue with RN 0.75.3, Expo 51 and @sentry/react-native 5.24.3

Kaveh-ap:
I'm getting the same error with RN 0.75.4 and @sentry/react-native: 5.31.1

gunnartorfis:
[...]

PR #4040, opened 2024-08-22
Use `install_modules_dependencies` for React iOS dependencies

oblador:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Fixes https://github.com/getsentry/sentry-react-native/issues/4035

`install_modules_dependencies` helper have been [available since 0.71](https://reactnative.dev/blog/2023/01/12/version-071#new-architecture) that determines what react dependencies to add depending on new arch flags etc. This is preferred and would make sure it's forwards and backwards compatible. 

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Use `install_modules_dependencies` for React iOS dependencies ([#4040](https://github.com/getsentry/sentry-react-native/pull/4040))
+
 ## 5.30.0
 
 ### Features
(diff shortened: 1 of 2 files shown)


krystofwoldrich:
Thank you @oblador for the fix. It looks good, I've returned the dependencies for the else branch when the `install_modules_dependencies` function is not available, not to introduce any breaking changes for older RN versions with new architecture. (Likely very small group is any, but we can't introduce a breaking change in fixup version)

Now just CI needs to pass and we can merge this.

Issue #3809, opened 2024-05-09
[Android] TimeToInitialDisplay doesn't track Tab screens from react-navigation


serzmerz:
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [ ]  iOS
- [x]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 5.22.2

**`react-native` version:** 0.73.6

**Are you using Expo?**
- [ ]  Yes
- [x]  No

**Are you using sentry.io or on-premise?**
- [ ]  sentry.io (SaaS)
- [x]  on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

[Link to issue]

**Configuration:**

(`@sentry/react-native`)
```
export const routingInstrumentation = new NativeSentry.ReactNavigationInstrumentation({
  enableTimeToInitialDisplay: true
});

Sentry.init({
  dsn: Config.get("REACT_APP_SENTRY_DSN"),
  environment: Config.get("ENV"),
  ignoreErrors,
  integrations: [
    new NativeSentry.ReactNativeTracing({
      routingInstrumentation
    })
  ],
  tracesSampleRate: Config.get("ENV") === "prod" && !__DEV__ ? 0.2 : 1,
  _experiments: {
    profilesSampleRate: Config.get("ENV") === "prod" && !__DEV__ ? 0 : 1
  }
});
```

---
**I have the following issue:**

On Android, when `enableTimeToInitialDisplay` flag is active, sentry doesn't capture initial time on screens from Tab navigator (react-navigation). It is happening in `@sentry/react-native/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java` because:
1. Tab screen has Class name `com.swmansion.rnscreens.ScreenFragment`, not `com.swmansion.rnscreens.ScreenStackFragment` and it doesnt subscribe to `eventDispatcher`.
2. Even if I remove this condition, EventDispatcher doesn't report `com.swmansion.rnscreens.events.ScreenAppearEvent` event, since this event(transitionEnd) only available for Stack screens.

My navigation setup is next: Root stack navigator and one of the screens is TabNavigator

```
<RootStackNavigator.Navigator screenOptions={rootScreenOptions}>
        {routes.map(({ name, component, options }) => (
          <RootStackNavigator.Screen
[...]

PR #4042, opened 2024-08-22
Fix: Implement fallback system to screens that aren't reporting on the native layer the time to display.

### LINKED ISSUE #3934: Manual TTID/TTFD API times out
**Platform:**
- [x]  iOS

**SDK:**

**SDK version:** 5.22.0

**`react-native` version:** 0.72.3

**Are you using Expo?**
- [ ]  Yes
- [x]  No

- [x]  sentry.io (SaaS)
- [ ]  on-premise

**Configuration:**

```
export const routingInstrumentation = new Sentry.ReactNavigationInstrumentation(
  {
  },
);


Sentry.init({
  environment: 'development'
  sampleRate: 1.0,
  tracesSampleRate: 0.1,
  enableAppHangTracking: false,
  integrations: [
      routingInstrumentation,
      tracePropagationTargets: [currentConfig.DOMAIN],
    }),
  ],
});
```
---
**I have the following issue:**

I am seeing some inconsistencies in the data reported and what should be reported based on the SDK behaviour. For example, it is odd to me that some transactions are being marked as '***deadline_exceeded***' when they do not exceed the default value for [[finalTimeoutMs](https://github.com/getsentry/sentry-react-native/blob/main/src/js/tracing/reactnativetracing.ts#L61)]. `finalTimeoutMs` is also not being overridden in the init code block

The default for `finalTimeoutMs` is 600000ms or 10mins, which is way above what the transactions being marked deadline_exceeded are reporting. The amount of failed transactions with this status seems to have increased after updating the SDK version from `5.3.0` to `5.22.0`

More details with links to examples in the [Customer Case](https://www.notion.so/sentry/Increase-in-deadline_exceeded-after-upgrade-9d472c310520417eb9ed4ab72a554dcc?pvs=4)

lucas-zimerman:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Context: the current native implementation doesn't always emit the required event in order to track the required time to display to be set on the time to display span. 

To fix that, I introduced an additional fallback emitter, in short, it reuses the same flow by emitting the original [event](https://github.com/getsentry/sentry-react-native/pull/4042/files#diff-2b6b8eb3a7668d3a9f076a37f76f1b330c5c4219ab8c08afa5ed4c9609535a76R61) when the fallback emitter received the notification that the page was rendered and the original emitter didn't emit any event on the following 3 seconds.

With those changes, we have the following benefits:
- Support for React Native web by also having the JavaScript fallback of [requestAnimationFrame](https://reactnative.dev/docs/timers).
- Greatly reduces the wrong time to display end timestamp by having a fallback when the original implementation doesn't return any event.
- Fallback to the JavaScript implementation in case both Native emitters fail.

Close #3934, https://github.com/getsentry/sentry-react-native/issues/3809

The playground tab on our sample is a good case for testing, since it doesn't use navigation stack on it, the events on this condition are limited, not generating and required event by us in order to track the time to display.

I used it and other tabs to compare the difference between the original implementation and also `requestAnimationFrame`, and the time difference between each other was quite low:

1St Tab
JS: requestAnimationFrame, "newFrameTimestampInSeconds" is 1724293664.8300002
Original: InitAsync Event received {"newFrameTimestampInSeconds":1724293664.8179998}

2Nd Tab
JS: requestAnimationFrame, "newFrameTimestampInSeconds" is 1724293761.688
Original: InitAsync Event received {"newFrameTimestampInSeconds":1724293761.689}

3Rd Tab
JS: requestAnimationFrame, "newFrameTimestampInSeconds" is 1724293824.887
Original: Android didn't emit an event for this page so it wasn't measured

Before this change on the playground screen:
https://sentry-sdks.sentry.io/performance/trace/10bd4bf28052404592f408f3cf58175a/?dataset=transactions&field=title&field=event.type&field=project&field=user.display&field=timestamp&field=replayId&fov=0%2C15046.88916015625&id=21705&name=&node=trace-root&project=5428561&query=&queryDataset=transaction-like&sort=-timestamp&source=discover&statsPeriod=1h&timestamp=1724358725&topEvents=5&yAxis=count%28%29

After this change on the playground screen:
https://sentry-sdks.sentry.io/performance/trace/8cb478b4370e444fa4a7b9be778d51ab/?dataset=transactions&field=title&field=event.type&field=project&field=user.display&field=timestamp&field=replayId&fov=0%2C235&id=21705&name=&node=txn-b5a8ba18aa254fe79743133baf9071b2&project=5428561&query=&queryDataset=transaction-like&sort=-timestamp&source=discover&statsPeriod=1h&timestamp=1724359145&topEvents=5&yAxis=count%28%29

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes


krystofwoldrich:
Hi @lucas-zimerman,
thank you for the detailed description, and opening the draft PR, this makes validation and discussing the ideas much easier.

I think waiting for the next drawn frame is a good fallback. But there is one party breaker hidden in the RN implementation. Compared to browsers implementation, in RN `requestAnimationFrame` is the same as `setTimeout(timeout, 0)`.

This means the callback runs after the React Native Node Tree is assembled and passed to native. So unless the JS loop is doing some heavy task, it will generally be executed before the render.

We can fix that by implementing our own `requestAnimationFrame` which will emit an event from the native layer when the next frame is rendered.

```
onNavigationStateChanged() {
  // executed in useEffect of the root navigator 
  SENTRY_NATIVE.requestAnimationFrame((frameTimestamp) {
  	// this will actually be called on the next native frame (linked to main app thread)
  })
}
```

```
// NATIVE IMPL
SENTRY_NATIVE.requestAnimationFrame(callback) {
	// Similar to RNSentryOnDrawReporterView implementation
	// https://github.com/getsentry/sentry-react-native/blob/cc9666e872837955125bd0a80492e57625fc16a0/android/src/main/java/io/sentry/react/RNSentryOnDrawReporterManager.java#L121
	// https://github.com/getsentry/sentry-react-native/blob/2b83fc5d3de33c5eab1c8a08d762ac813e3a7c9b/ios/RNSentryOnDrawReporter.m#L44
	hookToNextFrame(() {
		callback(Date.now())
	})
}
```

The native implementation would the same method to hook into the render loop as RNSentryOnDrawReporterView but instead of component it would be native method of RNSentry.

I think we can still use the JS `requestAnimationFrame` to add support to `react-native-web` and maybe good fall back for `Expo Go`. 

It should work well with the current draft, replacing `requestAnimationFrame` with custom method.

Let me know what you think, if my though process make sense, or I've missed something.

samples/react-native/src/Screens/PerformanceScreen.tsx
```diff
@@ -62,6 +72,787 @@ const PerformanceScreen = (props: Props) => {
...
+  title="Redux Example"
+  onPress={() => {
+    props.navigation.navigate('Redux');
+  }}
+/>        <Button
+          title="Redux Example"
+          onPress={() => {
+            props.navigation.navigate('Redux');
+          }}
+        />
```
lucas-zimerman, line 855:
```suggestion
```

samples/react-native/src/Screens/PerformanceScreen.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
```
lucas-zimerman, line 1:
```suggestion
import React from 'react';
```

samples/react-native/src/Screens/PerformanceScreen.tsx
```diff
@@ -32,6 +32,16 @@ const PerformanceScreen = (props: Props) => {
...
+  const [isLoading, setIsLoading] = useState(true);
+  useEffect(() => {
+    // Simulate heavy computation for 5 seconds
+    const start = Date.now();
+    while (Date.now() - start < 5000) {
+      // Perform some meaningless computation to occupy the CPU
+      Math.sqrt(Math.random() * Math.random());
+    }
+    setIsLoading(false);
+  }, []);
```
lucas-zimerman, line 44:
```suggestion
```

lucas-zimerman:
iOS integration should also be done and tests are also completed.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -135,10 +135,16 @@ public class RNSentryModuleImpl {
     /** Max trace file size in bytes. */
     private long maxTraceFileSize = 5 * 1024 * 1024;
 
+
+
+  private final RNSentryTimeToDisplay timeToDisplay;
+
+
```
lucas-zimerman, line 142:
```suggestion
  private final RNSentryTimeToDisplay timeToDisplay;
```

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -1,5 +1,5 @@
 package io.sentry.react;
-
+import android.view.Choreographer;
```
lucas-zimerman, line 2:
```suggestion

import android.view.Choreographer;
```

android/src/main/java/io/sentry/react/RNSentryPackage.java
```diff
@@ -20,9 +20,9 @@ public class RNSentryPackage extends TurboReactPackage {
   @Nullable
   @Override
   public NativeModule getModule(String name, ReactApplicationContext reactContext) {
-    if (name.equals(RNSentryModuleImpl.NAME)) {
-      return new RNSentryModule(reactContext);
-    } else {
+      if (name.equals(RNSentryModuleImpl.NAME)) {
+        return new RNSentryModule(reactContext);
+      } else {
```
lucas-zimerman, line 25:
```suggestion
    if (name.equals(RNSentryModuleImpl.NAME)) {
      return new RNSentryModule(reactContext);
    } else {
```

android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
```diff
@@ -0,0 +1,36 @@
...
+import android.view.Choreographer;
+
+import androidx.annotation.NonNull;
+
+import org.jetbrains.annotations.NotNull;
+import io.sentry.SentryDate;
+import io.sentry.SentryDateProvider;
+import io.sentry.android.core.SentryAndroidDateProvider;
+
+
```
lucas-zimerman, line 18:
```suggestion
```

ios/RNSentry.mm
```diff
@@ -776,4 +780,10 @@ - (NSDictionary*) fetchNativeStackFramesBy: (NSArray<NSNumber*>*)instructionsAdd
 }
 #endif
 
+RCT_EXPORT_METHOD(getNewScreenTimeToDisplay:(RCTPromiseResolveBlock)resolve
+                  rejecter:(RCTPromiseRejectBlock)reject) {
+  [_timeToDisplay getTimeToDisplay:resolve];
+}
+
+
```
lucas-zimerman, line 788:
```suggestion
```

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -1,5 +1,6 @@
 package io.sentry.react;
 
+import android.view.Choreographer;
```
antonis, line 3: I think it is safe to remove this import since the class is only used inside the `RNSentryTimeToDisplay` implementation.
```suggestion
```

antonis: I've also noticed a few more imports unrelated with this PR that can be safely removed. I'm attaching a patch since it escapes the scope GH allows me to comment on. I can handle this on a separate PR if it is more proper.
[imports.patch](https://github.com/user-attachments/files/17113794/imports.patch)

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -135,10 +136,12 @@ public class RNSentryModuleImpl {
     /** Max trace file size in bytes. */
     private long maxTraceFileSize = 5 * 1024 * 1024;
 
+  private final RNSentryTimeToDisplay timeToDisplay;
```
antonis, line 139: `l`: Minor formating nitpick
```suggestion
     private final RNSentryTimeToDisplay timeToDisplay;
     
```

android/src/newarch/java/io/sentry/react/RNSentryModule.java
```diff
@@ -173,4 +173,9 @@ public String getCurrentReplayId() {
     public void crashedLastRun(Promise promise) {
         this.impl.crashedLastRun(promise);
     }
+
+      @Override
+     public void getNewScreenTimeToDisplay(Promise promise) {
+        this.impl.getNewScreenTimeToDisplay(promise);
+     }
```
antonis, line 180: `l`: Minor formating nitpick
```suggestion
    @Override
    public void getNewScreenTimeToDisplay(Promise promise) {
        this.impl.getNewScreenTimeToDisplay(promise);
    }
```

android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
```diff
@@ -0,0 +1,35 @@
+package io.sentry.react;
+
+import com.facebook.react.bridge.ReactContextBaseJavaModule;
+import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.bridge.ReactMethod;
+import com.facebook.react.bridge.Promise;
+import com.facebook.react.turbomodule.core.interfaces.TurboModule;
+
+import android.view.Choreographer;
```
antonis, line 9: `l`: We could remove the unused imports
```suggestion
import com.facebook.react.bridge.Promise;

import android.view.Choreographer;
```

ios/RNSentry.mm
```diff
@@ -1,5 +1,6 @@
 #import <dlfcn.h>
 #import "RNSentry.h"
+#import  "RNSentryTimeToDisplay.h"
```
antonis, line 3: `l`: Minor code formatting
```suggestion
#import "RNSentryTimeToDisplay.h"
```

ios/RNSentry.mm
```diff
@@ -776,4 +780,9 @@ - (NSDictionary*) fetchNativeStackFramesBy: (NSArray<NSNumber*>*)instructionsAdd
 }
 #endif
 
+RCT_EXPORT_METHOD(getNewScreenTimeToDisplay:(RCTPromiseResolveBlock)resolve
+                  rejecter:(RCTPromiseRejectBlock)reject) {
+  [_timeToDisplay getTimeToDisplay:resolve];
```
antonis, line 785: `l`: Minor code formatting
```suggestion
    [_timeToDisplay getTimeToDisplay:resolve];
```

ios/RNSentryTimeToDisplay.m
```diff
@@ -0,0 +1,43 @@
+#import "RNSentryTimeToDisplay.h"
+#import <QuartzCore/QuartzCore.h>
+#import <React/RCTLog.h>
+
+@implementation RNSentryTimeToDisplay
```
antonis, line 5: Q: I wonder if we could reuse the implementation of [`SentryDisplayLinkWrapper`](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/include/SentryDisplayLinkWrapper.m) on this class.

antonis (approved):
Awesome work @lucas-zimerman!
LGTM! I've left some suggestions/comments on the native side mainly related to formatting.
Thank you for also adding extensive test cases 🙇

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
lucas-zimerman: Thank you for checking this!
Would you mind making another PR for the other unused imports? (this way you also get another PR here hehe)

android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
```diff
@@ -0,0 +1,29 @@
...
+
+        // Invoke the callback after the frame is rendered
+        choreographer.postFrameCallback(frameTimeNanos -> {
+                final @NotNull SentryDateProvider dateProvider = new SentryAndroidDateProvider();
+
+                final SentryDate endDate = dateProvider.now();
+
+                promise.resolve(endDate.nanoTimestamp() / 1e9);
+            }
+        });
```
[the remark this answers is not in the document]
lucas-zimerman: ah no worries, thanks for the checking it :D

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
antonis: > Would you mind making another PR for the other unused imports? (this way you also get another PR here hehe)

@lucas-zimerman Closing the loop here I've opened a PR at https://github.com/getsentry/sentry-react-native/pull/4119
Feel free to merge on top of your PR or after your PR is merged.

lucas-zimerman:
@stefanosiano or @brustolin could you guys help to review the native code?

ios/RNSentryTimeToDisplay.m
brustolin: SentryDisplayLinkWrapper is not available for RN, we could do it, but I think this Is good enough.

antonis: Makes sense 👍 Thank you for your response 🙇

brustolin (approved):
The objc code looks good to me. I did not review the rest.

android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -135,10 +131,13 @@ public class RNSentryModuleImpl {
     /** Max trace file size in bytes. */
     private long maxTraceFileSize = 5 * 1024 * 1024;
 
+     private final RNSentryTimeToDisplay timeToDisplay;
```
krystofwoldrich, line 134:
```suggestion
    private final RNSentryTimeToDisplay timeToDisplay;
```

android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
```diff
@@ -0,0 +1,26 @@
...
+import io.sentry.android.core.SentryAndroidDateProvider;
+
+public class RNSentryTimeToDisplay {
+
+    public void GetTimeToDisplay(Promise promise) {
+        Choreographer choreographer = Choreographer.getInstance();
+
+        // Invoke the callback after the frame is rendered
+        choreographer.postFrameCallback(frameTimeNanos -> {
+                final @NotNull SentryDateProvider dateProvider = new SentryAndroidDateProvider();
```
krystofwoldrich, line 19: Can we reuse the `dateProvider`, maybe even share it with the main Impl.

So we avoid creating it on every call.

android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
```diff
@@ -0,0 +1,26 @@
...
+import android.view.Choreographer;
+
+import org.jetbrains.annotations.NotNull;
+import io.sentry.SentryDate;
+import io.sentry.SentryDateProvider;
+import io.sentry.android.core.SentryAndroidDateProvider;
+
+public class RNSentryTimeToDisplay {
+
+    public void GetTimeToDisplay(Promise promise) {
```
krystofwoldrich, line 14: Since we don't have any state, this could be a static function, getting the promise and dataProvider in arguments.

src/js/utils/sentryeventemitterfallback.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  function defaultFallbackEventEmitter(): void {
+    // Schedule the callback to be executed when all UI Frames have flushed.
+    requestAnimationFrame(() => {
+      if (NativeEmitterCalled) {
+        NativeEmitterCalled = false;
+        isListening = false;
+        return;
+      }
+      const timestampInSeconds = timeNowNanosecond();
```
krystofwoldrich, line 39: Let's use `timestampInSeconds` from `@sentry/utils`. It uses more precise and monotonic clock.

src/js/utils/sentryeventemitterfallback.ts
```diff
@@ -0,0 +1,103 @@
...
+ * Creates emitter that allows to listen to UI Frame events when ready.
+ */
+export function createSentryFallbackEventEmitter(): SentryEventEmitterFallback {
+  let NativeEmitterCalled: boolean = false;
+  let subscription: EmitterSubscription | undefined = undefined;
+  let isListening = false;
+
+  function defaultFallbackEventEmitter(): void {
+    // Schedule the callback to be executed when all UI Frames have flushed.
+    requestAnimationFrame(() => {
```
krystofwoldrich, line 33: We should include here link to the RN explanation and also mentioned that currently its not the measuring the UI frames but it's implemented as `setTimeout(() => {}, 0)`

src/js/utils/sentryeventemitterfallback.ts
```diff
@@ -0,0 +1,103 @@
...
+      const timestampInSeconds = timeNowNanosecond();
+      waitForNativeResponseOrFallback(timestampInSeconds, 'JavaScript');
+    });
+  }
+
+  function waitForNativeResponseOrFallback(fallbackSeconds: number, origin: string): void {
+    const maxRetries = 3;
+    let retries = 0;
+
+    const retryCheck = (): void => {
```
krystofwoldrich, line 48: Can we replace the retry with one 3s timeout (have I missed why we use the retry)?

If the new frame event is emitted correctly we can use `clearTimeout`.

src/js/tracing/reactnavigation.ts
```diff
@@ -101,7 +102,9 @@ export class ReactNavigationInstrumentation extends InternalRoutingInstrumentati
 
     if (this._options.enableTimeToInitialDisplay) {
       this._newScreenFrameEventEmitter = createSentryEventEmitter();
+      this._newFallbackEventEmitter = createSentryFallbackEventEmitter();
```
krystofwoldrich, line 105: I think `sentryEventEmitter` could be internally used in the `sentryFallbackEventEmitter`. To isolate the fallback logic and remove it from ReactNavigation.

src/js/utils/sentryeventemitterfallback.ts
```diff
@@ -0,0 +1,103 @@
...
+        if (isListening) {
+          NativeEmitterCalled = true;
+        }
+      });
+    },
+
+    startListenerAsync() {
+      isListening = true;
+
+      NATIVE.getNewScreenTimeToDisplay()
```
krystofwoldrich, line 85: I think the timeout should start here and if the none of the new display time to displays (not react nav, not native fallback, not JS fallback) was retried it should just log error., otherwise it's 3s + the time of NewScreeTimeToDisplay.

test/utils/sentryeventemitterfallback.test.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  it('should start listener and use fallback when native call returned undefined/null', async () => {
+    jest.useFakeTimers();
+    const fallbackTime = Date.now() / 1000;
+
+    const animation = NATIVE.getNewScreenTimeToDisplay as jest.Mock;
+    animation.mockReturnValue(Promise.resolve());
+
+    emitter.startListenerAsync();
+    await animation; // wait for the Native execution to be completed.
```
krystofwoldrich, line 67: The current `await animation` allows the next event loop, but doesn't wait for the execution to be completed. I would rename this to be more accurate, but the test works correct in both cases.

```suggestions
    (NATIVE.getNewScreenTimeToDisplay as jest.Mock).mockReturnValue(Promise.resolve());

    emitter.startListenerAsync();

    // Wait for the next event loop to allow startListenerAsync to call NATIVE.getNewScreenTimeToDisplay
    await Promise.resolve();
```

krystofwoldrich:
The PR is looking good, but I have some questions about the fallbacks timing and some small refactoring ideas.

test/utils/sentryeventemitterfallback.test.ts
lucas-zimerman: Thank you for the suggestion! I will apply it to the other tests too.

src/js/utils/sentryeventemitterfallback.ts
```diff
@@ -0,0 +1,103 @@
...
+            defaultFallbackEventEmitter();
+          }
+        })
+        .catch((reason: Error) => {
+          logger.error('Failed to recceive Native fallback timestamp.', reason);
+          defaultFallbackEventEmitter();
+        });
+    },
+
+    closeAll() {
```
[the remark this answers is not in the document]
lucas-zimerman: Should we also remove the `closeAll` on `sentryeventemitter`? It seems not used too.

src/js/utils/sentryeventemitterfallback.ts
lucas-zimerman: One retry with 3s also works too

src/js/tracing/reactnavigation.ts
lucas-zimerman: I think you want to move the fallback inside of sentryEventEmitter? if so, I moved it.

android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
lucas-zimerman: Also made `createEmitNewFrameEvent` to use the shared dateProvider

android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
```diff
@@ -0,0 +1,23 @@
...
+import android.view.Choreographer;
+
+import org.jetbrains.annotations.NotNull;
+import io.sentry.SentryDate;
+import io.sentry.SentryDateProvider;
+import io.sentry.android.core.SentryAndroidDateProvider;
+
+public class RNSentryTimeToDisplay {
+    public static void GetTimeToDisplay(Promise promise, SentryDateProvider dateProvider) {
+        Choreographer choreographer = Choreographer.getInstance();
```
stefanosiano, line 14: Shouldn't it be in a try catch block?
accessing Choreographer may result in an exception (we had some reports in the SDK console)

src/js/utils/sentryeventemitterfallback.ts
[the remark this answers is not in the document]
krystofwoldrich: Yes, we can clean that up, after the fallback. I agree.

src/js/tracing/reactnavigation.ts
krystofwoldrich: I was thinking about something like this: https://github.com/getsentry/sentry-react-native/compare/test/deadline...kw-fallback-emitter-suggestion

Using the `sentryEventEmitter` inside of `sentryFallbackEventEmitter` and than use the `sentryFallbackEventEmitter` in `reactnavigation`.

Let me know what do you think about the suggestion.

android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.jetbrains.annotations.NotNull;
+import io.sentry.SentryDate;
+import io.sentry.SentryDateProvider;
+import io.sentry.android.core.SentryAndroidDateProvider;
+
+public class RNSentryTimeToDisplay {
+    public static void GetTimeToDisplay(Promise promise, SentryDateProvider dateProvider) {
+        try {
+            Choreographer choreographer = Choreographer.getInstance();
```
krystofwoldrich, line 15: We should run the `Choreographer.getInstance()` on the Main Thread as this function returns instance for the current thread.

https://github.com/getsentry/sentry-java/blob/b11dc55cfc01379e481be60de78b512191128a25/sentry-android-core/src/main/java/io/sentry/android/core/internal/util/SentryFrameMetricsCollector.java#L121
https://developer.android.com/reference/android/view/Choreographer#getInstance()

krystofwoldrich:
Thank you @lucas-zimerman, it's looking good. 

I've noticed one last thing with the choreographer. See https://github.com/getsentry/sentry-react-native/pull/4042#discussion_r1800711808

krystofwoldrich:
I've changed the target to `v5` to avoid conflicts.

But we should include this also in version 6.

First we release it as beta on v5 and then also as a hotfix on `6.0.1`

krystofwoldrich (approved):
Thank you! @lucas-zimerman 

Large improvement for TTD.

krystofwoldrich:
@lucas-zimerman 

Can you port this also to v6 (main)?

lucas-zimerman:
Sure thing, I will port it to V6

PR #4057, opened 2024-08-29
misc: Move to monorepo structure and Yarn V3

krystofwoldrich:

- [x] Refactoring

## :scroll: Description

This PR change the repository structure to use monorepo setup.

The new structure:
- / -> monorepo root private package
- /samples -> sampel application, expo, rn...
- /packages -> RN SDK packages
- /dev-packages -> dev packages, ts-3.8 test runner, e2e tests components and runner
- /performance-tests -> applications used for measuring performance in CI

This PR updates the repository to use Yarn V3, as with Yarn V1 the sample applications would not run due to dependencies conflicts. And the original `nohoist` would not fix the issue. With `nmHoistingLimits: workspaces` the RN and Expo application are working correctly.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- [x] Test the publish of `@sentry/react-native` does introduce unexpected (no) changes compared to version `6.0.0-alpha.0`
- [x] Verify version gets correctly bumped in all dependencies

#skip-changelog 

This is a internal change the SDK package published should be unchanged.


packages/core/ios/RNSentry.h
krystofwoldrich, line 1: Should be reverted

packages/core/ios/RNSentry.mm
krystofwoldrich, line 1: Should be reverted

.github/workflows/buildandtest.yml
```diff
@@ -97,112 +77,99 @@ jobs:
...
       - name: Archive Expo Plugin
         uses: actions/upload-artifact@v4
         with:
           name: expo-plugin
-          path: plugin/build
+          path: packages/core/plugin/build
       - name: Pack
-        run: yarn pack
+        run: |
+          cp README.md packages/core
```
Lms24, line 107: l (no action required):
Are you keeping the readme of the SDK top level and you copy it to the package to get it into the tarball? No objections btw, just curious.

In the JS SDK monorepo we keep a Readme per package but I guess that's only gonna become necessary if you have more than one package.

.craft.yml
Lms24, line 1: No immediate action required since you only publish one package at the moment. Once you have more than one pkg, I'd recommend to create one NPM target per package and release in package dependency order.

Lms24 (approved):
Nice work! LGTM from monorepo and publishing perspective. Noted down some things to keep in mind when we add more packages but we can do this whenever it's necessary.

Btw: We added this [New SDK (package) Checklist](https://github.com/getsentry/sentry-javascript/blob/develop/docs/new-sdk-release-checklist.md) to our monorepo docs to keep in mind what to do in which order. Some things are very JS SDK-specific but the general oder of actions and things to keep in mind should apply to this monorepo as well.

PR #4073, opened 2024-09-06
misc(deprecations): Add Tracing deprecations and replacements

krystofwoldrich:

- [x] Enhancement

## :scroll: Description

This PR add deprecations and simple replacements which will be dropped in V6.

sample app, unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Changes
 
-- React Native Tracing Deprecations  ([#4073](https://github.com/getsentry/sentry-react-native/pull/4073))
+- React Native Tracing Deprecations ([#4073](https://github.com/getsentry/sentry-react-native/pull/4073))
   - `new ReactNativeTracing` to `reactNativeTracingIntegration()`
   - `new ReactNavigationInstrumentation` to `reactNativeTracingIntegration()`.
   - `new ReactNativeNavigationInstrumentation` to `reactNativeTracingIntegration()`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -644,7 +644,7 @@ This release contains upgrade of `sentry-android` dependency to major version 7.
 ## 5.16.0
 
 This release ships with a beta version of our new built-in Expo SDK 50 support,
-which replaces the `sentry-expo`package. To learn more,
+which replaces the deprecated `sentry-expo` package. To learn more,
 see [the Expo guide](https://docs.sentry.io/platforms/react-native/manual-setup/expo/).
 
 ### Features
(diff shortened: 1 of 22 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## Unreleased
+
+### Changes
+
+- React Native Tracing Deprecations  ([#4073](https://github.com/getsentry/sentry-react-native/pull/4073))
```
lucas-zimerman, line 7: NIT:
```suggestion
- React Native Tracing Deprecations ([#4073](https://github.com/getsentry/sentry-react-native/pull/4073))
```

samples/react-native/src/App.tsx
```diff
@@ -38,11 +38,10 @@ LogBox.ignoreAllLogs();
 
 const isMobileOs = Platform.OS === 'android' || Platform.OS === 'ios';
 
-const reactNavigationInstrumentation =
-  new Sentry.ReactNavigationInstrumentation({
-    routeChangeTimeoutMs: 500, // How long it will wait for the route change to complete. Default is 1000ms
-    enableTimeToInitialDisplay: isMobileOs,
-  });
+const reactNavigationInstrumentation = Sentry.reactNavigationIntegration({
+  routeChangeTimeoutMs: 500, // How long it will wait for the route change to complete. Default is 1000ms
```
lucas-zimerman, line 42: Q: why using another value other than the default?

src/js/options.ts
```diff
@@ -204,15 +204,52 @@ export interface BaseReactNativeOptions {
...
+
+  /**
+   * Track the app start time by adding measurements to the first route transaction. If there is no routing instrumentation
+   * an app start transaction will be started.
+   *
+   * Requires performance monitoring to be enabled.
+   *
+   * @default true
+   */
+  enableAppStartTracking?: boolean;
```
lucas-zimerman, line 221: Q: Why nullable if we already set a default value as true, will null behave the same as setting it to false?

Same question applies to the below nullable boolean with null support.

test/client.test.ts
```diff
@@ -688,6 +688,19 @@ describe('Tests ReactNativeClient', () => {
...
+    test('register user interactions tracing root option', () => {
+      const client = new ReactNativeClient(
+        mockedOptions({
+          enableUserInteractionTracing: true,
+          dsn: EXAMPLE_DSN,
+          integrations: [new ReactNativeTracing({})],
+        }),
+      );
+      client.setupIntegrations();
+
```
lucas-zimerman, line 700:
```suggestion
```

src/js/options.ts
krystofwoldrich: To allow users to not specify the option in the init.

samples/react-native/src/App.tsx
krystofwoldrich: I have not updated the value here, just format. But the shorter value is here just for convenience when debugging.

PR #4090, opened 2024-09-13
misc(xcode): Use foreground upload by default

krystofwoldrich:

- [x] Enhancement

## :scroll: Description

Change default to blocking foreground upload of debug files and source maps.

This helps with bug reports, as we can see the whole log by default. 

Also this change mitigates issue with sentry-cli and curl library which in newer versions fail in background.

Related issues
- 

ci, sample app

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


antonis (approved):
LGTM 🎉

PR #4094, opened 2024-09-16
Add flag to disable native profilers

### LINKED ISSUE #4085: Add option to `hermesProfilingIntegration` to only profile Hermes JS
### Description

When `profilesSampleRate` set in `Sentry.init` of the React Native SDK both Hermes JS and Android JVM or iOS native profiler are recording.

This issue is proposing an option to only enable Hermes JS profiling if user doesn't want to profile the native threads.

```js
Sentry.init({
  profilesSampleRate: 1.0,
  integration: [
    Sentry.hermesProfilingIntegration({
      platformProfilers: false
    }),
  ],
})
```

romtsn:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

* Adds a flag `platformProviders` for the hermes profling integration to disable profiling of native threads, which only leaves JS profiling enabled and avoids potential crashes that the native SDKs profiling can introduce. Defaults to `true`

I ran the samples locally and verified that enabling/disabling the flag adds/removes native threads from the profiles, see:

**Flag enabled:**

**Flag disabled:**
Google Chrome 2024-09-16 17 02 24

Closes #4085 

**See links for Android:**

Flag enabled: https://sentry-sdks.sentry.io/profiling/profile/sentry-react-native/f95eae4eda2b413b9802c2bee89582da/flamegraph/
Flag disabled: https://sentry-sdks.sentry.io/profiling/profile/sentry-react-native/0e07464e9c784d408525cd118cdade00/flamegraph/

**See links for iOS:**

Flag enabled: https://sentry-sdks.sentry.io/profiling/profile/sentry-react-native/8cb25ce8b911427485d831fe01109f38/flamegraph/
Flag disabled: https://sentry-sdks.sentry.io/profiling/profile/sentry-react-native/7bbaf5bdff8b46ca85063a98f38200c6/flamegraph/

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,18 @@
 
 - Add an option to disable native (iOS and Android) profiling for the `HermesProfiling` integration ([#4094](https://github.com/getsentry/sentry-react-native/pull/4094))
 
+  To disable native profilers add the `hermesProfilingIntegration`.
+
+  ```js
+  import * as Sentry from '@sentry/react-native';
+
+  Sentry.init({
+    integrations: [
+      Sentry.hermesProfilingIntegration({ platformProfilers: false }),
+    ],
+  });
+  ```
+
 ## 5.32.0
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add an option to disable native (iOS and Android) profiling for the `HermesProfiling` integration ([#4094](https://github.com/getsentry/sentry-react-native/pull/4094))
+
 ## 5.32.0
 
 ### Features
--- a/test/profiling/integration.test.ts
+++ b/test/profiling/integration.test.ts
@@ -9,7 +9,7 @@ import type { Envelope, Event, Profile, ThreadCpuProfile, Transaction, Transport
 import * as Sentry from '../../src/js';
 import type { NativeDeviceContextsResponse } from '../../src/js/NativeRNSentry';
 import { getDebugMetadata } from '../../src/js/profiling/debugid';
-import { hermesProfilingIntegration } from '../../src/js/profiling/integration';
+import { hermesProfilingIntegration, HermesProfilingOptions } from '../../src/js/profiling/integration';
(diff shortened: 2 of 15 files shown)


test/profiling/integration.test.ts
```diff
@@ -351,6 +351,26 @@ describe('profiling integration', () => {
       jest.runAllTimers();
     });
   });
+
+  test('platformProviders flag passed down to native', () => {
+    mock = initTestClient({ withProfiling: false });
```
antonis, line 356: I've noticed that this test is failing. My understanding is that this is only a test issue and does not have impact on the functional code. I experimented with changing the `withProfiling` parameter to `true` and added a new parameter in the `initTestClient` to avoid using the default value for the `HermesProfilingOptions`.
I'm attaching a patch with the above.

[integration.test.patch](https://github.com/user-attachments/files/17024543/integration.test.patch)

Feel free to ignore if this is in the wrong direction 😅

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add an option to disable native (iOS and Android) profiling for the `HermesProfiling` integration ([#4094](https://github.com/getsentry/sentry-react-native/pull/4094))
+
```
antonis, line 8: This is a suggestion and not a request for a change. Being new to the project I wonder if we could also add a usage example in the changelog.

test/profiling/integration.test.ts
romtsn: hi, and welcome to Sentry 🙂 interestingly the test works when running it in isolation, so I assume there's some shared state that breaks it when running the whole suite. I applied your patch in https://github.com/getsentry/sentry-react-native/pull/4094/commits/841d5b70f133984afd61fb71188c5aeeaa7ab32b, thanks for this!

CHANGELOG.md
romtsn: certainly a great suggestion, I'll add it!

antonis (approved):
Awesome work @romtsn 🎉 
The code LGTM! Thank you for including my suggestions 🙇

Issue #2146, opened 2022-03-18
(Android) breadcrumb removed if set on error's context/scope.


lucas-zimerman:
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)
Which SDK and version?
e.g: Sentry React Native 3.3.3

### Steps to Reproduce

Sample code:
```js
import { StatusBar } from 'expo-status-bar';
import React from 'react';
import { StyleSheet, Text, View } from 'react-native';
import * as Sentry from '@sentry/react-native';
Sentry.init({
  dsn:   'https://d870ad989e7046a8b9715a57f59b23b5@o447951.ingest.sentry.io/5428561',
  debug: true,
});

Sentry.captureException(new Error(`${Date.now()}: a test error occurred`),
context => context.addBreadcrumb({ message: 'error with breadcrumbs' }));

export default function App() {
  return (
    <View style={styles.container}>
      <Text>Open up App.js to start working on your app!</Text>
      <StatusBar style="auto" />
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    backgroundColor: '#fff',
    alignItems: 'center',
    justifyContent: 'center',
  },
});

```

### Expected Result

The exception captured with the breadcrumb `error with breadcrumbs`

### Actual Result

Breadcrumb got removed
https://sentry.io/organizations/sentry-sdks/issues/3114951206/?project=5428561&query=is%3Aunresolved&statsPeriod=1h

https://github.com/getsentry/sentry-react-native/blob/a085ec4ef6e9506cca568d8841e1e1bc8dfb63a2/src/js/wrapper.ts#L108-L110
This code doesnt differentiate from synced breadcrumbs and local breadcrumbs, resulting on local breadcrumbs getting removed when setting an isolated scope/context.

marandaneto:
I believe the right solution here would be to align the way how we capture events, which is what iOS does, on RN, we call the native bridge to get all device context and assemble the event on RN with already all the context, so this won't need to be done on Android, after the envelope is written to the disk, and because of that, breadcrumbs won't be duplicated.
This is a side-effect of this hacky solution of enriching the events on the Android side.
We can patch the current version to fix it, but ideally, we'd do it properly.

lucas-zimerman:
I agree with this solution

marandaneto:
@krystofwoldrich have you tested that this is fixed after https://github.com/getsentry/sentry-react-native/pull/3170?

[...]

PR #4124, opened 2024-09-27
fix(core): Breadcrumbs added on forked context are now captured

antonis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Uses `breadcrumb.origin` field to prevent exception capture context from being overwritten by native scope sync.

* [Adds Capture exception with breadcrumb in the sample](https://github.com/getsentry/sentry-react-native/pull/4124/commits/37553f296a99d532ff69a63c1953247c67ec7d1e)
* [Sets `react-native` as breadcrumb origin when not set](https://github.com/getsentry/sentry-react-native/pull/4124/commits/24a527be83850ad4a63b47920b025fc1f7fe7003)
* [Filters out breadcrumbs with react-native origin from the native layer](https://github.com/getsentry/sentry-react-native/pull/4124/commits/fabf0526a8b9d70a75c703b2d29accbc09e4fe93)
* [Merges event breadcrumbs with native context](https://github.com/getsentry/sentry-react-native/pull/4124/commits/d0b163342380980a878c340f8ad55df08e86d1a5)
* Refactors and tests RNSentryModuleImpl.fetchNativeDeviceContexts (reviewed separately in https://github.com/getsentry/sentry-react-native/pull/4253)

Depends on:
- https://github.com/getsentry/sentry-java/pull/3727
- https://github.com/getsentry/sentry-cocoa/pull/4358
- https://github.com/getsentry/sentry-docs/pull/11449

Fixes https://github.com/getsentry/sentry-react-native/issues/2146
The discussion in https://github.com/getsentry/sentry-capacitor/pull/629#discussion_r1617321563 adds some context on the need of a new field to be used internally to solve a technical problem.

Manual testing with the `Capture exception with breadcrumb` button, CI

|Before|After iOS|After Android|
|---|---|---|
|main|ios|android|

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


lucas-zimerman:
> Hey @lucas-zimerman, @krystofwoldrich 👋
> 
> > Fixes #2146
> > The discussion in [getsentry/sentry-capacitor#629 (comment)](https://github.com/getsentry/sentry-capacitor/pull/629#discussion_r1617321563) adds some context on the need of a new field to be used internally to solve a technical problem.
> 
> Since some time has passed since the creation of the issue and the discussion above I'd like to verify that the direction I've chosen with this implementation is valid:
> 
>     * [Set react-native as breadcrumb origin](https://github.com/getsentry/sentry-react-native/pull/4124/commits/24a527be83850ad4a63b47920b025fc1f7fe7003)
> 
>     * [Filters out breadcrumbs with `react-native` origin from the native layer](https://github.com/getsentry/sentry-react-native/pull/4124/commits/fabf0526a8b9d70a75c703b2d29accbc09e4fe93)
> 
>     * [Merges event breadcrumbs with native context](https://github.com/getsentry/sentry-react-native/pull/4124/commits/d0b163342380980a878c340f8ad55df08e86d1a5)
> 
> 
> Any feedback is welcome 🙏

The approach looks good indeed, thank you!
One thing to note is we should respect `maxBreadcrumbs` and limit the merged breadcrumbs to the defined value.

krystofwoldrich:
Thank you @antonis for checking this.

> [Set react-native as breadcrumb origin](https://github.com/getsentry/sentry-react-native/pull/4124/commits/24a527be83850ad4a63b47920b025fc1f7fe7003)
[Filters out breadcrumbs with react-native origin from the native layer](https://github.com/getsentry/sentry-react-native/pull/4124/commits/fabf0526a8b9d70a75c703b2d29accbc09e4fe93)
[Merges event breadcrumbs with native context](https://github.com/getsentry/sentry-react-native/pull/4124/commits/d0b163342380980a878c340f8ad55df08e86d1a5)

Yes, all tree steps look good.

+ What @lucas-zimerman mentioned.

packages/core/ios/RNSentry.mm
```diff
@@ -382,6 +382,17 @@ - (NSDictionary*) fetchNativeStackFramesBy: (NSArray<NSNumber*>*)instructionsAdd
 
     [serializedScope setValue:contexts forKey:@"contexts"];
     [serializedScope removeObjectForKey:@"context"];
+        
+    // Remove react-native breadcrumbs
+    NSMutableArray<NSDictionary<NSString *, id> *> *breadcrumbs = [serializedScope[@"breadcrumbs"] mutableCopy];
+    for (NSInteger i = breadcrumbs.count - 1; i >= 0; i--) {
```
krystofwoldrich, line 388: This is unsafe as we are removing the items based on index while using the index to iterate over the array.

samples/react-native/src/Screens/ErrorsScreen.tsx
```diff
@@ -88,6 +88,13 @@ const ErrorsScreen = (_props: Props) => {
             Sentry.captureException(error);
           }}
         />
+        <Button
+          title="Capture exception with breadcrumb"
+          onPress={() => {
+            Sentry.captureException(new Error('Captured exception with breadcrumb'),
+              context => context.addBreadcrumb({ message: 'error with breadcrumb' }));
```
krystofwoldrich, line 95: This is a good way to try the breadcrumbs merging. 👍

packages/core/ios/RNSentry.mm
krystofwoldrich: True, the backward loop should be ok. Thank you for changing it for the predicate.

packages/core/src/js/integrations/devicecontext.ts
```diff
@@ -83,7 +84,8 @@ async function processEvent(event: Event): Promise<Event> {
     ? native['breadcrumbs'].map(breadcrumbFromObject)
     : undefined;
   if (nativeBreadcrumbs) {
-    event.breadcrumbs = nativeBreadcrumbs;
+    const maxBreadcrumbs = getClient()?.getOptions().maxBreadcrumbs ?? 100; // Default is 100.
+    event.breadcrumbs = nativeBreadcrumbs.concat(event.breadcrumbs || []).slice(0, maxBreadcrumbs);
```
krystofwoldrich, line 88: h: We have to ensure we keep the last `maxBreadcrumbs` (the last 100). Please also add a test that capture this edge case.

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -905,6 +907,17 @@ public void fetchNativeDeviceContexts(Promise promise) {
...
+    if (currentScope != null) {
+      // Remove react-native breadcrumbs
+      Iterator<Breadcrumb> breadcrumbsIterator = currentScope.getBreadcrumbs().iterator();
+      while (breadcrumbsIterator.hasNext()) {
+        Breadcrumb breadcrumb = breadcrumbsIterator.next();
+        if ("react-native".equals(breadcrumb.getOrigin())) {
+          breadcrumbsIterator.remove();
+        }
+      }
+    }
```
krystofwoldrich, line 919: m: I'm not sure how testable `fetchNativeDeviceContexts` is at the moment, but can we add tests for this method. If it requires refactor first lets do it in a follow up PR.

packages/core/src/js/integrations/devicecontext.ts
antonis: Good point @krystofwoldrich 👍 
I've updated the logic and tests with c4246d789b690da34fb47056f38c63e25064a9ee

packages/core/src/js/integrations/devicecontext.ts
```diff
@@ -83,7 +84,11 @@ async function processEvent(event: Event): Promise<Event> {
     ? native['breadcrumbs'].map(breadcrumbFromObject)
     : undefined;
   if (nativeBreadcrumbs) {
-    event.breadcrumbs = nativeBreadcrumbs;
+    const maxBreadcrumbs = getClient()?.getOptions().maxBreadcrumbs ?? 100; // Default is 100.
```
lucas-zimerman, line 87: @krystofwoldrich Do you think it's worth caching this value? It is quite unlikely that this option will change after the SDK initializes.

packages/core/src/js/integrations/devicecontext.ts
```diff
@@ -83,7 +84,11 @@ async function processEvent(event: Event): Promise<Event> {
...
+    event.breadcrumbs = nativeBreadcrumbs
```
lucas-zimerman, line 88: thats a nice clean solution

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
antonis: Good idea @krystofwoldrich 👍 
I did a small refactoring and added tests with https://github.com/getsentry/sentry-react-native/pull/4253

packages/core/src/js/integrations/devicecontext.ts
krystofwoldrich: I think it's fine without, it's just reading value from object.

But what we should do it read the client from the function args. To avoid the lookup of the global client.

We currently use only the first arg of the function.

https://github.com/getsentry/sentry-javascript/blob/79818ffc304041238b948c20db2aa9a25458e798/packages/types/src/integration.ts#L53 

```js
processEvent?(event: Event, hint: EventHint, client: Client): Event | null | PromiseLike<Event | null>;
```

krystofwoldrich:
@antonis Thank you for the updates. It looks great, I have one last small comment regarding @lucas-zimerman caching question. Then it's good to 🚀

krystofwoldrich (approved):
LG 🚀 Thank you!

krystofwoldrich:
This PR doesn't trigger our dangerous file warning, but I thought anyway we might want to release this as a shot beta before latest release.

lucas-zimerman (approved):
Looks good too, yeah, it would be great if this PR goes first on a beta version.

antonis:
>  I thought anyway we might want to release this as a shot beta before latest release.

This sounds like a good idea ➕

PR #4160, opened 2024-10-10
Fix: Add check version to new arch on TimetoTisplay

lucas-zimerman:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

As shown on the [Docs](https://docs.sentry.io/platforms/react-native/tracing/instrumentation/time-to-display/#troubleshooting) and also in code, We currently dont support the new React Native Architecture for Time to Display.

The idea is to show the following warning to the users:
`TimeToInitialDisplay and TimeToFullDisplay are not supported on the web, Expo Go and New Architecture. Run native build or report an issue at https://github.com/getsentry/sentry-react-native` 
But this wasn't being shown on the new architecture, so I added a new condition for checking if we are running on the new architecture in order to show the warning.

This gives the idea that our integration is broken, this PR will correctly warn users that it is curently not supported on the new architecture of React Native.

### Context

isTurboModuleEnabled is just  `  return RN_GLOBAL_OBJ.__turboModuleProxy != null;`
`__turboModuleProxy` was introduced on the new architecture, so it is a fast way to detect the new architecture with no extra code.
 
It is another case found on the following issue: #3934.

I added a new page to the sample app where we can test the normal time to display VS manual one:

image

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


antonis (approved):
LGTM 🚀 
I verified that the warning is displayed only with the manual option in the new architecture.
Thank you for adding tests and extending the sample 🙇

samples/react-native/src/Screens/HeavyNatigationScreen.tsx
```diff
@@ -0,0 +1,97 @@
...
+    justifyContent: 'space-between',
+  },
+  statisticTitle: {
+    fontSize: 16,
+    fontWeight: '500',
+  },
+  statisticCount: {
+    fontSize: 16,
+    fontWeight: '700',
+  },
```
krystofwoldrich, line 96: Let's remove the unused styles.

samples/react-native/src/Screens/HeavyNatigationScreen.tsx
```diff
@@ -0,0 +1,97 @@
...
+const buttonTitles = Array.from(
+  { length: 500 },
+  (_, index) => `Sample button ${index + 1}`,
+);
+
+/**
+ * this page takes around 300ms to initially display, we navigate to another page in 100ms.
+ * The time to initial display will never be finished on this page.
+ */
+const TrackerScreen = (props: Props) => {
```
krystofwoldrich, line 24:
```suggestion
const HeavyNavigationScreen = (props: Props) => {
```

samples/react-native/src/Screens/HeavyNatigationScreen.tsx
```diff
@@ -0,0 +1,97 @@
+import * as React from 'react';
```
krystofwoldrich, line 1: Small typo in the file name `Natigation -> Navigation`

samples/react-native/src/Screens/PerformanceScreen.tsx
```diff
@@ -62,6 +65,24 @@ const PerformanceScreen = (props: Props) => {
             props.navigation.navigate('Redux');
           }}
         />
+        <Spacer />
+        <Text>
+          Heavy Page Navigation that navigates back automatically, right after
+          finishing the display.
```
krystofwoldrich, line 71: `right after finishing the display` -> I thought it's after timeout before recording TTID.

test/tracing/timetodisplay.test.tsx
```diff
@@ -373,6 +379,67 @@ describe('TimeToDisplay', () => {
...
+        const testSpan = startTimeToInitialDisplaySpan();
+        TestRenderer.create(<TimeToInitialDisplay record={true} />);
+
+        emitNativeInitialDisplayEvent();
+
+        activeSpan?.end();
+
+        return [testSpan, activeSpan];
+      },
+    );
```
krystofwoldrich, line 402: Since we are only testing the warning, lest only render the component, the spans are not relevant and they are tested in the other tests in the file.

```suggestion
TestRenderer.create(<TimeToInitialDisplay record={true} />);
```

krystofwoldrich (changes requested):
Thank you for the tests and the test screen in the sample this will be super useful for development and debugging. 

I just have a few comment on the unit tests and some nits on the sample.

lucas-zimerman:
Added an extra spam `before effect` to help understanding the span durations.
example:
Without navigating back while the page is being rendered
image

With navigation (automatic time to display times out)
image

samples/react-native/src/Screens/PerformanceScreen.tsx
lucas-zimerman: The naming is indeed wrong, we return to the previous page before TTID is finished, I changed the text to better clarify it.

krystofwoldrich:
What does the `before effect` span should measure, because right now it measures the time from `the JS bundle started executing` to the component render. 

I think it should be only from the component started mounting to the effect, which is already measured by the profiler.

samples/react-native/src/Screens/HeavyNavigationScreen.tsx
```diff
@@ -0,0 +1,74 @@
...
+      manualTrack: boolean;
+    };
+  };
+}
+const buttonTitles = Array.from(
+  { length: 500 },
+  (_, index) => `Sample button ${index + 1}`,
+);
+
+const beforeEffectSpanStart = Date.now();
```
krystofwoldrich, line 20: This will be only set once when the JS bundle start executing. I think this should be in the component function body, but then its duplicate of the mount span.

lucas-zimerman: To not extend the PR review for too long I have opted to remove the `before effect` span

krystofwoldrich (approved):
Thank you! Looks great!

Issue #4165, opened 2024-10-11
com.facebook.react.bridge.ReadableNativeArray cannot be cast to com.facebook.react.bridge.ReadableNativeMap


ziga-hvalec:
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [ ]  iOS
- [x]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 0.0.0

**`react-native` version:** 0.73.8

**Are you using Expo?**
- [ ]  Yes
- [x]  No

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)
- [ ]  on-premise

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
      dsn: Config.SENTRY_DSN,
      release: `${Platform.OS}@${Config.VERSION_NUMBER}+${Config.BUILD_VERSION}`,
      dist: Config.BUILD_VERSION,
      environment: Config.NAME,
      debug: Config.NAME === 'development',
      appHangTimeoutInterval: 10000,
      enableWatchdogTerminationTracking: false,
      ignoreErrors: IGNORE_ERRORS_LIST,
    });
```

---
**I have the following issue:**

When passing an array to setExtra, setExtras or setContext(possibly others), I get the above error
com.facebook.react.bridge.ReadableNativeArray cannot be cast to com.facebook.react.bridge.ReadableNativeMap
But i only get this on Android. iOS works correctly and also shows data in sentry. This is also only on release build. On debug build both iOS and Android work correctly and i can see context or extra correctly in sentry.

**Steps to reproduce:**
- pass array to setExtra or setContext
- observer crash

**Actual result:**

App crashes with above error

**Expected result:**

app not crash :D

PR #4168, opened 2024-10-14
fix: setContext correctly processes non-plain JS object

krystofwoldrich:

- [x] Bugfix
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

This PR adds more check and conversions for `setExtra` and `setContext` to allow sync from JS to Platform runtime without crashes.

fixes: https://github.com/getsentry/sentry-react-native/issues/4165

unit tests, sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


src/js/wrapper.ts
```diff
@@ -395,10 +397,25 @@ export const NATIVE: SentryNativeWrapper = {
...
+    if (typeof extra === 'undefined') {
+      return RNSentry.setExtra(key, 'undefined');
+    }
+
+    let stringifiedExtra: string | undefined;
+    try {
+      const normalizedExtra = normalize(extra);
+      stringifiedExtra = JSON.stringify(normalizedExtra);
+    } catch (e) {
+      logger.error('Extra not passed to native SDK, because it contains non-stringifiable values', e);
```
antonis, line 412: l: We could add the `key` to the error for easier debugging.
```suggestion
      logger.error('Extra for key ${key} not passed to native SDK, because it contains non-stringifiable values', e);
```

antonis, line 412: Q: Should we also `return` here or let it fall-through and set `**non-stringifiable**` ?

test/wrapper.test.ts
```diff
@@ -623,4 +623,120 @@ describe('Tests Native Wrapper', () => {
...
+
+    test('converts boolean primitive to object with "value" key before passing to native method', () => {
+      NATIVE.setContext('key', true as unknown as object);
+      expect(RNSentry.setContext).toHaveBeenCalledWith('key', { value: true });
+      expect(RNSentry.setContext).toHaveBeenCalledOnce();
+    });
+
+    test('handles null value by passing null to native method', () => {
+      NATIVE.setContext('key', null);
+      expect(RNSentry.setContext).toHaveBeenCalledWith('key', null);
```
antonis, line 726: l: we could also check the number of calls for consistency with the other tests.
```suggestion
      expect(RNSentry.setContext).toHaveBeenCalledWith('key', null);
      expect(RNSentry.setContext).toHaveBeenCalledOnce();
```

test/wrapper.test.ts
```diff
@@ -623,4 +623,120 @@ describe('Tests Native Wrapper', () => {
...
+      NATIVE.setContext('key', undefined as unknown as object);
+      expect(RNSentry.setContext).toHaveBeenCalledWith('key', { value: undefined });
+      expect(RNSentry.setContext).toHaveBeenCalledOnce();
+    });
+
+    test('handles non-serializable value by converting to normalized object', () => {
+      const circular: { self?: unknown } = {};
+      circular.self = circular;
+      NATIVE.setContext('key', circular);
+      expect(RNSentry.setContext).toHaveBeenCalledWith('key', { self: '[Circular ~]' });
```
antonis, line 739: l: as above
```suggestion
      expect(RNSentry.setContext).toHaveBeenCalledWith('key', { self: '[Circular ~]' });
      expect(RNSentry.setContext).toHaveBeenCalledOnce();
```

src/js/wrapper.ts
```diff
@@ -439,15 +456,31 @@ export const NATIVE: SentryNativeWrapper = {
...
-    RNSentry.setContext(key, context !== null ? normalize(context) : null);
+    if (context === null) {
+      return RNSentry.setContext(key, null);
+    }
+
+    let normalizedContext: Record<string, unknown> | undefined;
+    try {
+      normalizedContext = convertToNormalizedObject(context);
+    } catch (e) {
+      logger.error('Context not passed to native SDK, because it contains non-serializable values', e);
```
antonis, line 476: l: We could add the key to the error for easier debugging.
```suggestion
      logger.error('Context for key ${key} not passed to native SDK, because it contains non-serializable values', e);
```

src/js/wrapper.ts
```diff
@@ -439,15 +456,31 @@ export const NATIVE: SentryNativeWrapper = {
...
+    }
+
+    if (normalizedContext) {
+      return RNSentry.setContext(key, normalizedContext);
+    } else {
+      return RNSentry.setContext(key, { error: '**non-serializable**' });
```
antonis, line 482: Q: Wdyt of omitting the `return` statements in the `void` function and only only using it for early return above?

antonis (approved):
LGTM 🎉 
Added some NP suggestions.

src/js/wrapper.ts
```diff
@@ -439,15 +456,31 @@ export const NATIVE: SentryNativeWrapper = {
    * @param key string
    * @param context key-value map
    */
-  setContext(key: string, context: { [key: string]: unknown } | null): void {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  setContext(key: string, context: { [key: string]: any } | null): void {
```
antonis, line 460: l: The JSDoc comment above mentions `Not implemented in Android yet.`. Since `RNSentryModuleImpl` implements the `setContext` method I think this is not accurate.

src/js/wrapper.ts
krystofwoldrich: I let it fall thru to `**non-stringifiable**` so the extra doesn't silently disappear from the native events.

src/js/wrapper.ts
krystofwoldrich: Yeah, thanks. That's not valid for a while.

lucas-zimerman (approved):
Looking good, LGTM!

Issue #3706, opened 2024-03-21
Enable Spotlight in Android and iOS SDKs 


krystofwoldrich:
### Description

At the moment Spotlight is only enabled in the JS layer of React Native. Since Android and iOS also support sending events to Spotlight we can enable the native integrations from React Native.

When Spotlight is enabled in RN, it should be automatically enabled also on Android and iOS.

PR #4211, opened 2024-10-28
Enable Spotlight in Android and iOS SDKs

antonis:

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Passes `spotlight` initialisation option to the native Android and iOS SDKs. When the a `url` is not passed the RN JS `defaultSidecarUrl` is used.

When spotlight is enabled on both JS an platform SDKs, JS event will be send twice. The duplication only affects the `Developer Info` tab and the rest of the UI remains uncluttered.

|Tab| RN | RN + Cocoa | RN + Java |
|---|---|---|---|
|SDKs|js1|sp1|a1|
|Errors|js2|sp2|a2|
|Developer Info|js3|sp3|a3|

Fixes https://github.com/getsentry/sentry-react-native/issues/3706

Manual testing, CI

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


lucas-zimerman (approved):
PR Looks good to me, Thank you for the contribution!

packages/core/RNSentryCocoaTester/RNSentryCocoaTesterTests/RNSentryTests.mm
```diff
@@ -122,6 +122,36 @@ - (void)testCreateOptionsWithDictionaryAutoPerformanceTracingDisabled
     XCTAssertEqual(actualOptions.enableAutoPerformanceTracing, false, @"Did not disable Auto Performance Tracing");
 }
 
+- (void)testCreateOptionsWithDictionarySpotlightEnabled
```
krystofwoldrich, line 125: The same as in Android, let's add test with a custom url.

krystofwoldrich:
RN Default URL is based on the Metro dev server host, this is slightly different from the platform SDK, which use localhost as default.

Depending on the effort it would be good to use the same default. 
 
https://github.com/getsentry/sentry-react-native/blob/ddcabffc8bd015392f28deb4f685ed358889c77e/packages/core/src/js/integrations/spotlight.ts#L86

krystofwoldrich:
When spotlight is enabled on both JS an platform SDKs, JS event will be send twice. This will be super useful for debugging the SDK, but for Spotlight customers this should not result in duplicates in the Spotlight UI (only in Developer Tab).

Maybe you already check this. But let's confirm that it works this way.

packages/core/RNSentryCocoaTester/RNSentryCocoaTesterTests/RNSentryTests.mm
antonis: Good catch @krystofwoldrich 👍 
I now realised that [the value may be a the url string](https://docs.sentry.io/platforms/react-native/migration/v5-to-v6/#enablespotlight-and-spotlightsidecarurl-replaced-by-spotlight). I've updated the cocoa code with cc5383f61b71bb09f7797f666e2c58e8e0009273 and will iterate with the rest of your feedback.

packages/core/RNSentryAndroidTester/app/src/test/java/io/sentry/react/RNSentryModuleImplTest.kt
```diff
@@ -93,4 +94,11 @@ class RNSentryModuleImplTest {
         val capturedMap = writableMapCaptor.value
         assertEquals(false, capturedMap.getBoolean("has_fetched"))
     }
+
```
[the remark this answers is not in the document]
antonis: Fixed with 563588582cbc6bea454b97428a9c342fb06dfaa9

antonis:

> RN Default URL is based on the Metro dev server host, this is slightly different from the platform SDK, which use localhost as default.

[iOS uses `http://localhost:8969/stream`](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/SentryOptions.m#L144) and [Android may use `http://10.0.2.2:8969/stream` as default](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SpotlightIntegration.java#L87).

> Depending on the effort it would be good to use the same default.

If I understand correctly, we want to use the same default values on the RN side (or vice versa)?

> When spotlight is enabled on both JS an platform SDKs, JS event will be send twice. This will be super useful for debugging the SDK, but for Spotlight customers this should not result in duplicates in the Spotlight UI (only in Developer Tab).

I will check this 🙏

krystofwoldrich:
> >RN Default URL is based on the Metro dev server host, this is slightly different from the platform SDK, which use localhost as default.
>
>[iOS uses http://localhost:8969/stream](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/SentryOptions.m?rgh-link-date=2024-10-29T16%3A43%3A29Z#L144) and [Android may use http://10.0.2.2:8969/stream as default](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SpotlightIntegration.java?rgh-link-date=2024-10-29T16%3A43%3A29Z#L87).

Those are the default of the native SDKs. I was not exact, `the same default` meant the default set in RN. 

In RN JS we take the host of the Metro dev server as host of Spotlight (the assumption is that the dev server and spotlight run on the same computer). We should pass the RN JS default to the platform layers when Spotlight is enabled.

antonis:
> When spotlight is enabled on both JS an platform SDKs, JS event will be send twice. This will be super useful for debugging the SDK, but for Spotlight customers this should not result in duplicates in the Spotlight UI (only in Developer Tab).

> Maybe you already check this. But let's confirm that it works this way.

Rechecking this I think the behaviour is the expected. The errors are not duplicated and only the events on the `Developer Info` screen are recorded twice.

|Tab| RN | RN + Cocoa |
|---|---|---|
|SDKs|js1|sp1|
|Errors|js2|sp2|
|Developer Info|js3|sp3|

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 - Enhanced accuracy of time-to-display spans. ([#4189](https://github.com/getsentry/sentry-react-native/pull/4189))
 
+### Changes
```
krystofwoldrich, line 15: l: I think Features is a better fit. We use Changes mainly in major version for breaking changes and others.

packages/core/ios/RNSentry.mm
```diff
@@ -161,6 +161,22 @@ - (SentryOptions *_Nullable)createOptionsWithDictionary:(NSDictionary *_Nonnull)
             sentryOptions.integrations = integrations;
         }
     }
+    
+    // Set spotlight option
+    if ([mutableOptions valueForKey:@"spotlight"] != nil) {
+        id spotlightValue = [mutableOptions valueForKey:@"spotlight"];
+        if ([spotlightValue isKindOfClass:[NSString class]]) {
+            NSLog(@"The value is a string: %@", spotlightValue);
```
lucas-zimerman, line 169: Suggestion: Let's specify that we are using this string for Spotlight, so users don't get confused when reading the logs.

```suggestion
            NSLog(@"Using Spotlight on address: %@", spotlightValue);
```

lucas-zimerman:
This PR is also a reminder to me that I need to do the same on Capacitor

packages/core/ios/RNSentry.mm
antonis: Good catch Lucas 🙇

lucas-zimerman: I tested the branch locally and its doing what it should do :D 

This is the only issue I spotted after the review.

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -175,135 +176,143 @@ public void initNativeReactNavigationNewFrameTracking(Promise promise) {
...
+      options.setSendDefaultPii(rnOptions.getBoolean("sendDefaultPii"));
+    }
+    if (rnOptions.hasKey("maxQueueSize")) {
+      options.setMaxQueueSize(rnOptions.getInt("maxQueueSize"));
+    }
+    if (rnOptions.hasKey("enableNdk")) {
+      options.setEnableNdk(rnOptions.getBoolean("enableNdk"));
+    }
+    if (rnOptions.hasKey("spotlight")) {
+      if (rnOptions.getType("spotlight") == ReadableType.Boolean) {
```
krystofwoldrich, line 265: Nice, we should use this type check more here.

krystofwoldrich (approved):
Awesome, thanks for adding the screens from testing and more platform tests. 🚀

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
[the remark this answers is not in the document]
antonis: Good point @krystofwoldrich especially given that the commits are squashed for merge. I'll use this approach in the future 🙇

BYK:
Hey @antonis, thanks so much for this! 🥳

PR #4233, opened 2024-11-04
fix(ci): Skip e2e metrics steps when the auth token is not available

antonis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Checks if the auth token is accessible on a reusable workflow to fix:
```
Invalid workflow file: .github/workflows/e2e.yml#L27
The workflow is not valid. .github/workflows/e2e.yml (Line: 27, Col: 9): Unrecognized named-value: 'env'. Located at position 47 within expression: needs.diff_check.outputs.skip_ci != 'true' && env.SENTRY_AUTH_TOKEN != null
```
([example ci failure](https://github.com/getsentry/sentry-react-native/actions/runs/11668469539))

This solution [inherits the secret to the reusable workflow](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow) that [accesses it via an env variable](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow). The output is then used in the parent workflow to `skip_ci` when the token is not available.

Follow up on https://github.com/getsentry/sentry-react-native/pull/4227

CI

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


krystofwoldrich (approved):
LGTM 🚀 

Thank you for the fix.

PR #4224, opened 2024-10-31
feat(replay): Add SentryMask and SentryUnmask iOS components

krystofwoldrich:
## :loudspeaker: Type of change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This PR adds custom redact/mask components for iOS. 

The implementation includes noop fallback for Android and other unsupported platforms and fallback Web implementation. These will be added in a follow up PRs. 

```jsx
import * as Sentry from '@sentry/react-native'

const Example = () => (
  <View>
    <Sentry.Unmask>
      <Text>This is unmasked because it's direct child of Sentry.Unmask (can be masked if Sentry.Masked is used higher in the hierarchy)</Text>
      <Sentry.Mask>
        <Text>This is masked always because it's a child of a Sentry.Mask</Text>
        <Sentry.Unmask>
          {/* Sentry.Unmask does not override the Sentry.Mask from above in the hierarchy */}
          <Text>This is masked always because it's a child of Sentry.Mask</Text>
        </Sentry.Unmask>
      </Sentry.Mask>
    </Sentry.Unmask>
  </View>
);
```

| Device | Replay |
|--------|--------|
| Screenshot 2024-11-11 at 16 24 03 | Screenshot 2024-11-11 at 16 24 41 | 

```[tasklist]
### Todo
- [x] Requires https://github.com/getsentry/sentry-cocoa/releases/tag/8.40.0
- [ ] https://github.com/getsentry/sentry-cocoa/pull/4522
- [x] iOS Legacy Native Components
- [x] iOS New Arch Components
- [x] Android Legacy Components (in a follow up PR)
- [x] Android New Arch Components (in a follow up PR)
- [x] Web Component (in a follow up PR)
```

## :bulb: Motivation and Context

This is a part of RN Replay GA tasks 
- https://github.com/getsentry/sentry/issues/74441

## :green_heart: How did you test it?
ci, local,

## :pencil: Checklist

- [ ] I reviewed submitted code
- [ ] I added tests to verify changes
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] All tests passing
- [ ] No breaking changes

PR #4238, opened 2024-11-06
test(ts3.8): fix ts3.8 type check test to not include dom types

krystofwoldrich:
#skip-changelog 

This PR changes the type check test project to only include the es2015/es6 types excluding dom, which would cause duplicate declarations when the project would import React or React Native.

- This is needed for https://github.com/getsentry/sentry-react-native/pull/4224
  - Where we want to export React Native HostComponent Type

### DIFF
--- a/dev-packages/type-check/ts3.8-test/package.json
+++ b/dev-packages/type-check/ts3.8-test/package.json
@@ -9,7 +9,7 @@
   },
   "license": "MIT",
   "devDependencies": {
-    "@types/react": "17.0.58",
+    "@types/react": "17.0.83",
     "@types/react-native": "0.65.30",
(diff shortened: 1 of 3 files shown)


dev-packages/type-check/ts3.8-test/package.json
```diff
@@ -9,7 +9,7 @@
   },
   "license": "MIT",
   "devDependencies": {
-    "@types/react": "17.0.58",
+    "@types/react": "17.0.83",
```
krystofwoldrich, line 12: This is has to to match with dependency of the react native types.

antonis (approved):
LGTM 🎉

PR #4245, opened 2024-11-07
fix(sample): 'codegen didn't run for RN*' warning caused by incorrect…

krystofwoldrich:
#skip-changelog 

Fixes sample app warning when using the new architecture.

Originally changed when switching monorepo structure.
- https://github.com/getsentry/sentry-react-native/pull/4057

The `react-native` must be first for the codegen to load in the ts files and generate correct React JS components based on it when new architecture is enabled.


antonis (approved):
LGTM!
I confirmed that the RN* warning is fixed with the change.

PR #3513, opened 2024-01-10
chore(ci): Temporarily disable new arch E2E test on ios

krystofwoldrich:
E2E on RN 0.73.0,1,2 New Architecture on iOS timeout due to race condition in the RN framework.

- More details https://github.com/facebook/react-native/issues/42123#issuecomment-1881203719

This PR disables the tests.

#skip-changelog

PR #4246, opened 2024-11-07
test(e2e): Return new architecture

krystofwoldrich:
#skip-changelog 

The tests were disabled due to a bug in RN new arch, which caused the tests to fail is CI.

Since then the bug was fixed so we can reenabled the tests.

- https://github.com/getsentry/sentry-react-native/pull/3513

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -326,10 +326,7 @@ jobs:
             engine: 'hermes'
           - rn-version: '0.76.0'
             engine: 'jsc'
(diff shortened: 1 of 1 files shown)


krystofwoldrich:
The E2E test of the new architecture on 0.76 passes correctly. I also checked the event reported to sentry and they are correct.

Sadly our sample app is not sending any events when the new architecture is enabled. The sample app is using 0.75.

antonis (approved):
LGTM

lucas-zimerman (approved):
LGTM!

PR #4249, opened 2024-11-08
fix(sample): Disable global promise patch

krystofwoldrich:
#skip-changelog 

This is a temporary fix to unblock the SDK development using the sample app.

Based on our E2E tests which use minimal RN application this is specific to our sample app setup.

We have not seen any reports of this from our users, but it's possible that application using similar setup might experience the same issue.


antonis (approved):
Thank you for this fix @krystofwoldrich 🙇 
I confirmed that the events are sent and the animation works as expected 🎉

PR #4260, opened 2024-11-11
fix(core): Export missing SeverityLevel from types

lucas-zimerman:
This PR adds the missing export SeverityLevel from @sentry/types.


krystofwoldrich (approved):
LGTM 🚀

Issue #3210, opened 2023-07-24
Don't create spans for Dev and Sentry Servers requests


krystofwoldrich:
### Description

This is the same issue we have with breadcrumbs.

- https://github.com/getsentry/sentry-react-native/issues/3105

But creating spans for these requests leads to transactions exceeding timeout because there are spans constantly being created.

Also in dev builds with debug on this leads to log spam, which makes it hard to debug the apps.

-   https://github.com/getsentry/sentry-react-native/issues/3205

nandorojo:
Hi @krystofwoldrich, thanks for remaking my issue here. Question for you: in dev mode, I have my app crashing, saying `Malformed JS: field sizes are different`. I traced the issue to `RNSentry.addBreadcrumb`. Is that related to this issue which you linked above for not adding to breadcrumbs? And if so, do you know of a recommended workaround?

PR #4271, opened 2024-11-13
feat(core): Drop dev server spans

antonis:

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Filters out development server spans using [the `shouldCreateSpanForRequest` option](https://docs.sentry.io/platforms/react-native/tracing/instrumentation/automatic-instrumentation/#shouldcreatespanforrequest).

Fixes https://github.com/getsentry/sentry-react-native/issues/3210

Manual testing, Unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


packages/core/src/js/tracing/reactnativetracing.ts
```diff
@@ -97,6 +98,22 @@ export const reactNativeTracingIntegration = (
...
+  const devServerUrl = getDevServer()?.url || '';
+  const finalShouldCreateSpanForRequest = (url: string): boolean => {
+    if (url.includes(devServerUrl)) {
+      return false;
+    }
+    if (userShouldCreateSpanForRequest) {
+      return userShouldCreateSpanForRequest(url);
+    }
+    return true;
+  };
```
lucas-zimerman, line 113: i: We could avoid this logic and rely only on the customer logic when devserver isn't found.
This can also avoid issues with `url.includes(''))` that can return true to any kind of string
```suggestion
  const devServerUrl = getDevServer()?.url;
  const finalShouldCreateSpanForRequest =
    devServerUrl === undefined
      ? userShouldCreateSpanForRequest
      : (url: string): boolean => {
          if (url.includes(devServerUrl)) {
            return false;
          }
          if (userShouldCreateSpanForRequest) {
            return userShouldCreateSpanForRequest(url);
          }
          return true;
        };
```

packages/core/test/tracing/reactnavigation.test.ts
```diff
@@ -32,6 +33,9 @@ const dummyRoute = {
 };
 
 jest.mock('../../src/js/wrapper.ts', () => jest.requireActual('../mockWrapper.ts'));
+jest.mock('./../../src/js/integrations/debugsymbolicatorutils', () => ({
+  getDevServer: jest.fn(),
```
lucas-zimerman, line 37: i: We should also add tests for when getDevServer() returns undefined

lucas-zimerman:
Overall, the PR looks good, but I have some concerns when `devserver` doesn't exist. Thank you for the contribution!

packages/core/src/js/tracing/reactnativetracing.ts
antonis: Thank you for the suggestion @lucas-zimerman and for catching this case 🙇

packages/core/test/tracing/reactnavigation.test.ts
antonis: Good idea @lucas-zimerman 👍 Added test cases with 5291fb1d01fc801bd09d4b2061fda65a5c5121df

PR #4273, opened 2024-11-14
fix(e2e): Bump tests to iOS 18.1  and macos-15

krystofwoldrich:
The E2E Test of RN 0.65 was failing on iPhone 14 simulator lookup.

Corepack is not available in macos-15 at the moment.
- https://github.com/actions/runner-images/issues/10905

#skip-changelog 

This PR doesn't fix the sample app provisioning profile issues not the native tests fail.


antonis (approved):
LGTM. Thank you for handling this 🙇

PR #4287, opened 2024-11-18
feat(metro): Add Sentry Middleware for source context in debug builds

krystofwoldrich:

- [x] Bugfix
- [x] New feature

## :scroll: Description

This PR sdds Sentry Metro Server Middleware which is enables the SDK to fetch source context in development build using one request.

This improves the resolution logic which originally required multiple requests per frame to get source context.

Another improvement is that request to a middleware does not trigger web builds in Expo applications. This makes the event processing faster and less confusing to the end user, who might be confused by why the web build are triggered (side note the if the web build would fail it might look like there is an error in the app).

In Expo SDK52 the dev server stopped serving the project file, thus the Sentry SDK could not retried the source context at all. Error in debug build would be locally symbolicated but without source context.

RN 0.76 still serves the files as expected.

sample expo app, sample rn app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- [x] Add unit tests
- [x] Test with older metro version (SDK 50)


antonis (approved):
The code LGTM and the RN and Expo samples work as expected with the `enableSourceContextInDevelopment` flag enabled and disabled 🚀

Issue #4121, opened 2024-09-26
Fix concurrent runs with Sentry Babel transformer


feidaZhang:
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [x]  iOS
- [ ]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 5.30.0

**`react-native` version:** 0.72.12

**Are you using Expo?**
- [ ]  Yes
- [x]  No

---
**I have the following issue:**

After integrating `withSentryConfig` in the `metro.config.js`, `annotateReactComponents` is true, we found that when building iOS application through Xcode in debug mode, the `.sentry/. defaultBabelTransformerPath` file was first created and then deleted before the build was completed. **However, when loading the JS Bundle, sentry tried to read the file again and threw an error.**

**Actual result:**

![Image](https://github.com/user-attachments/assets/f5ab5ffc-9e97-4c0d-9f10-d9439b95c2dd)

**Expected result:**

no error

feidaZhang:
if i comment this line. There is no error but the file can not be deleted automatically.
```
function withSentryBabelTransformer(config) {
    const defaultBabelTransformerPath = config.transformer && config.transformer.babelTransformerPath;
    utils_1.logger.debug('Default Babel transformer path from `config.transformer`:', defaultBabelTransformerPath);
    if (!defaultBabelTransformerPath) {
        // This has to be console.warn because the options is enabled but won't be used
        // eslint-disable-next-line no-console
        console.warn('`transformer.babelTransformerPath` is undefined.');
        // eslint-disable-next-line no-console
        console.warn('Sentry Babel transformer cannot be used. Not adding it...');
        return config;
    }
    if (defaultBabelTransformerPath) {
        (0, sentryBabelTransformerUtils_1.saveDefaultBabelTransformerPath)(defaultBabelTransformerPath);
        process.on('exit', () => {
             // comment this line
            // (0, sentryBabelTransformerUtils_1.cleanDefaultBabelTransformerPath)();
        });
    }
    return Object.assign(Object.assign({}, config), { transformer: Object.assign(Object.assign({}, config.transformer), { babelTransformerPath: require.resolve('./sentryBabelTransformer') }) });
}
```

feidaZhang:
It's my device info.
  Model Name:	MacBook Pro
[...]

PR #4298, opened 2024-11-20
fix(metro): Use env for default Babel transformer path

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

This PR remove use of tmp `.sentry` dir. And adds dedicated Sentry default Babel transformer env instead.

The use of ENV ensures that multiple concurent builds are not sharing the same resource (the tmp dir). The env is by default scoped to the current process and its children.

- Fixes: https://github.com/getsentry/sentry-react-native/issues/4121

sample app, updated unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


antonis (approved):
LGTM 🚀

Issue #4311, opened 2024-11-26
`lastEventId` is removed in v6


thespacemanatee:
### What React Native libraries do you use?

React Navigation, Hermes, Expo Application Services (EAS), Expo (mobile only)

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### @sentry/react-native SDK Version

6.3.0

### How does your development environment look like?

expo-env-info 1.2.1 environment info:
    System:
      OS: macOS 15.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 20.9.0 - ~/.asdf/installs/nodejs/20.9.0/bin/node
      Yarn: 1.22.22 - ~/.asdf/installs/nodejs/20.9.0/bin/yarn
      npm: 10.1.0 - ~/.asdf/plugins/nodejs/shims/npm
      Watchman: 2024.11.18.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.15.2 - /Users/honeycomb/.asdf/shims/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 24.1, iOS 18.1, macOS 15.1, tvOS 18.1, visionOS 2.1, watchOS 11.1
    IDEs:
      Android Studio: EAP AI-242.23726.103.2422.12630895 AI-242.23726.103.2422.12630895
      Xcode: 16.1/16B40 - /usr/bin/xcodebuild
    npmGlobalPackages:
      eas-cli: 13.3.0
    Expo Workflow: bare

### Sentry.init()

Sentry.init({
  dsn: 'https://...ingest.sentry.io/...',
  autoInitializeNativeSdk: false,
  integrations: [reactNavigationIntegration],
  beforeBreadcrumb: (breadcrumb) => {
    if (SENTRY_LOG_BREADCRUMBS) {
      console.log('[Sentry] Breadcrumb:', breadcrumb)
    }
    return breadcrumb
  },
})

### Steps to Reproduce

Refer to https://github.com/getsentry/sentry-javascript/issues/11951#issuecomment-2499563835 on why `lastEventId` should be accessible.

1. Update to v6.
2. Restart editor

### Expected Result

Should be able to use `lastEventId`.

### Actual Result

Lint error when trying to use `lastEventId`.

krystofwoldrich:
Thank you @thespacemanatee

I confirm this was removed during the JS Core v8 upgrade, but since JS v8 returned the function RN SDK can return it as well.

[...]

PR #4315, opened 2024-11-26
fix(exports): Return `lastEventId`

krystofwoldrich:
Same as JS https://github.com/getsentry/sentry-javascript/issues/11951#issuecomment-2499563835

Fixes https://github.com/getsentry/sentry-react-native/issues/4311


antonis (approved):
LGTM 🎉

Issue #3859, opened 2024-06-05
Add new captureFeedback API to RN SDK


dalnoki:
### Description

This function is the base functionality needed to allow capturing user feedback while including tags, `replay_id`, `trace_id`, etc.
It's part of the 'new user feedback' product (https://github.com/getsentry/sentry/issues/64671) that was shipped for Web. That included a built-in widget, that allows screenshot attachment, and under the hood relies on this new function that deprecates the old `captureUserFeedback` which required an `event_id`. The new function optionally takes a related `event_id` and will contain relevant context stand alone.

**Since RN is built on top of the JS SDK we might be able to re-export the function. But we need to verify that connection to tracing and replays will work properly. See:**
* https://docs.sentry.io/platforms/javascript/user-feedback/#user-feedback-api

This means a new envelope format for feedback was introduced. 
The new Feedback product architecture has thoroughly documented by @aliu39 here:

* https://develop.sentry.dev/application/feedback-architecture/

This new function will allow outgoing feedbacks to be tied to Session Replay, which is currently Open Beta and we're working on GA: 
* https://github.com/getsentry/sentry/issues/74441

The `develop` docs have not been added yet, but the ticket with details about it is here:
* https://github.com/getsentry/develop/issues/1206

Relates to:
* https://github.com/getsentry/sentry-java/issues/3599
* https://github.com/getsentry/sentry-cocoa/issues/4187
* https://github.com/getsentry/sentry-go/issues/842
* https://github.com/getsentry/sentry-dart/issues/1593
* https://github.com/getsentry/sentry-javascript/issues/10953

bruno-garcia:
See also:
* https://github.com/sentry-demos/sentry_react_native/pull/72

krystofwoldrich:
- Depends on https://github.com/getsentry/sentry-java/pull/3687

PR #4320, opened 2024-11-27
feat: Add new captureFeedback API to RN SDK

antonis:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

* [Updates the client implementation to use the new capture feedback js api](https://github.com/getsentry/sentry-react-native/commit/817eac82107f668f71ab51e8d071ca62c472e62d)
* [Updates SDK API](https://github.com/getsentry/sentry-react-native/commit/5370a990006b5d53b738ebc250c7e240a7bffa36)
* [Adds new feedback button in the sample](https://github.com/getsentry/sentry-react-native/commit/42e2fa14823f067a6b4088770f5ff0aea9a22b69)

Fixes https://github.com/getsentry/sentry-react-native/issues/3859

CI, Manual testing with the sample app:
* [Feedback with event id](https://sentry-sdks.sentry.io/feedback/?feedbackSlug=sentry-react-native%3A6094082511&mailbox=unresolved&project=-1&referrer=feedback_list_page&statsPeriod=1h)
* [Feedback without event id](https://sentry-sdks.sentry.io/feedback/?feedbackSlug=sentry-react-native%3A6094091123&mailbox=unresolved&project=-1&referrer=feedback_list_page&statsPeriod=1h)

Screenshot 2024-11-27 at 1 58 32 PM

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/packages/core/src/js/client.ts
+++ b/packages/core/src/js/client.ts
@@ -1,4 +1,4 @@
-import { eventFromException, eventFromMessage } from '@sentry/browser';
+import { captureFeedback as captureFeedbackApi, eventFromException, eventFromMessage } from '@sentry/browser';
 import { BaseClient } from '@sentry/core';
 import type {
   ClientReportEnvelope,
@@ -7,9 +7,9 @@ import type {
   Event,
   EventHint,
   Outcome,
+  SendFeedbackParams,
   SeverityLevel,
   TransportMakeRequestResponse,
-  UserFeedback,
 } from '@sentry/types';
 import { dateTimestampInSeconds, logger, SentryError } from '@sentry/utils';
 import { Alert } from 'react-native';
@@ -20,7 +20,7 @@ import { getDefaultSidecarUrl } from './integrations/spotlight';
 import type { ReactNativeClientOptions } from './options';
 import type { mobileReplayIntegration } from './replay/mobilereplay';
 import { MOBILE_REPLAY_INTEGRATION_NAME } from './replay/mobilereplay';
-import { createUserFeedbackEnvelope, items } from './utils/envelope';
+import { items } from './utils/envelope';
 import { ignoreRequireCycleLogs } from './utils/ignorerequirecyclelogs';
 import { mergeOutcomes } from './utils/outcome';
 import { ReactNativeLibraries } from './utils/rnlibraries';
@@ -86,14 +86,8 @@ export class ReactNativeClient extends BaseClient<ReactNativeClientOptions> {
   /**
    * Sends user feedback to Sentry.
    */
-  public captureUserFeedback(feedback: UserFeedback): void {
-    const envelope = createUserFeedbackEnvelope(feedback, {
-      metadata: this._options._metadata,
-      dsn: this.getDsn(),
-      tunnel: undefined,
-    });
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    this.sendEnvelope(envelope);
+  public captureFeedback(feedback: SendFeedbackParams): void {
+    captureFeedbackApi(feedback);
   }
 
   /**
--- a/packages/core/src/js/sdk.tsx
+++ b/packages/core/src/js/sdk.tsx
@@ -4,7 +4,7 @@ import {
   defaultStackParser,
   makeFetchTransport,
 } from '@sentry/react';
-import type { Breadcrumb, BreadcrumbHint, Integration, Scope, UserFeedback } from '@sentry/types';
+import type { Breadcrumb, BreadcrumbHint, Integration, Scope, SendFeedbackParams, UserFeedback } from '@sentry/types';
 import { logger, stackParserFromStackParserOptions } from '@sentry/utils';
 import * as React from 'react';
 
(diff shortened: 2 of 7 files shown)


lucas-zimerman (approved):
All seems good, thank you for splitting the PR!

packages/core/src/js/client.ts
```diff
@@ -86,14 +86,8 @@ export class ReactNativeClient extends BaseClient<ReactNativeClientOptions> {
...
-  public captureUserFeedback(feedback: UserFeedback): void {
-    const envelope = createUserFeedbackEnvelope(feedback, {
-      metadata: this._options._metadata,
-      dsn: this.getDsn(),
-      tunnel: undefined,
-    });
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    this.sendEnvelope(envelope);
+  public captureFeedback(feedback: SendFeedbackParams): void {
+    captureFeedbackApi(feedback);
```
krystofwoldrich, line 90: [`captureFeedbackApi` dynamically gets client](https://github.com/getsentry/sentry-javascript/blob/dafd51054d8b2ab2030fa0b16ad0fd70493b6e08/packages/core/src/feedback.ts#L8), we should not call in in a client as it could cause unexpected result.

Client methods has to use this to captureEvent.

Maybe we could just deprecate this old API?

CHANGELOG.md
```diff
@@ -8,6 +8,26 @@
...
+  // OR: const eventId = Sentry.lastEventId();
+
+  const userFeedback: SendFeedbackParams = {
+    name: "John Doe",
+    email: "john@doe.com",
+    message: "Hello World!",
+    associatedEventId: eventId,// Optional
+  };
+  Sentry.captureFeedback(userFeedback);
+  ```
```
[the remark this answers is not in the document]
antonis: Good point @krystofwoldrich 👍 
I used the [current doc](https://docs.sentry.io/platforms/react-native/user-feedback/#user-feedback-api) as a starting point but I agree with your suggestion 🙇

packages/core/src/js/sdk.tsx
```diff
@@ -219,9 +219,23 @@ export async function close(): Promise<void> {
...
+  };
+  captureFeedback(feedbackParams);
+}
+
+/**
+ * Captures user feedback and sends it to Sentry.
+ */
+export function captureFeedback(feedbackParams: SendFeedbackParams): void {
+  getClient<ReactNativeClient>()?.captureFeedback(feedbackParams);
 }
```
krystofwoldrich, line 239: Could we directly use `captureFeedback` from `@sentry/core`.

packages/core/src/js/client.ts
```diff
@@ -1,4 +1,4 @@
-import { eventFromException, eventFromMessage } from '@sentry/browser';
```
krystofwoldrich, line 1: Let's use the export from `@sentry/core`, it's more safe as core should not take any browser assumptions. (In this case it's the same function, but it's a good practice for RN)

packages/core/src/js/client.ts
antonis: Makes sense. I will remove the client implementation and move this on the sdk as per [your other comment](https://github.com/getsentry/sentry-react-native/pull/4320#discussion_r1862033816).

krystofwoldrich:
https://github.com/getsentry/sentry-javascript/blob/dafd51054d8b2ab2030fa0b16ad0fd70493b6e08/packages/core/test/lib/feedback.test.ts#L14

Let's copy this tests of the API to our RN repository, to ensure it behaves as expected. I would like to start doing this going forward, as we encountered in the past bug caused by change of behavior of the API which was not obvious from the changelog and thus we missed it.

packages/core/src/js/sdk.tsx
antonis: Updated with da0d4ac091e20b0232f841ff9a5597597e47070c

packages/core/src/js/client.ts
antonis: Updated with 3e36c6dcfcc20c32f8d0d0bc5585cf1d6dd4df30
Note that `eventFromException` is only exported from '@sentry/browser'

antonis:
> https://github.com/getsentry/sentry-javascript/blob/dafd51054d8b2ab2030fa0b16ad0fd70493b6e08/packages/core/test/lib/feedback.test.ts#L14
> Let's copy this tests of the API to our RN repository, to ensure it behaves as expected. I would like to start doing this going forward, as we encountered in the past bug caused by change of behavior of the API which was not obvious from the changelog and thus we missed it.

Copied the tests with f9d2b59500ea58956f42f5250d395270a385ae31

packages/core/src/js/client.ts
antonis: Reverted the change since it resulted in [a build error](https://github.com/getsentry/sentry-react-native/actions/runs/12070129197/job/33658926517). I suggest to revisit this separately.

krystofwoldrich: Yeah, should work, but the implementation are sadly not the exactly same like in case of the `captureFeedback`. :/

packages/core/src/js/client.ts
```diff
@@ -83,19 +82,6 @@ export class ReactNativeClient extends BaseClient<ReactNativeClientOptions> {
     });
   }
 
-  /**
-   * Sends user feedback to Sentry.
-   */
-  public captureUserFeedback(feedback: UserFeedback): void {
```
krystofwoldrich, line 89: I would only deprecate this, to avoid breaking changes. Since the Client class is public interface.

https://github.com/getsentry/sentry-react-native/blob/8ab11b68f5637b8a7886341bcc06214f8cd48ebb/packages/core/src/js/index.ts#L60

PR #4326, opened 2024-11-28
fix(span): Navigation Span should have no parent by default

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

This PR fixes a bug introduced in RN SDK V6. Where unfinished navigations spans are set as parents of newly created navigations spans.

This is unexpected as navigations span should be a active root span (aka transaction in the SDK V5).

This issue was reported by the sales team working with clients trying out the new SDK V6.

added unit test for the behaviour, tests failed before the fix.

Also tested this with the Sentry RN Demo App.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Features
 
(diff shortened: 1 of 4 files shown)


antonis (approved):
LGTM 🚀 
Thank you for adding tests 🙇

Issue #4337, opened 2024-12-02
Add attachment/screenshot handling in the feedback form UI


antonis:
Similar to the web the user should be able to add attachments/screenshots via the feedback form

antonis:
Implemented in https://github.com/getsentry/sentry-react-native/pull/4338

PR #4328, opened 2024-11-29
(1) feat: Add Feedback Form Component 

antonis:
## :loudspeaker: Type of change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Adds a basic feedback form with the following fields: name, email, description.
The form text and styles can be customised.

|Feedback Form Sample Button|Feedback Form iOS|Feedback Form Android|
|---|---|---|
|Simulator Screenshot - iPhone SE (3rd generation) - 2024-11-29 at 23 16 21|Simulator Screenshot - iPhone SE (3rd generation) - 2024-12-11 at 16 44 55|Screenshot_1733913880|

### Subtask PRs:
- https://github.com/getsentry/sentry-react-native/pull/4364
  - https://github.com/getsentry/sentry-react-native/pull/4383
- https://github.com/getsentry/sentry-react-native/pull/4357
- ~https://github.com/getsentry/sentry-react-native/pull/4370~
  - https://github.com/getsentry/sentry-react-native/pull/4378
- https://github.com/getsentry/sentry-react-native/pull/4338
- https://github.com/getsentry/sentry-react-native/pull/4379

## :bulb: Motivation and Context

Fixes https://github.com/getsentry/sentry-react-native/issues/4336

## :green_heart: How did you test it?
CI, Manual testing ([example](https://sentry-sdks.sentry.io/feedback/?feedbackSlug=sentry-react-native%3A6105538110&mailbox=unresolved&project=5428561&referrer=feedback_list_page&statsPeriod=1h))

## :pencil: Checklist

- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- [Handle network or other errors](https://github.com/getsentry/sentry-react-native/issues/4359)
- [Implement attachment handling](https://github.com/getsentry/sentry-react-native/issues/4337)
- [Add branding logo](https://github.com/getsentry/sentry-react-native/pull/4357)
- [Show success toast](https://github.com/getsentry/sentry-react-native/issues/4356)
- [Autoinject feedback widget](https://github.com/getsentry/sentry-react-native/issues/4358)
- [Support themes](https://github.com/getsentry/sentry-react-native/issues/4360)

PR #4338, opened 2024-12-03
(2.4) feat(feedback-ui): Add screenshots 

antonis:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

**Based on https://github.com/getsentry/sentry-react-native/pull/4328**

## :scroll: Description

This PR adds attachment handling in the SDK capture feedback form [leaving the image picking to be implemented on the client](https://github.com/getsentry/sentry-react-native/pull/4338/files#diff-44444b02440ef6ec263ca8330a83217645baeb2ea657b03aee3aad50a54a672aR35):
* [Adds attachment button](https://github.com/getsentry/sentry-react-native/pull/4338/files#diff-4ea1bc6415dbb3c3f1761351b30bfa6745197e54c6b65515392be2ca3985ac37R152)
* [Adds SDK callbacks for handling attachments](https://github.com/getsentry/sentry-react-native/pull/4338/files#diff-e95c4cea1ecbca9dd81fec4e312a2327ffe196c11c0fca2fbbe5ff6488f700f4R144)
* [Implements feedback attachments in the sample app](https://github.com/getsentry/sentry-react-native/pull/4338/files#diff-1c1adb2dc16ba38641aaeb0f0239b7fc1596dddd4574b835cec7689834592ee3R180)

_Note that [string attachments are converted to UInt8Array before sending to the server](https://github.com/getsentry/sentry-react-native/pull/4338#issuecomment-2514610752)._

|Android|iOS|
|---|---|
|||

Fixes https://github.com/getsentry/sentry-react-native/issues/4337

CI, Manual testing ([example](https://sentry-sdks.sentry.io/feedback/?feedbackSlug=sentry-react-native%3A6134548994&mailbox=ignored&project=5428561&referrer=feedback_list_page&statsPeriod=1h))

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog

[As discussed](https://github.com/getsentry/sentry-react-native/pull/4328#discussion_r1871802610) we will use the no props configuration in the changelog.


antonis:
Hey @krystofwoldrich, @lucas-zimerman 👋 

thought this PR is still draft (trying to figure out [why images are not uploaded correctly](https://sentry-sdks.sentry.io/feedback/?feedbackSlug=sentry-react-native%3A6107486325&mailbox=ignored&project=5428561&referrer=feedback_list_page&statsPeriod=1h)) I'd like your feedback since my approach might be a bit opinionated.
To avoid adding a library in Sentry core (and [issues with hooks](https://github.com/getsentry/sentry-react-native/pull/4328#discussion_r1866149551)) I've left the actual image picking on the client app:
* [Sentry adds the necessary callbacks](https://github.com/getsentry/sentry-react-native/pull/4338/commits/eabacb4b956e72d2d7f1a67e070504a92c7fa691)
* [The client app (sample) implements image picking](https://github.com/getsentry/sentry-react-native/pull/4338/commits/34bf6a853513a7a01a93fdd91e2dd196d6aa867e)

Wdyt of this approach? Should the library offer the full functionality or leave it to the sample app 🙇

antonis:
> trying to figure out [why images are not uploaded correctly](https://sentry-sdks.sentry.io/feedback/?feedbackSlug=sentry-react-native%3A6107486325&mailbox=ignored&project=5428561&referrer=feedback_list_page&statsPeriod=1h)

I wasn't able to correctly upload a base64 image string thus I [always convert to Uint8Array before sending](https://github.com/getsentry/sentry-react-native/pull/4338/commits/0cd3982fa5da1cced2debf2d50c5077597599bf0).

antonis:
Converting to draft similar to https://github.com/getsentry/sentry-react-native/pull/4328#issuecomment-2531311367

packages/core/src/js/feedback/FeedbackForm.types.ts
```diff
@@ -33,6 +33,12 @@ export interface FeedbackGeneralConfiguration {
    */
   showName?: boolean;
 
+  /**
+   * Should the screen shots field be included?
+   * Screen shots cannot be marked as required
+   */
```
lucas-zimerman, line 39: i: we should suggest the default set
```suggestion
   * @default false
   */
```

packages/core/src/js/feedback/FeedbackForm.types.ts
```diff
@@ -33,6 +33,12 @@ export interface FeedbackGeneralConfiguration {
    */
   showName?: boolean;
 
+  /**
+   * Should the screen shots field be included?
+   * Screen shots cannot be marked as required
```
lucas-zimerman, line 38:
```suggestion
   * This flag determines whether the "Add Screenshot" button is displayed
```

packages/core/src/js/feedback/FeedbackForm.types.ts
```diff
@@ -33,6 +33,12 @@ export interface FeedbackGeneralConfiguration {
...
+   */
+  enableScreenshot?: boolean;
```
lucas-zimerman, line 40: i: The name could change, since `enableScreenshot` doesn't give the impression that it will impact on the button visibility.
I thought about a bit verbose name, but when you read the property name, you know exactly what it will do without reading the description.
```suggestion
  showAddScreenshotButton?: boolean;
```

antonis: That's indeed a good suggestion @lucas-zimerman 👍 
I'm a bit torn between [aligning](https://github.com/getsentry/sentry-react-native/pull/4328#discussion_r1871764316) with [the JS api](https://github.com/getsentry/sentry-javascript/blob/develop/packages/core/src/types-hoist/feedback/config.ts#L49) and using a more clear name for this 🤔

krystofwoldrich: I also think the name is too generic, but let stick with what users already know from JS.

packages/core/src/js/utils/base64ToUint8Array.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    if (bits >= 8) {
+      // Add a byte when we have 8 or more bits
+      bits -= 8;
+      bytes.push((buffer >> bits) & 0xff);
+    }
+  }
+
+  return new Uint8Array(bytes);
+};
```
krystofwoldrich, line 29: This is quite resource intensive, we do some of these conversion in the JS already, but it should be avoided as much as possible.

---

I this case I would leave that up to the users even thought it makes it a bit harder to use the API (depending on how they get the screenshot)

But it enables user to use conversion lib that are implemented in cpp with much better performance.

krystofwoldrich:
The impl looks good, but I would scale the scope down to only allow bytes array for now and remove the conversion. We can extend it later.

krystofwoldrich:
@antonis Have you explored the option of taking screenshot of the app and attaching it?

krystofwoldrich:
> I haven't explored this option on the SDK side. My assumption was that the developer may choose to implement this. E.g to capture a screenshot with react-native-view-shot instead of selecting an image with react-native-image-picker as the sample implementation.

Got it. This is nice and flexible, but provides less out of the box value.

> Looking at it now I see that we already have [`captureScreenshot`](https://github.com/getsentry/sentry-react-native/blob/b6f8ea26ccc93d7a6ce3aacc0faafecdd16b3d3d/packages/core/src/js/wrapper.ts#L79) (that convenietly returns [`Uint8Array`](https://github.com/getsentry/sentry-react-native/blob/b6f8ea26ccc93d7a6ce3aacc0faafecdd16b3d3d/packages/core/src/js/wrapper.ts#L47)) functionality exposed by the native SDKs. I'm not sure if this would be useful though since it will be a screenshot of the feedback form unless we use a screenshot taken before 🤔  Maybe I can explore this on this or a separate PR.

Yes, exactly we already have that function, but as you said taking the screenshot of the form is not useful. But especially when combined with the auto inject, since we will be responsible for showing and hiding the form we can, hide the form take the screenshot and then show it again.

Note: We might anyway also need to create a base64 representation of the screenshot to be able to show a preview using RN Image tag. Or create a tmp file.

samples/react-native/src/App.tsx
```diff
@@ -139,6 +141,53 @@ const Stack = isMobileOs
...
+
+    if (bits >= 8) {
+      // Add a byte when we have 8 or more bits
+      bits -= 8;
+      bytes.push((buffer >> bits) & 0xff);
+    }
+  }
+
+  return new Uint8Array(bytes);
+};
```
krystofwoldrich, line 172: l: Since this is in the sample app we could use a lib to convert the data.

https://www.npmjs.com/package/react-native-quick-base64

packages/core/src/js/feedback/FeedbackForm.tsx
```diff
@@ -90,6 +99,17 @@ export class FeedbackForm extends React.Component<FeedbackFormProps, FeedbackFor
     }
   };
 
+  public addRemoveScreenshot: () => void = () => {
```
krystofwoldrich, line 102: l: I find the name a bit confusing. Does `toggleScreenshot` or `onScreenshotButtonPress` should better to you?

krystofwoldrich (approved):
Look good! 🚀 

Just left two small nits in the comments.

packages/core/src/js/feedback/FeedbackForm.tsx
antonis: Good catch 👍 
Updated with 91a4b8ad95e4d539a134bb232f8d1fccc54bbfe1 to use `onScreenshotButtonPress`

antonis:
> Just left two small nits in the comments.

Thank you once more for the feedback @krystofwoldrich 🙇 
I fixed the nits and I'll merge on the feature branch as soon as the CI is 🟢

PR #4354, opened 2024-12-10
feat(release): Replace release bot with GH app

Jeffreyhung:
The `getsentry-release` is a GitHub bot account that is used in various automation, and the `${{ secrets.GH_RELEASE_PAT }}` is a personal access token from that bot account. We are using a regular GitHub account as a bot while it should be a non-human account since there are no humans behind it. Hence, we are replacing it with a GitHub App. 

Functionality wise, there will be no difference.

More details: https://www.notion.so/sentry/DACI-Replace-GitHub-bot-accounts-with-GitHub-Apps-getsentry-release-15109965d1204a91b9be71c49e8b66e0?pvs=4

#skip-changelog

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,10 +17,16 @@ jobs:
     runs-on: ubuntu-latest
     name: "Release a new version"
     steps:
+      - name: Get auth token
+        id: token
+        uses: actions/create-github-app-token@5d869da34e18e7287c1daad50e0b8ea0f506ce69 # v1.11.0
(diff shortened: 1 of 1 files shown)


lucas-zimerman (approved):
@krystofwoldrich Do also need the same PR for the branch `v5`?

In regard to the main, LGTM!

Jeffreyhung:
@lucas-zimerman @krystofwoldrich I can have another PR up if we figured we need one for the other branch too!
Thank you both!

Issue #4359, opened 2024-12-11
Handle feedback form submision errors


antonis:

antonis:
Implemented in https://github.com/getsentry/sentry-react-native/pull/4364

PR #4364, opened 2024-12-12
(2.1) Handles captureFeedback errors

antonis:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

**Based on https://github.com/getsentry/sentry-react-native/pull/4328**

## :scroll: Description

* Handles captureFeedback errors
* Aligns with the JS callbacks API by adding the `onFormOpen`, `onSubmitSuccess`, `onSubmitError`, `onFormSubmitted`

Fixes https://github.com/getsentry/sentry-react-native/issues/4359

Manual testing, Unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


packages/core/src/js/feedback/FeedbackForm.tsx
```diff
@@ -53,11 +54,15 @@ export class FeedbackForm extends React.Component<FeedbackFormProps, FeedbackFor
...
-    Alert.alert(text.successMessageText);
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    checkInternetConnection(() => {
+      onFormClose();
+      this.setState({ isVisible: false });
+      captureFeedback(userFeedback);
+      Alert.alert(text.successMessageText);
+    }, () => {
+      Alert.alert(text.errorTitle, text.networkError);
+    });
```
lucas-zimerman, line 65: I think the only negative side with this implementation is that the user will not have an opportunity to retry sending the feedback in case of internet failure.

In this case, we'd need to invert our logic again by closing the form only after captureFeedback succeeds.

packages/core/src/js/feedback/FeedbackForm.tsx
```diff
@@ -53,11 +54,15 @@ export class FeedbackForm extends React.Component<FeedbackFormProps, FeedbackFor
...
-    captureFeedback(userFeedback);
-    Alert.alert(text.successMessageText);
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    checkInternetConnection(() => {
+      onFormClose();
+      this.setState({ isVisible: false });
+      captureFeedback(userFeedback);
+      Alert.alert(text.successMessageText);
+    }, () => {
+      Alert.alert(text.errorTitle, text.networkError);
```
lucas-zimerman, line 64: i: We should use logger to also log the feedback was capture/failed.

packages/core/src/js/feedback/FeedbackForm.tsx
antonis: Good point 👍  

The current implementation should close the form only on success. In case there is a network issue the error alert pops on top of the form and the user can close it and retry sending the feeback. Do you mean adding a 'Retry' button?

packages/core/src/js/feedback/FeedbackForm.tsx
antonis: Good idea 👍 
Added with fe7b14cd6901f944891a8e191f6709a2813d3e7a

packages/core/src/js/feedback/FeedbackForm.tsx
lucas-zimerman: after reading the other part of the code, this will not be an issue since we will only try to send the feedback and closing it only after making sure we can communicate with the server.

packages/core/src/js/feedback/utils.ts
```diff
@@ -0,0 +1,17 @@
+export const checkInternetConnection = async (onConnected: () => void, onDisconnected: () => void): Promise<void> => {
+  try {
+    const response = await fetch('https://sentry.io', { method: 'HEAD' });
+    if (response.ok) {
+      onConnected();
+    } else {
+      onDisconnected();
+    }
+  } catch (error) {
+    onDisconnected();
```
lucas-zimerman, line 10: i: We should add an onError call and pass error as a parameter, so users know they are facing an issue that isn't unrelated to connectivity issues (for example, wrong dsn, server rejecting the feedback, some sdk error,...)

a suggestion for the error message: "Unable to send feedback due to an unexpected error."

antonis: Good idea 👍 
Added an onError method to communicate the unexpected error with 40b4fd3be9d9f11bee401c972e1efa07aaad1dcf

antonis:
I plan to iterate with the alternative approach since [the latest JS SDK](https://github.com/getsentry/sentry-react-native/pull/4382) [includes the SendFeedback fix](https://github.com/getsentry/sentry-javascript/releases/tag/8.45.1)

lucas-zimerman:
> I plan to iterate with the alternative approach sinc

So this PR is blocked by https://github.com/getsentry/sentry-react-native/pull/4383 ?

antonis:
>> I plan to iterate with the alternative approach sinc
> So this PR is blocked by https://github.com/getsentry/sentry-react-native/pull/4383 ?

I think we can start with the current approach and then iterate with an improvement with https://github.com/getsentry/sentry-react-native/pull/4383

krystofwoldrich (changes requested):
The transport in mobile SDK works differently than the JS transport, and we currently don't have the response available, so the `sendFeedback` will still throw event after the fix in JS.

See impl of sendEnvelope in RN https://github.com/getsentry/sentry-react-native/blob/18ce5e87b6e981e23ef9493207fb152d9c6aa8e7/packages/core/src/js/client.ts#L132

---

I think we should avoid checking the connection before send, because the native transport handles that (it stores the envelope offline and tries to send it later).

---

We can just try catch our current `captureFeedback` api for the on error call back. And note in the docs that if the device is offline the envelope is sent later.

antonis:
Thank you for the feedback @krystofwoldrich and the explanation on why not to use `sendFeedback` 🙇 

> We can just try catch our current captureFeedback api for the on error call back. And note in the docs that if the device is offline the envelope is sent later.

Makes sense 👍 Updated with c014ccc3ab2eca8fbb8480697cb48544e4fbe878

lucas-zimerman (approved):
Looks good to me! Thank you for the PR!

krystofwoldrich (approved):
Thank you for all the changes and refs. 

Looks good! 🚀

Issue #4375, opened 2024-12-14
No releases detected in sentry.io


knro:
### What React Native libraries do you use?

React Native without Frameworks

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### @sentry/react-native SDK Version

6.4.0

### How does your development environment look like?

```
info Fetching system and libraries information...
System:
  OS: Linux 6.8 KDE neon 6.2 6.2
  CPU: (24) x64 AMD Ryzen 9 5900X 12-Core Processor
  Memory: 28.22 GB / 62.71 GB
  Shell:
    version: 5.2.21
    path: /bin/bash
Binaries:
  Node:
    version: 20.9.0
    path: ~/.nvm/versions/node/v20.9.0/bin/node
  Yarn:
    version: 4.4.0
    path: ~/.nvm/versions/node/v20.9.0/bin/yarn
  npm:
    version: 10.1.0
    path: ~/.nvm/versions/node/v20.9.0/bin/npm
  Watchman: Not Found
SDKs:
  Android SDK:
    API Levels:
      - "23"
      - "28"
      - "29"
      - "30"
      - "31"
      - "33"
      - "34"
    Build Tools:
      - 30.0.2
      - 30.0.3
      - 31.0.0
      - 33.0.0
      - 33.0.1
      - 34.0.0
    System Images:
      - android-29 | Google APIs Intel x86 Atom
      - android-29 | Google Play Intel x86 Atom
      - android-30 | Google APIs Intel x86 Atom
      - android-30 | Google APIs Intel x86_64 Atom
      - android-33 | Google APIs Intel x86_64 Atom
    Android NDK: Not Found
IDEs:
  Android Studio: AI-231.9392.1.2311.11255304
Languages:
  Java:
    version: 21.0.5
    path: /usr/bin/javac
  Ruby:
    version: 3.2.3
    path: /usr/bin/ruby
npmPackages:
  "@react-native-community/cli":
    installed: 11.3.8
    wanted: 11.3.8
  react:
    installed: 18.2.0
    wanted: 18.2.0
  react-native:
    installed: 0.74.2
    wanted: 0.74.2
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: true
iOS:
  hermesEnabled: true
  newArchEnabled: false
```

### Sentry.init()

```
            dsn: "....",
			environment: __DEV__ ? "staging" : "production",
[...]

PR #4240, opened 2024-11-06
feat(core): Exclude Dev Server and Sentry Dsn request from Breadcrumbs

antonis:
## :loudspeaker: Type of change

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Filters out DevServer and DSN related breadcrumbs using the `beforeBreadcrumb` option on [iOS](https://docs.sentry.io/platforms/apple/enriching-events/breadcrumbs/#customize-breadcrumbs), [Android](https://docs.sentry.io/platforms/java/enriching-events/breadcrumbs/#customize-breadcrumbs) and [JS](https://docs.sentry.io/platforms/javascript/enriching-events/breadcrumbs/#customize-breadcrumbs):
- [Passes development server url as an option to the native sdks](https://github.com/getsentry/sentry-react-native/pull/4240/commits/3811047f24770a0e9ebed5ea7459a6d4886c48a1)
- [Filter out Dev Server and Sentry Dsn breadcrumbs on Android](https://github.com/getsentry/sentry-react-native/pull/4240/commits/c3da88b56b97fd7e3241826f9f01ad8ba7c34cfe)
- [Filter out Dev Server and Sentry Dsn breadcrumbs on iOS](https://github.com/getsentry/sentry-react-native/pull/4240/commits/d67de6cea5239c4ee79777d6014a47f17cb050a1)
- [Filter out Dev Server and Sentry Dsn breadcrumbs on JS](https://github.com/getsentry/sentry-react-native/pull/4240/commits/6ddc665e5755c42f5eeb7ae52a46d71ac1e94587)

_Note: The current implementation excludes breadcrumbs with the dev server including the port. For example `http://localhost:8081` breadcrumbs are excluded but `http://localhost:8969/stream` are not. A possible enhancement would be to [get the host name from the dev server url](https://github.com/getsentry/sentry-react-native/blob/main/packages/core/src/js/integrations/spotlight.ts#L102) and exclude all requests. This might have the side effect of missing other requests from a local development environment._

_⚠️ This PR was initially based on https://github.com/getsentry/sentry-react-native/pull/4124 to avoid conflicts. Since the implementation [changed direction](https://github.com/getsentry/sentry-react-native/pull/4240#discussion_r1832355293) this is no longer needed._

## :bulb: Motivation and Context

Fixes https://github.com/getsentry/sentry-react-native/issues/3105

## :green_heart: How did you test it?
CI, Manual testing

[...]

PR #4395, opened 2024-12-19
fix(core): Use makeDsn from core to extract the URL from DSN for filtering breadcrumbs

antonis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Uses `makeDsn` from `sentry/core` to extract the URL from DSN for filtering breadcrumbs instead of relying on the implementation of the URL protocol https://github.com/getsentry/sentry-react-native/pull/4240 that may result in errors like the following on some environments:
```
Failed to extract url from DSN [Error: URL.protocol is not implemented] [Component Stack]
```

See https://github.com/getsentry/sentry-react-native/issues/4375#issuecomment-2553285614

Manual testing, Existing unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


packages/core/src/js/sdk.tsx
```diff
@@ -67,8 +67,15 @@ export function init(passedOptions: ReactNativeOptions): void {
       return undefined;
     }
     try {
-      const url = new URL(dsn);
-      return `${url.protocol}//${url.host}`;
+      const regex = /^(https?):\/\/(?:[^@]+@)?([^/]+)(?:\/.*)?$/;
+      const matches = dsn.match(regex);
+
+      if (matches) {
+        const [, protocol, host] = matches;
```
krystofwoldrich, line 74: I've missed that in the original review as well, but let's use the `makeDsn` function from `sentry/core`. It already uses regex and is used in the client impl, so it 100% works with RN.

https://github.com/getsentry/sentry-javascript/blob/6f56d063e479f90ce07a5ddc0636dc0bc3b3bb2e/packages/core/src/utils-hoist/dsn.ts#L122

krystofwoldrich:
Just for completeness this is the default RN implementation.

https://github.com/facebook/react-native/blob/51ffc5cc12fb476a16af30d9aaadd34a6b625439/packages/react-native/Libraries/Blob/URL.js#L142

packages/core/src/js/sdk.tsx
antonis: Good idea @krystofwoldrich 👍 
Updated with a1ecb49a4165bed7106f84327303529d844b57f1

packages/core/src/js/sdk.tsx
```diff
@@ -66,13 +66,13 @@ export function init(passedOptions: ReactNativeOptions): void {
...
-      const url = new URL(dsn);
-      return `${url.protocol}//${url.host}`;
-    } catch (e) {
-      logger.error('Failed to extract url from DSN', e);
+    const dsnComponents = makeDsn(dsn);
+    if (!dsnComponents) {
+      logger.error('Failed to extract url from DSN: ', dsn);
       return undefined;
     }
+    const port = dsnComponents.port ? `:${dsnComponents.port}` : '';
```
lucas-zimerman, line 74: This DSN return is used on the following code logic:
```typescript
(dsn && url.startsWith(dsn)))) {
      return null;
    }
```

Q: Isn't the port here redundant?  Since we check the initial DSN path, it should filter if we set nor not the port, making the port parameter redundant.

I: Consider removing the port parameter, unless it's really required.

krystofwoldrich:
Looks good, just a changelog entry is missing.

antonis:
Thank you for reviewing @krystofwoldrich 🙇 
Added a changelog with 32943e9c7800e04effaad22da2bbbea70c395c43

krystofwoldrich (approved):
LGTM 🚀

Issue #4418, opened 2025-01-07
ci: metrics (new, android) fails


antonis:
### Description

[`metrics (new, android)`](https://github.com/getsentry/sentry-react-native/actions/runs/12628837574/job/35195892800) fails on CI consistently due to undeclared `uint8_t` in `hermes/Support/SHA1.h`

```
[ 53%] Building CXX object lib/Support/CMakeFiles/hermesSupport.dir/SHA1.cpp.o
In file included from /home/runner/work/sentry-react-native/sentry-react-native/performance-tests/TestAppSentry/node_modules/react-native/sdks/hermes/lib/Support/SHA1.cpp:8:
/home/runner/work/sentry-react-native/sentry-react-native/performance-tests/TestAppSentry/node_modules/react-native/sdks/hermes/include/hermes/Support/SHA1.h:17:25: error: ‘uint8_t’ was not declared in this scope
   17 | using SHA1 = std::array<uint8_t, SHA1_NUM_BYTES>;
      |                         ^~~~~~~
/home/runner/work/sentry-react-native/sentry-react-native/performance-tests/TestAppSentry/node_modules/react-native/sdks/hermes/include/hermes/Support/SHA1.h:13:1: note: ‘uint8_t’ is defined in header ‘<cstdint>’; did you forget to ‘#include <cstdint>’?
[ 53%] Building CXX object lib/CMakeFiles/hermesOptimizer.dir/Optimizer/Scalar/InstructionEscapeAnalysis.cpp.o
[ 56%] Building CXX object lib/CMakeFiles/hermesOptimizer.dir/Optimizer/Scalar/TDZDedup.cpp.o
   12 | #include <string>
  +++ |+#include <cstdint>
   13 | 
/home/runner/work/sentry-react-native/sentry-react-native/performance-tests/TestAppSentry/node_modules/react-native/sdks/hermes/include/hermes/Support/SHA1.h:17:48: error: template argument 1 is invalid
   17 | using SHA1 = std::array<uint8_t, SHA1_NUM_BYTES>;
      |                                                ^
/home/runner/work/sentry-react-native/sentry-react-native/performance-tests/TestAppSentry/node_modules/react-native/sdks/hermes/include/hermes/Support/SHA1.h:19:32: error: ‘SHA1’ does not name a type
   19 | std::string hashAsString(const SHA1 &hash);
      |                                ^~~~
[...]

PR #4423, opened 2025-01-08
fix(react-navigation): Make `navigation.processing` span title more explicit

krystofwoldrich:
Based on feedback (in Slack) making the span title more expensive exactly describing what is being measured.


antonis (approved):
LGTM 🚀

antonis:
The failing `End-to-End Tests / metrics (new, android)` check is unrelated with this PR. See https://github.com/getsentry/sentry-react-native/issues/4418

PR #4424, opened 2025-01-08
chore(ci): Pin the CI Environment to Ubuntu 22.04 for metrics android

antonis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Pins the CI Environment to Ubuntu 22.04 for metrics Android to avoid the CI failure as described in https://github.com/getsentry/sentry-react-native/issues/4418#issuecomment-2577512319

Fix https://github.com/getsentry/sentry-react-native/issues/4418

CI

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- [Upgrade the performance test apps](https://github.com/getsentry/sentry-react-native/issues/4426) (TestAppPlain/TestAppSentry) on the latest RN version so that they run with latest Ubuntu.

#skip-changelog

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -43,7 +43,7 @@ jobs:
             name: iOS
             appPlain: performance-tests/test-app-plain.ipa
           - platform: android
-            runs-on: ubuntu-latest
+            runs-on: ubuntu-22.04
             name: Android
             appPlain: performance-tests/TestAppPlain/android/app/build/outputs/apk/release/app-release.apk
     steps:
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -43,6 +43,7 @@ jobs:
             name: iOS
             appPlain: performance-tests/test-app-plain.ipa
           - platform: android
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -43,7 +43,7 @@ jobs:
             name: iOS
             appPlain: performance-tests/test-app-plain.ipa
           - platform: android
-            runs-on: ubuntu-latest
+            runs-on: ubuntu-22.04
```
lucas-zimerman, line 46: I would add a comment linking to the reason why we did this so in the future we have a reference on what to look for when updating the runner

```suggestion
            # Not using the latest version due to a known issue: https://github.com/getsentry/sentry-react-native/issues/4418
            runs-on: ubuntu-22.04
```

lucas-zimerman (approved):
just a suggestion but other than that LGTM!

.github/workflows/e2e.yml
antonis: Good idea to document this @lucas-zimerman 🙇

Issue #3608, opened 2024-02-19
Initialize native SDKs before JS


krystofwoldrich:
## Available in https://github.com/getsentry/sentry-react-native/releases/tag/6.7.0-alpha.0

Create `sentry.options.json` in the React Native project root and set options the same as you currently have in `Sentry.init` in JS.

```json
{
    "dsn": "https://key@example.io/value",
}
```

Initialize Sentry on the native layers by newly provided native methods.

```kotlin
import io.sentry.react.RNSentrySDK

class MainApplication : Application(), ReactApplication {
    override fun onCreate() {
        super.onCreate()
        RNSentrySDK.init(this)
    }
}
```

```obj-c
#import <RNSentry/RNSentry.h>

@implementation AppDelegate
- (BOOL)application:(UIApplication *)application
    didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
    [RNSentrySDK start];
    return [super application:application didFinishLaunchingWithOptions:launchOptions];
}
@end
```

### Tasks
- [x] Create RNSentry.init for Android
- [x] Create RNSentry.init for Cocoa
- [ ] Update `@sentry/wizard` to optionally use `sentry.options.json`
- [ ] Add new manual init docs
- [ ] Add JSON file schema
- [ ] Add E2E tests using manual init
- [ ] Update Sample App to use manual init by default

### Native RNSentry.init tasks
- [x] Drop JS Errors https://github.com/getsentry/sentry-react-native/blob/942161665edbac984baf81bc4e1ec289ebb0e58c/ios/RNSentry.mm#L127
- [x] Add event origin and environment https://github.com/getsentry/sentry-react-native/blob/942161665edbac984baf81bc4e1ec289ebb0e58c/ios/RNSentry.mm#L177

realkosty:
@krystofwoldrich customer is asking: 

> do you have any timeline/estimate around when this issue could be fixed?

krystofwoldrich:
@realkosty We don't have a timeline yet. But it's possible to capture the startup crashes when initializing the native SDKs manually. The `How to set up for start up crashes at the moment` part of the issue.

kahest:
@krystofwoldrich we should also update at least one sample app to use this IMO :) (may be part of the e2e test task above, so I didn't add an item, please go ahead if you want to track it separately)

PR #4443, opened 2025-01-14
chore: Initialize `RNSentryTimeToDisplay` during native module `init` on iOS

krystofwoldrich:
Part of https://github.com/getsentry/sentry-react-native/issues/3608

This PR moved the `RNSentryTimeToDisplay` init to the module init, so the time to display functionality is not dependent on the auto init from JS.


antonis (approved):
The changes LGTM 🚀

PR #4451, opened 2025-01-15
feat(experimental): Initialize Android SDK from json configuration

antonis:
## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

**Depends on https://github.com/getsentry/sentry-react-native/pull/4445 and https://github.com/getsentry/sentry-react-native/pull/4479**

## :scroll: Description

Adds the following methods that allow initialisation of Sentry Android in RN applications:
- `RNSentrySDK.init(Context context)`: Initialises using the `sentry.options.json` file located in the root of the RN project.
- `RNSentrySDK.init(Context context, Sentry.OptionsConfiguration<SentryAndroidOptions> configuration)`: Initialises with the passed `configuration` and the `sentry.options.json` if it exists.

## :bulb: Motivation and Context

Part of https://github.com/getsentry/sentry-react-native/issues/3608

## :green_heart: How did you test it?
CI, Manual testing

## :pencil: Checklist

- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

PR #4445, opened 2025-01-14
ref(android): Extracts Android native initialization to standalone structures

antonis:
## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Extracts Android native initialization to standalone structures

### ⛓️ PR Chain
- (this PR) https://github.com/getsentry/sentry-react-native/pull/4445
- https://github.com/getsentry/sentry-react-native/pull/4451
- https://github.com/getsentry/sentry-react-native/pull/4469
- https://github.com/getsentry/sentry-react-native/pull/4470
- https://github.com/getsentry/sentry-react-native/pull/4474

## :bulb: Motivation and Context

Part of https://github.com/getsentry/sentry-react-native/issues/3608

## :green_heart: How did you test it?
Manual, CI

## :pencil: Checklist

- [ ] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

PR #4469, opened 2025-01-21
Ref: Adds a Java only implementation in RNSentryMapConverter.convertToWritable

antonis:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring

**Based on https://github.com/getsentry/sentry-react-native/pull/4451**

## :scroll: Description

Refactors RNSentryMapConverter.convertToWritable and adds a Java only implementation

### ⛓️ PR Chain
- https://github.com/getsentry/sentry-react-native/pull/4445
- https://github.com/getsentry/sentry-react-native/pull/4451
- (this PR) https://github.com/getsentry/sentry-react-native/pull/4469
- https://github.com/getsentry/sentry-react-native/pull/4470
- https://github.com/getsentry/sentry-react-native/pull/4474

See https://github.com/getsentry/sentry-react-native/pull/4451#discussion_r1922630314
Part of https://github.com/getsentry/sentry-react-native/issues/3608

CI, Manual

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/packages/core/android/src/main/java/io/sentry/react/RNSentryMapConverter.java
+++ b/packages/core/android/src/main/java/io/sentry/react/RNSentryMapConverter.java
@@ -15,6 +15,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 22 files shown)


packages/core/android/src/main/java/io/sentry/react/RNSentryMapConverter.java
```diff
@@ -28,21 +30,85 @@ private RNSentryMapConverter() {
...
+  public interface WritableArrayCreator {
+    WritableArray create();
+  }
+
+  public interface WritableMapCreator {
+    WritableMap create();
+  }
+
+  public static class WritableJavaArrayCreator implements WritableArrayCreator {
+    @Override
```
krystofwoldrich, line 42: Is the annotation needed? We don't override already implemented method here.

packages/core/android/src/main/java/io/sentry/react/RNSentryMapConverter.java
```diff
@@ -139,19 +205,7 @@ private static void addValueToWritableMap(WritableMap writableMap, String key, O
...
-    // We are not directly using `convertToWritable` since `Arguments.createArray()`
-    // fails before bridge initialisation
-    Object[] keysAndValues = new Object[map.size() * 2];
-    int index = 0;
-    for (Map.Entry<String, Object> entry : map.entrySet()) {
-      keysAndValues[index++] = entry.getKey();
-      keysAndValues[index++] = entry.getValue();
-    }
-    return JavaOnlyMap.of(keysAndValues);
+    return (WritableMap) convertToJavaWritable(map);
```
krystofwoldrich, line 208: I see that base on the current code the return value can't be null, but rather than forcing the compiler to the non-null value, let's check the output of `convertToJavaWritable` and if null we can return new empty `JavaReadableMap`.

krystofwoldrich:
I realized we haven't done it in the past, but we should mark all the publics with `@internal`

packages/core/android/src/main/java/io/sentry/react/RNSentryMapConverter.java
antonis: Technically it is not needed but it is usually recommended ([example style guide](https://google.github.io/styleguide/javaguide.html#s6.1-override-annotation)) to ensure that the method correctly matches the method signature in the interface.

antonis:
> I realized we haven't done it in the past, but we should mark all the publics with @internal

@krystofwoldrich do you mean a javadoc tag or a custom annotation?
I will also make sure to lower the access level of classes/methods were this is possible.

krystofwoldrich:
I meant access level if possible if not annotation.

antonis:
Closing in favour of a simpler implementation in https://github.com/getsentry/sentry-react-native/pull/4479

PR #4482, opened 2025-01-23
fix(samples): Only load browser replay when Expo sample is running on web

krystofwoldrich:
Small fix as browser replay causes a crash due to https://github.com/getsentry/sentry-javascript/blob/de2c1ad309b850c1254c69890c96e869e297fa4c/packages/replay-internal/src/replay.ts#L764

#skip-changelog


antonis (approved):
LGTM 🚀

Issue #4302, opened 2024-11-22
[EPIC] Feedback Widget for React Native


krystofwoldrich:
Sentry currently offers only Feedback API for React Native applications and the widget to collect feedback needs to be implemented by the app developers. The goal of this task is to provide pre-made widget, which can be directly used in React Native applications.

![](https://docs.sentry.io/mdx-images/user-feedback-widget-F3HHPXRV.png)

[Current browser widget docs.](https://docs.sentry.io/platforms/javascript/user-feedback/) 
[Current React Native Feedback api docs.](https://docs.sentry.io/platforms/react-native/user-feedback/)
[Example React Native implementation in Sentry Demo app](https://github.com/sentry-demos/sentry_react_native/pull/72)

### [Feature Branch /feedback-ui](https://github.com/getsentry/sentry-react-native/pull/4435)
- https://github.com/getsentry/sentry-react-native/pull/4435

### [Feature Branch /feedback-ui-2](https://github.com/getsentry/sentry-react-native/pull/4726)
 - https://github.com/getsentry/sentry-react-native/pull/4726

### Currently in progress for GA release
- [x] Support themes (light/dark mode) https://github.com/getsentry/sentry-react-native/pull/4677
- [x] Take screenshot button in the UI https://github.com/getsentry/sentry-react-native/pull/4714
- [x] Review Accessibility - VoiceOver (iOS) and TalkBack (Android) https://github.com/getsentry/sentry-react-native/issues/4302#issuecomment-2788927285
- [x] Add feedback widget to wizard (ask user if they would like to setup Feedback widget) https://github.com/getsentry/sentry-wizard/pull/969
- [x] Add Feedback Auto-Inject E2E test (maestro flow) https://github.com/getsentry/sentry-react-native/pull/4604

### Post GA Features
- [ ] https://github.com/getsentry/sentry-react-native/issues/4735
- [ ] https://github.com/getsentry/sentry-react-native/issues/4736
- [ ] #4737

### Done
- [x] https://github.com/getsentry/sentry-react-native/pull/4378
- [x] https://github.com/getsentry/sentry-docs/pull/12503
- [x] Fix Expo Web issues https://github.com/getsentry/sentry-react-native/pull/4558
- [x] https://github.com/getsentry/sentry-docs/pull/11980
- [x] https://github.com/getsentry/sentry-react-native/pull/4524
- [x] https://github.com/getsentry/sentry-react-native/issues/3859
[...]

PR #4483, opened 2025-01-23
Autoinject feedback widget

antonis:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Adds `Sentry.showFeedbackForm()` method that presents the feedback form in a `Modal`.

Part of https://github.com/getsentry/sentry-react-native/issues/4302

Manual, CI

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


samples/react-native/src/App.tsx
```diff
@@ -318,7 +318,6 @@ function App() {
   return (
     <>
       <RootNavigationContainer />
-      <RunningIndicator />
```
antonis, line 321: Hey @krystofwoldrich 👋 

though this is still draft I'd like your feedback since I'm facing an issue with the sample app when the `RunningIndicator` is rendered. I guess this might be too specific to our sample but I was wondering if you have any ideas on how to overcome this since similar issues might occur for other apps too.

When the running indicator is removed (as in the current state of this PR) the form is showing/hiding as expected. When the indicator is added back the `showFeedbackForm` does nothing 🤔

krystofwoldrich: I've noticed there is open issue in the Reanimated repository, with a similar issue.

The recommendation is to wrap the modal in a view. 

I've tried that and it works for our case as well.

https://github.com/software-mansion/react-native-reanimated/issues/6035#issuecomment-2579469189

packages/core/src/js/feedback/FeedbackFormManager.tsx
```diff
@@ -0,0 +1,92 @@
...
+        {this.props.children}
+        {isVisible && (
+          <Modal visible={isVisible} transparent animationType="slide">
+            <View style={styles.modalBackground}>
+              <FeedbackForm
+                onFormClose={this._handleClose}
+                onFormSubmitted={this._handleClose}
+              />
+            </View>
+          </Modal>
```
krystofwoldrich, line 64: Just wrapping this in `<View></View>` no styles nothing fixes the issue.

packages/core/src/js/feedback/FeedbackFormManager.tsx
```diff
@@ -0,0 +1,92 @@
...
+   * Renders the feedback form modal.
+   */
+  public render(): React.ReactNode {
+    const { isVisible } = this.state;
+
+    return (
+      <>
+        {this.props.children}
+        {isVisible && (
+          <Modal visible={isVisible} transparent animationType="slide">
```
krystofwoldrich, line 57: I think on Android it make sense to enable close by the back button by default.
 
````suggestion
            <Modal visible={isVisible} transparent animationType="slide" onRequestClose={this._handleClose}>
````

File: packages/core/src/js/feedback/FeedbackFormManager.tsx [hunk 4]
```diff
@@ -0,0 +1,92 @@
...
+
+const showFeedbackForm = (): void => {
+  FeedbackFormManager.show();
+};
+
+const styles = StyleSheet.create({
+  modalBackground: {
+    flex: 1,
+    justifyContent: 'center',
+    backgroundColor: 'rgba(0,0,0,0.5)',
```
krystofwoldrich, line 88: Is this background ever visible?

krystofwoldrich:
I know this is still a draft, but after the modal is working let's add a test for the public function to ensure calling it displays the form when provider is used and doesn't throw error when provider is not used.

krystofwoldrich:
We should also enable users to change the forms options/props. Similar to JS setting them in the integration options `feedbackIntegration({})`. But that can be a follow up PR to keep this consise.

krystofwoldrich:
Modal is implemented in the legacy architecture for all versions the SDK supports.

But for new architecture only for RN 0.71 and up -> https://github.com/facebook/react-native/issues/33652#issuecomment-1328217904

I think we could print error when we notice RN 0.70 and older and Fabric renderer (new arch).

I'm also adding this comment so we include this information in the docs.

samples/react-native/src/App.tsx
antonis: Thank you for all your help on this Krystof 🙇 
Wrapping the modal in a `View` [as suggested](https://github.com/getsentry/sentry-react-native/pull/4483#discussion_r1929211735) worked perfectly :)

packages/core/src/js/feedback/FeedbackFormManager.tsx
antonis: This works as expected on both platforms :) Thank you Krystof 🙇 
Updated with 011fea7fafab764ac49e240a7c21cbffb541da97

packages/core/src/js/feedback/FeedbackFormManager.tsx
antonis: Good catch Krystof. Updated with a4962b9694dadc50ee6ca24e49440e652ee20a45

packages/core/src/js/feedback/FeedbackFormManager.tsx
antonis: Not really since the modal is fullscreen. I've removed the background color and moved the modalBackground along with the other styles with 2d4abb37a3a0612d740b98d75539adf3184cc7d2

antonis:
Thank you for bringing this to my attention @krystofwoldrich 🙇 

> I think we could print error when we notice RN 0.70 and older and Fabric renderer (new arch).

I've added a check with 84108f151b0e4dcfe8c24ca402f98cc09086020a

antonis:
Thank you for your help and feedback @krystofwoldrich 🙇 

> I know this is still a draft, but after the modal is working let's add a test for the public function to ensure calling it displays the form when provider is used and doesn't throw error when provider is not used.

Added tests with f4df57c2c52ac2c6709863ad4fab39e0fecaec44

> We should also enable users to change the forms options/props. Similar to JS setting them in the integration options feedbackIntegration({}). But that can be a follow up PR to keep this consise.

Makes sense 👍 I'll follow up with a separate PR for this.

packages/core/src/js/feedback/utils.ts
```diff
@@ -1,3 +1,18 @@
+import { isFabricEnabled } from '../utils/environment';
+import { ReactNativeLibraries } from './../utils/rnlibraries';
+
+/**
+ * Modal is not supported in React Native < 0.71 with Fabric renderer.
+ * ref: https://github.com/facebook/react-native/issues/33652
+ */
+export function isModalSupported(): boolean {
+  const rnVersion = ReactNativeLibraries.ReactNativeVersion?.version;
```
lucas-zimerman, line 9: i: We could get the versions on separated variables here.
```suggestion
  const { major, minor } = ReactNativeLibraries.ReactNativeVersion?.version || {};
```

lucas-zimerman (approved):
Overall the PR Looks good! Added a few suggestions but they are non blocking

packages/core/src/js/feedback/utils.ts
antonis: Thank you for the suggestion Lucas 🙇

packages/core/src/js/feedback/utils.ts
```diff
@@ -1,3 +1,18 @@
...
+  if (isFabricEnabled() && rnVersion && rnVersion.major === 0 && rnVersion.minor < 71) {
```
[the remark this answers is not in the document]
antonis: Thank you for the suggestions Lucas bow 🙇‍♂️ 
Closing this one as I committed the one above.

packages/core/test/feedback/FeedbackFormManager.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+  isModalSupported: jest.fn(),
+}));
+
+beforeEach(() => {
+  logger.error = jest.fn();
+});
+
+describe('FeedbackFormManager', () => {
+  it('showFeedbackForm displays the form when FeedbackFormProvider is used', () => {
+    require('../../src/js/feedback/utils').isModalSupported.mockReturnValue(true);
```
krystofwoldrich, line 18: l: `import { isModalSupported } from ''../../src/js/feedback/utils'';` and then `(isModalSupported as jest.MockedFunction<typeof isModalSupported>)` should also work and it will be typesafe.

krystofwoldrich (approved):
Thank you. Looks good.

I left one small TS improvement comment in the test file.

packages/core/test/feedback/FeedbackFormManager.test.tsx
antonis: Good idea Krystof 👍 
Updated with d8ab9146f4fc130dd1a5b5c1c46af6fc8b98a831

PR #4492, opened 2025-01-30
Feedback modal UI tweaks

antonis:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/packages/core/src/js/feedback/FeedbackForm.styles.ts
+++ b/packages/core/src/js/feedback/FeedbackForm.styles.ts
@@ -79,7 +79,21 @@ const defaultStyles: FeedbackFormStyles = {
   },
   modalBackground: {
     flex: 1,
-    justifyContent: 'center',
+    justifyContent: 'flex-end',
+    backgroundColor: 'rgba(0, 0, 0, 0.8)',
+  },
+  modalSheetContainer: {
+    backgroundColor: BACKROUND_COLOR,
+    borderTopLeftRadius: 16,
+    borderTopRightRadius: 16,
+    padding: 16,
+    alignSelf: 'stretch',
+    height: '98%',
+    shadowColor: '#000',
+    shadowOffset: { width: 0, height: -3 },
+    shadowOpacity: 0.1,
+    shadowRadius: 4,
+    elevation: 5,
   },
 };
 
--- a/packages/core/src/js/feedback/FeedbackFormManager.tsx
+++ b/packages/core/src/js/feedback/FeedbackFormManager.tsx
@@ -1,6 +1,6 @@
 import { logger } from '@sentry/core';
 import * as React from 'react';
-import { Modal, View } from 'react-native';
+import { Modal, SafeAreaView, View } from 'react-native';
 
 import { FeedbackForm } from './FeedbackForm';
 import defaultStyles from './FeedbackForm.styles';
@@ -70,12 +70,14 @@ class FeedbackFormProvider extends React.Component<FeedbackFormProviderProps> {
         {isVisible && (
           <View>
             <Modal visible={isVisible} transparent animationType="slide" onRequestClose={this._handleClose} testID="feedback-form-modal">
-              <View style={styles.modalBackground}>
-                <FeedbackForm {...getFeedbackOptions()}
-                  onFormClose={this._handleClose}
-                  onFormSubmitted={this._handleClose}
-                />
-              </View>
+              <SafeAreaView style={styles.modalBackground}>
+                <View style={styles.modalSheetContainer}>
+                  <FeedbackForm {...getFeedbackOptions()}
+                    onFormClose={this._handleClose}
+                    onFormSubmitted={this._handleClose}
+                    />
+                </View>
+              </SafeAreaView>
             </Modal>
           </View>
         )}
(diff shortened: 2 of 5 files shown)


packages/core/src/js/feedback/FeedbackForm.styles.ts
```diff
@@ -79,7 +79,22 @@ const defaultStyles: FeedbackFormStyles = {
   },
   modalBackground: {
     flex: 1,
-    justifyContent: 'center',
+    justifyContent: 'flex-end',
+    backgroundColor: 'rgba(0, 0, 0, 0.8)',
+  },
+  modalSheetContainer: {
```
antonis, line 85: Note: If we proceed with this approach the new style should be added [in the docs](https://github.com/getsentry/sentry-docs/pull/12503/files#diff-f152431fe6cca40aaef7918e3e1b44445e7b7c56506201274fa128c5a6bbf024R119)

packages/core/src/js/feedback/FeedbackForm.styles.ts
```diff
@@ -79,7 +79,22 @@ const defaultStyles: FeedbackFormStyles = {
...
+    backgroundColor: BACKROUND_COLOR,
+    borderTopLeftRadius: 16,
+    borderTopRightRadius: 16,
+    padding: 16,
+    marginBottom: -32,
```
antonis, line 90: This is a bit of a hack to cover the iOS home indicator at the bottom with the background color. It seems to work fine on different devices both on iOS and Android for the given form.

<details><summary>Screenshot</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-01-30 at 17 38 35

</details> 

I wanted to avoid importing external libraries like [react-native-safe-area-context](https://www.npmjs.com/package/react-native-safe-area-context/) and other solutions like tweaking padding didn't really work.

krystofwoldrich: Yes, make sense we have to use only the RN defaults, or we would have to include the third-party code in our lib, vendor it. But better to avoid it due to maintenance costs.

krystofwoldrich:
The auto-injected form has double the border of the manually created one. That seems to me like a bug.

krystofwoldrich:
I would like to iterate over the possible implementations of this. 

I think the shadow should not slide up with the modal, it doesn't look natural.

Could we make use of the fact we wrap the modal in view and make the wrapper view fade in out the shadow as the modal slides up?

krystofwoldrich:
I've tried the code on iOS and the form does not avoid the on-screen keyboard.

antonis:
> The auto-injected form has double the border of the manually created one. That seems to me like a bug.

Good catch 👍 
Removed the extra border 05f94f85ceaad44f88c199b3371098c3065f1842 and adjusted the height 9ecd8a20f2011b84f748694e7517611ff9e1b535 to match the manual

|Manually created form|Modal|
|---|---|
|)||

> Could we make use of the fact we wrap the modal in view and make the wrapper view fade in out the shadow as the modal slides up?

Good idea 👍 
Updated with ce1de86607d31502c42f827010b4bc9a49105da0

<details><summary>iOS Animation</summary>

https://github.com/user-attachments/assets/c55ecba5-f7c0-4379-b005-080912800076

</details> 

<details><summary>Android Animation</summary>

https://github.com/user-attachments/assets/214720e2-b8ed-4901-9960-9342a91c2f7e

</details>

lucas-zimerman (approved):
Looking good to me, and it worked fine for me on a real device
https://sentry-sdks.sentry.io/feedback/?feedbackSlug=sentry-react-native%3A6286398787&mailbox=ignored&project=5428561&referrer=feedback_list_page&statsPeriod=1h

packages/core/src/js/feedback/FeedbackFormManager.tsx
```diff
@@ -59,25 +80,34 @@ class FeedbackFormProvider extends React.Component<FeedbackFormProviderProps> {
       return <>{this.props.children}</>;
     }
 
-    const { isVisible } = this.state;
-    const styles: FeedbackFormStyles = { ...defaultStyles, ...this.props.styles };
+    const { isVisible, backgroundOpacity } = this.state;
+
+    const backgroundColor = backgroundOpacity.interpolate({
+      inputRange: [0, 1],
+      outputRange: ['rgba(0, 0, 0, 0)', 'rgba(0, 0, 0, 0.9)'],
```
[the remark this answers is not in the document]
antonis: Good question. I mainly tried to match the darker background of the [manually created form](https://github.com/getsentry/sentry-react-native/pull/4492#issuecomment-2627562230) and still keep the transparency.

krystofwoldrich:
Looks perfect, thank you!

---

Also [this comment](https://github.com/getsentry/sentry-react-native/pull/4492#issuecomment-2627562230) is great, it makes it super easy to follow the development and changes in the PR. 👏 👏 👏

krystofwoldrich (approved):

PR #4510, opened 2025-02-05
feat(init): Load options from `sentry.options.json` in JS

krystofwoldrich:
This PR adds merging of the file options loaded by the Sentry Metro Plugin and user-supplied options in the `Sentry.init`.


antonis (approved):
LGTM and works as expected 🚀 
Thank you for the detailed test cases 🙇

PR #4522, opened 2025-02-10
misc(sample): Change RN Sample to use native file init by default

krystofwoldrich:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

This PR changes the RN sample to use the new native init apis by default.

Secondary this PR adds simple build/start options to change to the auto init behavior from JS.

This will be used for testing, where we will need to test that both ways of init result in the same SDK behaviour.

sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


samples/react-native/src/utils.ts
```diff
@@ -1,7 +1,37 @@
...
+      }
+    ).SENTRY_DISABLE_NATIVE_START;
+  } else if (Platform.OS === 'ios') {
+    const args = LaunchArguments.value<{
+      sentrydisablenativestart?: boolean;
+    }>();
+    return !!args.sentrydisablenativestart;
+  } else {
+    return false;
+  }
```
lucas-zimerman, line 27: I: Since each if condition returns a value, we could use if on each case and leave return false outside an if condition, what do you think?

antonis (approved):
Thank you for adding this option in the sample @krystofwoldrich 🙇 
The code changes LGTM and I didn't notice any weird behaviour testing the sample app on iOS and Android 🚀

PR #4538, opened 2025-02-14
feat(feedback): Save form state for un-submitted data

antonis:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Saves form content when closed, failed (reopening the form keeps the content), only clear after sent

See https://github.com/getsentry/sentry-react-native/issues/4302

Manual, CI

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/packages/core/src/js/feedback/FeedbackForm.tsx
+++ b/packages/core/src/js/feedback/FeedbackForm.tsx
@@ -33,6 +33,15 @@ export class FeedbackForm extends React.Component<FeedbackFormProps, FeedbackFor
     ...defaultConfiguration
   }
 
+  private static _savedState: FeedbackFormState = {
+    isVisible: false,
+    name: '',
+    email: '',
+    description: '',
+    filename: undefined,
+    attachment: undefined,
+  };
+
   public constructor(props: FeedbackFormProps) {
     super(props);
 
@@ -45,9 +54,11 @@ export class FeedbackForm extends React.Component<FeedbackFormProps, FeedbackFor
 
     this.state = {
       isVisible: true,
-      name: currentUser.useSentryUser.name,
-      email: currentUser.useSentryUser.email,
-      description: '',
+      name: FeedbackForm._savedState.name || currentUser.useSentryUser.name,
+      email: FeedbackForm._savedState.email || currentUser.useSentryUser.email,
+      description: FeedbackForm._savedState.description || '',
+      filename: FeedbackForm._savedState.filename || undefined,
+      attachment: FeedbackForm._savedState.attachment || undefined,
     };
   }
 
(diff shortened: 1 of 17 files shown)


packages/core/src/js/feedback/FeedbackForm.tsx
```diff
@@ -33,6 +33,15 @@ export class FeedbackForm extends React.Component<FeedbackFormProps, FeedbackFor
     ...defaultConfiguration
   }
 
+  private static _savedState: FeedbackFormState = {
```
krystofwoldrich, line 36: The `isVisible` doesn't seem to be used, we can omit it like this `Omit<FeedbackFormState, 'isVisible'>`.

packages/core/src/js/feedback/FeedbackForm.tsx
```diff
@@ -254,4 +266,19 @@ export class FeedbackForm extends React.Component<FeedbackFormProps, FeedbackFor
     </SafeAreaView>
     );
   }
+
+  private _saveFormState = (): void => {
+    FeedbackForm._savedState = { ...this.state };
+  };
```
krystofwoldrich, line 272: Would it work if we only save it when closing the form instead of on every state change? Just to avoid the object recreation potentially on every keystroke when typing.

packages/core/src/js/feedback/FeedbackForm.tsx
antonis: Good point 👍 
Updated with 1efb23e290f3aeef6977e8e714d1c191a1e8342c

packages/core/src/js/feedback/FeedbackForm.tsx
antonis: That's a good idea @krystofwoldrich 👍 
Updated 59f2f570ae0e6bb6744bb59566cdc13e549add75.

packages/core/src/js/feedback/FeedbackWidget.tsx
```diff
@@ -33,6 +33,7 @@ export class FeedbackWidget extends React.Component<FeedbackWidgetProps, Feedbac
     ...defaultConfiguration
   }
 
+  private static _didSubmitForm: boolean = false;
```
krystofwoldrich, line 36: `_didSubmitForm` should not be static since it's related to the form instance.

antonis: Good catch @krystofwoldrich 🙇 
Updated with f17106e502694606c1428201f0b568706dddf15d

krystofwoldrich (approved):
Looks good. Thank you!

PR #4545, opened 2025-02-14
feat(feedback): Show selected screenshot

antonis:
## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

Based on https://github.com/getsentry/sentry-react-native/pull/4538 to avoid conflicts

## :scroll: Description

Shows selected screenshot in the UI. The screenshot was positioned at the left of the screenshot button similar to Cocoa.

|Cocoa|RN Android|RN iOS|
|---|---|---|
|cocoa|android|ios|

⚠️ The docs PR needs to be updated with the added styles https://github.com/getsentry/sentry-docs/pull/12503

## :bulb: Motivation and Context

See https://github.com/getsentry/sentry-react-native/issues/4302

## :green_heart: How did you test it?
Manual, CI

## :pencil: Checklist

- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog

PR #4546, opened 2025-02-14
feat(feedback): Use only image uri in the onAddScreenshot callback

antonis:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

Based on https://github.com/getsentry/sentry-react-native/pull/4545

## :scroll: Description

Simplifies `onAddScreenshot` callback by requiring image uri `string` instead of `UInt8Array` of for data. 

Rendering screenshots for which [we only have `Uint8Array` data](https://github.com/getsentry/sentry-react-native/pull/4545/files#diff-4ea1bc6415dbb3c3f1761351b30bfa6745197e54c6b65515392be2ca3985ac37R159) as part of https://github.com/getsentry/sentry-react-native/pull/4545 requires conversion to `base64` which adds processing overhead and is not straightforward without an external library. Given that acquiring just the the image uri should be easier for developers that want to use a custom implementation with the `onAddScreenshot` callback, it might make sense to deprecate `Uint8Array` on the api level and just use it to send the data.

⚠️ The docs PR needs to be updated with the added styles https://github.com/getsentry/sentry-docs/pull/12503

Manual

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


packages/core/src/js/feedback/FeedbackForm.types.ts
```diff
@@ -171,7 +171,7 @@ export interface FeedbackCallbacks {
   /**
    * Callback when a screenshot is added
    */
-  onAddScreenshot?: (attachFile: (filename: string, data: Uint8Array) => void) => void;
+  onAddScreenshot?: (attachFile: (filename: string, fileUri: string) => void) => void;
```
krystofwoldrich, line 174: Reflecting the recent changes, absolutely makes sense to use uri. 

I'm thinking that maybe we can simplify the callback to only accept `uri`. I don't think the file name is important, to satisfy the sentry API we can use static name like `feedback_screenshot`

````suggestion
  onAddScreenshot?: (attachFile: (uri: string) => void) => void;
````

Comment (antonis) [reply packages/core/src/js/feedback/FeedbackForm.types.ts]: Good idea @krystofwoldrich 👍 
Updated with f14197d9a352b8af3d1784abdea53f36ddd99946

File: packages/core/src/js/feedback/FeedbackWidget.types.ts [hunk 2]
```diff
@@ -171,7 +171,7 @@ export interface FeedbackCallbacks {
   /**
    * Callback when a screenshot is added
    */
-  onAddScreenshot?: (attachFile: (filename: string, fileUri: string) => void) => void;
+  onAddScreenshot?: (screenshotAdded: (uri: string) => void) => void;
```
krystofwoldrich, line 174: I think the present-tense verb was right since users will call this to add screenshot `addScreenshot('my-screenshot')`

````suggestion
  onAddScreenshot?: (addScreenshot: (uri: string) => void) => void;
````

krystofwoldrich (approved):
Just a small comment about the callback name, otherwise looks good. Thank you.

antonis:
> ⚠️ The docs PR needs to be updated with the added styles https://github.com/getsentry/sentry-docs/pull/12503

Updated with https://github.com/getsentry/sentry-docs/commit/48a79723e92ba22b14e9618e9f58de69935ec536

PR #4555, opened 2025-02-17
chore(feedback): Improve widget animations

krystofwoldrich:
This PR removes duplicate animations of the feedback widget. 

Because modal slide and animated slide, the modal would flash in an upper position when sliding out, see the video below.

https://github.com/user-attachments/assets/074a62ca-3c85-484c-865a-f5740f54a29d

After the changes, the widget plays only one slide animation and slides once. Als,o the background opacity, animates now on close. Before was canceled by unmounting the widget.

### Behaviour change

if onClose or onSucess callbacks are supplied the widget doesn't unmount users get to decide what to do with the widget. This allows us in case of the auto-injection keep the form on the modal which is sliding out instead of sliding out a blank modal.

https://github.com/user-attachments/assets/eee1aa64-8f7a-4e9a-b7f1-7754800a3f22

### DIFF
--- a/packages/core/src/js/feedback/FeedbackWidget.tsx
+++ b/packages/core/src/js/feedback/FeedbackWidget.tsx
@@ -88,7 +88,9 @@ export class FeedbackWidget extends React.Component<FeedbackWidgetProps, Feedbac
     };
 
     try {
-      this.setState({ isVisible: false });
+      if (!onFormSubmitted) {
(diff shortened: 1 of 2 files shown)


antonis (approved):
thank you for fixing this @krystofwoldrich 🙇 
The code changes LGMT and I didn't notice any side effects on my tests on Android or iOS 🚀

PR #4558, opened 2025-02-17
feat(feedback): Support web environments

antonis:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

Based on https://github.com/getsentry/sentry-react-native/pull/4546 to avoid conflicts

## :scroll: Description

This PR handles the following issues on the web:
- Adds support for attachments
- Fixes web alerts using `window.alert`
- Fixes input field focus issue

See https://github.com/getsentry/sentry-react-native/issues/4302#issuecomment-2659973105

Manual

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


packages/core/src/js/feedback/utils.ts
```diff
@@ -14,3 +21,29 @@ export const isValidEmail = (email: string): boolean => {
...
+    throw new Error('atob is not available in this environment.');
+  }
+
+  const binaryString = atob(base64);
+  return new Uint8Array([...binaryString].map(char => char.charCodeAt(0)));
+};
+
+export const feedbackAlertDialog = (title: string, message: string): void => {
+  /* eslint-disable @typescript-eslint/ban-ts-comment, no-restricted-globals, no-alert, @typescript-eslint/no-unsafe-member-access */
+  // @ts-ignore
```
antonis, line 41: I used `ts-ignore` to avoid touching the global state and compilerOptions to use `window`. I'll be happy to iterate if there is better suggestion for this 🙇

lucas-zimerman: One thing you could try is altering the worldwide.ts file at packages/core/src/js/utils/worldwride.ts

Change the ReactNativeInternalGlobal and add the following parameter:
`  alert?: (message: string) => void;`

So with that you could use the following code

```typescript
  if (isWeb() && typeof RN_GLOBAL_OBJ.alert !== 'undefined') {
    RN_GLOBAL_OBJ.alert(`${title}\n${message}`);
  } else {
```

antonis: Thank you for the suggestion @lucas-zimerman 🙇 
This actually seems to work pretty well and there is no need to suppress any typescript checks.
Updated with 28738bf40e7d3e82c2549f59d1e93120e837e230

krystofwoldrich (approved):
Thank you! Works as expected with Expo Web.

Issue #4542, opened 2025-02-14
[React-Native]: v7 Main Tasks


krystofwoldrich:
### Description

Covers the main work to be done in preparation for React Native v7.

lucas-zimerman:
@krystofwoldrich  @antonis Should we clear the deprecation/warnings on swift for V7? I noticed them when reviwing this PR

PR #4568, opened 2025-02-19
feat: Bump to Sentry Javascript V9

lucas-zimerman:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This PR bumps JavaScript to version V9 and also fixes any break changes found.
In short, normal users not experience any changes, as shown on the sample apps, there were no required changes for this bump.

Changes handled:
- [feat(core)!: Remove deprecated Request type](https://github.com/getsentry/sentry-javascript/pull/14858)
- [feat!: Remove metrics API from the JS SDK](https://github.com/getsentry/sentry-javascript/pull/14745)
- [feat!: Remove enableTracing](https://github.com/getsentry/sentry-javascript/pull/15078)
- [ref(core)!: Move shutdownTimeout option type from core to node](https://github.com/getsentry/sentry-javascript/pull/15217)
- [feat(core): Deprecate spanId on propagationContext](https://github.com/getsentry/sentry-javascript/pull/14482)
- [ref(core)!: Remove backwards compatible SentryCarrier type](https://github.com/getsentry/sentry-javascript/pull/14697)
- [feat(core): Streamline SpanJSON type](https://github.com/getsentry/sentry-javascript/pull/14693)
- [feat(core): Add inheritOrSampleWith helper to traceSampler](https://github.com/getsentry/sentry-javascript/pull/15277)

In regard to shutdownTimeout, we can keep it as an additional parameter or completely remove it if that is the plan.

Sample App, CI, Sentry.io discover

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

- [x]  Update Changelog with breakchanges.
- [ ]  Create Migration Doc.

#skip-changelog for now


lucas-zimerman:
I will wait for some information in regard to https://github.com/getsentry/sentry-javascript/issues/15471 in case it's a problem on the JavaScript SDK

krystofwoldrich:
@lucas-zimerman The `yarn.lock` seems to be completely regenerated.

How have you upgraded to the JS v9? 

We should review it and ensure only the `@sentry/*` are upgraded to avoid issues like https://github.com/getsentry/sentry-react-native/pull/4592

The `yarn.lock` change should be readable as only sentry deps should change.

lucas-zimerman:
> @lucas-zimerman The `yarn.lock` seems to be completely regenerated.
> 
> How have you upgraded to the JS v9?
> 
> We should review it and ensure only the `@sentry/*` are upgraded to avoid issues like #4592
> 
> The `yarn.lock` change should be readable as only sentry deps should change.

I fixed on https://github.com/getsentry/sentry-react-native/pull/4568/commits/028dbf97cee1e4c97aa6ed0fdfd8f056a32e3021, At the time there were a conflict with yarn.lock, and I created a new one. Next time I will get the target branch yarn.lock and install again on the PR

krystofwoldrich:
Thank you @lucas-zimerman, now the `yarn.lock` looks good.

packages/core/src/js/tracing/span.ts
```diff
@@ -104,7 +105,7 @@ export const startIdleSpan = (
     return new SentryNonRecordingSpan();
   }
 
-  getCurrentScope().setPropagationContext(generatePropagationContext());
+  getCurrentScope().setPropagationContext({ traceId: generateTraceId(), sampleRand: Math.random() });
```
lucas-zimerman, line 108: Q: Should we use Math.random() here?

packages/core/package.json
```diff
@@ -66,22 +66,21 @@
   },
   "dependencies": {
     "@sentry/babel-plugin-component-annotate": "3.1.2",
-    "@sentry/browser": "8.54.0",
-    "@sentry/cli": "2.42.1",
```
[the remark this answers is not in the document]
lucas-zimerman: Fixed on https://github.com/getsentry/sentry-react-native/pull/4568/commits/55c19617a0543b4c1ba5b979a4e68b11cfd3acd0

lucas-zimerman:
@antonis  @krystofwoldrich I'd say now it's ready for review :D

packages/core/package.json
```diff
@@ -66,22 +66,21 @@
...
     "@mswjs/interceptors": "^0.25.15",
     "@react-native/babel-preset": "0.77.1",
-    "@sentry-internal/eslint-config-sdk": "8.54.0",
-    "@sentry-internal/eslint-plugin-sdk": "8.54.0",
-    "@sentry-internal/typescript": "8.54.0",
-    "@sentry/wizard": "3.42.0",
+    "@sentry-internal/eslint-config-sdk": "9.1.0",
+    "@sentry-internal/eslint-plugin-sdk": "9.1.0",
+    "@sentry-internal/typescript": "9.1.0",
+    "@sentry/wizard": "3.40.0",
```
antonis, line 83: Should we keep `3.42.0` from the base branch?

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -155,7 +155,8 @@ function setupErrorUtilsGlobalHandler(): void {
       return;
     }
 
-    void client.flush(client.getOptions().shutdownTimeout || 2000).then(
+    // shutdownTimeout was removed: https://github.com/getsentry/sentry-javascript/pull/15217
```
antonis, line 158: l: We can probably also remove `shutdownTimeout` from [RNSentryModuleImpl.java](https://github.com/getsentry/sentry-react-native/pull/4566/files#diff-fe8d5afb81105308bdadba53d94f67ee2b7142f0f003d85a44dcd8d65ee2e2e5)

packages/core/src/js/tracing/span.ts
antonis: I think this should work given how `sampleRand` [is defined](https://github.com/getsentry/sentry-javascript/blob/317782c38206b775fcc5cdf42a2fc8c30f52c9ac/packages/core/src/types-hoist/tracing.ts#L23). It is [also used on the javascript sdk](https://github.com/search?q=repo%3Agetsentry%2Fsentry-javascript%20%22sampleRand%3A%20Math.random()%22&type=code).

packages/core/package.json
lucas-zimerman: sounds  good to me! fixed on https://github.com/getsentry/sentry-react-native/pull/4568/commits/9a0b21ecf6b7f52098035b36837d41296502532f

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
lucas-zimerman: Nice catch! fixed on https://github.com/getsentry/sentry-react-native/pull/4568/commits/9a0b21ecf6b7f52098035b36837d41296502532f

CHANGELOG.md
```diff
@@ -44,6 +47,18 @@
   - [changelog](https://github.com/getsentry/sentry-cli/blob/master/CHANGELOG.md#2421)
   - [diff](https://github.com/getsentry/sentry-cli/compare/2.41.1...2.42.1)
 
+
+### Major Changes
+
+- Remove autoSessionTracking option
```
krystofwoldrich, line 53: The Browser SDK now recommends directly using https://github.com/getsentry/sentry-javascript/blob/defeaaf99ff6180be714ae770ce1ed0da6243676/packages/browser/src/integrations/browsersession.ts#L12

RN SDK has also `enableAutoSessionTracking` https://github.com/getsentry/sentry-react-native/blob/f8289de4acf2755f5a9180dd33772fc38d59c7ea/packages/core/src/js/options.ts#L45

`autoSessionTracking` and `enableAutoSessionTracking` were two unrelated flags.

We should make it clear to the users.

CHANGELOG.md
```diff
@@ -44,6 +47,18 @@
...
+- Remove `enableTracing` Instead, set  tracesSampleRate: 0 to disable, or a higher value to enable it.
```
krystofwoldrich, line 54: Let's explain here the difference between `tracesSampleRate: undefined` and `0`.

Because `0` means the integrations are enabled and spans are created, but not sampled and `undefined` means performance integrations are disabled and no spans are created.

krystofwoldrich: From the browser changelog: https://github.com/getsentry/sentry-javascript/blob/defeaaf99ff6180be714ae770ce1ed0da6243676/CHANGELOG.md#L1667-L1672

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -239,9 +239,6 @@ protected void getSentryAndroidOptions(
     if (rnOptions.hasKey("sessionTrackingIntervalMillis")) {
       options.setSessionTrackingIntervalMillis(rnOptions.getInt("sessionTrackingIntervalMillis"));
     }
-    if (rnOptions.hasKey("shutdownTimeout")) {
-      options.setShutdownTimeoutMillis(rnOptions.getInt("shutdownTimeout"));
-    }
```
krystofwoldrich, line 244: Let's define the `shutdownTimeout` option in the RN TS Options since it's still relevant for closing the native SDKs.

packages/core/package.json
```diff
@@ -100,7 +99,7 @@
     "eslint-plugin-react-native": "^3.8.1",
     "expo": "^52.0.0",
     "expo-module-scripts": "3.1.0",
-    "jest": "^29.6.2",
+    "jest": "^29.7.0",
```
krystofwoldrich, line 102: L: Were there some issues with the older jest and the new JS SDK? We can keep it, I'm just curious.

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
krystofwoldrich: I think we should move the timeout type to RN since the native SDKs still use it.

https://github.com/getsentry/sentry-react-native/pull/4568/files#r1992043223

packages/core/src/js/sdk.tsx
```diff
@@ -157,6 +157,7 @@ export function wrap<P extends Record<string, unknown>>(
   const profilerProps = {
     ...(options?.profilerProps ?? {}),
     name: RootComponent.displayName ?? 'Root',
+    updateProps: {}
```
krystofwoldrich, line 160: Is this required? I don't know the prop.

packages/core/test/feedback.test.ts
```diff
@@ -261,10 +261,13 @@ describe('captureFeedback', () => {
 
     expect(typeof eventId).toBe('string');
 
-    expect(mockTransport).toHaveBeenCalledWith([
+    const [[envelope]] = mockTransport.mock.calls;
```
krystofwoldrich, line 264: I don't understand the change, the original code should do the same check.

````suggestion
    expect(mockTransport).toHaveBeenCalledWith([
````

File: packages/core/test/feedback.test.ts [hunk 11]
```diff
@@ -322,12 +326,15 @@ describe('captureFeedback', () => {
...
     expect(span).toBeDefined();
 
-    const { spanId, traceId } = span!.spanContext();
+    const traceId = span!.spanContext().traceId;
 
-    expect(mockTransport).toHaveBeenCalledWith([
+    const [[envelope]] = mockTransport.mock.calls;
+    expect(mockTransport).toHaveBeenCalledWith(envelope);
+
+    expect(envelope).toEqual([
```
krystofwoldrich, line 334: The same as above, I think this change is not necessary.

packages/core/test/tracing/addTracingExtensions.test.ts
```diff
@@ -70,10 +70,5 @@ describe('Tracing extensions', () => {
         }),
       }),
     );
-    expect(spanToJSON(span!)).toEqual(
-      expect.objectContaining({
-        parent_span_id: transaction!.spanContext().spanId,
-      }),
-    );
```
krystofwoldrich, line 77: We should still check that the span has the correct parent.

````suggestion
    expect(spanToJSON(span!)).toEqual(
      expect.objectContaining({
        parent_span_id: spanToJSON(transaction!).span_id,
      }),
    );
````

File: packages/core/test/tracing/timetodisplaynative.web.test.tsx [hunk 13]
```diff
@@ -1,9 +1,3 @@
-jest.mock('react-native', () => {
-  const RN = jest.requireActual('react-native');
-  RN.UIManager = {};
-  return RN;
-});
-
```
krystofwoldrich, line 6: Was it just not needed or does it break the test after the JS SDK upgrade?

packages/core/test/utils/safe.test.ts
```diff
@@ -38,14 +38,30 @@ describe('safe', () => {
...
+        inheritOrSampleWith: function (fallbackSampleRate: number): number {
+          return fallbackSampleRate;
+        },
+      });
       expect(mockFn).toBeCalledTimes(1);
-      expect(mockFn).toBeCalledWith({ name: 'foo', transactionContext: { name: 'foo' } });
+      expect(mockFn).toBeCalledWith({
+        name: 'foo',
+        transactionContext: { name: 'foo' },
+        inheritOrSampleWith: expect.any(Function),
```
krystofwoldrich, line 52: Let's check that it's the same function passed to the original. That will be more robust than any function.

packages/core/test/tracing/integrations/userInteraction.test.ts
```diff
@@ -255,15 +255,20 @@ describe('User Interaction Tracing', () => {
     });
 
     test('do not start UI event transaction if active transaction on scope', () => {
-      const activeTransaction = startSpanManual(
-        { name: 'activeTransactionOnScope', scope: getCurrentScope() },
-        (span: Span) => span,
-      );
-      expect(activeTransaction).toBeDefined();
-      expect(activeTransaction).toBe(getActiveSpan());
```
krystofwoldrich, line 263: Was there a change in how `startSpanManual` works? 

Was `expect(activeTransaction).toBe(getActiveSpan());` failing? 

I'm trying to understand why the change was needed.

krystofwoldrich:
Overall the PR looks good. Thank you for all the changes.

I have just a few suggestions and some questions to understand the v9 changes.

packages/core/test/tracing/timetodisplaynative.web.test.tsx
lucas-zimerman: The test was failing with the mock
```
       FAIL  test/tracing/timetodisplaynative.web.test.tsx
        ● Test suite failed to run
      
          TypeError: Cannot set property UIManager of #<Object> which has only a getter
      
            1 | jest.mock('react-native', () => {
            2 |   const RN = jest.requireActual('react-native');
          > 3 |   RN.UIManager = {};
              |               ^
            4 |   return RN;
            5 | });
            6 | jest.mock('../../src/js/utils/rnlibraries', () => {
      
            at test/tracing/timetodisplaynative.web.test.tsx:3:15
            at Object.require (src/js/tracing/timetodisplaynative.tsx:3:1)
            at Object.<anonymous> (test/tracing/timetodisplaynative.web.test.tsx:13:1)
```
Since removing the mock made the error to not happen again I opted to not mock it.

packages/core/src/js/sdk.tsx
lucas-zimerman: It was always required, but for for some reason it only started to complain once I started working on V9
https://github.com/getsentry/sentry-javascript/blob/6914a2453472b3683fe76e67e248414d7ec4ec06/packages/react/src/profiler.tsx#L24

packages/core/test/feedback.test.ts
lucas-zimerman: True, it wasn't before, but now it is, so I roll back the original implementation here 
  https://github.com/getsentry/sentry-react-native/pull/4568/commits/45ee6fadd66e5d745627d4eccdbab5698cbfb0c7

packages/core/test/feedback.test.ts
lucas-zimerman: True, it wasn't before, but now it is, so I roll back the original implementation here 
  https://github.com/getsentry/sentry-react-native/pull/4568/commits/45ee6fadd66e5d745627d4eccdbab5698cbfb0c7

packages/core/package.json
lucas-zimerman: There were, but not anymore, so I am returning to the older version

packages/core/test/tracing/integrations/userInteraction.test.ts
lucas-zimerman: the startSpan behavior has changed https://github.com/getsentry/sentry-javascript/blob/develop/MIGRATION.md#sentrycore--all-sdks

```
The startSpan behavior was changed if you pass a custom scope: While in v8, the passed scope was set active directly on the passed scope, in v9, the scope is cloned. This behavior change does not apply to @sentry/node where the scope was already cloned. This change was made to ensure that the span only remains active within the callback and to align behavior between @sentry/node and all other SDKs. As a result of the change, span hierarchy should be more accurate. However, modifying the scope (for example, setting tags) within the startSpan callback behaves a bit differently now.

startSpan({ name: 'example', scope: customScope }, () => {
  getCurrentScope().setTag('tag-a', 'a'); // this tag will only remain within the callback
  // set the tag directly on customScope in addition, if you want to to persist the tag outside of the callback
  customScope.setTag('tag-a', 'a');
});
```

With that `expect(activeTransaction).toBe(getActiveSpan());` was failing

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
lucas-zimerman: fixed on https://github.com/getsentry/sentry-react-native/pull/4568/commits/816b506dc3c8377405470eea8300041624acb0a3

CHANGELOG.md
lucas-zimerman: Added explanation here:  https://github.com/getsentry/sentry-react-native/pull/4568/commits/25bf1471978b9b24a5537ec411fc3077843e1a64

CHANGELOG.md
lucas-zimerman: In regard to this, should we remove the integration `autoSessionTracking` when running natively?

CHANGELOG.md
```diff
@@ -44,6 +47,18 @@
...
+- Remove `getCurrentHub()`, `Hub`, and `getCurrentHubShim()`
+- Remove `spanId` from propagation `context`
+- Remove metrics API
+- Remove `transactionContext` from `samplingContext`
+- Remove deprecated `Request` type
+- Remove `@sentry/utils` package, the exports were moved to `@sentry/core`
+
```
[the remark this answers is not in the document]
lucas-zimerman: I have updated the list with more information that only impacts React Native

CHANGELOG.md
krystofwoldrich: Yes, makes sense, let's add the `browserSessionIntegration` when `enableAutioSessionTracking` and we are running in a browser.

packages/core/src/js/sdk.tsx
krystofwoldrich: Okay, I see, thank you for the details.

Could you open a follow-up PR with `options?.profilerProps` type update to exclude the update props and children. https://github.com/getsentry/sentry-javascript/blob/1048a437b09955d31118960624b6d58242389c45/packages/react/src/profiler.tsx#L158

packages/core/test/tracing/addTracingExtensions.test.ts
krystofwoldrich: Got it, seems like the behavior of the JS SDK changed.

packages/core/test/tracing/timetodisplaynative.web.test.tsx
krystofwoldrich: The issue with removing the mock is that this test is now not ensuring the component is not throwing in react-native-web environment.

RN Web UIManager https://github.com/necolas/react-native-web/blob/fcbe2d1e9225282671e39f9f639e2cb04c7e1e65/packages/react-native-web/src/exports/UIManager/index.js#L55

has a different interface from the RN UIManager, for example, the `hasViewManagerConfig` doesn't exist there.

I would prefer to update the mock in a way so we avoid the type error but still ensure we test for the web env. 

Maybe we can mock the whole RN object, instead of overwritting the UIManager property.

packages/core/test/tracing/integrations/userInteraction.test.ts
krystofwoldrich: Thank you for the details, I understand it now, I didn't know that.

CHANGELOG.md
krystofwoldrich: We can also do this in a follow up PR.

krystofwoldrich:
Looks good, thank you for all the comments, changes and explanations.

Two things we should do in a follow-up PRs:
- Update the type of `options?.profilerProps` https://github.com/getsentry/sentry-react-native/pull/4568#discussion_r2005203769
- Auto add `browserSessionIntegration` https://github.com/getsentry/sentry-react-native/pull/4568#discussion_r2005088065

One item to finalize this PR:
- Update the mock of the UIManager to test the RN Web version of the components. https://github.com/getsentry/sentry-react-native/pull/4568#discussion_r2005309689

packages/core/src/js/transports/encodePolyfill.ts
```diff
@@ -6,7 +8,7 @@ export const useEncodePolyfill = (): void => {
     (RN_GLOBAL_OBJ.__SENTRY__ as Partial<(typeof RN_GLOBAL_OBJ)['__SENTRY__']>) = {};
   }
 
-  RN_GLOBAL_OBJ.__SENTRY__.encodePolyfill = encodePolyfill;
+  RN_GLOBAL_OBJ.__SENTRY__[SDK_VERSION].encodePolyfill = encodePolyfill;
```
krystofwoldrich, line 11: The `SDK_VERSION` is created as a side effect of creating Sentry Logger (https://github.com/getsentry/sentry-javascript/blob/bf323b111b7e7b781e60e6299e432ce89f144970/packages/core/src/utils-hoist/logger.ts#L102)

But it's not guaranteed that it will exist.

We should use https://github.com/getsentry/sentry-javascript/blob/d773cb7324480ed3cffc14504f0e41951e344d19/packages/core/src/carrier.ts#L42

krystofwoldrich: Also can be a follow up PR.

packages/core/test/tracing/timetodisplaynative.web.test.tsx
lucas-zimerman: I came with a simple approach:
delete UIManager and then create it like in the original test, so it becomes:
```typescript
jest.mock('react-native', () => {
  const RN = jest.requireActual('react-native');
  // Fixes TypeError: Cannot set property UIManager of #<Object> which has only a getter
  delete RN.UIManager;
  RN.UIManager = {};

  return RN;
});
```

fixed on https://github.com/getsentry/sentry-react-native/pull/4568/commits/709697c559c5ef88422752636c79d841b5781c2f

lucas-zimerman:
> Looks good, thank you for all the comments, changes and explanations.
> 
> Two things we should do in a follow-up PRs:
> 
>     * Update the type of `options?.profilerProps` [feat: Bump to Sentry Javascript V9 #4568 (comment)](https://github.com/getsentry/sentry-react-native/pull/4568#discussion_r2005203769)
> 
>     * Auto add `browserSessionIntegration` [feat: Bump to Sentry Javascript V9 #4568 (comment)](https://github.com/getsentry/sentry-react-native/pull/4568#discussion_r2005088065)
> 
>     * Use `getMainCarrier` [feat: Bump to Sentry Javascript V9 #4568 (comment)](https://github.com/getsentry/sentry-react-native/pull/4568#discussion_r2005519564)
> 
> 
> One item to finalize this PR:
> 
>     * Update the mock of the UIManager to test the RN Web version of the components. [feat: Bump to Sentry Javascript V9 #4568 (comment)](https://github.com/getsentry/sentry-react-native/pull/4568#discussion_r2005309689)

I have added your notes here so we don't lose them https://github.com/getsentry/sentry-react-native/issues/4542

krystofwoldrich (approved):
Thank you for bumping RN to JS v9.

Looks good!

PR #4569, opened 2025-02-19
misc(samples): Update Gemfile missing x86 ffi

krystofwoldrich:
#skip-changelog 

Not sure why, but I keep seeing this Gemfile change after `bundle install`


krystofwoldrich:
Ahh, thanks, maybe my pod install is running in Rosetta by accident, I'll check that.

Issue #4562, opened 2025-02-18
Error loading metro config: Cannot find module 'tslib'


DaniFoldi:
### What React Native libraries do you use?

React Navigation, Expo (mobile only), Expo Application Services (EAS)

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### @sentry/react-native SDK Version

6.6.0

### How does your development environment look like?

OS: macOS 15.3
Node: 22.11.0
Expo: 52.0.35
react: 18.3.1
react-native: 0.77.0

although I think what is more relevant: we use pnpm for package manager, and it's set up as a workspace.
The app lives in `/packages/app`.

### Sentry.init()

```ts
Sentry.init({
  dsn: constants.expoConfig?.extra?.sentry?.dsn,
  attachStacktrace: true,
  enableTracing: true,
  tracesSampleRate: 1,
  sampleRate: 1,
  debug: __DEV__,
  integrations: [
    Sentry.httpClientIntegration(),
    Sentry.reactNativeInfoIntegration()
  ]
})
```

### Steps to Reproduce

1. create a pnpm workspace, with an expo app in say /packages/app - make sure hoisting is disabled
2. add sentry
3. observe that while building, the error below is thrown.

The reason (I suspect) is that `tslib` is not declared as a dependency, but somehow ends up in the built file `dist/js/tools/metroconfig.js` (and others, but this is the first one to be found by the build process).

### Expected Result

tslib is depended upon, and therefore found while importing it

### Actual Result

```
ERROR: Error loading .../packages/app/metro.config.cjs
Reason: Cannot find module 'tslib'
Require stack:
- .../node_modules/.pnpm/@sentry+react-native@6.6.0_expo@52.0.35_@babel+core@7.26.7_@babel+preset-env@7.26.0_@babel+co_r4te4muu5h7i3sl6odqu22xk5u/node_modules/@sentry/react-native/dist/js/tools/metroconfig.js
```

krystofwoldrich:
Hi @DaniFoldi,
thank you for the message, you are right the SDK doesn't depend on `tslib` and so shouldn't import it.

We will remove the `tslib` import -> https://github.com/getsentry/sentry-react-native/pull/4573

PR #4573, opened 2025-02-20
fix(metro): Avoid importing tslib in Sentry Metro Plugin

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

- fixes: https://github.com/getsentry/sentry-react-native/issues/4562

The Sentry RN Package doesn't depend on `tslib` so it should not import it in its dist files. 

The only file where `tslib` was imported was `metroconfig.js`.

`tslib` was removed in Sentry JS already in v7 -> https://github.com/getsentry/sentry-javascript/pull/9299

checked that tslib is not part of the output in dist. 

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


antonis (approved):
LGTM 🚀

Issue #4461, opened 2025-01-17
Tag all spans with thread info


krystofwoldrich:
### Description

- https://github.com/getsentry/team-mobile/issues/153

antonis:
The `thread.name` was added with https://github.com/getsentry/sentry-react-native/pull/4579
It is still pending to [investigate and add the `thread.id`](https://github.com/getsentry/sentry-react-native/pull/4579#discussion_r1969712601).

krystofwoldrich:
We can consider this done since there is no thread id given by the platform.

antonis:
> We can consider this done since there is no thread id given by the platform.

Makes sense 👍

PR #4579, opened 2025-02-21
feat: Adds span thread info

antonis:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Adds span thread info. This would be:
- `name: main` for UI threads
- `name: javascript` for JS threads

Ref: https://reactnative.dev/architecture/threading-model

https://github.com/getsentry/sentry-react-native/issues/4461

Manual, CI

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/src/js/tracing/span.ts
+++ b/packages/core/src/js/tracing/span.ts
@@ -154,3 +154,29 @@ export function addDefaultOpForSpanFrom(client: Client): void {
     }
   });
 }
+
+const SPAN_THREAD_ID = 'thread.id';
+const SPAN_THREAD_ID_MAIN = 1;
+const SPAN_THREAD_ID_JAVASCRIPT = 2;
+const SPAN_THREAD_NAME = 'thread.name';
+const SPAN_THREAD_NAME_MAIN = 'main';
+const SPAN_THREAD_NAME_JAVASCRIPT = 'javascript';
+const UI_THREAD_PREFIX = 'ui.';
+
+/**
+ * Adds thread info to spans.
+ * Ref: https://reactnative.dev/architecture/threading-model
+ */
+export function addThreadInfoToSpan(client: Client): void {
+  client.on('spanStart', (span: Span) => {
+    if (!spanToJSON(span).data?.[SPAN_THREAD_ID]) {
+      if (spanToJSON(span).op?.startsWith(UI_THREAD_PREFIX)) {
+        span.setAttribute(SPAN_THREAD_ID, SPAN_THREAD_ID_MAIN);
+        span.setAttribute(SPAN_THREAD_NAME, SPAN_THREAD_NAME_MAIN);
+      } else {
+        span.setAttribute(SPAN_THREAD_ID, SPAN_THREAD_ID_JAVASCRIPT);
+        span.setAttribute(SPAN_THREAD_NAME, SPAN_THREAD_NAME_JAVASCRIPT);
+      }
+    }
+  });
+}
--- a/packages/core/src/js/tracing/span.ts
+++ b/packages/core/src/js/tracing/span.ts
@@ -155,9 +155,6 @@ export function addDefaultOpForSpanFrom(client: Client): void {
   });
 }
 
-export const SPAN_THREAD_ID = 'thread.id';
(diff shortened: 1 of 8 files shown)


packages/core/src/js/tracing/span.ts
```diff
@@ -154,3 +154,30 @@ export function addDefaultOpForSpanFrom(client: Client): void {
...
 }
+
+export const SPAN_THREAD_ID = 'thread.id';
+export const SPAN_THREAD_ID_MAIN = 1;
+export const SPAN_THREAD_ID_JAVASCRIPT = 2;
+export const SPAN_THREAD_NAME = 'thread.name';
+export const SPAN_THREAD_NAME_MAIN = 'main';
+export const SPAN_THREAD_NAME_JAVASCRIPT = 'javascript';
+
+const UI_THREAD_PREFIX = 'ui.';
```
krystofwoldrich, line 165: The `ui.` prefix doesn't indicate that the span is from the main thread. Currently all the spans created in RN are from the JS thread.

The only exception are app start spans which are fetched from the native SDKs. https://github.com/getsentry/sentry-react-native/blob/18ce5e87b6e981e23ef9493207fb152d9c6aa8e7/packages/core/src/js/tracing/integrations/appStart.ts#L310

These should be marked as main thread.

antonis: Thank you for pointing me to the right direction @krystofwoldrich 🙇 
Updated with 377fa6a5bfda9f0b52df139d6e2bbf28441328f0

packages/core/test/tracing/reactnavigation.ttid.test.tsx
```diff
@@ -80,6 +81,8 @@ describe('React Navigation - TTID', () => {
               data: {
                 'sentry.op': 'ui.load.initial_display',
                 'sentry.origin': SPAN_ORIGIN_AUTO_UI_TIME_TO_DISPLAY,
+                [SPAN_THREAD_ID]: SPAN_THREAD_ID_JAVASCRIPT,
+                [SPAN_THREAD_NAME]: SPAN_THREAD_NAME_JAVASCRIPT,
```
krystofwoldrich, line 85: When considering what TTID and TTFD measure it includes both JS and main threads. But the majority of the work should be in JS so I think this is OK.

I just want to note this down.

packages/core/src/js/tracing/span.ts
```diff
@@ -154,3 +154,33 @@ export function addDefaultOpForSpanFrom(client: Client): void {
     }
   });
 }
+
+export const SPAN_THREAD_ID = 'thread.id';
+export const SPAN_THREAD_ID_MAIN = 1;
```
krystofwoldrich, line 159: Depending on the scenario and where the value is read from `sentry-cocoa` sets MAIN_ID as `0` or `259`.

I don't know the situation on Android, but I think we should leave the IDs for now until we establish how to read them so they map correctly to what the native SDKs are reporting.

packages/core/src/js/tracing/span.ts
```diff
@@ -154,3 +154,33 @@ export function addDefaultOpForSpanFrom(client: Client): void {
...
+export const SPAN_THREAD_ID_JAVASCRIPT = 2;
```
krystofwoldrich, line 160: Same as for main I think it's better not to send the ID than hard code 2.

packages/core/src/js/tracing/span.ts
antonis: Makes sense @krystofwoldrich 👍 
Removed the thread id with 91632023dfd92652a1d2391741901ecadf4a12c0 and I'll investigate how to get this information from the native SDKs separately

packages/core/src/js/tracing/span.ts
antonis: Updated with 91632023dfd92652a1d2391741901ecadf4a12c0

lucas-zimerman (approved):
Overall the Pr looks good!

krystofwoldrich (approved):
LGTM 🚀 

Thank you!

PR #4580, opened 2025-02-21
feat(feedback): Add integrations to client to allow for usage tracking

krystofwoldrich:
#skip-changelog 

This PR ensures we can track the feedback usage.

If widget only is used `MobileFeedback` integration is added.

If auto inject show is called `AutoInjectMobileFeedback` dummy integration is added.

This PR also removed the static options of the feedback integration, so users have to correctly register the integration and so we can track it.


antonis (approved):
LGTM 🎸 
Thank you for fixing usage tracking Krystof 🙇

PR #4583, opened 2025-02-24
fix(ci): Consider build environment variables in expo sample app configuration

antonis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Consider the `SENTRY_DISABLE_AUTO_UPLOAD` and `SENTRY_DISABLE_NATIVE_DEBUG_UPLOAD` environment variables in the Sentry Android Gradle Plugin Expo plugin configuration.

This is done by using [`shouldSentryAutoUpload`](https://github.com/getsentry/sentry-react-native/blob/60ad48cdd1e6882c9512161f0dfa85fd7026b764/packages/core/sentry.gradle#L14) in the generated gradle script similar [to the gradle script of the react native sample](https://github.com/getsentry/sentry-react-native/blob/60ad48cdd1e6882c9512161f0dfa85fd7026b764/samples/react-native/android/app/build.gradle#L29).

Fixes `dependabot` CI failure https://github.com/getsentry/sentry-react-native/pull/4582#issuecomment-2677968196

CI, Manual

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 
   ```js
   import Sentry from "@sentry/react-native";
-  
+
   Sentry.showFeedbackWidget();
 
   Sentry.wrap(RootComponent);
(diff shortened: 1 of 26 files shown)


CHANGELOG.md
```diff
@@ -41,6 +41,30 @@
 
   To learn more about the available configuration options visit [the documentation](https://docs.sentry.io/platforms/react-native/user-feedback/configuration).
 
+- Adds the ability to consider the `SENTRY_DISABLE_AUTO_UPLOAD` and `SENTRY_DISABLE_NATIVE_DEBUG_UPLOAD` environment variables in the Sentry Android Gradle Plugin Expo plugin configuration `app.config.js` ([#4583](https://github.com/getsentry/sentry-react-native/pull/4583))
```
antonis, line 44: Probably this is too extended and can be included just in [the documentation](https://docs.sentry.io/platforms/react-native/manual-setup/expo/gradle/)

krystofwoldrich:
Adding `shouldSentryAutoUpload` definitely makes sense, but I think we should not read the variable in the Expo settings, but we should add it to the generated code.

The difference is that Expo settings apply during the native project generation and not during the build.

antonis:

> The difference is that Expo settings apply during the native project generation and not during the build.

I've updated the PR changing the generated code.

krystofwoldrich (approved):
LGTM 🚀 

Thank you for the fix. 

This is a good catch we missed in the initial implementation.

PR #4588, opened 2025-02-24
test(e2e): Add auto init from JS tests

krystofwoldrich:
Run the sample E2E test with auto init from JS. 

The test suit is the same as we don't expect any differences between the manual native init and auto init from JS.

#skip-changelog


.github/workflows/sample-application.yml
```diff
@@ -200,11 +200,20 @@ jobs:
             rn-architecture: 'new'
             ios-use-frameworks: 'no-frameworks'
             build-type: 'production'
+            test-command: 'yarn test-ios-auto' # tests native auto init from JS
```
antonis, line 203: l: due to the test name (`name: Test ${{ matrix.platform }} ${{ matrix.build-type }}` we have a duplicate entry `Test ios production` in the test results.
Wdyt of adding the `test-command` or a new matrix field (e.g. `test-type`) in the name to separate those and easily detect the root of a test failure?

antonis (approved):
Added [a nit pick suggestion](https://github.com/getsentry/sentry-react-native/pull/4588/files#r1969220178) but overall LGTM 🎸 
Thank you for adding those tests Krystof 🙇

PR #4602, opened 2025-02-26
test(e2e): Add auto init from JS tests for Android

krystofwoldrich:
#skip-changelog 

Runs the sample e2e test suit with the new manual init and the original auto init from JS.


krystofwoldrich:
I'm currently investigating why Android tests with auto init are not returning app start.

I'm getting an error message that the app was in the background on start.

```
Invalid app start data: app not launched in foreground.
'Sentry Logger [warn]:', '[AppStart] Failed to retrieve the app start metrics from the native layer.'
```

krystofwoldrich:
Closing this because the branch is outdated. And the sample app is using different test suit. 

Maestro instead of Detox. (See https://github.com/getsentry/sentry-react-native/pull/4619 for context)

PR #4631, opened 2025-03-06
fix(tracing): Use route name instead of route key for current route tracking

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

Current route must be a name, not a key, as keys contain navigation specific hashes and thus generated transaction names would always be unique.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


krystofwoldrich:
Closing in favor of https://github.com/getsentry/sentry-react-native/pull/4650 which includes tests

Issue #4625, opened 2025-03-05
Add `RNSentrySDK` APIs support to `@sentry/react-native/expo` plugin


krystofwoldrich:
### Description

The goal of this task is to add support for the new manual APIs in Sentry Expo Plugin.

- Initially as opt-in. `useNativeInit: true`
- The plugin should warn users if native init is already present in the native code.
- The plugin should patch `AppDelegate.mm` for iOS and `MainApplication.kt` for Android.

renchap:
Are you planning for this to be in 7.0? This would be a very nice improvement for Expo users, and the work seems to be nearly finished?

antonis:
Thank you for the ping @renchap. Yes, this is planned for v7. It probably won’t make it into the first release, which is currently in [RC](https://github.com/getsentry/sentry-react-native/releases/tag/7.0.0-rc.1), but it will be included in a subsequent release.

Lucienzera:
Thanks for the work on this! 
Do you have any rough timeline for when it might be released? Since it looks almost done, we’re hesitant to implement manual initialization ourselves if an official update is coming soon.

antonis:
Thank you for bumping this @Lucienzera 🙇 
Our plan is to release this in this quarter.

Lucienzera:
Awesome @antonis thanks for the timeline! we will keep that in mind in this quarter!

antonis:
Thank you! I'll iterate here as soon as there is an update on this.

antonis:
Fixed with https://github.com/getsentry/sentry-react-native/pull/4633

PR #4472, opened 2025-01-21
feat: Capture app start errors before JS 

krystofwoldrich:
## Create merge commit. Do not squash.

This PR contains reviewed intermediate commits for capturing error before JS start feature. 

#skip-changelog

PR #4633, opened 2025-03-07
feat(expo): Add RNSentrySDK APIs support to @sentry/react-native/expo plugin

antonis:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

~~**Based on [feat: Capture app start errors before JS](https://github.com/getsentry/sentry-react-native/pull/4472/)**~~
**Based on https://github.com/getsentry/sentry-react-native/pull/5470**

## :scroll: Description

Adds RNSentrySDK APIs support to @sentry/react-native/expo plugin by importing sentry and adding injecting `RNSentrySDK.init/start` in the Android MainApplication (Kotlin or Java) or AppDelegate (Objective-C or Swift). 

  This feature is opt-out to enable it set `useNativeInit` to `true` in your `@sentry/react-native/expo` plugin configuration.

  ```js
  "plugins": [
    [
      "@sentry/react-native/expo",
      {
        "useNativeInit": true
      }
    ],
  ```

Fixes https://github.com/getsentry/sentry-react-native/issues/4625

CI, Manual

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/packages/core/RNSentry.podspec
+++ b/packages/core/RNSentry.podspec
@@ -46,6 +46,10 @@ Pod::Spec.new do |s|
 
   s.compiler_flags = other_cflags
 
+  s.pod_target_xcconfig = {
+    'DEFINES_MODULE' => 'YES'
+  }
+
   s.dependency 'Sentry/HybridSDK', '8.58.0'
 
   if defined? install_modules_dependencies
@@ -56,10 +60,10 @@ Pod::Spec.new do |s|
 
     if is_new_arch_enabled then
       # New Architecture on React Native 0.70 and older
-      s.pod_target_xcconfig = {
+      s.pod_target_xcconfig.merge!({
           "HEADER_SEARCH_PATHS" => "\"$(PODS_ROOT)/boost\"",
           "CLANG_CXX_LANGUAGE_STANDARD" => "c++17"
-      }
+      })
 
       s.dependency "React-RCTFabric" # Required for Fabric Components (like RCTViewComponentView)
       s.dependency "React-Codegen"


CHANGELOG.md
```diff
@@ -10,6 +10,20 @@
 
 ### Features
 
+- Add RNSentrySDK APIs support to @sentry/react-native/expo plugin ([#4633](https://github.com/getsentry/sentry-react-native/pull/4633))
+
+  This feature is opt-out to enable it set `useNativeInit` to `true` in your `@sentry/react-native/expo` plugin configuration.
```
lucas-zimerman, line 15: NIT: It may make it easier for reading.
```suggestion
  This feature is opt-out, to enable it set `useNativeInit` to `true` in your `@sentry/react-native/expo` plugin configuration.

File: packages/core/plugin/src/withSentryAndroid.ts [hunk 2]
```diff
@@ -49,3 +56,50 @@ export function modifyAppBuildGradle(buildGradle: string): string {
...
+        // Insert import statement after package declaration
+        config.modResults.contents = config.modResults.contents.replace(
+          /(package .*\n\n?)/,
+          `$1import io.sentry.react.RNSentrySDK\n`,
+        );
+      }
+      // Add RNSentrySDK.init
+      config.modResults.contents = config.modResults.contents.replace(
+        /(super\.onCreate\(\)[;\n]*)([ \t]*)/,
+        `$1\n$2RNSentrySDK.init(this)\n$2`,
```
lucas-zimerman, line 99: q: I think we should warn the user if `RNSentrySDK.init` wasn't inserted, to avoid giving the impression the SDK was set in some edge case where it wasn't

lucas-zimerman (commented):
Overall the PR looks good and thank you for the tests!
I still think we could fine tune a bit more the android part and after it we could get ready for merge

packages/core/plugin/src/withSentryAndroid.ts
```diff
@@ -49,3 +56,50 @@ export function modifyAppBuildGradle(buildGradle: string): string {
...
+      if (!config.modResults.contents.includes('import io.sentry.react.RNSentrySDK')) {
+        // Insert import statement after package declaration
+        config.modResults.contents = config.modResults.contents.replace(
+          /(package .*\n\n?)/,
+          `$1import io.sentry.react.RNSentrySDK\n`,
+        );
+      }
+      // Add RNSentrySDK.init
+      config.modResults.contents = config.modResults.contents.replace(
+        /(super\.onCreate\(\)[;\n]*)([ \t]*)/,
```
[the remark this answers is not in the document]
antonis: Good question @lucas-zimerman 👍 I've started implementing with parameters but then realised that the [Application.onCreate](https://developer.android.com/reference/android/app/Application#onCreate()) doesn't have parameters (the [Activity.onCreate](https://developer.android.com/guide/components/activities/activity-lifecycle#oncreate) does)

packages/core/plugin/src/withSentryAndroid.ts
antonis: Good catch @lucas-zimerman 🙇 
I've added warning for both platforms with d25db305aaed11e2a8d5e9657ffae4fcc5287eee

antonis:
_Note: The [Build & Test / Type Check Typescript 3.8 (pull_request)](https://github.com/getsentry/sentry-react-native/actions/runs/14262614941/job/39977478017?pr=4633) failure should be fixed when we merge https://github.com/getsentry/sentry-react-native/pull/4673 from `main`_

CHANGELOG.md
```diff
@@ -10,6 +10,20 @@
 
 ### Features
 
+- Add RNSentrySDK APIs support to @sentry/react-native/expo plugin ([#4633](https://github.com/getsentry/sentry-react-native/pull/4633))
+
+  This feature is opt-out, to enable it set `useNativeInit` to `true` in your `@sentry/react-native/expo` plugin configuration.
```
krystofwoldrich, line 15: What's the reason for making this opt-out?

This could break user apps if they don't have the sentry options JSON in their project; the added init would cause a runtime error.  

I think this should be opt-in first.

antonis: @krystofwoldrich You are right! I think I've misread the [issue](https://github.com/getsentry/sentry-react-native/issues/4625) 😓

antonis: Updated with adc81a54d475203b1c69e92855a5e7631339cb5f. I also removed the note from the changelog since it is opt-in first. I can describe how to opt-out on a docs PR.

packages/core/plugin/src/withSentryAndroid.ts
```diff
@@ -49,3 +56,58 @@ export function modifyAppBuildGradle(buildGradle: string): string {
 
   return buildGradle.replace(pattern, match => `${applyFrom}\n\n${match}`);
 }
+
+export function modifyMainApplication(config: ExpoConfig): ExpoConfig {
+  return withMainApplication(config, async config => {
+    if (!config.modResults || !config.modResults.path) {
+      warnOnce('Skipping MainApplication modification because the file does not exist.');
```
krystofwoldrich, line 63: Let's try to be more explicit about what this means for the user -> the native Android init won't be added.

````suggestion
      warnOnce("Can't add `RNSentrySDK.init` to Android MainApplication, because the file was not found.");
````

File: packages/core/plugin/src/withSentryAndroid.ts [hunk 6]
```diff
@@ -49,3 +56,58 @@ export function modifyAppBuildGradle(buildGradle: string): string {
...
+      return config;
+    }
+
+    const fileName = config.modResults.path.split('/').pop();
+
+    if (config.modResults.contents.includes('RNSentrySDK.init')) {
+      warnOnce(`Your '${fileName}' already contains 'RNSentrySDK.init'.`);
```
krystofwoldrich, line 70:
````suggestion
      warnOnce(`Your '${fileName}' already contains 'RNSentrySDK.init', the native code won't be updated.`);
````

File: packages/core/plugin/src/withSentryAndroid.ts [hunk 7]
```diff
@@ -49,3 +56,58 @@ export function modifyAppBuildGradle(buildGradle: string): string {
...
 }
+
+export function modifyMainApplication(config: ExpoConfig): ExpoConfig {
+  return withMainApplication(config, async config => {
+    if (!config.modResults || !config.modResults.path) {
+      warnOnce('Skipping MainApplication modification because the file does not exist.');
+      return config;
+    }
+
+    const fileName = config.modResults.path.split('/').pop();
```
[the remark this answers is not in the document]
antonis: Good idea 👍 
Updated with 5f4f7c59856acb85d0261fe859acfdf7debc2504

packages/core/plugin/src/withSentryAndroid.ts
```diff
@@ -49,3 +56,58 @@ export function modifyAppBuildGradle(buildGradle: string): string {
...
+      // Add RNSentrySDK.init
+      const originalContents = config.modResults.contents;
+      config.modResults.contents = config.modResults.contents.replace(
+        /(super\.onCreate\(\)[;\n]*)([ \t]*)/,
+        `$1\n$2RNSentrySDK.init(this);\n$2`,
+      );
+      if (config.modResults.contents === originalContents) {
+        warnOnce(`Failed to insert 'RNSentrySDK.init'.`);
+      }
+    } else {
```
krystofwoldrich, line 91: Can we explicitly check for `lang == kotlin`? It will make the script more resilient

packages/core/plugin/src/withSentryAndroid.ts
```diff
@@ -49,3 +56,58 @@ export function modifyAppBuildGradle(buildGradle: string): string {
...
+        );
+      }
+      // Add RNSentrySDK.init
+      const originalContents = config.modResults.contents;
+      config.modResults.contents = config.modResults.contents.replace(
+        /(super\.onCreate\(\)[;\n]*)([ \t]*)/,
+        `$1\n$2RNSentrySDK.init(this);\n$2`,
+      );
+      if (config.modResults.contents === originalContents) {
+        warnOnce(`Failed to insert 'RNSentrySDK.init'.`);
```
krystofwoldrich, line 89: Let's also log the file name.

packages/core/plugin/src/withSentryAndroid.ts
```diff
@@ -49,3 +56,58 @@ export function modifyAppBuildGradle(buildGradle: string): string {
...
+    }
+
+    if (config.modResults.language === 'java') {
+      if (!config.modResults.contents.includes('import io.sentry.react.RNSentrySDK;')) {
+        // Insert import statement after package declaration
+        config.modResults.contents = config.modResults.contents.replace(
+          /(package .*;\n\n?)/,
+          `$1import io.sentry.react.RNSentrySDK;\n`,
+        );
+      }
```
krystofwoldrich, line 81: Let's do this only if the init call added to avoid used code.

packages/core/plugin/src/withSentryAndroid.ts
```diff
@@ -49,3 +56,58 @@ export function modifyAppBuildGradle(buildGradle: string): string {
...
+      config.modResults.contents = config.modResults.contents.replace(
+        /(super\.onCreate\(\)[;\n]*)([ \t]*)/,
+        `$1\n$2RNSentrySDK.init(this);\n$2`,
+      );
+      if (config.modResults.contents === originalContents) {
+        warnOnce(`Failed to insert 'RNSentrySDK.init'.`);
+      }
+    } else {
+      // Kotlin
+      if (!config.modResults.contents.includes('import io.sentry.react.RNSentrySDK')) {
```
krystofwoldrich, line 93: Let do this after init added same as Java comment.

packages/core/plugin/src/withSentryAndroid.ts
```diff
@@ -49,3 +56,58 @@ export function modifyAppBuildGradle(buildGradle: string): string {
...
+        );
+      }
+      // Add RNSentrySDK.init
+      const originalContents = config.modResults.contents;
+      config.modResults.contents = config.modResults.contents.replace(
+        /(super\.onCreate\(\)[;\n]*)([ \t]*)/,
+        `$1\n$2RNSentrySDK.init(this)\n$2`,
+      );
+      if (config.modResults.contents === originalContents) {
+        warnOnce(`Failed to insert 'RNSentrySDK.init'.`);
```
krystofwoldrich, line 107: Let's add the file name to the log.

packages/core/plugin/src/withSentryIOS.ts
```diff
@@ -79,3 +85,58 @@ export function addSentryWithBundledScriptsToBundleShellScript(script: string):
...
+      const originalContents = config.modResults.contents;
+      config.modResults.contents = config.modResults.contents.replace(
+        /(func application\([^)]*\) -> Bool \{)/s,
+        `$1\n    RNSentrySDK.start()`,
+      );
+      if (config.modResults.contents === originalContents) {
+        warnOnce(`Failed to insert 'RNSentrySDK.start()'.`);
+      }
+    } else {
+      // Objective-C
```
krystofwoldrich, line 117: Can we explicitly check for obj-c?

packages/core/plugin/src/withSentryIOS.ts
```diff
@@ -79,3 +85,58 @@ export function addSentryWithBundledScriptsToBundleShellScript(script: string):
     (match: string) => `/bin/sh ${SENTRY_REACT_NATIVE_XCODE_PATH} ${match}`,
   );
 }
+
+export function modifyAppDelegate(config: ExpoConfig): ExpoConfig {
+  return withAppDelegate(config, async config => {
+    if (!config.modResults || !config.modResults.path) {
+      warnOnce('Skipping AppDelegate modification because the file does not exist.');
```
krystofwoldrich, line 92: Same as for Android let's state that the 'RNSentrySDK.start()' won't be added to the AppDelegate.

packages/core/plugin/src/withSentryIOS.ts
```diff
@@ -79,3 +85,58 @@ export function addSentryWithBundledScriptsToBundleShellScript(script: string):
...
+    }
+
+    const fileName = path.basename(config.modResults.path);
+
+    if (config.modResults.language === 'swift') {
+      if (config.modResults.contents.includes('RNSentrySDK.start()')) {
+        warnOnce(`Your '${fileName}' already contains 'RNSentrySDK.start()'.`);
+        return config;
+      }
+      if (!config.modResults.contents.includes('import RNSentry')) {
```
krystofwoldrich, line 103: Let's add the imports after the start call is added.

packages/core/test/expo-plugin/modifyAppDelegate.test.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+}`;
+
+describe('modifyAppDelegate', () => {
+  let config: MockedExpoConfig;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    // Reset to a mocked Swift config after each test
+    config = {
```
krystofwoldrich, line 98: Let's deep copy or use func to create the config to avoid comparing the same reference int he test with no modification.

packages/core/test/expo-plugin/modifyMainApplication.test.ts
```diff
@@ -0,0 +1,188 @@
...
+}
+`;
+
+describe('modifyMainApplication', () => {
+  let config: MockedExpoConfig;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    // Reset to a mocked Java config after each test
+    config = {
```
krystofwoldrich, line 128: Same as for iOS let's avoid comparing the same ref.

packages/core/test/expo-plugin/modifyAppDelegate.test.ts
antonis: Good catch 👍 Updated in 744993c8255ca160e41673f01cc09f7e58eb3c69

packages/core/plugin/src/withSentryIOS.ts
antonis: Updated with 0ffd26c8f16ee2e1a1bb04bca13be7a95ff2726b

packages/core/plugin/src/withSentryIOS.ts
antonis: Updated with dce74b25792f6d9b1b1da7e894642b365c9890ed

packages/core/plugin/src/withSentryIOS.ts
antonis: Updated with a53c7f46a14361e8f3b2b1da7f52b4f404799751

packages/core/plugin/src/withSentryAndroid.ts
antonis: Updated with 235f3efbe6180b41d45a748b04fd4d95393873b7

packages/core/plugin/src/withSentryAndroid.ts
antonis: Updated with 369cce774d10f252e53399b708f7c3d6c6042190

packages/core/plugin/src/withSentryAndroid.ts
antonis: Good point 👍 Updated with 369cce774d10f252e53399b708f7c3d6c6042190

packages/core/plugin/src/withSentryAndroid.ts
antonis: Updated with 235f3efbe6180b41d45a748b04fd4d95393873b7

packages/core/plugin/src/withSentryAndroid.ts
antonis: Good idea 👍 Updated with 62d39ccb4f99530bd9ee3188cf4b1220c3d5ab4d

packages/core/plugin/src/withSentryIOS.ts
```diff
@@ -79,3 +85,57 @@ export function addSentryWithBundledScriptsToBundleShellScript(script: string):
...
+        return config;
+      }
+      // Add RNSentrySDK.start() at the beginning of application method
+      const originalContents = config.modResults.contents;
+      config.modResults.contents = config.modResults.contents.replace(
+        /(func application\([^)]*\) -> Bool \{)/s,
+        `$1\n    RNSentrySDK.start()`,
+      );
+      if (config.modResults.contents === originalContents) {
+        warnOnce(`Failed to insert 'RNSentrySDK.start()' in '${fileName}.`);
```
lucas-zimerman, line 110: NIT: missing  '
```suggestion
        warnOnce(`Failed to insert 'RNSentrySDK.start()' in '${fileName}'.`);
```

lucas-zimerman (approved):
I have re-checked the PR and overall, I only found a nit on a string. After fixing it and if @krystofwoldrich doesn't find anything else, LGTM!

packages/core/plugin/src/withSentryIOS.ts
antonis: Good catch 🙇

packages/core/plugin/src/withSentryAndroid.ts
```diff
@@ -49,3 +56,57 @@ export function modifyAppBuildGradle(buildGradle: string): string {
 
   return buildGradle.replace(pattern, match => `${applyFrom}\n\n${match}`);
 }
+
+export function modifyMainApplication(config: ExpoConfig): ExpoConfig {
+  return withMainApplication(config, async config => {
```
krystofwoldrich: There are no async operations in the function.

````suggestion
  return withMainApplication(config, config => {
````

File: packages/core/plugin/src/withSentryAndroid.ts [hunk 20]
```diff
@@ -49,3 +56,57 @@ export function modifyAppBuildGradle(buildGradle: string): string {
...
+      warnOnce(`Your '${fileName}' already contains 'RNSentrySDK.init', the native code won't be updated.`);
+      return config;
+    }
+
+    if (config.modResults.language === 'java') {
+      // Add RNSentrySDK.init
+      const originalContents = config.modResults.contents;
+      config.modResults.contents = config.modResults.contents.replace(
+        /(super\.onCreate\(\)[;\n]*)([ \t]*)/,
+        `$1\n$2RNSentrySDK.init(this);\n$2`,
```
krystofwoldrich: Nice use of the regex, I really like how it matches the existing code white space prefix.

packages/core/plugin/src/withSentryIOS.ts
```diff
@@ -12,8 +13,11 @@ const SENTRY_REACT_NATIVE_XCODE_PATH =
 const SENTRY_REACT_NATIVE_XCODE_DEBUG_FILES_PATH =
   "`${NODE_BINARY:-node} --print \"require('path').dirname(require.resolve('@sentry/react-native/package.json')) + '/scripts/sentry-xcode-debug-files.sh'\"`";
 
-export const withSentryIOS: ConfigPlugin<string> = (config, sentryProperties: string) => {
-  const cfg = withXcodeProject(config, config => {
+export const withSentryIOS: ConfigPlugin<{ sentryProperties: string; useNativeInit: boolean | undefined }> = (
+  config,
+  { sentryProperties, useNativeInit = true },
```
krystofwoldrich: I believe `useNativeInit` must be `false` by default, otherwise it would break existing users.

By adding the native init the SDK would behave differently for user without the init file. The options would not be synced.

packages/core/plugin/src/withSentryIOS.ts
```diff
@@ -79,3 +85,57 @@ export function addSentryWithBundledScriptsToBundleShellScript(script: string):
...
+    if (config.modResults.language === 'swift') {
+      if (config.modResults.contents.includes('RNSentrySDK.start()')) {
+        warnOnce(`Your '${fileName}' already contains 'RNSentrySDK.start()'.`);
+        return config;
+      }
+      // Add RNSentrySDK.start() at the beginning of application method
+      const originalContents = config.modResults.contents;
+      config.modResults.contents = config.modResults.contents.replace(
+        /(func application\([^)]*\) -> Bool \{)/s,
+        `$1\n    RNSentrySDK.start()`,
```
krystofwoldrich: Would it make sense to also use a second capture group to dynamically fill in the tabs/spaces?

packages/core/plugin/src/withSentryIOS.ts
antonis: Good catch 👍 
Updated with c356288dbbb4afb1b82f897ecc720bf1e457b015

packages/core/plugin/src/withSentryIOS.ts
antonis: Good idea. Updated with 8e32556b26c27a334e47b425356602b003240cb2

packages/core/plugin/src/withSentryIOS.ts
```diff
@@ -79,3 +85,57 @@ export function addSentryWithBundledScriptsToBundleShellScript(script: string):
...
+        warnOnce(`Failed to insert '[RNSentrySDK start]' in '${fileName}.`);
+      } else if (!config.modResults.contents.includes('#import <RNSentry/RNSentry.h>')) {
+        // Add import after AppDelegate.h
+        config.modResults.contents = config.modResults.contents.replace(
+          /(#import "AppDelegate.h"\n)/,
+          `$1#import <RNSentry/RNSentry.h>\n`,
+        );
+      }
+    } else {
+      warnOnce(`Unsupported language detected in '${fileName}', the native code won't be updated.`);
```
krystofwoldrich: I've tried to run the plugin with our sample and got this message for iOS.

Might be worth to log out the value of the language.

packages/core/plugin/src/withSentryIOS.ts
```diff
@@ -79,3 +85,57 @@ export function addSentryWithBundledScriptsToBundleShellScript(script: string):
...
+        /(func application\([^)]*\) -> Bool \{)\s*\n(\s*)/s,
+        `$1\n$2RNSentrySDK.start()\n$2`,
+      );
+      if (config.modResults.contents === originalContents) {
+        warnOnce(`Failed to insert 'RNSentrySDK.start()' in '${fileName}'.`);
+      } else if (!config.modResults.contents.includes('import RNSentry')) {
+        // Insert import statement after UIKit import
+        config.modResults.contents = config.modResults.contents.replace(/(import UIKit\n)/, `$1import RNSentry\n`);
+      }
+    } else if (config.modResults.language === 'objc') {
```
krystofwoldrich: Depending on RN version, we can encounter also `objcpp`.

```suggestion
    } else if (['objcpp', 'objc'].includes(config.modResults.language)) {
```

krystofwoldrich:
I'm currently updating the base branch with the latest main since there was quite a lot of development that was not synced to the feature branch.

antonis:
> I'm currently updating the base branch with the latest main since there was quite a lot of development that was not synced to the feature branch.

Thank you for the heads up @krystofwoldrich 🙇 
I'll update this branch after that along with the review feedback.

packages/core/plugin/src/withSentryIOS.ts
antonis: Good idea 👍

packages/core/plugin/src/withSentryIOS.ts
antonis: Good idea 👍 
Added the language in the warning for both iOS and Android with a20984c8fdc30deff9a04c8eac7dc9501404e81d

krystofwoldrich (approved):
Thank you for the fixes. 🚀 It looks great!

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### Features
 
+- Add RNSentrySDK APIs support to @sentry/react-native/expo plugin ([#4633](https://github.com/getsentry/sentry-react-native/pull/4633))
```
krystofwoldrich, line 17: Nit. It would be nice to include an example code snippet and a small summary of what will the flag do.

lucas-zimerman:
@sentry review

lucas-zimerman:
@antonis should we restart this PR?

antonis:
> @antonis should we restart this PR?

I don't expect this to have major conflicts with main/v7.

packages/core/plugin/src/withSentryIOS.ts
```diff
@@ -78,3 +85,59 @@ export function addSentryWithBundledScriptsToBundleShellScript(script: string):
...
+      if (config.modResults.contents.includes('RNSentrySDK.start()')) {
+        warnOnce(`Your '${fileName}' already contains 'RNSentrySDK.start()'.`);
+        return config;
+      }
+      // Add RNSentrySDK.start() at the beginning of application method
+      const originalContents = config.modResults.contents;
+      config.modResults.contents = config.modResults.contents.replace(
+        /(func application\([^)]*\) -> Bool \{)\s*\n(\s*)/s,
+        '$1\n$2RNSentrySDK.start()\n$2',
+      );
```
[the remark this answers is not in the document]
antonis: There IS a theoretical edge case where the regex could fail: if function parameters contain closing parentheses (like closure types):                                                                                                                                             
                                                                                                                                                                                                                                                                                    
  func application(_ app: UIApplication, handler: ((Int) -> Void)?) -> Bool {                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                    
  Here, [^)]* would stop at the first ) in ((Int) -> Void), not the function's closing paren.                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                    
  However: This is not a concern because:                                                                                                                                                                                                                                           
  - Expo-generated AppDelegates don't use such parameter types                                                                                                                                                                                                                      
  - The standard application(_:didFinishLaunchingWithOptions:) signature doesn't have nested parentheses                                                                                                                                                                            
  - All our test cases pass                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                    
 No fix needed.

antonis:
@sentry review

packages/core/ios/RNSentry.h
```diff
@@ -10,7 +10,7 @@
 #import <Sentry/Sentry.h>
 
 // This import exposes public RNSentrySDK start
-#import "RNSentrySDK.h"
+#import <RNSentry/RNSentrySDK.h>
```
lucas-zimerman: q: Why this changed?

lucas-zimerman:
@sentry review

packages/core/plugin/src/logger.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Log a warning message only once per run.
+ * This is used to avoid spamming the console with the same message.
+ */
+export function warnOnce(message: string): void {
+  if (!warningMap.has(message)) {
+    warningMap.set(message, true);
+    // eslint-disable-next-line no-console
+    console.warn(yellow(prefix(message)));
```
lucas-zimerman, line 11: q: should we temporarily disable the capture of console warnings as events here?

packages/core/plugin/src/version.ts
```diff
@@ -0,0 +1,8 @@
+const packageJson: {
+  name: string;
+  version: string;
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+} = require('../../package.json');
+
+export const PLUGIN_NAME = `${packageJson.name}/expo`;
+export const PLUGIN_VERSION = packageJson.version;
```
lucas-zimerman, line 8: shouldn't we protect these variables in case `packagejson` is not found?

packages/core/plugin/src/withSentry.ts
```diff
@@ -79,6 +81,6 @@ ${authToken ? `${existingAuthTokenMessage}\nauth.token=${authToken}` : missingAu
 }
 
 // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-const withSentry = createRunOncePlugin(withSentryPlugin, sdkPackage.name, sdkPackage.version);
+const withSentry = createRunOncePlugin(withSentryPlugin, PLUGIN_NAME, PLUGIN_VERSION);
```
lucas-zimerman, line 84: might be a good idea to use the sdkPackage value as a fallback in case PLUGIN_NAME/VERSION are undefined.

lucas-zimerman:
` I updated the docs if needed.` was marked and this is a feature. Do we already have a PR to document this?

lucas-zimerman:
q: What will happen when an expo go app sets `useNativeInit: true`?

antonis:
> ` I updated the docs if needed.` was marked and this is a feature. Do we already have a PR to document this?

Good catch @lucas-zimerman. I've unchecked those.

> q: What will happen when an expo go app sets `useNativeInit: true`?

My understanding is that it would fail similar to other features on runtime. The code injection would also fail with errors similar to [this](https://github.com/getsentry/sentry-react-native/pull/4633/changes#diff-e2bb8c37eebce5184fa1eefa69f6f8548aceab620e2409a2cdfc58a1c0ffc02dR63). I'll test more extensively this case though.

packages/core/ios/RNSentry.h
antonis: My understanding is that the latest Swift changes on the iOS side require angle bracket imports

packages/core/plugin/src/logger.ts
antonis: I think this would appear during the build phase. Why should we disable it?

packages/core/plugin/src/version.ts
antonis: If package.json doesn't exist, this import statement fails during build/compilation

packages/core/plugin/src/withSentry.ts
antonis: I think this should fail again at compile time if the import above fails.

PR #4661, opened 2025-03-14
feat(ttid): Add support for measuring Time to Initial Display for already seen routes

krystofwoldrich:

- [x] Enhancement

## :scroll: Description

- Introduced `enableTimeToInitialDisplayForPreloadedRoutes` option to the React Navigation integration.
- Updated logic to measure Time to Initial Display for routes that have already been seen.
- Added tests to verify functionality for preloaded routes in the tracing module.

Depending on an application, this can improve the number of transactions with TTID measurements. 

Originally we excluded already seen routes with the idea that the seen route is already loaded.

But routes that have been seen, can on the next visit refresh their context and this PR enables users to measure the time this next visit takes to load.

integration test, manually in sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


antonis (approved):
LGTM 🚀

PR #4669, opened 2025-03-19
001 fix(ttd): Add TimeToDisplay integration to force fetch the data

krystofwoldrich:

- [x] Bugfix
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

This PR changes how TTID and TTFD data are retrieved from the native layer. The new approach relies on the retrieval of the data during the transaction processing. This ensures the data are added to the transaction if recorded on the native layer. This is more robust than reading data from the event emitted on the native layer which could be received in the JS after the transaction ended and thus the data are lost. 

How often the data are lost depends on a given application and if transactions are often ended early or ended manually.  

The timestamps are linked to the active span id (the transaction span id) and saved in a map with a max size to avoid leaking memory. If the timestamp would not be retrieved by JS (should only happen in unexpected scenarios).

sample app, e2e tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


antonis (approved):
LGTM 🚀

PR #4604, opened 2025-02-26
test(e2e): Adds Feedback Widget Maestro E2E tests

antonis:
## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

**Based on: https://github.com/getsentry/sentry-react-native/pull/4726**

PR Chain:
- [feat(feedback): Report a Bug button](https://github.com/getsentry/sentry-react-native/pull/4378)
- [feat(feedback): Theming](https://github.com/getsentry/sentry-react-native/pull/4677)
- [feat(feedback): Screenshot button](https://github.com/getsentry/sentry-react-native/pull/4714)
- [ref(feedback): Extracts FeedbackWidgetProvider in a separate file](https://github.com/getsentry/sentry-react-native/pull/4718)
- [fix(feedback): Fixes accessibility issue on iOS](https://github.com/getsentry/sentry-react-native/pull/4739)
- [feat(feedback): Screenshot button error flow](https://github.com/getsentry/sentry-react-native/pull/4757)
- 👉 [test(e2e): Adds Feedback Widget Maestro E2E tests](https://github.com/getsentry/sentry-react-native/pull/4604)

## :scroll: Description

Adds Feedback Widget Maestro E2E tests for the happy path flow:
- Show feedback button
- Open the feedback widget
- Fill and submit the form

## :bulb: Motivation and Context

Part of https://github.com/getsentry/sentry-react-native/issues/4302

## :green_heart: How did you test it?
CI ([submitted E2E test feedback](https://sentry-sdks.sentry.io/feedback/?end=2025-04-28T23%3A59%3A59&feedbackSlug=sentry-react-native%3A6572821416&project=5428561&referrer=feedback_list_page&start=2025-04-28T00%3A00%3A00&utc=true)), Manual

_Note: To run Android locally you might need to change the build architecture [android.patch](https://github.com/user-attachments/files/19668775/android.patch)_

## :pencil: Checklist

- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- Add more tests outside the happy path

#skip-changelog

PR #4378, opened 2024-12-16
feat(feedback): Report a Bug button

antonis:
## :loudspeaker: Type of change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

**Targeting [`feedback-ui-2` branch](https://github.com/getsentry/sentry-react-native/tree/feedback-ui-2)**

PR Chain:
- 👉 [feat(feedback): Report a Bug button](https://github.com/getsentry/sentry-react-native/pull/4378)
- [feat(feedback): Theming](https://github.com/getsentry/sentry-react-native/pull/4677)
- [feat(feedback): Screenshot button](https://github.com/getsentry/sentry-react-native/pull/4714)
- [ref(feedback): Extracts FeedbackWidgetProvider in a separate file](https://github.com/getsentry/sentry-react-native/pull/4718)

## :scroll: Description

Implements feedback button that launches the feedback widget using [the auto-injected feedback widget](https://github.com/getsentry/sentry-react-native/pull/4483)

The button is shown with `Sentry.showFeedbackButton()` and hidden with  `Sentry.hideFeedbackButton()` provided that the app is wrapped (`Sentry.wrap(RootComponent)`). The button can be styled like:
```
Sentry.feedbackIntegration({
  buttonOptions: {
    styles: {
      triggerButton: {
        marginBottom: 75,
      },
    },
  },
})
```

## :bulb: Motivation and Context

Part of https://github.com/getsentry/sentry-react-native/issues/4302

## :green_heart: How did you test it?
Manual testing with the sample app

## :pencil: Checklist

- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

PR #4677, opened 2025-03-21
feat(feedback): Theming

antonis:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

**Based on [feat(feedback): Report a Bug button](https://github.com/getsentry/sentry-react-native/pull/4378)**

PR Chain:
- [feat(feedback): Report a Bug button](https://github.com/getsentry/sentry-react-native/pull/4378)
- 👉 [feat(feedback): Theming](https://github.com/getsentry/sentry-react-native/pull/4677)
- [feat(feedback): Screenshot button](https://github.com/getsentry/sentry-react-native/pull/4714)
- [ref(feedback): Extracts FeedbackWidgetProvider in a separate file](https://github.com/getsentry/sentry-react-native/pull/4718)
- [fix(feedback): Fixes accessibility issue on iOS](https://github.com/getsentry/sentry-react-native/pull/4739)
- [test(e2e): Adds Feedback Widget Maestro E2E tests](https://github.com/getsentry/sentry-react-native/pull/4604)

## :scroll: Description

Adds theming options [similar to JS](https://docs.sentry.io/platforms/javascript/guides/nextjs/user-feedback/configuration/#css-customization)
```
      Sentry.feedbackIntegration({
        colorScheme: 'system',
        themeLight: {
          foreground: '#ff0000',
          background: '#00ff00',
        },
        themeDark: {
          foreground: '#00ff00',
          background: '#ff0000',
        },
      }),
```

<details><summary>Demo Video</summary>
<p>

web

https://github.com/user-attachments/assets/72a65d2c-765f-4867-831d-aea3a56fe96f

</p>
</details> 

|Light|Dark|
|---|---|
|Simulator Screenshot - iPhone 16 Pro - 2025-03-27 at 12 03 19|Simulator Screenshot - iPhone 16 Pro - 2025-03-27 at 12 03 43|

Sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


packages/core/src/js/feedback/FeedbackWidget.styles.ts
```diff
@@ -63,30 +59,30 @@ const defaultStyles: FeedbackWidgetStyles = {
...
-    backgroundColor: PURPLE,
+    backgroundColor: getTheme().accent,
     paddingVertical: 15,
     borderRadius: 5,
     alignItems: 'center',
     marginBottom: 10,
   },
   submitText: {
-    color: BACKGROUND_COLOR,
+    color: getTheme().background,
```
krystofwoldrich, line 73: Just a small oversight.
 
The submit button text should be in contrast with the button, not the widget background.

Screenshot 2025-03-26 at 18 15 50

packages/core/src/js/feedback/integration.ts
```diff
@@ -42,3 +54,36 @@ export const getFeedbackButtonOptions = (): Partial<FeedbackButtonProps> => {
...
+  return integration.colorScheme;
+};
+
+export const getFeedbackLightTheme = (): Partial<FeedbackWidgetTheme> => {
+  const integration = getClient()?.getIntegrationByName<ReturnType<typeof feedbackIntegration>>(
+    MOBILE_FEEDBACK_INTEGRATION_NAME,
+  );
+  if (!integration) {
+    return {};
+  }
```
krystofwoldrich, line 75: The helper functions look good, but since we are using them more and more, let's extract the core (integration getter) to a helper to avoid repeating it.

packages/core/src/js/feedback/FeedbackWidget.theme.ts
```diff
@@ -0,0 +1,41 @@
+import { Appearance } from 'react-native';
+
+import { getColorScheme, getFeedbackDarkTheme, getFeedbackLightTheme } from './integration';
+
+/**
+ * Get the theme for the feedback widget based on the current color scheme
+ */
+export function getTheme(): FeedbackWidgetTheme {
```
krystofwoldrich, line 8: The get theme is being called when the styles `consts` are created, which is before `Sentry.init`, and thus, the `feedbackIntegration` theme options are ignored.

I assume this might be solved in the reactive follow up, but it seems like a bug.

packages/core/src/js/feedback/FeedbackWidget.theme.ts
```diff
@@ -0,0 +1,41 @@
...
+  const userTheme = getColorScheme();
+  const colorScheme = userTheme === 'system' ? Appearance.getColorScheme() : userTheme;
+  const lightTheme = { ...LightTheme, ...getFeedbackLightTheme() };
+  const darkTheme = { ...DarkTheme, ...getFeedbackDarkTheme() };
+  return colorScheme === 'dark' ? darkTheme : lightTheme;
+}
+
+export interface FeedbackWidgetTheme {
+  accent: string;
```
krystofwoldrich, line 17: I think we are missing some properties compared to JS.

https://github.com/getsentry/sentry-javascript/blob/a3e44f222caef518ea1f006dc75c1a0c98b2617e/packages/core/src/types-hoist/feedback/theme.ts#L13-L31

krystofwoldrich:
Overall, it's looking good, but I've noticed some issues with customizing the theme.

We should add tests to ensure that the correct values are applied when using custom styles and custom themes.

packages/core/src/js/feedback/FeedbackWidget.theme.ts
antonis: Good catch 👍 

I tried to use just one accent for simplicity but this resulted in the [issue reported above](https://github.com/getsentry/sentry-react-native/pull/4677#discussion_r2014646665).

> The rest is probably not applicable to the mobile version, right?

I think too the rest do not apply to the mobile version.

Updated with c831eeeac0886bed343c0755c7d8bed37cf8d984

packages/core/src/js/feedback/FeedbackWidget.styles.ts
antonis: Good catch Krystof 👍 
This escaped my attention. Updated with c831eeeac0886bed343c0755c7d8bed37cf8d984 and also updated the screenshots in the PR description to reflect the latest state.

packages/core/src/js/feedback/integration.ts
antonis: Good idea 👍 
Extracted a helper function with 6e7b61ba78de9f3cebc6056f7dc2f826fd4b5500

packages/core/src/js/feedback/FeedbackWidget.theme.ts
antonis: Thank you for the feedback @krystofwoldrich 🙇 

I wasn't able to reproduce with the sample app in dev mode but I agree that this is a bug and the code needs restructuring. I'll iterate on this asap.

<details><summary>Recording</summary>
<p>

https://github.com/user-attachments/assets/e429d993-bb90-458f-bd15-b0f73a85d7ce

</p>
</details>

krystofwoldrich: Thank you for giving it a try. It might have been my machine issue, I'try it again.

antonis:
Thank you for all your feedback @krystofwoldrich 🙇 

This should be ready for another go. Note that I now also cover the dynamic theme changes on this PR with 7a331b6c9c892dd4ccb773f7485c819aa66e1bce as part of the refactoring discussed [above](https://github.com/getsentry/sentry-react-native/pull/4677#discussion_r2016241991).

> We should add tests to ensure that the correct values are applied when using custom styles and custom themes.

I've added snapshot tests to cover this with 6c63d038b3f9d6aba6c135e71c34a53c3c694ffe

krystofwoldrich:
Looks and works great! Reacts to the changes live.

However, in the tests, only the custom dark and light scheme is checked.

Can you also add a test for the system theme option, to check that the correct scheme is used, and a test for the live change if possible?

antonis:

> Can you also add a test for the system theme option, to check that the correct scheme is used, and a test for the live change if possible?

Added test for the system theme option with ba9868565804f89c0c0fec662b8b6ae33105fd75
and tests for a 'live' change with 3b4a1de358182d5b7f118978ab95b2a08e11d70c

krystofwoldrich (approved):
Thank you for adding the extra tests!

Issue #4430, opened 2025-01-09
[ReactNavigation] Don't start navigation spans for non-navigation actions


krystofwoldrich:
### Description

The ReactNavigationIntegration doesn't read dispatch action data and starts a new navigation span on every dispatch.

---

[current integration implementation](https://github.com/getsentry/sentry-react-native/blob/f2d10b7bd12e10b57209f3a1651d1f37b33e08b1/packages/core/src/js/tracing/reactnavigation.ts#L160)

---

Related issue: 
- https://github.com/getsentry/sentry-react-native/issues/4429

krystofwoldrich:
We need to ensure adding this doesn't break support for ReactNavigation V5 and V6.

PR #4684, opened 2025-03-25
003 feat(tracing): Use React Navigation dispatched action data to discard noops

krystofwoldrich:

- [x] New feature
- [x] Enhancement

## :scroll: Description

This PR adds `useDispatchedActionData` when set to `true` the integration will use the action data to improve the created transactions. Mainly to avoid creating transactions for noop and non-navigation actions.

We only stop creation for known non-navigation to avoid issues with not creating for unknown navigation actions. This might be subject to change in the next major, depending on the feedback on the feature.

```js
Sentry.reactNavigationIntegration({
});
```

fixes: https://github.com/getsentry/sentry-react-native/issues/4430

integration tests, sample app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


packages/core/test/tracing/reactnavigation.test.ts
```diff
@@ -210,6 +210,41 @@ describe('ReactNavigationInstrumentation', () => {
     );
   });
 
+  test('navigation action not ', async () => {
```
antonis, line 213: m: This test case name and implementation looks incomplete. Should this test a different scenario than the above?

krystofwoldrich: Thank you, no just a bad copy paste.

antonis (approved):
LGTM 🚀

PR #4690, opened 2025-03-26
fix(xcode): Avoid silent failure when JS bundle creation fails due to Sentry Xcode scripts

krystofwoldrich:
Before: finishes with errors, which means the app might be missing the js bundle

Screenshot 2025-03-26 at 11 10 09

After: build fails (and only if explicitly allowed to continue silently then it continues)

Screenshot 2025-03-26 at 11 09 33


antonis (approved):
The code changes LGTM and the build now fails as it should 🚀 


I took the liberty to merge from main and revert a script error which I guessed was introduced while testing with https://github.com/getsentry/sentry-react-native/pull/4690/commits/dafde95660656554cfcdf682efd23bf6fbc78dce

lucas-zimerman (approved):
It would be great if apple replaced the error sign as warning on finished builds with errors haha.
Other than that, LGTM!

PR #4725, opened 2025-04-07
chore(expo-sample): Add Expo Updates

krystofwoldrich:
This PR adds the Expo Updates config to make it easier to test the sample application with EAS Update.

To use the EAS Update, follow https://docs.expo.dev/eas-update/getting-started/ and update the null identifiers using the EAS commands.


antonis (approved):
Thank you for this enhancement Krystof 🙇 
The changes LGMT and the EAS update worked as expected after setting the configuration for my test project (`slug`, `package`, `projectId`, `owner`, `url`).

Issue #4728, opened 2025-04-07
V7: JavaScript V9 - update type of profiler Props


lucas-zimerman:
### Description

This is a new issue create to track the status of https://github.com/getsentry/sentry-react-native/pull/4568#discussion_r2005203769

PR #4740, opened 2025-04-09
Fix: V7 - JavaScript V9 - set undefined for updateProd when not set

lucas-zimerman:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

- Added tests to the wrap function that were missing.
- Made `updateProps`  always set but with undefined value, avoiding break changes and making the compiler happy.
- Hide `updateProps` and `children` from wrap options.

Fixes: https://github.com/getsentry/sentry-react-native/issues/4728

Unit test, Sample app

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog.


lucas-zimerman:
As a follow-up from https://github.com/getsentry/sentry-react-native/pull/4568#discussion_r2005203769
@krystofwoldrich I have hidden the updateProps and children from the options.
When not set, I set `updateProps` as `undefined` since this was the behavior the PR isn't changed and this way the compiler gets happy with it.

https://github.com/getsentry/sentry-javascript/blob/56aac537ea2627baec9025291b73f46b48670abf/packages/react/src/profiler.tsx#L72-L76
As seen on the above code, we have a tiny margin that this could break (same behavior could happen before this PR) where we set props before triggering `shouldComponentUpdate` and then set it as undefined and call `shouldComponentUpdate`, could make it go inside the if condition and break Object.keys since updateProps is undefined.

That could be addressed on a follow up issue since this issue wasn't introduced by this PR.

antonis (approved):
The code changes and the tests LGMT and aligned with the previous discussion https://github.com/getsentry/sentry-react-native/pull/4568#discussion_r2005203769 🚀

packages/core/test/sdk.wrap.test.ts
```diff
@@ -0,0 +1,97 @@
+import { render } from '@testing-library/react-native';
+import * as React from 'react';
+
+import { wrap } from '../src/js/sdk';
+
+jest.mock('../src/js/touchevents', () => ({
+  __esModule: true,
```
krystofwoldrich, line 7: Is the ESM flag required in this case?

krystofwoldrich:
There is already `wrap.test.tsx` in the tests. We should add the new tests there.

Since it's TSX, we can use the JSX syntax, which will make the test more readable.

packages/core/test/sdk.wrap.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    const renderResult = render(element);
+
+    const { getByTestId } = renderResult;
+    expect(getByTestId('touch-boundaryID')).toBeDefined();
+    expect(getByTestId('profilerID')).toBeTruthy();
+    expect(getByTestId('feedback-widgetID')).toBeTruthy();
+  });
+
+  describe('ReactNativeProfiler', () => {
+    it('uses given options when set', () => {
```
krystofwoldrich, line 45: Since we removed the `updateProps` from the type, I think the option should be ignored.

packages/core/test/sdk.wrap.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    it('sets updateProps when not defined', () => {
+      const Wrapped = wrap(DummyComponent);
+      const element = React.createElement(Wrapped, { value: 'wrapped' });
+      render(element);
+
+      expect(ReactNativeProfiler).toHaveBeenCalledWith(
+        expect.objectContaining({
+          name: 'Root',
+          updateProps: undefined,
```
krystofwoldrich, line 70: It would be safer to default to an empty object to avoid reading props from undefined, as you mentioned.

packages/core/src/js/sdk.tsx
```diff
@@ -153,12 +153,11 @@ export function init(passedOptions: ReactNativeOptions): void {
  */
 export function wrap<P extends Record<string, unknown>>(
   RootComponent: React.ComponentType<P>,
-  options?: ReactNativeWrapperOptions
+  options?:  Pick<Partial<ReactNativeWrapperOptions>, Exclude<keyof ReactNativeWrapperOptions, 'updateProps' | 'children'>>
```
krystofwoldrich, line 156: This won't work; we need to update the `profilerProps` in the `ReactNativeWrapperOptions` type.

The `ReactNativeWrapperOptions` don't have `updateProps`.

packages/core/test/sdk.wrap.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      );
+    });
+
+    it('overwrites Root when component has a name', () => {
+      const NamedDummyComponent: React.FC<{ value?: string }> = ({ value }) =>
+        React.createElement('div', { displayName: 'custom Name' }, value);
+      NamedDummyComponent.displayName = 'custom Name';
+
+      const options: ReactNativeWrapperOptions = {
+        profilerProps: { name: 'custom Name', updateProps: { update: true } },
```
krystofwoldrich, line 82: Let's also drop the `name` from the profiler type since it's defined inside of the `wrap` implementation and the user-set option is ignored.

krystofwoldrich (changes requested):
I noticed a few things that should be addressed before merging.

packages/core/test/sdk.wrap.test.ts
lucas-zimerman: it wasn't required, I have removed it from the tests on https://github.com/getsentry/sentry-react-native/pull/4740/commits/a38cbcb2b699583c1bcc8daf3a3efcac1aef825c 🙏

packages/core/test/sdk.wrap.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  __esModule: true,
+  FeedbackWidgetProvider: ({ children }: { children: React.ReactNode }) =>
+    React.createElement('div', { testID: 'feedback-widgetID' }, children),
+}));
+
+import type { ReactNativeWrapperOptions } from 'src/js/options';
+
+import { ReactNativeProfiler } from '../src/js/tracing';
+
+// Test
```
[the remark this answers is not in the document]
lucas-zimerman: fixed on https://github.com/getsentry/sentry-react-native/pull/4740/commits/a38cbcb2b699583c1bcc8daf3a3efcac1aef825c

packages/core/test/sdk.wrap.test.ts
lucas-zimerman: fixed on https://github.com/getsentry/sentry-react-native/pull/4740/commits/a38cbcb2b699583c1bcc8daf3a3efcac1aef825c

packages/core/src/js/sdk.tsx
lucas-zimerman: I have updated on https://github.com/getsentry/sentry-react-native/pull/4740/commits/a38cbcb2b699583c1bcc8daf3a3efcac1aef825c to to update the ReactNativeWrapperOptions

packages/core/test/sdk.wrap.test.ts
lucas-zimerman: Fixed on https://github.com/getsentry/sentry-react-native/pull/4740/commits/359153bdf5d5d6bbce7f52f127e09bdd064052de

lucas-zimerman:
> I noticed a few things that should be addressed before merging.

I can't change `ReactNativeProfiler` to make `updateProps` optional, since it requires me to make a large rewrite or alter `Profiler` from `@sentry/react` to make it optional.

As a way to avoid a big change, I added a flag to remove this parameter when it is included by our wrap function.
Let me know if you have any opinion in regard to it

krystofwoldrich:
> I can't change ReactNativeProfiler to make updateProps optional, since it requires me to make a large rewrite or alter Profiler from @sentry/react to make it optional.

> As a way to avoid a big change, I added a flag to remove this parameter when it is included by our wrap function.
Let me know if you have any opinion in regard to it

Thank you, yes, makes sense, let's not change the types in JS React, just updating the types in the `wrap` is good.

packages/core/src/js/sdk.tsx
```diff
@@ -153,18 +153,18 @@ export function init(passedOptions: ReactNativeOptions): void {
  */
 export function wrap<P extends Record<string, unknown>>(
   RootComponent: React.ComponentType<P>,
-  options?: ReactNativeWrapperOptions
+  options?:  ReactNativeWrapperOptions
```
krystofwoldrich, line 156:
````suggestion
  options?: ReactNativeWrapperOptions
````

File: packages/core/src/js/sdk.tsx [hunk 8]
```diff
@@ -153,18 +153,18 @@ export function init(passedOptions: ReactNativeOptions): void {
...
     name: RootComponent.displayName ?? 'Root',
-    updateProps: {}
+    updateProps: options?.profilerProps
   };
 
   const RootApp: React.FC<P> = (appProps) => {
     return (
       <TouchEventBoundary {...(options?.touchEventBoundaryProps ?? {})}>
-        <ReactNativeProfiler {...profilerProps}>
+        <ReactNativeProfiler {...profilerProps }>
```
krystofwoldrich, line 167:
````suggestion
        <ReactNativeProfiler {...profilerProps}>
````

File: packages/core/src/js/sdk.tsx [hunk 9]
```diff
@@ -153,18 +153,18 @@ export function init(passedOptions: ReactNativeOptions): void {
...
   RootComponent: React.ComponentType<P>,
-  options?: ReactNativeWrapperOptions
+  options?:  ReactNativeWrapperOptions
 ): React.ComponentType<P> {
   const profilerProps = {
-    ...(options?.profilerProps ?? {}),
+    ...(options?.profilerProps ?? { removeUpdateProps: true }),
     name: RootComponent.displayName ?? 'Root',
-    updateProps: {}
+    updateProps: options?.profilerProps
```
krystofwoldrich, line 161: I thought we would keep the `updateProps: {}`?

lucas-zimerman: Sorry, probably a misunderstanding, fixed here https://github.com/getsentry/sentry-react-native/pull/4740/commits/517161cb2485951ec396690a41df37d412b0e3cd

lucas-zimerman:
@krystofwoldrich I had to split the mocked and real tests from the wrap due to internal conflicts with the internal libraries when interacting with both on the same file (I can't properly remove the mocks otherwise it breaks the React Native testing hooks).

packages/core/src/js/options.ts
```diff
@@ -294,7 +294,7 @@ export interface ReactNativeClientOptions
 
 export interface ReactNativeWrapperOptions {
   /** Props for the root React profiler */
-  profilerProps?: ProfilerProps;
+  profilerProps?: Pick<Partial<ProfilerProps>, Exclude<keyof ProfilerProps, 'updateProps' | 'children' | 'name'>>;
```
krystofwoldrich, line 297: Both variants have the same result with the current `ProfilerProps`, but with Omit is a bit shorter.

````suggestion
profilerProps?: Omit<ProfilerProps, 'updateProps' | 'children' | 'name'>
````

File: packages/core/src/js/tracing/reactnativeprofiler.tsx [hunk 11]
```diff
@@ -12,13 +12,24 @@ const ReactNativeProfilerGlobalState = {
   },
 };
 
+type ProfilerConstructorProps = ConstructorParameters<typeof Profiler>[0];
+type ReactNativeProfilerConstructorProps = ProfilerConstructorProps &
```
krystofwoldrich, line 16: We don't use the extra `removeUpdateProps` anymore.

packages/core/test/wrap.mocked.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+import type { ReactNativeWrapperOptions } from 'src/js/options';
+
+  jest.doMock('../src/js/touchevents', () => {
+    return {
+          TouchEventBoundary: ({ children }: { children: React.ReactNode }) => (
+            // eslint-disable-next-line react/no-unknown-property
+            <div testID="touch-boundaryID">{children}</div>
+          ),
+        }
+  });
```
krystofwoldrich, line 13:
````suggestion
jest.doMock('../src/js/touchevents', () => {
  return {
        TouchEventBoundary: ({ children }: { children: React.ReactNode }) => (
          // eslint-disable-next-line react/no-unknown-property
          <div testID="touch-boundaryID">{children}</div>
        ),
      }
});
````

File: packages/core/test/wrap.mocked.test.tsx [hunk 13]
```diff
@@ -0,0 +1,117 @@
...
+
+  jest.doMock('../src/js/feedback/FeedbackWidgetManager', () => {
+    return {
+          FeedbackWidgetProvider: ({ children }: { children: React.ReactNode }) => (
+            // eslint-disable-next-line react/no-unknown-property
+            <div testID="feedback-widgetID">{children}</div>
+          ),
+      };
+  });
```
krystofwoldrich, line 31:
````suggestion
        ReactNativeProfiler: jest.fn(({ children }: { children: React.ReactNode }) => (
          // eslint-disable-next-line react/no-unknown-property
          <div testID="profilerID">{children}</div>
        )),
      }
});

jest.doMock('../src/js/feedback/FeedbackWidgetManager', () => {
  return {
        FeedbackWidgetProvider: ({ children }: { children: React.ReactNode }) => (
          // eslint-disable-next-line react/no-unknown-property
          <div testID="feedback-widgetID">{children}</div>
        ),
    };
});
````

packages/core/src/js/tracing/reactnativeprofiler.tsx
lucas-zimerman: Good catch! I have removed it on https://github.com/getsentry/sentry-react-native/pull/4740/commits/cf5c90fa29082e70b0d1ac905f54cbc34e92e886

krystofwoldrich (approved):
Thank you for all the updates and fixing these props. Looks good.

PR #4746, opened 2025-04-11
fix(app-start): Use ms to calculate app start age threshold

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

The app start age check, compared seconds (transaction start) with milliseconds (app start).

The test were passing because the test app start was also too long, so it was rejected by the next check in order.

unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


antonis (approved):
LGTM 🚀

PR #4753, opened 2025-04-15
Kw-remove-sythetic-default-imports

krystofwoldrich:
This flag is not necessary, we are not importing non existing default anywhere in the SDK code.

#skip-changelog


lucas-zimerman (approved):
LGTM!

PR #4756, opened 2025-04-16
fix(changelog): Mode JS and Android deps to the unreleased section

krystofwoldrich:
#skip-changelog


antonis (approved):
Thank you for catching that Krystof 🙇 
Just noticed this while bumping Android in https://github.com/getsentry/sentry-react-native/pull/4755 but it makes sense to be a separate PR.

PR #4760, opened 2025-04-16
chore(typescript): Enable strictNullChecks for the SDK source

krystofwoldrich:
These are enabled in JS Core since the last major v8. 

Let's enable them in RN as well.

This PR should not introduce any behavior changes, only fix the null checks.

#skip-changelog


antonis (approved):
Thank you for enabling `strictNullChecks` and improving the code along the way 🎸

PR #4809, opened 2025-05-06
fix(expo-context): All values should be lowercase

krystofwoldrich:

- [x] Bugfix
- [x] Enhancement

## :scroll: Description

From testing we noticed that on iOS the values are upper case.

unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/src/js/integrations/expocontext.ts
+++ b/packages/core/src/js/integrations/expocontext.ts
@@ -89,16 +89,16 @@ export function getExpoUpdatesContext(): ExpoUpdatesContext {
   };
 
   if (typeof expoUpdates.updateId === 'string' && expoUpdates.updateId) {
-    updatesContext.update_id = expoUpdates.updateId.toLocaleLowerCase();
+    updatesContext.update_id = expoUpdates.updateId.toLowerCase();
   }
   if (typeof expoUpdates.channel === 'string' && expoUpdates.channel) {
-    updatesContext.channel = expoUpdates.channel.toLocaleLowerCase();
+    updatesContext.channel = expoUpdates.channel.toLowerCase();
   }
   if (typeof expoUpdates.runtimeVersion === 'string' && expoUpdates.runtimeVersion) {
-    updatesContext.runtime_version = expoUpdates.runtimeVersion.toLocaleLowerCase();
+    updatesContext.runtime_version = expoUpdates.runtimeVersion.toLowerCase();
   }
   if (typeof expoUpdates.checkAutomatically === 'string' && expoUpdates.checkAutomatically) {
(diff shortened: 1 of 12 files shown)


packages/core/src/js/integrations/expocontext.ts
```diff
@@ -85,19 +88,19 @@ function getExpoUpdatesContext(): ExpoUpdatesContext {
     is_using_embedded_assets: !!expoUpdates.isUsingEmbeddedAssets,
   };
 
-  if (typeof expoUpdates.updateId === 'string') {
-    updatesContext.update_id = expoUpdates.updateId;
+  if (typeof expoUpdates.updateId === 'string' && expoUpdates.updateId) {
+    updatesContext.update_id = expoUpdates.updateId.toLocaleLowerCase();
```
lucas-zimerman, line 92: q: Why not just toLowerCase? It should be a GUID only and be language-independent.
```suggestion
    updatesContext.update_id = expoUpdates.updateId.toLowerCase();
```

packages/core/src/js/integrations/expocontext.ts
```diff
@@ -85,19 +88,19 @@ function getExpoUpdatesContext(): ExpoUpdatesContext {
...
+  if (typeof expoUpdates.runtimeVersion === 'string' && expoUpdates.runtimeVersion) {
+    updatesContext.runtime_version = expoUpdates.runtimeVersion.toLocaleLowerCase();
   }
-  if (typeof expoUpdates.checkAutomatically === 'string') {
-    updatesContext.check_automatically = expoUpdates.checkAutomatically;
+  if (typeof expoUpdates.checkAutomatically === 'string' && expoUpdates.checkAutomatically) {
+    updatesContext.check_automatically = expoUpdates.checkAutomatically.toLocaleLowerCase();
   }
-  if (typeof expoUpdates.emergencyLaunchReason === 'string') {
+  if (typeof expoUpdates.emergencyLaunchReason === 'string' && expoUpdates.emergencyLaunchReason) {
```
lucas-zimerman, line 103: I may risk saying the same applies to the other fields, would it be a specific reason to use toLocaleLowerCase on the altered code above compared to toLowerCase?

packages/core/src/js/integrations/expocontext.ts
krystofwoldrich: Thank you, yes that the local was not intentional.

packages/core/src/js/integrations/expocontext.ts
krystofwoldrich: Yes, exactly applies to all of the lowercased items.

lucas-zimerman (approved):
LGTM!

PR #4810, opened 2025-05-06
chore(expo-sample): Add Expo Dev Client

krystofwoldrich:
The Dev Client makes it easier to test different EAS Updates as it allows developers to loging to their EAS Account and pick an update in the Dev Menu.

https://docs.expo.dev/versions/latest/sdk/dev-client/

#skip-changelog

### DIFF
--- a/samples/expo/app/(tabs)/index.tsx
+++ b/samples/expo/app/(tabs)/index.tsx
@@ -9,6 +9,7 @@ import { setScopeProperties } from '@/utils/setScopeProperties';
 import React from 'react';
 import * as WebBrowser from 'expo-web-browser';
 import { useUpdates } from 'expo-updates';
+import { isWeb } from '../../utils/isWeb';
 
 const isRunningInExpoGo = Constants.appOwnership === 'expo';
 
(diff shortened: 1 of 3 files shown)


samples/expo/app/(tabs)/index.tsx
```diff
@@ -20,6 +21,12 @@ export default function TabOneScreen() {
       <Text>Update ID: {currentlyRunning.updateId}</Text>
       <Text>Channel: {currentlyRunning.channel}</Text>
       <Text>Runtime Version: {currentlyRunning.runtimeVersion}</Text>
+      <Button
+        title="Open DevMenu"
+        onPress={() => {
+          DevClient.openMenu();
```
lucas-zimerman, line 27: Q: what if we disable this button on the web? It seems to only run on mobile
image

antonis: I think we should since [the documentation](https://docs.expo.dev/versions/latest/sdk/dev-client/) does not include `Web` as a supported platform 👍

krystofwoldrich: Thank you that's a good catch.

antonis (approved):
LGTM 🚀

PR #4815, opened 2025-05-07
ref(session-replay): iOS: Use enableViewRendererV2 instead of the dep…

antonis:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Use `enableViewRendererV2` which defaults to `true` instead of the renamed/deprecated `enableExperimentalViewRenderer`

`enableExperimentalViewRenderer` was [renamed/deprecated](https://github.com/getsentry/sentry-cocoa/blob/fb7eac99f7e336ec9c5b9cb74b28ff50b3aa409d/Sources/Swift/Integrations/SessionReplay/SentryReplayOptions.swift#L115) in [8.50.0](https://github.com/getsentry/sentry-react-native/pull/4807) with https://github.com/getsentry/sentry-cocoa/pull/5054

CI, Manual

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


packages/core/src/js/replay/mobilereplay.ts
```diff
@@ -40,10 +40,24 @@ export interface MobileReplayOptions {
    *
    * - Experiment: This is an experimental feature and is therefore disabled by default.
    *
-   * @default false
+   * @deprecated Use `enableViewRendererV2` instead.
    */
   enableExperimentalViewRenderer?: boolean;
 
+  /**
+   * Enables the up to 5x faster new view renderer used by the Session Replay integration on iOS.
```
lucas-zimerman, line 48: NIT: 
```suggestion
   * Enables up to 5x faster new view renderer used by the Session Replay integration on iOS.
```

lucas-zimerman (approved):
Nice addition! 
I added a small wording that will help on the reading, after resolving that comment, LGTM!

packages/core/src/js/replay/mobilereplay.ts
antonis: Good catch 🙇

packages/core/ios/RNSentryReplay.mm
```diff
@@ -27,8 +27,7 @@ + (void)updateOptions:(NSMutableDictionary *)options
         @"errorSampleRate" : options[@"replaysOnErrorSampleRate"] ?: [NSNull null],
         @"maskAllImages" : replayOptions[@"maskAllImages"] ?: [NSNull null],
         @"maskAllText" : replayOptions[@"maskAllText"] ?: [NSNull null],
-        @"enableExperimentalViewRenderer" : replayOptions[@"enableExperimentalViewRenderer"]
-            ?: [NSNull null],
+        @"enableViewRendererV2" : replayOptions[@"enableViewRendererV2"] ?: @YES,
```
krystofwoldrich, line 30: I would keep here `?: [NSNull null]`, to avoid baking in any logic in here. 

If null cocoa uses its default if not values is coming from RN (user or default).

krystofwoldrich: I see `sentry-cocoa` set's it to `false`. That was an oversight, I've talked to @philprime and we will update this in cocoa. 

So I would hold off this PR until cocoa hotfix release.

antonis: > So I would hold off this PR until cocoa hotfix release.

Sounds good @krystofwoldrich 👍 

> I would keep here ?: [NSNull null], to avoid baking as little logic as possible in here.

Updated with 88280eb76756d1cb99474f615f1b5fcb80a0abc3

philprime: Related PR to fix the inconsistency in sentry-cocoa:
https://github.com/getsentry/sentry-cocoa/pull/5210

antonis: The Cocoa change was merged with `8.50.2` https://github.com/getsentry/sentry-react-native/pull/4830

Issue #3957, opened 2024-07-17
Promise.allSettled is undefined


ignaciosantise:
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [x]  iOS
- [x]  Android

**SDK:**
- [x]  `@sentry/react-native` (>= 1.0.0)
- [ ]  `react-native-sentry` (<= 0.43.2)

**SDK version:** 0.0.0

**`react-native` version:** 5.26.0

**Are you using Expo?**
- [x]  Yes
- [ ]  No

**Are you using sentry.io or on-premise?**
- [x]  sentry.io (SaaS)
- [ ]  on-premise

**If you are using sentry.io, please post a link to your issue so we can take a look:**

[Link](https://walletconnect.sentry.io/issues/5237253717/?project=4506067860717568&query=is:unresolved+issue.priority:%5Bhigh,+medium%5D&statsPeriod=14d&stream_index=0)

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  dsn: 'https://...@sentry.io/...'
});
```

---
**I have the following issue:**

`Promise.allSettled` is undefined when Sentry is enabled in the app

**Steps to reproduce:**
- call `Promise.allSettled` function in any part of the app
- See that it throws the error "Promise.allSettled is not a function"
- Disable Sentry (by commenting `Sentry.init` code)
- call `Promise.allSettled` function in any part of the app
- Check that now it works as expected

**Actual result:**

error "Promise.allSettled is not a function"

**Expected result:**

The function works as expected

**Minimal Reproducible Example**
I created a new project with Expo and added Sentry using the Wizard. Check the readme to reproduce the issue
https://github.com/ignaciosantise/sentry-issue

**Screenshots**

#### Sentry enabled

https://github.com/user-attachments/assets/29427948-19ed-4260-b5a4-68994c55161f

#### Sentry disabled

https://github.com/user-attachments/assets/af4502b6-0d8c-4558-b404-014ac1be6c49

krystofwoldrich:
Hi @ignaciosantise,
you appear to have multiple versions of the "promise" package installed. This may cause unexpected behavior like undefined `Promise.allSettled`. Please install the `promise` package manually using the exact version as the React Native package.

In your case that is:
```bash
npm i promise@8.3.0
```

ignaciosantise:
Hi @krystofwoldrich 👋 thanks for the info

[...]

PR #4457, opened 2025-01-16
Fix: Add unhandled rejection mechanism

lucas-zimerman:
## :loudspeaker: Type of change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Unhandled rejection were lacking the Mechanism field, so I added it.

## :bulb: Motivation and Context

https://github.com/getsentry/sentry-react-native/issues/4093

## :green_heart: How did you test it?

Unit test and https://sentry-sdks.sentry.io/issues/6018946733/events/91e85dfdbd5340c487e7d279378441d4/?project=5428561

## :pencil: Checklist

- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] All tests passing
- [ ] No breaking changes

Issue #4141, opened 2024-10-02
Mark unhandled promise rejections as `handled: false`


krystofwoldrich:
### 🛑 Blocked by backend Release Health changes

This change would require change of calculating crash rate independent of `handled: false`. As rejected promises do not cause a crash in RN.

### Description

We can't do this at the moment because the `handled` flag is used to calculate the Crash Rate in the Release Health product.

- Related to https://github.com/getsentry/sentry-react-native/issues/4093
- Related to Stale RFC https://github.com/getsentry/rfcs/pull/10
- The same issue as Flutter Unhandled Errors which do not cause crash of the app

PR #4826, opened 2025-05-13
fix(promise): Use engine-specific promise rejection tracking

antonis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

- Uses Hermes's native tracker when available
- Leverages Sentry Browser SDK for React Native Web
- Falls back to existing implementation for JSC

Fixes https://github.com/getsentry/sentry-react-native/issues/3957

CI, Manual ([mobile](https://sentry-sdks.sentry.io/issues/6018946733/events/latest/?project=5428561&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=latest-event&stream_index=0), [web](https://sentry-sdks.sentry.io/issues/6604676621/events/latest/?project=5428561&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=latest-event&stream_index=0))

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/src/js/integrations/reactnativeerrorhandlers.ts
+++ b/packages/core/src/js/integrations/reactnativeerrorhandlers.ts
@@ -74,8 +74,7 @@ function setupUnhandledRejectionsTracking(patchGlobalPromise: boolean): void {
         },
         onHandled: (id: string) => {
           if (__DEV__) {
-            // eslint-disable-next-line no-console
-            console.warn(
+            logger.warn(
               `Promise Rejection Handled (id: ${id})\n` +
                 'This means you can ignore any previous messages of the form ' +
                 `"Possible Unhandled Promise Rejection (id: ${id})"`,
--- a/packages/core/src/js/integrations/reactnativeerrorhandlers.ts
+++ b/packages/core/src/js/integrations/reactnativeerrorhandlers.ts
@@ -96,7 +96,10 @@ function setupUnhandledRejectionsTracking(patchGlobalPromise: boolean): void {
         });
       });
 
-      logger.log('Unhandled promise rejections will be caught by Sentry.');
+      logger.warn(
+        'Unhandled promise rejections will not be caught by Sentry. ' +
+          'See https://docs.sentry.io/platforms/react-native/troubleshooting/ for more details.',
+      );
     } else if (patchGlobalPromise) {
       // For JSC and other environments, use the existing approach
       polyfillPromise();
@@ -107,14 +110,11 @@ function setupUnhandledRejectionsTracking(patchGlobalPromise: boolean): void {
       logger.log('Promise rejection tracking is disabled by configuration.');
     }
   } catch (e) {
-    // Use setTimeout to avoid issues with early logging
-    setTimeout(() => {
-      logger.warn(
-        'Failed to set up promise rejection tracking. ' +
-          'Unhandled promise rejections will not be caught by Sentry. ' +
-          'Read about how to fix this on our troubleshooting page.',
-      );
-    }, 0);
+    logger.warn(
+      'Failed to set up promise rejection tracking. ' +
+        'Unhandled promise rejections will not be caught by Sentry.' +
+        'See https://docs.sentry.io/platforms/react-native/troubleshooting/ for more details.',
+    );
   }
 }
 
--- a/packages/core/src/js/integrations/reactnativeerrorhandlers.ts
+++ b/packages/core/src/js/integrations/reactnativeerrorhandlers.ts
@@ -64,6 +64,8 @@ function setupUnhandledRejectionsTracking(patchGlobalPromise: boolean): void {
             console.warn(`Possible Unhandled Promise Rejection (id: ${id}):\n${error}`);
           }
 
+          // Marking the rejection as handled to avoid breaking crash rate calculations.
+          // See: https://github.com/getsentry/sentry-react-native/issues/4141
           captureException(error, {
             data: { id },
             originalException: error,
@@ -143,6 +145,8 @@ function attachUnhandledRejectionHandler(): void {
         promiseRejectionTrackingOptions.onUnhandled(id, error);
       }
 
+      // Marking the rejection as handled to avoid breaking crash rate calculations.
+      // See: https://github.com/getsentry/sentry-react-native/issues/4141
       captureException(error, {
         data: { id },
         originalException: error,
--- a/packages/core/src/js/integrations/reactnativeerrorhandlers.ts
+++ b/packages/core/src/js/integrations/reactnativeerrorhandlers.ts
@@ -62,29 +62,8 @@ function setupUnhandledRejectionsTracking(patchGlobalPromise: boolean): void {
 
       RN_GLOBAL_OBJ.HermesInternal.enablePromiseRejectionTracker({
         allRejections: true,
-        onUnhandled: (id: string, error: unknown) => {
-          if (__DEV__) {
-            logger.warn(`Possible Unhandled Promise Rejection (id: ${id}):\n${error}`);
-          }
-
-          // Marking the rejection as handled to avoid breaking crash rate calculations.
-          // See: https://github.com/getsentry/sentry-react-native/issues/4141
-          captureException(error, {
-            data: { id },
-            originalException: error,
-            syntheticException: isErrorLike(error) ? undefined : createSyntheticError(),
-            mechanism: { handled: true, type: 'onunhandledrejection' },
-          });
-        },
-        onHandled: (id: string) => {
-          if (__DEV__) {
-            logger.warn(
-              `Promise Rejection Handled (id: ${id})\n` +
-                'This means you can ignore any previous messages of the form ' +
-                `"Possible Unhandled Promise Rejection (id: ${id})"`,
-            );
-          }
-        },
+        onUnhandled: promiseRejectionTrackingOptions.onUnhandled,
+        onHandled: promiseRejectionTrackingOptions.onHandled,
       });
 
       logger.log('Unhandled promise rejections will be caught by Sentry.');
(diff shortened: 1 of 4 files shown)


packages/core/src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -44,12 +52,70 @@ function setup(options: ReactNativeErrorHandlersOptions): void {
...
+          if (__DEV__) {
+            // eslint-disable-next-line no-console
+            console.warn(`Possible Unhandled Promise Rejection (id: ${id}):\n${error}`);
+          }
+
+          captureException(error, {
+            data: { id },
+            originalException: error,
+            syntheticException: isErrorLike(error) ? undefined : createSyntheticError(),
+            mechanism: { handled: true, type: 'onunhandledrejection' },
```
lucas-zimerman, line 71: why handled true here?

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -44,12 +52,70 @@ function setup(options: ReactNativeErrorHandlersOptions): void {
...
+  try {
+    if (isHermesEnabled() && RN_GLOBAL_OBJ.HermesInternal?.enablePromiseRejectionTracker) {
+      logger.log('Using Hermes native promise rejection tracking');
+
+      RN_GLOBAL_OBJ.HermesInternal.enablePromiseRejectionTracker({
+        allRejections: true,
+        onUnhandled: (id: string, error: unknown) => {
+          if (__DEV__) {
+            // eslint-disable-next-line no-console
+            console.warn(`Possible Unhandled Promise Rejection (id: ${id}):\n${error}`);
```
lucas-zimerman, line 64: I would use logger.warn here so that users can opt out from seeing this log in production.

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -44,12 +52,70 @@ function setup(options: ReactNativeErrorHandlersOptions): void {
...
+            data: { id },
+            originalException: error,
+            syntheticException: isErrorLike(error) ? undefined : createSyntheticError(),
+            mechanism: { handled: true, type: 'onunhandledrejection' },
+          });
+        },
+        onHandled: (id: string) => {
+          if (__DEV__) {
+            // eslint-disable-next-line no-console
+            console.warn(
```
lucas-zimerman, line 77: same here

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -44,12 +52,70 @@ function setup(options: ReactNativeErrorHandlersOptions): void {
...
+        'Unhandled promise rejections will not be caught by Sentry. ' +
+          'See https://docs.sentry.io/platforms/react-native/troubleshooting/ for more details.',
+      );
+    } else if (patchGlobalPromise) {
+      // For JSC and other environments, use the existing approach
+      polyfillPromise();
+      attachUnhandledRejectionHandler();
+      checkPromiseAndWarn();
+    } else {
+      // Patching was disabled by user configuration
```
lucas-zimerman, line 109: how users can disable this configuration on web? it will always enter on the else if on line 87

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
antonis: Good question @lucas-zimerman 👍 
I kept [the same `captureException` call from the existing `attachUnhandledRejectionHandler` below](https://github.com/getsentry/sentry-react-native/blob/11f67eb3cf750fae2a7182d0948450b6fb288791/packages/core/src/js/integrations/reactnativeerrorhandlers.ts#L84).

lucas-zimerman: For reference https://github.com/getsentry/sentry-react-native/issues/4141
And https://github.com/getsentry/sentry-react-native/pull/4457#discussion_r1922179466
```
We have to mark the rejection handled: true to avoid breaking crash rate calculations.

Rejections do not crash the app in RN at the moment.
```
Might be worth adding a small comment in the comment pointing to https://github.com/getsentry/sentry-react-native/issues/4141

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
antonis: Good point. On the web we always use the [browser rejection tracking](https://github.com/getsentry/sentry-react-native/issues/3957#issuecomment-2262487656). I updated the comment and log to be more clear. Let me know it that makes sense 🙏

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
antonis: Updated with 4e2149c7029e18e0a130a6341b589b8688b0d97f

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
antonis: Thank you for the context @lucas-zimerman 🙇 
I didn't know that. I've added comments with 92ae70abd4b415f60c72dd8b60a234e5c208e4de

lucas-zimerman (approved):
LGTM!

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -44,12 +52,70 @@ function setup(options: ReactNativeErrorHandlersOptions): void {
  * Setup unhandled promise rejection tracking
  */
 function setupUnhandledRejectionsTracking(patchGlobalPromise: boolean): void {
-  if (patchGlobalPromise) {
-    polyfillPromise();
-  }
+  try {
+    if (isHermesEnabled() && RN_GLOBAL_OBJ.HermesInternal?.enablePromiseRejectionTracker) {
```
krystofwoldrich, line 56: I think we should also check the `hasPromise` same as RN does it to ensure we are adding the tracker to the correct Promise.

https://github.com/facebook/react-native/blob/c89c5d7e3d981e5e11ae926bdd83567cbe61201a/packages/react-native/Libraries/Core/polyfillPromise.js#L25

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -44,12 +52,70 @@ function setup(options: ReactNativeErrorHandlersOptions): void {
...
+          }
+
+          // Marking the rejection as handled to avoid breaking crash rate calculations.
+          // See: https://github.com/getsentry/sentry-react-native/issues/4141
+          captureException(error, {
+            data: { id },
+            originalException: error,
+            syntheticException: isErrorLike(error) ? undefined : createSyntheticError(),
+            mechanism: { handled: true, type: 'onunhandledrejection' },
+          });
```
krystofwoldrich, line 73: Since the handling logic is the same for Hermes Promise and JSC Promise Polyfill I think we should share the impl to avoid discrepancies.

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -44,12 +52,70 @@ function setup(options: ReactNativeErrorHandlersOptions): void {
...
+          }
+        },
+      });
+
+      logger.log('Unhandled promise rejections will be caught by Sentry.');
+    } else if (isWeb()) {
+      logger.log('Using Browser JS promise rejection tracking for React Native Web');
+
+      // Use Sentry's built-in global unhandled rejection handler
+      addGlobalUnhandledRejectionInstrumentationHandler((error: unknown) => {
```
krystofwoldrich, line 91: Nice, make sense. Thank you for adding it.

krystofwoldrich:
Thank you for the improvement. It looks great, I only have a two small comments, on reusability of the handlers code and check of the `hasPromise` method.

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
antonis: Good idea Krystof 👍 
Updated with 9fd5226f326a7bbd131c8dc7db77e4c06566e5ac

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
antonis: Good point 👍 
Updated with 4ed5c880ff10f40a566a4c16915760244da91367 to share the handling logic

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -44,49 +52,88 @@ function setup(options: ReactNativeErrorHandlersOptions): void {
...
+          originalException: error,
+          syntheticException: isErrorLike(error) ? undefined : createSyntheticError(),
+          mechanism: { handled: false, type: 'onunhandledrejection' },
+        });
+      });
+
+      logger.warn(
+        'Unhandled promise rejections will not be caught by Sentry. ' +
+          'See https://docs.sentry.io/platforms/react-native/troubleshooting/ for more details.',
+      );
```
krystofwoldrich, line 85: I think this is a oversight, we setup the global rejections handler above, so the rejections will be captured.

````suggestion
````

antonis: Good catch @krystofwoldrich 👍 
This is indeed an oversight. No warning is needed there. Removed with a2fbf2bc6ad990d9653bd32ff6c7e03a935cb36b

krystofwoldrich (approved):
LGTM 🚀

PR #4827, opened 2025-05-13
(WIP) Log Tracing

lucas-zimerman:
## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] I added tests to verify changes
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] All tests passing
- [ ] No breaking changes

PR #4860, opened 2025-05-26
 chore(deps-v7): update Javascript SDK to v9.22.0

lucas-zimerman:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

This PR updates  the JavaScript SDK from 9.12.0 to 9.22.0.
It also fixes the new lint issues that were introduced when upgrading the SDK.

I also ignored the rule 'import/first' on tests since it impact the way mock works for the SDK.

https://github.com/getsentry/sentry-react-native/pull/4827 requires a newer SDK and since there were lots of lint issues I decided to split the PR so it doesn't get polluted with unrelated changes.

CI

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/.eslintrc.js
+++ b/packages/core/.eslintrc.js
@@ -38,6 +38,7 @@ module.exports = {
         '@typescript-eslint/no-empty-function': 'off',
         '@typescript-eslint/no-explicit-any': 'off',
         '@typescript-eslint/unbound-method': 'off',
+        'import/first': 'off',
       },
     },
     {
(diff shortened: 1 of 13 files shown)


lucas-zimerman:
Moved as draft since it broke some tests

antonis (approved):
LGTM and the sample app worked as expected in my tests on both platforms 🚀 

> It also fixes the new lint issues that were introduced when upgrading the SDK.

Thank you for fixing those 🙇 

> I also ignored the rule 'import/first' on tests since it impact the way mock works for the SDK.

👍

PR #4872, opened 2025-05-29
fix(sec): Reduct sensitive data from Maestro logs

antonis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Reduct sensitive data from Maestro logs.

#### Other considered approaches:
* Checked if [Maestro](https://github.com/mobile-dev-inc/Maestro) has a way to not emit this kind of data in [the command logs](https://docs.maestro.dev/troubleshooting/debug-output#maestro-logs) but there is no such option ❌ 
* Tried other ways of [passing the token in the Maestro cli command](https://github.com/getsentry/sentry-react-native/blob/bfbb0a1c3232341eb3cc8759bdab540d7f4873ee/dev-packages/e2e-tests/cli.mjs#L240). Unfortunately, Maestro is always logging the `defineVariablesCommand.env` section which shows all resolved environment variables. ❌
* A working alternative is [to remove the logs artifact completely](https://github.com/getsentry/sentry-react-native/blob/bfbb0a1c3232341eb3cc8759bdab540d7f4873ee/.github/workflows/e2e.yml#L417), but it is sometimes useful for debugging failures since Maestro tests may behave differently on CI ⚠️ 

Sensitive data from Maestro logs

* Tested the command locally with an actual maestro zip
* Tested the secret reduction script with 431eccd81aba2d87dae0d71157d68f665beb3db5 and [verified that the produced artifacts had the dummy secret redacted](https://github.com/getsentry/sentry-react-native/actions/runs/15346060387/job/43182418600?pr=4872) and the Github command logs (my [initial approach](https://github.com/getsentry/sentry-react-native/pull/4872/commits/9a2bfa836dd50d384af1f965132e49ba5ac5a7aa) leaked there) did not contain the secret either

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


antonis:
> I'd say with the ci test you did, it's enough validation for it to be approved without depending on a real secret validation.

Thank you for suggesting this approach for testing on CI @lucas-zimerman 🙇 
I agree that this should validate that the secret won't leak in the maestro logs.
I'll wait for a real secret to be deployed to make sure all tests are green before merging.

krystofwoldrich:
Thank you for fixing this so quickly!

.github/workflows/e2e.yml
```diff
@@ -414,6 +414,19 @@ jobs:
         if: ${{ matrix.platform == 'ios' }}
         run: ./dev-packages/e2e-tests/cli.mjs ${{ matrix.platform }} --test
 
+      - name: Redact sensitive data from logs
```
krystofwoldrich: The fix looks good to me.

I have one thought, should we include it directly in 

https://github.com/getsentry/sentry-react-native/blob/77680ecf266a860bfb896a48f2e1beb05d774bb2/dev-packages/e2e-tests/cli.mjs#L237-L247

to run the redaction every time the test command is run, not only in CI?

antonis: That's a great idea @krystofwoldrich 👍 
Added with 5542a0fd2d903619c411497256fb1fe934d32296 and reverted the workflow change

dev-packages/e2e-tests/cli.mjs
```diff
@@ -234,16 +234,35 @@ if (actions.includes('test')) {
...
-        --env=APP_ID="${appId}" \
-        --env=SENTRY_AUTH_TOKEN="${sentryAuthToken}" \
-        --debug-output maestro-logs \
-        --flatten-debug-output`,
-      {
-        stdio: 'inherit',
-        cwd: e2eDir,
-      },
-    );
+  try {
```
krystofwoldrich:
````suggestion
    try {
````

krystofwoldrich (approved):
LGTM 🚀

dev-packages/e2e-tests/cli.mjs
antonis: Good catch 🙏

PR #4874, opened 2025-05-30
chore(encoding): Use global TextEncode when available

krystofwoldrich:

- [x] Enhancement
- [X] Refactoring

## :scroll: Description

This PR uses global TextEncode when available. This change mainly improves performance of the envelope encoding before passing the envelope to the native layers. 

Since https://github.com/facebook/hermes/commit/8fb0496d426a8e50d00385148d5fb498a6daa312 Hermes for RN 0.74 includes native implementation of TextEncoder which is more performant than the JS polyfill we use at the moment.

sample apps, unit tests

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes ~ If users have a broken polyfill in their environment this could cause issues, but that's the case with all global classes (that why this should rather be included in a major release)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 ### Major Changes
 
 - `Sentry.captureUserFeedback` removed, use `Sentry.captureFeedback` instead ([#4855](https://github.com/getsentry/sentry-react-native/pull/4855))
-- Use global `TextEncoder` when available ([#4874](https://github.com/getsentry/sentry-react-native/pull/4874))
+- Use global `TextEncoder` (available with Hermes in React Native 0.74 or higher) to greatly improve envelope encoding performance. ([#4874](https://github.com/getsentry/sentry-react-native/pull/4874))
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 ### Major Changes
 
 - `Sentry.captureUserFeedback` removed, use `Sentry.captureFeedback` instead ([#4855](https://github.com/getsentry/sentry-react-native/pull/4855))
+- Use global `TextEncoder` when available ([#4874](https://github.com/getsentry/sentry-react-native/pull/4874))
 
 ### Changes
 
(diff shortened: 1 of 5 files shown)


packages/core/test/transports/encodePolyfill.test.ts
```diff
@@ -0,0 +1,33 @@
+import { SDK_VERSION } from '@sentry/core';
+import { encodePolyfill, globalEncodeFactory, useEncodePolyfill } from '../../src/js/transports/encodePolyfill';
```
lucas-zimerman: NIT: this is not used on the test and should fix the lint failure
```suggestion
import { encodePolyfill, useEncodePolyfill } from '../../src/js/transports/encodePolyfill';
```

lucas-zimerman:
The PR looks great! I think it would be helpful to include a changelog entry highlighting the performance improvement
something like:
`Improved envelope encoding performance by up to 35% when the global 'TextEncoder' is available.`

krystofwoldrich:
The speed comparison is not that simple, it can be 4x to 100x times, it depends.

... https://github.com/maksimlya/react-native-fast-encoder/

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 ### Major Changes
 
 - `Sentry.captureUserFeedback` removed, use `Sentry.captureFeedback` instead ([#4855](https://github.com/getsentry/sentry-react-native/pull/4855))
+- Use global `TextEncoder` when available ([#4874](https://github.com/getsentry/sentry-react-native/pull/4874))
```
lucas-zimerman: i: We could mention that this operation will increase encoding performance and also encourages folks to migrate to RN 0.74 or newer in order to enjoy this feature 
```suggestion
- Use global `TextEncoder` (available with Hermes in React Native 0.74 or higher) to greatly improve envelope encoding performance. ([#4874](https://github.com/getsentry/sentry-react-native/pull/4874))
```

lucas-zimerman (approved):
Added a changelog suggestion to mention the performance improvement without mentioning any big number of performance gain.
Other than that, LGTM!

PR #4876, opened 2025-06-02
fix(e2e): Remove auth token from launch params

krystofwoldrich:
This PR Remove unused token launch param from RN E2E tests.

#skip-changelog

### DIFF
--- a/dev-packages/e2e-tests/maestro/utils/launchTestAppClear.yml
+++ b/dev-packages/e2e-tests/maestro/utils/launchTestAppClear.yml
@@ -8,7 +8,6 @@ jsEngine: graaljs
 - launchApp:
     clearState: true
     arguments:
-      sentryAuthToken: ${SENTRY_AUTH_TOKEN}
       replaysOnErrorSampleRate: ${replaysOnErrorSampleRate}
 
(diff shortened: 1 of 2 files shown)


antonis (approved):
Thank you for looking into this @krystofwoldrich 🙇 

I've tested locally and the token still leaks in the `commands-*.json` files.

```
  "command" : {
    "defineVariablesCommand" : {
      "env" : {
        "APP_ID" : "org.reactjs.native.example.RnDiffApp",
        "SENTRY_AUTH_TOKEN" : "redacted-token-string"
      },
      "optional" : false
    }
  },
```

My understanding is that this leaks when [passing the token in the Maestro cli command](https://github.com/getsentry/sentry-react-native/blob/bfbb0a1c3232341eb3cc8759bdab540d7f4873ee/dev-packages/e2e-tests/cli.mjs#L240). but I haven't found a way to avoid this other than [redacting after the tests](https://github.com/getsentry/sentry-react-native/pull/4872).

Removing the auth token from the launch parameters does not affect the test results and it makes sense to remove them 👍

krystofwoldrich:
@antonis Thank you for testing this. And catching that the token is still in the `commands-*.json`. 🙏

Issue #4875, opened 2025-05-30
Crash on RNSentryOnDrawReporterView createEmitNewFrameEvent


pierricklete:
### What React Native libraries do you use?

React Native without Frameworks

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### @sentry/react-native SDK Version

6.14.0

### How does your development environment look like?

````
⬇  Place the `npx react-native@latest info` output here. ⬇ 

 react-native:
    installed: 0.76.6
    wanted: 0.76.6


````

### Sentry.init()

````js
Sentry.init({
  dsn: 'https://...@sentry.io/...'
  // other options
});
````

### Steps to Reproduce

Since I have upgraded `@sentry/react-native` from `6.10.0` to `6.14.0`, I got few crash reports caused  by `RNSentryOnDrawReporterView createEmitNewFrameEvent`

I can't reproduce it myself on simulator and real devices

from the report crash happens when app starts

We are using `<Sentry.TimeToInitialDisplay/>` and `<Sentry.TimeToFullDisplay/>` components on top level component (higher than `NavigationContainer` from `@react-navigation/native`)

### Expected Result

no crash as it was the case with `6.10.0`

### Actual Result

```
OS Version: iOS 18.5 (22F76)
Report Version: 104

Exception Type: EXC_CRASH (SIGABRT)
Crashed Thread: 0

Application Specific Information:
*** -[__NSCFConstantString stringByAppendingString:]: nil argument

Thread 0 Crashed:
0   CoreFoundation                  0x310b3c21c         __exceptionPreprocess
1   libobjc.A.dylib                 0x30b559ab8         objc_exception_throw
2   Foundation                      0x30e33e59c         -[NSString stringByAppendingString:]
3   Taxfix                          0x2057b5860         __53-[RNSentryOnDrawReporterView createEmitNewFrameEvent]_block_invoke
4   Taxfix                          0x2057b55b8         -[RNSentryFramesTrackerListener framesTrackerHasNewFrame:] (RNSentryFramesTrackerListener.m:25)
5   Taxfix                          0x2059d13b8         -[SentryFramesTracker reportNewFrame] (SentryFramesTracker.m:264)
6   Taxfix                          0x2059d12a4         -[SentryFramesTracker displayLinkCallback] (SentryFramesTracker.m:254)
```

krystofwoldrich:
Hi @pierricklete, thank you for the message,

[...]

PR #4881, opened 2025-06-02
fix(ttd): Gracefully handle missing `parentSpanId` when measuring TTID/TTFD on iOS

krystofwoldrich:

- [x] Bugfix

## :scroll: Description

fixes https://github.com/getsentry/sentry-react-native/issues/4875

I was not able to reproduce the crash by using the SDK, but the code allowed the emit new frame callback could be called without the `parentSpanId`, which this PR fixes. 

sample app, native tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/scripts/update-maestro.sh
+++ b/scripts/update-maestro.sh
@@ -8,7 +8,7 @@ fi
 
 # yml files to search and update MAESTRO_VERSION
 files=(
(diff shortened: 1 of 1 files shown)


antonis (approved):
LGTM 🚀 
I was not able to reproduce either but the fix makes sense 👍

Issue #4896, opened 2025-06-05
User Feedback and useSentryUser expected functionality


iffa:
### What React Native libraries do you use?

Expo Router

### Are you using sentry.io or on-premise?

on-premise (Self-Hosted)

### @sentry/react-native SDK Version

6.15.0

### How does your development environment look like?

````
Expo SDK 53 + React Native 0.79.3
````

### Sentry.init()

````js
Sentry.init({
  enabled: SENTRY_ENABLED,
  dsn: process.env.EXPO_PUBLIC_SENTRY_DSN,
  tracesSampleRate: 0.5,
  profilesSampleRate: 1.0,
  integrations: [
    navigationIntegration,
    Sentry.feedbackIntegration({
      useSentryUser: {
        email: "email",
        name: "username",
      },
    }),
  ],
  enableNativeFramesTracking: !isRunningInExpoGo(),
});
````

### Steps to Reproduce

Open feedback modal with `showFeedbackWidget` & see the email and name inputs being prefilled with `email` and `username` instead of actual data available from `Sentry.setUser`.

Even if I remove useSentryUser block completely, I get empty email & username fields even though we call `Sentry.getUser` when we receive user data (way before opening the feedback modal with `showFeedbackWidget`)

### Expected Result

Fields should be populated with actual values from `Sentry.setUser`, right? If not, how can I manage that? The docs are vague.

### Actual Result

![Image](https://github.com/user-attachments/assets/7dfec1cf-1f93-4b46-97bc-2264bc3457f2)

**Comment by iffa** [comment]
Right after I wrote this ticket I went to browse the code of the feedback stuff and found that indeed `useSentryUser` values are passed as-is to the feedback widget. BUT it has a fallback to get the current user from the context, which is not working for me or so it seems.

**Comment by antonis** [comment]
Thank you for your report @iffa 🙇 
I was able to reproduce the reported behaviour and we will iterate with a fix and updated documentation.
Currently `setSentryUser` [sets the default values for the email and name fields](https://docs.sentry.io/platforms/react-native/user-feedback/configuration/#general). 

What worked for me is passing the user data like below and skip the `setSentryUser` in the `feedbackIntegration`
```
Sentry.getCurrentScope().setUser({
  email: 'the.user@email.com',
  name: 'The User Name',
});
```

Let us know if that works for you 🙏

iffa:
[...]

PR #4901, opened 2025-06-06
fix(feedback): Align get user with JS implementation

antonis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Align get user [with JS implementation](https://github.com/getsentry/sentry-javascript/blob/e0400020d9acddc4b9291d88006edf1c77311bc0/packages/feedback/src/modal/integration.tsx#L9)

Fixes https://github.com/getsentry/sentry-react-native/issues/4896

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/packages/core/src/js/feedback/FeedbackWidget.tsx
+++ b/packages/core/src/js/feedback/FeedbackWidget.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable max-lines */
-import type { SendFeedbackParams } from '@sentry/core';
-import { captureFeedback, getCurrentScope, lastEventId, logger } from '@sentry/core';
+import type { SendFeedbackParams, User } from '@sentry/core';
+import { captureFeedback, getCurrentScope, getIsolationScope, getGlobalScope, lastEventId, logger } from '@sentry/core';
 import * as React from 'react';
 import type { KeyboardTypeOptions ,
   NativeEventSubscription} from 'react-native';
@@ -55,8 +55,8 @@ export class FeedbackWidget extends React.Component<FeedbackWidgetProps, Feedbac
 
     const currentUser = {
       useSentryUser: {
-        email: this.props?.useSentryUser?.email || getCurrentScope()?.getUser()?.email || '',
-        name: this.props?.useSentryUser?.name || getCurrentScope()?.getUser()?.name || '',
+        email: this.props?.useSentryUser?.email || this._getUser()?.email || '',
+        name: this.props?.useSentryUser?.name || this._getUser()?.name || '',
       }
     }
 
@@ -401,5 +401,18 @@ export class FeedbackWidget extends React.Component<FeedbackWidgetProps, Feedbac
 
   private _hasScreenshot = (): boolean => {
     return this.state.filename !== undefined && this.state.attachment !== undefined && this.state.attachmentUri !== undefined;
+  }
+
+  private _getUser = (): User | undefined =>{
+    const currentUser = getCurrentScope().getUser();
+    const isolationUser = getIsolationScope().getUser();
+    const globalUser = getGlobalScope().getUser();
+    if (currentUser && Object.keys(currentUser).length) {
+      return currentUser;
+    }
+    if (isolationUser && Object.keys(isolationUser).length) {
+      return isolationUser;
+    }
+    return globalUser;
   }
 }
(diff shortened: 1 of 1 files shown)


packages/core/src/js/feedback/FeedbackWidget.tsx
```diff
@@ -406,6 +406,18 @@ export class FeedbackWidget extends React.Component<FeedbackWidgetProps, Feedbac
     return this.state.filename !== undefined && this.state.attachment !== undefined && this.state.attachmentUri !== undefined;
   }
 
+  private _getUser = (): User | undefined => {
+    const currentUser = getCurrentScope().getUser();
+    if (currentUser && Object.keys(currentUser).length) {
```
krystofwoldrich: Is there any specific edge case for which `&& Object.keys(currentUser).length` is needed?

My thinking is if there is a user set (empty, or without email and name) we should take it.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Consider isolation and global scope for prefilling feedback widget user data([#4901](https://github.com/getsentry/sentry-react-native/pull/4901))
```
krystofwoldrich: I think it might be easier to understand for users what is fixed, if we mention the `Sentry.setUser`.

````suggestion
- User set by `Sentry.setUser` is prefilled in Feedback Widget ([#4901](https://github.com/getsentry/sentry-react-native/pull/4901))
  - User data are considered from all scopes in the following order current, isolation and global.
````

File: packages/core/src/js/feedback/FeedbackWidget.tsx (hunk 1 above)
Comment (antonis) [reply packages/core/src/js/feedback/FeedbackWidget.tsx]: Good point @krystofwoldrich 👍 
I just used the [JS logic on this](https://github.com/getsentry/sentry-javascript/blob/e0400020d9acddc4b9291d88006edf1c77311bc0/packages/feedback/src/modal/integration.tsx#L9) but I agree with you and I'll iterate on this.

Comment (antonis) [reply packages/core/src/js/feedback/FeedbackWidget.tsx]: Fixed with beaa541b63d9cc8e49e31212f5f8fc652453506e

An alternative would be to check just the user email or name like:
```
    if (currentUser && (currentUser.email || currentUser.name)) {
```

File: packages/core/src/js/feedback/FeedbackWidget.tsx [hunk 3]
```diff
@@ -406,6 +406,18 @@ export class FeedbackWidget extends React.Component<FeedbackWidgetProps, Feedbac
...
 
+  private _getUser = (): User | undefined => {
+    const currentUser = getCurrentScope().getUser();
+    if (currentUser) {
+      return currentUser;
+    }
+    const isolationUser = getIsolationScope().getUser();
+    if (isolationUser) {
+      return isolationUser;
+    }
```
lucas-zimerman, line 417: i: Most people set the user by using Sentry.setUser, with that we could invert the `currentScope` with `isolationScope` so it return in most of the cases on first try.

```suggestion
    const isolationUser = getIsolationScope().getUser();
    if (isolationUser {
      return isolationUser;
    }
    const currentUser = getCurrentScope().getUser();
    if (currentUser) {
      return currentUser;
    }
```

antonis: Your suggestion makes sense to me @lucas-zimerman and also [aligns with the documentation](https://docs.sentry.io/platforms/react-native/enriching-events/scopes/#isolation-scope) given that we are handling user data. On the other hand [any data set on the current scope should take precedence over data set on the isolation and global scope](https://docs.sentry.io/platforms/react-native/enriching-events/scopes/#how-scope-data-is-applied-to-events), which is [the logic used in the JS feedback widget](https://github.com/getsentry/sentry-javascript/blob/e0400020d9acddc4b9291d88006edf1c77311bc0/packages/feedback/src/modal/integration.tsx#L9) 🤔

lucas-zimerman: we can skip the suggeston and talk about it on a follow up issue so the main issue doesn't get blocked for too long.
To me it sounds like, on react native context, where you'll only have one user, Sentry.SetUser should use the global scope instead of an isolation scope.

Maybe it could be a nice nit that we could change on V7 if others agree.

lucas-zimerman (approved):
LGTM!

packages/core/src/js/feedback/FeedbackWidget.tsx
antonis: Opened an issue for this https://github.com/getsentry/sentry-react-native/issues/4915

Issue #4894, opened 2025-06-05
Automatically detect Release name and version for Expo Web


krystofwoldrich:
On mobile release is automatically prefilled based on the native application metadata, `package@version+build` (for example, `my.project.name@2.3.12+1234`).<br><br>On web, release is expected to be set manually by the user. Since Expo Web project contain metadata about the project name and version, we can match the behavior from mobile and fill in release based on the ExpoConfig.<br><br>[https://docs.expo.dev/versions/latest/sdk/constants/](https://docs.expo.dev/versions/latest/sdk/constants/)

[https://github.com/expo/expo/blob/16f6d9e78ba28adae7bff034efbc34752f0e1719/packages/%40expo/config-types/src/ExpoConfig.ts#L44](https://github.com/expo/expo/blob/16f6d9e78ba28adae7bff034efbc34752f0e1719/packages/%40expo/config-types/src/ExpoConfig.ts#L44)

antonis:
Fixed with https://github.com/getsentry/sentry-react-native/pull/4967

PR #4910, opened 2025-06-11
feat(expo): Automatically detect Release name and version for Expo Web

antonis:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Automatically detect Release name and version for Expo Web

Fixes https://github.com/getsentry/sentry-react-native/issues/4894

[Manual](https://sentry-sdks.sentry.io/issues/6604676621/events/latest/?project=5428561&query=is%3Aunresolved&referrer=latest-event&stream_index=0), CI

Screenshot 2025-06-11 at 7 35 38 PM

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/src/js/integrations/release.ts
+++ b/packages/core/src/js/integrations/release.ts
@@ -10,6 +10,14 @@ interface ExpoConfig {
   version?: string;
 }
 
+let Constants: { default?: { manifest?: unknown }; manifest?: unknown } | null = null;
+try {
+  // eslint-disable-next-line import/no-extraneous-dependencies
+  Constants = require('expo-constants');
+} catch {
+  // expo-constants not available, do nothing
+}
+
 /** Release integration responsible to load release from file. */
 export const nativeReleaseIntegration = (): Integration => {
   return {
@@ -22,7 +30,7 @@ export const nativeReleaseIntegration = (): Integration => {
 };
 
 /**
- * Get Expo Web configuration name and version from environment variables
+ * Get Expo Web configuration name and version from environment variables or Constants
  */
 function getExpoWebConfig(): ExpoConfig | null {
   if (!isWeb() || !isExpo()) {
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Features
+
+- Automatically detect Release name and version for Expo Web ([#4910](https://github.com/getsentry/sentry-react-native/pull/4910))
```
lucas-zimerman, line 13: i: We should mention that it only works in development builds.
If you try it on a release build you'll not get the release number
https://sentry-sdks.sentry.io/issues/6673524305/events/04e3a3759a7c4cc5bd4a63038d0c2d9b/?project=5428561

antonis: Good point Lucas 👍 
I think this is something we should tackle for the enhancement to make sense.

krystofwoldrich: If the application name and version are not available in production, I would not added it at all, to avoid confusing users with differences between dev and prod.

If Expo doesn't include any of the information in production on web, we could include the release name during the build (bundling process), utilizing our Sentry Metro Plugin.

antonis: Hey @lucas-zimerman 👋 
I've updated the PR with f2648bae3dfa6bb866b0785ac01dd2b78a77fcea to handle the release build case ([example](https://sentry-sdks.sentry.io/issues/6673523996/events/latest/?project=5428561&query=is%3Aunresolved&referrer=latest-event&stream_index=0)). The downside is that it depends on [expo-constants](https://docs.expo.dev/versions/latest/sdk/constants/) and will fail if it is not available. I've tested other approaches like bundle injection, but either didn't work or they were overcomplicated for the task at hand.
I think this approach is simple and would work on most cases. As a fallback the user's can still set the release manually on the web as they do now. Let me know wdyt and I can clean up the PR and prepare it for review 🙇

krystofwoldrich: Hi @antonis and @lucas-zimerman,
I've taken a look at this again and I think the SDK should not depend on `expo-constants`, to keep compatibility with non-expo apps. I would also avoid the ENVs approach since it doesn't work both prod and dev.

I would propose to use a custom module to load a global variable with the release information. This will work for both prod and dev. 

Similar to how Sentry JS Webpack Plugin is doing it, [link](https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/c6b18cdbdc9e1b19a1efc6f936796308ea600b05/packages/bundler-plugin-core/src/utils.ts#L308).

This is, how it could be implemented: https://github.com/getsentry/sentry-react-native/compare/v7...krystofwoldrich:sentry-react-native:%40krystofwoldrich/inject-expo-app-release#diff-de4a1259c31e8b44b35788d85ff5eb3d27eef51e73325b3ee41532d9095722cd

Let me know, if this would work for you, I can open a PR.

antonis: Thank you for taking the time to look at this @krystofwoldrich. Your feedback is always welcome :)

> I've taken a look at this again and I think the SDK should not depend on `expo-constants`, to keep compatibility with non-expo apps. 

I agree. This makes sense only as a best effort approach for expo web.


> I would propose to use a custom module to load a global variable with the release information. This will work for both prod and dev. 
> 
> Similar to how Sentry JS Webpack Plugin is doing it, [link](https://github.com/getsentry/sentry-javascript-bundler-plugins/blob/c6b18cdbdc9e1b19a1efc6f936796308ea600b05/packages/bundler-plugin-core/src/utils.ts#L308).
> 
> This is, how it could be implemented: https://github.com/getsentry/sentry-react-native/compare/v7...krystofwoldrich:sentry-react-native:%40krystofwoldrich/inject-expo-app-release#diff-de4a1259c31e8b44b35788d85ff5eb3d27eef51e73325b3ee41532d9095722cd

Your approach looks great and of course a PR is more than welcome 🤗

krystofwoldrich: Let me know, if anything else is needed.

- https://github.com/getsentry/sentry-react-native/pull/4967

antonis:
Closing in favor of https://github.com/getsentry/sentry-react-native/pull/4967

PR #4930, opened 2025-06-17
feat(Logs): expose logger and console Integration

lucas-zimerman:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


hamoudaarfaoui:
Found an issue with implementation of `logger.error` it's not the same across.

I have an app using older version 6.x of sentry, where I can import
```
{logger} from '@sentry/core'
logger.error(new Error("..."), "my message", { ..properties })
```

no longer possible because `logger from `@sentry/react-native"` implementation is different

image

Issue #4924, opened 2025-06-16
Spamming of HTTP Client Error with status code: 500


JunPG:
### What React Native libraries do you use?

React Navigation, Hermes, RN New Architecture

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### @sentry/react-native SDK Version

6.14.0

### How does your development environment look like?

````
⬇  Place the `npx react-native@latest info` output here. ⬇ 

"@sentry/react-native": "~6.14.0",
"react-native": "0.79.3",

````

### Sentry.init()

````js
const sentryReactNavigationIntegration = reactNavigationIntegration();
Sentry.init({
    dsn: Config.SENTRY_DSN,
    environment: Platform.OS,
    sampleRate,
    tracesSampleRate: sampleRate,
    enableAutoSessionTracking: true,
    enableAutoPerformanceTracing: true,
    normalizeDepth: 10,
    integrations: [
      sentryReactNavigationIntegration,
      httpClientIntegration({
        failedRequestStatusCodes: [],
      }),
    ],
    ignoreErrors: [
      'HTTP Client Error with status code: 500',
      'HTTP Client Error with status code: 503',
    ],
    beforeSend: async (event) => {
      const {isInternetReachable} = await NetInfo.fetch();
      return isInternetReachable ? event : null;
    },
  });
````

### Steps to Reproduce

1. Build and run for iOS

### Expected Result

Error like HTTP Client Error with status code: 5xx should not be sent to sentry.io

### Actual Result

5xx errors are all sent to sentry.io

lucas-zimerman:
Hi and thank you for opening this issue! Would you be able to share an event that wasn't supposed to be captured?

JunPG:
> Hi and thank you for opening this issue! Would you be able to share an event that wasn't supposed to be captured?

Hi @lucas-zimerman, do you need the link to the event or just a screenshot would be suffice?

stefanosiano:
hi @JunPG a link would be better, as it gives us the full picture. If you're worried about personal data, you can send it to stefano.siano@sentry.io

JunPG:
> hi [@JunPG](https://github.com/JunPG) a link would be better, as it gives us the full picture. If you're worried about personal data, you can send it to [stefano.siano@sentry.io](mailto:stefano.siano@sentry.io)

Hi @stefanosiano , I've sent an email to the address above with title `Spamming of HTTP Client Error with status code: 500 #4924`

[...]

PR #4948, opened 2025-06-25
fix: ignoreErrors now works on Native

lucas-zimerman:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Before this change, ignoreErrors was only applied to JavaScript errors. With this PR, Native errors will also be able to be filtered by ignoreErrors.

Since different SDKs apply different approaches to ignoreErrors, I have unified them by following what the JavaScript SDK does.

Usage:

on JS SDK, strings are compared with contains, making it able to seach for substrings, so `1234` will ignore the error ` this is an error 1234`.
Regex filters will continue with the same behaviour on all platforms.

### Android

It has a build in ignoreError filter, but string errors doesn't compare substrings but the entire string, making the `1234` example invalid. To fix it, string asserts are transformed into Regex by doing the pattern `.*\Q this is my string\E.*`, making it behave like the JavaScript contains.

### iOS

Since there are no ignoreErrors, we apply those filters on the beforeSend callback, regex errors are filtered by regex objects where string filters behave like the JS implementation.

Fixes https://github.com/getsentry/sentry-react-native/issues/4924

Unit Tests, sample app

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- Document ignoreErrors on V7

### DIFF
--- a/packages/core/ios/RNSentry.mm
+++ b/packages/core/ios/RNSentry.mm
@@ -184,8 +184,7 @@ RCT_EXPORT_METHOD(initNativeSdk
     }
 
     for (NSString *str in _ignoreErrorPatternsStr) {
-        if ([message containsString:str])
-        {
+        if ([message containsString:str]) {
             return YES;
         }
     }


CHANGELOG.md
```diff
@@ -6,6 +6,23 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+- ignoreError now filters Native errors ([#4948](https://github.com/getsentry/sentry-react-native/pull/4948))
```
lucas-zimerman, line 13:
```suggestion
### Fixes

- ignoreError now filters Native errors ([#4948](https://github.com/getsentry/sentry-react-native/pull/4948))
```

CHANGELOG.md
```diff
@@ -6,6 +6,25 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Fixes
+
+- ignoreError now filters Native errors ([#4948](https://github.com/getsentry/sentry-react-native/pull/4948))
+
+you can use strings to filter errors that contains your string or RegEx for filtering for a pattern.
```
antonis: nit
```suggestion
You can use strings to filter errors or RegEx for filtering with a pattern.
```

packages/core/ios/RNSentry.mm
```diff
@@ -132,21 +133,98 @@ - (instancetype)init
...
+        if ([regex firstMatchInString:message options:0 range:range]) {
+            return YES;
+        }
+    }
+
+    for (NSString *str in _ignoreErrorPatternsStr) {
+        if ([message containsString:str]) {
+            return YES;
+        }
+    }
```
antonis: l: Wdyt of reverting the order since string checks might be faster?

```suggestion
    for (NSString *str in _ignoreErrorPatternsStr) {
        if ([message containsString:str]) {
            return YES;
        }
    }
    
    for (NSRegularExpression *regex in _ignoreErrorPatternsRegex) {
        NSRange range = NSMakeRange(0, message.length);
        if ([regex firstMatchInString:message options:0 range:range]) {
            return YES;
        }
    }
```

antonis (approved):
Awesome work @lucas-zimerman 🙇 
The changes LGTM and work as expected 🎉 

Just added two minor suggestions.

packages/core/ios/RNSentry.mm
lucas-zimerman: Sounds good to me. As a user, I believe I would avoid using Regex and only use normal strings since it's more convenient

PR #4995, opened 2025-07-16
fix(span): Skip idle span creation when app is in background

antonis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Skip idle span creation when app is in background.

This is a follow up to https://github.com/getsentry/sentry-react-native/pull/3307

Client reports of transactions starting in the background resulting in (a small number of) unexplainable long transactions skewing the app statistics.

Manual, CI tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/src/js/tracing/span.ts
+++ b/packages/core/src/js/tracing/span.ts
@@ -12,6 +12,7 @@ import {
   spanToJSON,
   startIdleSpan as coreStartIdleSpan,
 } from '@sentry/core';
+import { AppState } from 'react-native';
 
 import { isRootSpan } from '../utils/span';
 import { adjustTransactionDuration, cancelInBackground } from './onSpanEndUtils';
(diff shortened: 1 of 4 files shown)


packages/core/src/js/tracing/span.ts
```diff
@@ -104,6 +105,12 @@ export const startIdleSpan = (
     return new SentryNonRecordingSpan();
   }
 
+  const currentAppState = AppState.currentState;
+  if (currentAppState === 'background') {
+    logger.debug(`[startIdleSpan] App is already in background, not starting span for ${startSpanOption.name}`);
```
lucas-zimerman, line 110: qq: What is the startSpanOption.name value here?

lucas-zimerman (approved):
Looking good, LGTM!

packages/core/src/js/tracing/span.ts
antonis: I think there are two cases:
* for user interaction [it should be](https://github.com/getsentry/sentry-react-native/blob/13d9dc0f24ac6bebb4b283c30ced028cb9e2ebc0/packages/core/src/js/tracing/integrations/userInteraction.ts#L70) `${tracing.state.currentRoute}.${elementId}`
* or 'Route Change' which is [the default navigation span name](https://github.com/getsentry/sentry-react-native/blob/13d9dc0f24ac6bebb4b283c30ced028cb9e2ebc0/packages/core/src/js/tracing/span.ts#L119)

PR #5055, opened 2025-08-08
Ref(CI): Reduce processor/ram used by jest test locally.

lucas-zimerman:
This changes only impact local tests and not github tests.

If your device has too many CPU cores, our tests will spam too many node processes, making the test to use a lot of memory and CPU resources.
By limiting the amount of workers to 25% I was able to reduce locally the time to finish the tests in 16%, reduce memory usage by 54% on an environment with 32 CPU Cores and 32 GB of ram. Assuming that our GitHub actions are running with 2 core and only 8 GB of RAM, I expect those values to be even better.

Here is my local-measured data given the total amount of workers

Workers : how many workers jest can make
Duration: time to complete the test
Memory Avg: the average memory increase before - during the tests
| Workers | Duration Avg (s) | Duration Improvement % | Memory Avg (MB) | Memory Improvement % |
|---------|------------------|-----------------------|-----------------|----------------------|
| 100%    | 10.3067          | 0%                    | 6167.83         | 0%                   |
| 50%     | 9.176            | 11.00%                | 3954.21         | 35.91%               |
| 25%     | 8.6813           | 15.79%                | 2857.76         | 53.67%               |
| 20%     | 8.984            | 12.82%                | 2236.91         | 63.69%               |
| 10%     | 9.3193           | 9.62%                 | 1671.26         | 72.91%               |

#skip-changelog.


lucas-zimerman:
Not sure if it's worth adding in the project but here is the measuring script that I used 

```js
#!/usr/bin/env node
const { spawn } = require('child_process');
const os = require('os');

function getUsedMemory() {
  return os.totalmem() - os.freemem();
}

const startMem = getUsedMemory();
console.log(`Starting memory usage: ${(startMem / 1024 / 1024).toFixed(2)} MB`);

const startTime = Date.now();
let peakMemIncrease = 0;

const jest = spawn('npx', ['jest', '--maxWorkers=8'], { stdio: 'inherit' });

const memoryInterval = setInterval(() => {
  const currentUsed = getUsedMemory();
  const increase = currentUsed - startMem;
  if (increase > peakMemIncrease) {
    peakMemIncrease = increase;
  }
}, 1000);

jest.on('close', (code) => {
  clearInterval(memoryInterval);
  const endTime = Date.now();
  console.log(`\nTest duration: ${(endTime - startTime) / 1000}s`);
  console.log(`Peak memory increase: ${(peakMemIncrease / 1024 / 1024).toFixed(2)} MB`);
  process.exit(code);
});
```

All I had to do is place it on the core folder, play with the maxWorkers value and replace the `test:sdk` script for `node run-jest.js`

packages/core/package.json
```diff
@@ -30,7 +30,7 @@
     "clean:plugin": "expo-module clean plugin",
     "circularDepCheck": "madge --circular dist/js/index.js && madge --circular metro.js && madge --circular expo.js",
     "test": "yarn test:sdk && yarn test:tools",
-    "test:sdk": "npx jest",
+    "test:sdk": "npx jest --maxWorkers=25%",
```
lucas-zimerman: limit to 8 cores
```suggestion
    "test:sdk": "npx jest --maxWorkers=8",
```

packages/core/package.json
```diff
@@ -30,7 +30,7 @@
     "clean:plugin": "expo-module clean plugin",
     "circularDepCheck": "madge --circular dist/js/index.js && madge --circular metro.js && madge --circular expo.js",
     "test": "yarn test:sdk && yarn test:tools",
-    "test:sdk": "npx jest",
+    "test:sdk": "npx jest --maxWorkers=8",
```
lucas-zimerman: Change has no effect, lets limit it only on local devices.
```suggestion
    "test:sdk": "sh -c 'if [ \"$CI\" = \"true\" ]; then yarn test:sdk-ci; else yarn test:sdk-local; fi'",
    "test:sdk-ci": "npx jest",
    "test:sdk-local": "jest --maxWorkers=8",
```

antonis (approved):
Thank you for looking at this and running the metrics 🙇 
The changes LGTM 🎸 

Running this locally on a Macbook M1 Pro / 32GB, I didn't notice any difference, but I think we should experiment and see how this performs on CI.

> Not sure if it's worth adding in the project but here is the measuring script that I used

Thank you for sharing this. I'm not sure if we should add it but we can have this PR as a reference.

It will only be noticeable on the latest Pro Max where the cpu count is higher than 8 cores

PR #5064, opened 2025-08-12
fix(newArch): fixes how the availability of Turbo Modules is being detected

alwx:
During an investigation of our support of the new architecture I accidentally found that the way we detect turbo modules is outdated since `__turboModuleProxy` was removed from React Native in bridgeless mode. 
Here is where I found that: https://github.com/facebook/react-native/pull/48362
Here is more information about bridgeless mode in React Native: https://github.com/reactwg/react-native-new-architecture/discussions/154

And, finally, here is how React Native itself detects if Turbo Modules are available: https://github.com/facebook/react-native/blob/641a79dc5137b69a3c0813413b9fb82d0b9df783/packages/react-native/src/private/featureflags/ReactNativeFeatureFlagsBase.js#L110

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Basically this change implements the same mechanism for detecting Turbo Modules that's already used in React Native (see link above).

Solving how we detect the usage of the new architecture is crucial for correct usage of corresponding native modules.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/src/js/utils/environment.ts
+++ b/packages/core/src/js/utils/environment.ts
@@ -11,6 +11,7 @@ export function isHermesEnabled(): boolean {
 
 /** Checks if the React Native TurboModules are enabled */
 export function isTurboModuleEnabled(): boolean {
+  // Reference: https://github.com/facebook/react-native/blob/641a79dc5137b69a3c0813413b9fb82d0b9df783/packages/react-native/src/private/featureflags/ReactNativeFeatureFlagsBase.js#L110
   return RN_GLOBAL_OBJ.RN$Bridgeless === true || RN_GLOBAL_OBJ.__turboModuleProxy != null;
 }
 
--- a/packages/core/src/js/utils/environment.ts
+++ b/packages/core/src/js/utils/environment.ts
@@ -11,7 +11,7 @@ export function isHermesEnabled(): boolean {
 
 /** Checks if the React Native TurboModules are enabled */
 export function isTurboModuleEnabled(): boolean {
-  return RN_GLOBAL_OBJ.__turboModuleProxy != null;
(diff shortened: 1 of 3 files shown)


lucas-zimerman:
Thank you for finding and fixing this!

packages/core/src/js/utils/environment.ts
```diff
@@ -11,7 +11,7 @@ export function isHermesEnabled(): boolean {
 
 /** Checks if the React Native TurboModules are enabled */
 export function isTurboModuleEnabled(): boolean {
-  return RN_GLOBAL_OBJ.__turboModuleProxy != null;
+  return RN_GLOBAL_OBJ.RN$Bridgeless === true || RN_GLOBAL_OBJ.__turboModuleProxy != null;
```
lucas-zimerman, line 15: Nit: let's add a URL for a quick reference on the React Native implementation for future checks.
```suggestion
  // Reference: https://github.com/facebook/react-native/blob/641a79dc5137b69a3c0813413b9fb82d0b9df783/packages/react-native/src/private/featureflags/ReactNativeFeatureFlagsBase.js#L110
  return RN_GLOBAL_OBJ.RN$Bridgeless === true || RN_GLOBAL_OBJ.__turboModuleProxy != null;
```

lucas-zimerman:
The change should be mentioned on changelog, 

suggestion:
```md

### Fixes

- Performance improvements when initializing the SDK with turbomodule ([#5064](https://github.com/getsentry/sentry-react-native/pull/5064))

antonis (commented):
LGTM! Thank you for finding this and fixing it so quickly @alwx 🏅
Let's add a changelog and 🚢

antonis (approved):
LGTM 🚀 
Thanks again for fixing this Alex 🙇

Issue #4705, opened 2025-03-28
Implement TTID/TTFD as Turbo Modules


krystofwoldrich:
https://docs.sentry.io/platforms/react-native/tracing/instrumentation/time-to-display/

TimeToDisplay component is currently implemented as a legacy native component, and needs to be re-implemented as a turbo module the way it's done in this PR for Mask/Unmask: https://github.com/getsentry/sentry-react-native/pull/4265

# Where to look?
- iOS implementation: https://github.com/getsentry/sentry-react-native/blob/9de3d458dcde971c058944dd4da71be3ff2c228b/packages/core/ios/RNSentryOnDrawReporter.m
- Android implementation: https://github.com/getsentry/sentry-react-native/blob/9de3d458dcde971c058944dd4da71be3ff2c228b/packages/core/android/src/main/java/io/sentry/react/RNSentryOnDrawReporterManager.java
- js/tracing/timetodisplaynative.tsx
- js/tracing/integrations/timeToDisplayIntegration.ts
- js/tracing/timetodisplaynative.tsx
- RNSentryReplayMask is a good example of how to do that

PR #5081, opened 2025-08-14
fix(tracing): Time To Initial Display / Time To Full Display: remove warning that it's unsupported

alwx:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

Partially fixes https://github.com/getsentry/sentry-react-native/issues/4705

## :scroll: Description

1) Even though in the docs we clearly state that TTID/TTFD is not supported, apparently it works if you disable all the checks in the SDK. I suspect it works since this PR got merged: https://github.com/getsentry/sentry-react-native/pull/4669
2) However, it is NOT packed as a Turbo Module which means all the standard disadvantages like extra overhead and higher latency due to serialization/deserialization and native modules being loaded into memory during app startup.

The solution for now is the following: we can now remove the checks in the SDK and basically make it work with the new architecture. However, we will need to implement TTID/TTFD as turbo modules to make it more efficient: https://github.com/getsentry/sentry-react-native/issues/4705

Here is the corresponding docs change: https://github.com/getsentry/sentry-docs/pull/14662

This change is needed to not show message that it's not supported because it actually is supported :)

With the sample app:

Screenshot 2025-08-14 at 12 13 53

(not all the results are there because I did experiment with <Sentry.TimeToInitialDisplay /> and <Sentry.TimeToFullDisplay /> wrapper components on individual screens)

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/test/tracing/timetodisplay.test.tsx
+++ b/packages/core/test/tracing/timetodisplay.test.tsx
@@ -7,7 +7,6 @@ jest.mock('../../src/js/wrapper', () => mockWrapper);
 import * as mockedtimetodisplaynative from './mockedtimetodisplaynative';
 jest.mock('../../src/js/tracing/timetodisplaynative', () => mockedtimetodisplaynative);
 
-import { isTurboModuleEnabled } from '../../src/js/utils/environment';
 jest.mock('../../src/js/utils/environment', () => ({
   isWeb: jest.fn().mockReturnValue(false),
   isTurboModuleEnabled: jest.fn().mockReturnValue(false),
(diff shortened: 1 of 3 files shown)


packages/core/test/tracing/timetodisplay.test.tsx
```diff
@@ -290,24 +289,6 @@ describe('TimeToDisplay', () => {
...
-  });
-
-  test('should log a warning if in new architecture', async () => {
-    (isTurboModuleEnabled as jest.Mock).mockReturnValue(true);
-    TestRenderer.create(<TimeToInitialDisplay record={true} />);
-    await jest.runOnlyPendingTimersAsync(); // Flush setTimeout.
-
-    expect(logger.warn).toHaveBeenCalledWith(
-      'TimeToInitialDisplay and TimeToFullDisplay are not supported on the web, Expo Go and New Architecture. Run native build or report an issue at https://github.com/getsentry/sentry-react-native');
-  });
```
antonis, line 310: q: My understanding is that the existing tests already cover both architectures. Is that the case or we are planning to add more?

alwx: @antonis yes, existing tests are architecture-independent — they are basically UI tests that render components, mock certain behavior and verify spans after that.

antonis (approved):
LGTM 🚀 
Thank you for your work on this @alwx 🙇

PR #5086, opened 2025-08-15
ci: Disable iOS tests paralellization

itaybre:
Using parallelization on UI Tests requires cloning the simulator, which seems to cause slow tests on `sentry-cocoa` integrations tests.

I guess this change will also benefit this repo


lucas-zimerman:
Makes sense since the VM doens't have alot of resources.
LGTM!

PR #5092, opened 2025-08-20
Fix(V6): missing ip settings from SDK package

lucas-zimerman:
Based on https://github.com/getsentry/sentry-javascript/pull/17364

original context;
```
This PR fixes a long-standing problem in the SDK where it would set incorrect information about when Relay should (not) infer IP addresses from sent envelope requests.

Previously, this was thought to be controlled by setting event.user.ip_address: '{{auto}}'. However, after an incident in Relay, it was determined that this is in fact not a reliably way to control IP inference. Instead, SDKs should set event.sdk.settings.infer_ip: 'auto' | 'never' (see closes https://github.com/getsentry/sentry-javascript/issues/16252).
Unfortunately, this wasn't implemented immediately but is taken care of in this PR.

(FWIW, the only reason why Relay continued to infer IP addresses for the JS SDK was because it is excempt from logic that would infer IP addresses only if user.ip_address was set to '{{auto}}'. This is necessary to backwards compatibility with older SDKs.)

Follow-ups: We likely also need to adjust the logic in Electron and Lynx (at the very least remove setting user.ip_address).

closes https://github.com/getsentry/sentry-javascript/issues/17351
closes https://github.com/getsentry/sentry-javascript/issues/16252
```

Instead of bumping o V10, the fix was patched on the current release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,24 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Important Changes
+
+- **fix(browser): Ensure IP address is only inferred by Relay if `sendDefaultPii` is `true`** ([#???](https://github.com/getsentry/sentry-react-native/pull/??))
+
+This release includes a fix for a [behaviour change](https://docs.sentry.io/platforms/javascript/migration/v8-to-v9/#behavior-changes)
+that was originally introduced with v9 of the JavaScript SDK: User IP Addresses should only be added to Sentry events automatically,
+if `sendDefaultPii` was set to `true`.
+
+However, the change in v9 required further internal adjustment, which should have been included in v10 of the SDK.
+To avoid making a major bump, the fix was patched on the current version and not by bumping to V10.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,24 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Important Changes
+
+- **fix(browser): Ensure IP address is only inferred by Relay if `sendDefaultPii` is `true`** ([#???](https://github.com/getsentry/sentry-react-native/pull/??))
```
lucas-zimerman:
```suggestion
- **fix(browser): Ensure IP address is only inferred by Relay if `sendDefaultPii` is `true`** ([#5092](https://github.com/getsentry/sentry-react-native/pull/5092))
```

samples/react-native/src/App.tsx
```diff
@@ -63,6 +63,7 @@ Sentry.init({
   dsn: getDsn(),
   debug: true,
   environment: 'dev',
+  sendDefaultPii: true,
```
lucas-zimerman:
```suggestion
```

PR #5134, opened 2025-09-02
Fix: Fix compile error in v7

noahsmartin:
The latest v7 release is causing CI to break in sentry-cocoa, this should fix it

#skip-changelog

### DIFF
--- a/packages/core/ios/RNSentry.mm
+++ b/packages/core/ios/RNSentry.mm
@@ -467,7 +467,11 @@ RCT_EXPORT_METHOD(fetchNativeLogAttributes
             contexts[@"os"] = os;
         }
 
-        NSString *releaseName = [SentrySDK options].releaseName;
+#if CROSS_PLATFORM_TEST
+        NSString *releaseName = SentrySDKInternal.options.releaseName;
+#else
+              NSString *releaseName = [SentrySDK options].releaseName;
+#endif
         if (releaseName) {
             contexts[@"release"] = releaseName;
         }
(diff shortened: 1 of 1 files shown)


antonis (approved):
Thank you for fixing this @noahsmartin 🙇 
LGTM 🚀

packages/core/ios/RNSentry.mm
```diff
@@ -467,7 +467,11 @@ - (void)stopObserving
             contexts[@"os"] = os;
         }
 
-        NSString *releaseName = [SentrySDK options].releaseName;
+#if CROSS_PLATFORM_TEST
+        NSString *releaseName = SentrySDKInternal.options.releaseName;
+#else
+              NSString *releaseName = [SentrySDK options].releaseName;
```
antonis: nit: I think we should align this on the left though the linter is not complaining.
```suggestion
         NSString *releaseName = [SentrySDK options].releaseName;
```

lucas-zimerman: I applied the suggestion, if it breaks, lets revert it.

lucas-zimerman (approved):
LGTM!

packages/core/ios/RNSentry.mm
```diff
@@ -467,7 +467,11 @@ - (void)stopObserving
             contexts[@"os"] = os;
         }
 
-        NSString *releaseName = [SentrySDK options].releaseName;
+#if CROSS_PLATFORM_TEST
+        NSString *releaseName = SentrySDKInternal.options.releaseName;
+#else
+         NSString *releaseName = [SentrySDK options].releaseName;
+#endif
```
antonis:
```suggestion
#if CROSS_PLATFORM_TEST
        NSString *releaseName = SentrySDKInternal.options.releaseName;
#else
        NSString *releaseName = [SentrySDK options].releaseName;
#endif
```

antonis:
Thank you for fixing this @noahsmartin. This is now release with https://github.com/getsentry/sentry-react-native/releases/tag/7.0.1

PR #5174, opened 2025-09-16
fix(replay): Allow excluding the sentry-android-replay module from android targets

romtsn:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

- Introduce new module `replay-stubs` under the android target. This allows customers to exclude `sentry-android-replay` if they are not interested in replay and save some bundle size
- Introduce new step in the `update-deps` workflow to bump the current sentry version which the stubs are compiled against
- The stubs are generated as part of `yarn build` (via `./gradlew jar`) and checked into git so the android target can use it
- In addition, check at runtime if the replay classes are present via reflection and only instantiate replay-related classes if that's the case
- Please read more details under `README.md` of the respective module

Here's an example replay after the changes: https://sentry-sdks.sentry.io/explore/replays/f0a8d20f73d54ea4b553901edeb83c50/

Closes https://github.com/getsentry/sentry-java/issues/4718
Closes ANDROID-197

Only manually tested for now, there's no really a good way to automatically test it without setting up gradle testkit which seems like a lot of work for little benefit (considering this is not a default behavior and likely rarely needed).

## :pencil: Checklist

- [] I updated the docs if needed.
- [] I updated the wizard if needed.
- [ ] All tests passing
- [x] No breaking changes


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Session Replay: Allow excluding `sentry-android-replay` from android targets ([#5174](https://github.com/getsentry/sentry-react-native/pull/5174))
```
lucas-zimerman, line 13: It would be nice to include a simple mention on how to remove it with a sample diff

lucas-zimerman: Example of what we did for Web Replay: https://github.com/getsentry/sentry-react-native/blob/cf745081e2928fd7b1459e38989d54d8ef0815b9/CHANGELOG.md#features-25

And with that, it's optional, but we could also include an estimation on how much app size is reduced when users opt out for mobile replay.

romtsn: Added both 👍

antonis (approved):
LGTM 🚀

PR #5191, opened 2025-09-22
Drop outdated clang-format from npm

lucas-zimerman:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

For some reason, when running `yarn fix` on macOS with ARM processor, it relies on an X86 wrapper/install of clang-format, throwing build errors if you don't have Rosetta installed.

This PR will do the following changes:

## Package

- Remove deprecated yarn package that includes an old clang format that has no arm support.

## GH CI

- Install on Ubuntu VM `clang 20 and other dependencies` so it matches a the output that   `brew install` on macos produces.
- sentry-cocoa doesn't need this change since the lint step uses macos VM, making the setup easy by using brew.
 
## Clang Script
- Prefer brew version of clang-format so it uses the correct architecture for the current system.
- Rollback to the original path if we don't have a brew installation.
- Alert users on how to install clang-format if not found.
- Alert users about arch errors and how to manually run clang-format in case of architecture mismatch.

## iOS Folder

- Changes applied by clang-format after running `yarn fix` due to new changes generated by newer clang-format.

Annoying errors where I can't use clang-format locally due to wrong CPU Arch. ALSO, clang-format was always producing different formating than CI clang-format, wasting developer time by manually fixing the CI format.

CI/Locally

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog.

### DIFF
--- a/.github/workflows/buildandtest.yml
+++ b/.github/workflows/buildandtest.yml
@@ -35,7 +35,6 @@ jobs:
       - name: Test
         run: yarn test
 
-
   job_lint:
     name: Lint
     runs-on: ubuntu-latest


lucas-zimerman:
Oh that's why!
Clang used by the repo comes from npm, the same is 3 years old unmaintained.

.github/workflows/buildandtest.yml
```diff
@@ -35,6 +35,7 @@ jobs:
       - name: Test
         run: yarn test
 
+
```
lucas-zimerman:
```suggestion
```

PR #5202, opened 2025-09-24
 Fix the issue with changing immutable metadata structure in the contructor of `ReactNativeClient`

### LINKED ISSUE #5187: Error: You attempted to set the key `settings` with the value `{"infer_ip":"auto"}` on an object that is meant to be immutable and has been frozen., js engine: hermes

React Native without Frameworks, React Navigation, Hermes

sentry.io (SaS)

7.1.0

````

System:
  OS: macOS 14.6.1
  CPU: (8) arm64 Apple M1 Pro
  Memory: 122.03 MB / 16.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 23.11.0
    path: /opt/homebrew/bin/node
  Yarn:
    version: 1.22.19
    path: /usr/local/bin/yarn
  npm:
    version: 10.9.2
    path: /opt/homebrew/bin/npm
  Watchman:
    version: 2025.04.28.00
Managers:
  CocoaPods:
    version: 1.16.2
    path: /opt/homebrew/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 24.2
      - iOS 18.2
      - macOS 15.2
      - tvOS 18.2
      - visionOS 2.2
      - watchOS 11.2
  Android SDK: Not Found
IDEs:
  Android Studio: Not Found
  Xcode:
    version: 16.2/16C5032a
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 24.0.1
    path: /usr/bin/javac
  Ruby:
    version: 2.6.10
    path: /usr/bin/ruby
npmPackages:
    installed: 18.0.0
    wanted: 18.0.0
  react:
    installed: 19.0.0
    wanted: 19.0.0
  react-native:
    installed: 0.79.6
    wanted: 0.79.6
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: false
iOS:
  hermesEnabled: true
  newArchEnabled: false


````

### Sentry.init()

````js
Sentry.init({
  dsn: APIConstants.SENTRY_URL,
  enabled: APIConstants.SENTRY_URL && !__DEV__ ? true : false,
  // spotlight: __DEV__,
  tracesSampleRate: 1.0,
  sendDefaultPii: true
});
````

### Steps to Reproduce

````js
import MyApp from "MyApp";
import React, { useEffect } from "react";
import "./src/localization/i18n";
import { GestureHandlerRootView } from "react-native-gesture-handler";
import { SafeAreaProvider } from "react-native-safe-area-context";
import * as Sentry from "@sentry/react-native";
import APIConstants from "models/APIConstants/API_CONSTANTS";

export const navigationIntegration = Sentry.reactNavigationIntegration({
});

Sentry.init({
  dsn: APIConstants.SENTRY_URL,
  enabled: APIConstants.SENTRY_URL && !__DEV__ ? true : false,
  // spotlight: __DEV__,
  tracesSampleRate: 1.0,
  sendDefaultPii: true
});
import { GoogleSignin } from "@react-native-google-signin/google-signin";
import { IOS_AUTH_CLIENTID, WEB_CLIENT_AUTH_ID } from "@env";
import { useThemeStore } from "zustand/themes/useThemeStore";
import { KeyboardProvider } from "react-native-keyboard-controller";

const App = () => {
  const backgroundColor = useThemeStore((state) => state.colors.background);

  useEffect(() => {
    configureGoogleSignin();
  }, []);

  const configureGoogleSignin = () => {
    GoogleSignin.configure({
      webClientId: WEB_CLIENT_AUTH_ID,
      iosClientId: IOS_AUTH_CLIENTID,
    });
  };

  return (
    <SafeAreaProvider style={{ flex: 1, backgroundColor: backgroundColor }}>
      <GestureHandlerRootView>
        <KeyboardProvider>
          <MyApp />
        </KeyboardProvider>
      </GestureHandlerRootView>
    </SafeAreaProvider>
  );
};

export default Sentry.wrap(App);

````

### Expected Result

No error when modifying App.tsx while in development mode

### Actual Result

Whenever I modify somethig inside App.tsx I get

`Error: You attempted to set the key `settings` with the value `{"infer_ip":"auto"}` on an object that is meant to be immutable and has been frozen., js engine: hermes`

This wasn't happening prior to updating the library to version 7.x (From 6.x)

### PR DESCRIPTION
**Pull request by alwx**

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Fixes #5187

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/src/js/client.ts
+++ b/packages/core/src/js/client.ts
@@ -49,16 +49,17 @@ export class ReactNativeClient extends Client<ReactNativeClientOptions> {
    */
   public constructor(options: ReactNativeClientOptions) {
     ignoreRequireCycleLogs(ReactNativeLibraries.ReactNativeVersion?.version);
-    options._metadata = options._metadata || {};
-    options._metadata.sdk = options._metadata.sdk || defaultSdkInfo;
-
-    // Only allow IP inferral by Relay if sendDefaultPii is true
-    if (options._metadata?.sdk) {
-      options._metadata.sdk.settings = {
-        infer_ip: options.sendDefaultPii ? 'auto' : 'never',
-        ...options._metadata.sdk.settings,
-      };
-    }
+    options._metadata = {
+      ...options._metadata,
+      sdk: {
+        ...(options._metadata?.sdk || defaultSdkInfo),
+        settings: {
+          // Only allow IP inferral by Relay if sendDefaultPii is true
+          infer_ip: options.sendDefaultPii ? 'auto' : 'never',
+          ...options._metadata?.sdk?.settings,
+        }
+      }
+    };
 
     // We default this to true, as it is the safer scenario
     options.parentSpanIsAlwaysRootSpan =
--- a/packages/core/src/js/client.ts
+++ b/packages/core/src/js/client.ts
@@ -57,8 +57,8 @@ export class ReactNativeClient extends Client<ReactNativeClientOptions> {
           // Only allow IP inferral by Relay if sendDefaultPii is true
           infer_ip: options.sendDefaultPii ? 'auto' : 'never',
           ...options._metadata?.sdk?.settings,
-        }
-      }
+        },
+      },
     };
 
     // We default this to true, as it is the safer scenario
--- a/packages/core/test/client.test.ts
+++ b/packages/core/test/client.test.ts
@@ -679,7 +679,7 @@ describe('Tests ReactNativeClient', () => {
       );
     });
 
-    test('adds ip_address {{auto}} to user if set to undefined', () => {
+    test("doesn't change infer_ip if the ip_address is set to undefined", () => {
       client.captureEvent({
         user: {
           ip_address: undefined,
@@ -692,13 +692,13 @@ describe('Tests ReactNativeClient', () => {
       expect(mockTransportSend.mock.calls[0][firstArg][envelopeItems][0][envelopeItemPayload].sdk).toEqual(
         expect.objectContaining({
           settings: {
-            infer_ip: 'never',
+            infer_ip: 'auto',
           },
         }),
       );
     });
 
-    test('adds ip_address undefined to user if not set', () => {
+    test("doesn't change infer_ip if the user is not set", () => {
       client.captureEvent({
         user: {},
       });
@@ -707,20 +707,20 @@ describe('Tests ReactNativeClient', () => {
       expect(mockTransportSend.mock.calls[0][firstArg][envelopeItems][0][envelopeItemPayload].sdk).toEqual(
         expect.objectContaining({
           settings: {
-            infer_ip: 'never',
+            infer_ip: 'auto',
           },
         }),
       );
     });
 
-    test('leaves ip_address undefined to undefined user', () => {
+    test("doesn't change infer_ip if the event is empty", () => {
       client.captureEvent({});
 
       expect(mockTransportSend.mock.calls[0][firstArg][envelopeItems][0][envelopeItemPayload].user).toBeUndefined();
       expect(mockTransportSend.mock.calls[0][firstArg][envelopeItems][0][envelopeItemPayload].sdk).toEqual(
         expect.objectContaining({
           settings: {
-            infer_ip: 'never',
+            infer_ip: 'auto',
           },
         }),
       );
--- a/packages/core/test/client.test.ts
+++ b/packages/core/test/client.test.ts
@@ -679,7 +679,7 @@ describe('Tests ReactNativeClient', () => {
       );
     });
 
-    test('doesn\'t change infer_ip if the ip_address is set to undefined', () => {
+    test("doesn't change infer_ip if the ip_address is set to undefined", () => {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: packages/core/src/js/client.ts [hunk 1]
```diff
@@ -49,16 +49,17 @@ export class ReactNativeClient extends Client<ReactNativeClientOptions> {
...
+      ...options._metadata,
+      sdk: {
+        ...(options._metadata?.sdk || defaultSdkInfo),
+        settings: {
+          // Only allow IP inferral by Relay if sendDefaultPii is true
+          infer_ip: options.sendDefaultPii ? 'auto' : 'never',
+          ...options._metadata?.sdk?.settings,
+        },
+      },
+    };
```
lucas-zimerman, line 62: The changes follows the new implementation https://github.com/getsentry/sentry-javascript/pull/17364

alwx:
I've updated this PR and updated some tests that were not correct — all those tests were checking if `infer_ip` is set to `never` which should happen only if `setDefaultPii` is set to `false`.
There are no new tests added since there are already tests that check for `infer_ip` value. Those are the following:
- `does not add ip_address {{auto}} to undefined user if sendDefaultPii is false`
- `doesn't change infer_ip if the ip_address is set to undefined` (I've updated the name)
- `doesn't change infer_ip if the user is not set` (I've updated the name)
- `doesn't change infer_ip if the event is empty`

packages/core/src/js/client.ts
alwx: Yes, but the issue here is that the metadata structure is immutable so instead of modifying it we can construct a new one, and that's what I'm doing in this PR. Other than that, the logic stays pretty much the same.

I've also updated tests to reflect that change and indiciate the expected behaviour which is the following:
- If `sendDefaultPii` is `true`, Sentry will infer the IP address of users' devices to events (errors, traces, replays, etc) in all browser-based SDKs.
- If `sendDefaultPii` is `false` or not set, Sentry will not infer or collect IP address data.

antonis (commented):
The change LGTM and handle the reported issue. My understanding is that it also does not break the `sendDefaultPii` fix. That said, I'm leaving the final approval to @lucas-zimerman who has more context on that patch.
We should also add a changelog entry since this is user facing and fixes an issue.

packages/core/test/client.test.ts
```diff
@@ -679,7 +679,7 @@ describe('Tests ReactNativeClient', () => {
       );
     });
 
-    test('adds ip_address {{auto}} to user if set to undefined', () => {
+    test("doesn't change infer_ip if the ip_address is set to undefined", () => {
```
lucas-zimerman: NIT:

```suggestion
    test('doesn't change infer_ip if the ip_address is set to undefined', () => {
```

packages/core/test/client.test.ts
```diff
@@ -692,13 +692,13 @@ describe('Tests ReactNativeClient', () => {
...
           settings: {
-            infer_ip: 'never',
+            infer_ip: 'auto',
           },
         }),
       );
     });
 
-    test('adds ip_address undefined to user if not set', () => {
+    test("doesn't change infer_ip if the user is not set", () => {
```
lucas-zimerman: NIT:

```suggestion
    test('doesn't change infer_ip if the user is not set', () => {
```

packages/core/test/client.test.ts
```diff
@@ -707,20 +707,20 @@ describe('Tests ReactNativeClient', () => {
...
           settings: {
-            infer_ip: 'never',
+            infer_ip: 'auto',
           },
         }),
       );
     });
 
-    test('leaves ip_address undefined to undefined user', () => {
+    test("doesn't change infer_ip if the event is empty", () => {
```
lucas-zimerman: NIT:
```suggestion
    test('doesn't change infer_ip if the event is empty', () => {
```

antonis (approved):
LGTM 🎸 
Thank you for fixing this @alwx 🙇

antonis:
If this is not a CI issue 😓 there seems to be an issue with the linter (`ERROR: "lint:prettier" exited with 1.`) which might be fixed with a `yarn fix`.

lucas-zimerman:
> If this is not a CI issue 😓 there seems to be an issue with the linter (`ERROR: "lint:prettier" exited with 1.`) which might be fixed with a `yarn fix`.

probably related to
```
  Error while parsing /home/runner/work/sentry-react-native/sentry-react-native/packages/core/node_modules/react-native/Libraries/Utilities/codegenNativeComponent.js
  Line 19, column 26: Property or signature expected.
  `parseForESLint` from parser `/home/runner/work/sentry-react-native/sentry-react-native/packages/core/node_modules/@typescript-eslint/parser/dist/index.js` is invalid and will just be ignored
```

PR #5221, opened 2025-09-29
Update `sentry-xcode.sh` and the default settings for the project.pbxproj

alwx:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Updates the `sentry-xcode.sh` and the default configuration for iOS project to support the changes done in the recent versions of React Native (0.81+): https://github.com/facebook/react-native/pull/53194

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


packages/core/scripts/sentry-xcode.sh
```diff
@@ -39,7 +39,8 @@ if [ -z "$SENTRY_CLI_EXECUTABLE" ]; then
 fi
 [ -z "$SENTRY_CLI_EXECUTABLE" ] && SENTRY_CLI_EXECUTABLE="$SENTRY_CLI_PACKAGE_PATH"
 
-REACT_NATIVE_XCODE=$1
+REACT_NATIVE_XCODE_DEFAULT="../node_modules/react-native/scripts/react-native-xcode.sh"
```
lucas-zimerman, line 42: This may not work on multirepo projects where node_modules is not where the app is located.
Since this is the same behavior, we could keep it that way and open an issue to mitigate this in the future.

alwx: Yes but it's just the default value — the path could still be accepted as an argument to this script.

alwx:
@lucas-zimerman mentioned in the changelog!

lucas-zimerman (approved):
LGTM!

Irfanwani:
how is this change solving anything?

PR #5226, opened 2025-09-30
chore: Git-Ignore Android SDK replay stub

antonis:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


antonis:
@romtsn The file is recreated on every build but I'm adding you as a reviewer in case I'm missing something

romtsn:
@antonis hm, actually it should be checked into VCS, otherwise the customer builds will be failing I believe (they will include sentry-react-native, which includes our `android` folder, which references `replay-stubs` as `compileOnly` dependency, but with this change it wouldn't exist?).

Theoretically it shouldn't be recompiled everytime you run `yarn build` because it should re-use the existing Gradle cache (unless there were changes to the Android SDK).

antonis:
Thank you for the feedback @romtsn 🙇 

> Theoretically it shouldn't be recompiled everytime you run yarn build because it should re-use the existing Gradle cache 

Ok. I'll iterate with another approach to avoid changing the jar if not needed.

> (unless there were changes to the Android SDK).

Should the [`scripts/update-android-stubs.sh`](https://github.com/getsentry/sentry-react-native/pull/5174/files#diff-08d320f774ca122d5ee3b9d37dcdc923a3bc35ab7ef094dc8beda93c0909d398) also [update](https://github.com/getsentry/sentry-react-native/pull/5194/files) jar on SDK/stub bumps?

romtsn:
> Ok. I'll iterate with another approach to avoid changing the jar if not needed.

I guess you could remove that command from `yarn build` and just run it as a separate step in CI

> Should the [scripts/update-android-stubs.sh](https://github.com/getsentry/sentry-react-native/pull/5174/files#diff-08d320f774ca122d5ee3b9d37dcdc923a3bc35ab7ef094dc8beda93c0909d398) also [update](https://github.com/getsentry/sentry-react-native/pull/5194/files) jar on SDK/stub bumps?

ideally yes, but i wanted to avoid dealing with git from CI (would have to commit + push + it's not guaranteed that nothing's broken, so would still need some manual work after all). So this script is more-so to test if the new stubs are compatible with what RN expects and after that requires manually to regenerate them and adapt the RN usage site (if anything changed)

antonis:
> ideally yes, but i wanted to avoid dealing with git from CI (would have to commit + push + it's not guaranteed that nothing's broken, so would still need some manual work after all). So this script is more-so to test if the new stubs are compatible with what RN expects and after that requires manually to regenerate them and adapt the RN usage site (if anything changed)

Sounds good @romtsn 👍 Thank you for the clarification.
So after each version bump (like [this](https://github.com/getsentry/sentry-react-native/pull/5194/files)) we should also commit the new jar artifact.

romtsn:
> So after each version bump (like [this](https://github.com/getsentry/sentry-react-native/pull/5194/files)) we should also commit the new jar artifact.

Right, but as mentioned it shouldn't update too often (if ever). And Gradle should take care of caching and not generating a new .jar if nothing's changed for the stubbed classes.

antonis:
Closing as the .jar file shouldn't be ignored

Issue #5074, opened 2025-08-13
Unable to get replay session when integrating both reactNavigationIntegration() & mobileReplayIntegration()


InterstellarStella:
### What React Native libraries do you use?

React Navigation

### Are you using [sentry.io](<http://sentry.io>) or on-premise?

[sentry.io](<http://sentry.io>) (SaS)

### @sentry/react-native SDK Version

6.19.0

### How does your development environment look like?

```json
"dependencies": {
    "@react-navigation/native": "^6.1.9",
    "@sentry/react-native": "^6.19.0",
    "react": "19.0.0",
    "react-native": "0.78.2"
  },
  "devDependencies": {
    "@babel/core": "^7.25.2",
    "@babel/preset-env": "^7.25.3",
    "@babel/runtime": "^7.25.0",
    "@react-native-community/cli": "15.0.1",
    "@react-native-community/cli-platform-android": "15.0.1",
    "@react-native-community/cli-platform-ios": "15.0.1",
    "@react-native/babel-preset": "0.78.2",
    "@react-native/eslint-config": "0.78.2",
    "@react-native/metro-config": "0.78.2",
    "@react-native/typescript-config": "0.78.2",
    "@types/jest": "^29.5.13",
    "@types/react": "^19.0.0",
    "@types/react-test-renderer": "^19.0.0",
    "eslint": "^8.19.0",
    "jest": "^29.6.3",
    "prettier": "2.8.8",
    "react-test-renderer": "19.0.0",
    "typescript": "5.0.4"
  },
  "engines": {
    "node": ">=18"
  }



```

### Sentry.init()

```js
import React, {useRef} from 'react';
import {Button, View} from 'react-native';
import * as Sentry from '@sentry/react-native';
import {NavigationContainer} from '@react-navigation/native';

export const navigationIntegration = Sentry.reactNavigationIntegration({
  enableTimeToInitialDisplay: true,
});

const configureSentry = () => {
  Sentry.init({
    dsn: 'YOUR_DSN',
    tracesSampleRate: 0.2,
    environment: 'development',
    appHangTimeoutInterval: 10,
    enableUserInteractionTracing: true,
    replaysSessionSampleRate: 0.0, // No need to record full session
    replaysOnErrorSampleRate: 1.0, // Record a replay for 100% of error sessions
    integrations: [
      Sentry.mobileReplayIntegration({
        maskAllText: false,
        maskAllImages: false,
        maskAllVectors: false,
      }),
[...]

PR #5243, opened 2025-10-07
fix(session-replay): Leave captureReplay logic to the native implementation

antonis:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Leaves captureReplay logic to [the native implementation](https://github.com/getsentry/sentry-react-native/pull/5243#discussion_r2410678941)

Fixes https://github.com/getsentry/sentry-react-native/issues/5074

When `replaysSessionSampleRate: 0.0`, and with the steps below, I always got no recording attached to the event though I should have given that `replaysOnErrorSampleRate: 1.0`. I wasn't able to associate this behavior with the `reactNavigationIntegration` as the above issue states but the fix seems to make the the SR recording more reliable.

1. Set `replaysSessionSampleRate: 0.0` [in the sample app](https://github.com/getsentry/sentry-react-native/blob/0b647531c13884b6be72c6f652482d79c6d2d96f/samples/react-native/src/App.tsx#L160). Leave `replaysOnErrorSampleRate: 1.0`
2. Tap the `Uncaught Thrown Error`
3. Notice that a replay is attached to the crash

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/src/js/replay/mobilereplay.ts
+++ b/packages/core/src/js/replay/mobilereplay.ts
@@ -137,19 +137,25 @@ export const mobileReplayIntegration = (initOptions: MobileReplayOptions = defau
       return event;
     }
 
+    const hardCrash = isHardCrash(event);
+
     const recordingReplayId = NATIVE.getCurrentReplayId();
-    if (recordingReplayId) {
+    if (!hardCrash && recordingReplayId) {
       debug.log(
         `[Sentry] ${MOBILE_REPLAY_INTEGRATION_NAME} assign already recording replay ${recordingReplayId} for event ${event.event_id}.`,
       );
       return event;
     }
 
-    const replayId = await NATIVE.captureReplay(isHardCrash(event));
+    const replayId = await NATIVE.captureReplay(hardCrash);
     if (!replayId) {
       debug.log(`[Sentry] ${MOBILE_REPLAY_INTEGRATION_NAME} not sampled for event ${event.event_id}.`);
       return event;
     }
+
+    debug.log(
+      `[Sentry] ${MOBILE_REPLAY_INTEGRATION_NAME} Captured recording replay ${replayId} for event ${event.event_id}.`,
+    );
 
     return event;
   }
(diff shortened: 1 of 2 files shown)


antonis:
@romtsn I took the liberty of adding you as a reviewer on this one since you might have more context on how things work on the native side or a suggestion for a better approach 🙏

packages/core/src/js/replay/mobilereplay.ts
```diff
@@ -137,20 +137,32 @@ export const mobileReplayIntegration = (initOptions: MobileReplayOptions = defau
       return event;
     }
 
+    const hardCrash = isHardCrash(event);
+
     const recordingReplayId = NATIVE.getCurrentReplayId();
-    if (recordingReplayId) {
+    if (!hardCrash && recordingReplayId) {
```
romtsn: nice find! I'm actually not even sure if we need this `if` altogether.. When I look at the native implementations, it seems that they take care of not doing anything if the replay is running in session mode (=replay_id is on the scope).

iOS - https://github.com/getsentry/sentry-cocoa/blob/80538ca6089b2bb1c0c23efc78b62043454ee013/Sources/Swift/Integrations/SessionReplay/SentrySessionReplay.swift#L148-L152

Android - https://github.com/getsentry/sentry-java/blob/bdbe1f4058e77d9a8b9d11482a8aaec9b56ac0c2/sentry-android-replay/src/main/java/io/sentry/android/replay/capture/SessionCaptureStrategy.kt#L64-L72

Perhaps we should just rely on the native SDKs handling that here?

romtsn: the only important thing we have to make sure is that `isHardCrash` actually returns `true` when the program is about to terminate - this is needed on Android so we don't try to send anything and potentially lose segments due to the termination

antonis: Thank you for the feedback @romtsn 🙇 

> Perhaps we should just rely on the native SDKs handling that here?

Good point 👍 I've removed the unneeded RN logic with c467718e44119f1e4e02834dfb415d0cec2c6345. Only kept the `if` statements to keep the logging consistent with the previous versions. As expected I didn't notice any change in behavior with this change.

> the only important thing we have to make sure is that isHardCrash actually returns true when the program is about to terminate

[The function](https://github.com/getsentry/sentry-react-native/blob/0b647531c13884b6be72c6f652482d79c6d2d96f/packages/core/src/js/misc.ts#L11) uses a bit of a heuristic approach but it worked as expected in my tests and it seems to have served us well till now.

romtsn (approved):
nice!

Issue #5236, opened 2025-10-03
Conflict with react-native-legal on Android


ajarl:
### What React Native libraries do you use?

Expo (mobile only), RN New Architecture, React Navigation, Expo Router, Expo Web

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### @sentry/react-native SDK Version

7.0.1

### How does your development environment look like?

See sample project https://github.com/tiagocorreiaalmeida/ReactNativeLegalWithSentry

### Sentry.init()

See sample project https://github.com/tiagocorreiaalmeida/ReactNativeLegalWithSentry

### Steps to Reproduce

When using the package `react-native-legal` to show license attributions and the Sentry React Native SDK is included, no license attributions are shown for the Android app (works fine in iOS). It seems there is some conflict with the way resources are generated on the Android side related to something that Sentry's gradle plugin does (maybe). See the following thread for more details including a reproduction sample. According to the maintainer of `react-native-legal` it seems to be something on the Sentry side; maybe some gradle task is not being executed.

https://github.com/callstackincubator/react-native-legal/issues/108

Direct link to repro project: https://github.com/tiagocorreiaalmeida/ReactNativeLegalWithSentry

### Expected Result

Sentry shouldn't have any effect on the generation of resources by other libraries. The license attributions in `react-native-legal` should be shown in Android when the Sentry SDK is applied.

### Actual Result

`react-native-legal`'s expected json file with the license data is not generated on build on Android and hence there are no licenses to show. If Sentry is not included, licenses are in fact shown, so there is some conflict between these two.

alwx:
Hello @ajarl! Thanks for reporting this. We will investigate this issue as soon as possible (probably early next week) and will come with more information.

alwx:
I'm not an expert in Android but it seems like there is some conflict between Gradle tasks created and executed by [AboutLibraries](https://github.com/mikepenz/AboutLibraries) — the Android counterpart of `react-native-legal` — and Sentry. What complicates things as well is the fact that the version of `AboutLibraries` used by `react-native-legal` is 10.8.3 which is pretty old, it was released more than 2 years ago (the current one is 12.2.4). However, I tried updating `AboutLibraries` in `react-native-legal` — it's relatively straightforward but doen't solve the issue.

Here is one thing that is puzzling:
[...]

PR #5253, opened 2025-10-09
fix(gradle): Fix compatibility with react-native-legal

romtsn:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

- This moves the entire script logic from `afterEvaluate` (which is too late to register tasks) into `androidComponents.onVariants`. Since our tasks only run when there's an android app being build, it's a good place to hook our logic into. It also gives us direct access to the variant, so we don't have to search for it and build a `releases` map anymore. The tasks are configured per-variant, omitting the `debug` variant.
- Register our tasks lazily to avoid early resolution via `register` instead of `create`

I tested it locally both `debug` and `release` builds and the `react-native-legal` json file is correctly generated in both cases. Also, our tasks run correctly for the release build, you can see here I uploaded sourcemaps: https://sentry-sdks.sentry.io/settings/projects/sentry-react-native/source-maps/23317267-0d24-5f07-9852-28630b8cf98a/

The only thing that probably needs more testing is modules collection, because it generated an empty json locally for me, but perhaps I was doing something unexpected, because I was running gradle builds directly as opposed calling `yarn/npx`. If someone could verify it works, that'd be great.

Closes https://github.com/getsentry/sentry-react-native/issues/5236

manually

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -44,7 +44,7 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@64d10c13136e1c5bce3e5fbde8d4906eeaafc885 # pin@v3.29.5
+        uses: github/codeql-action/init@f443b600d91635bebf5b0d9ebc620189c0d6fba5 # pin@v3.29.5
         with:
           languages: ${{ matrix.language }}
           # If you wish to specify custom queries, you can do so here or in a config file.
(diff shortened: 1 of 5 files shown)


packages/core/sentry.gradle
```diff
@@ -22,11 +22,265 @@ interface InjectedExecOps {
 
 def config = project.hasProperty("sentryCli") ? project.sentryCli : [];
 
+plugins.withId('com.android.application') {
+    def ac = extensions.getByName("androidComponents")
+
+    ac.onVariants(androidComponents.selector().all()) { v ->
```
lucas-zimerman: Q: v is the release?

lucas-zimerman:
Thank you for the PR! is it ready for review?

romtsn:
@lucas-zimerman not yet, I will mark it as ready soonish, still sorting out some stuff

romtsn:
@lucas-zimerman this is ready for review now.

packages/core/sentry.gradle
romtsn: v is `variant`, but since we have one down there already, I just named it `v`

PR #5291, opened 2025-10-22
chore: Remove deprecated user property

noahsmartin:
This property is deprecated and the next version of the iOS SDK will remove it entirely, so this deletes it from the RN code so that it will be compatible with the next iOS version 

#skip-changelog

### DIFF
--- a/packages/core/RNSentryCocoaTester/RNSentryCocoaTesterTests/RNSentryUserTests.m
+++ b/packages/core/RNSentryCocoaTester/RNSentryCocoaTesterTests/RNSentryUserTests.m
@@ -14,7 +14,6 @@
     [expected setIpAddress:@"192.168.1.1"];
     [expected setEmail:@"test@example.com"];
     [expected setUsername:@"testuser"];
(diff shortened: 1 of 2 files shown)


lucas-zimerman:
Shouldn't we apply this change to a major version since it removes the segment field from the current version?

noahsmartin:
@lucas-zimerman it's in a major version of the iOS SDK, it didn't seem like it was part of the public API of the RN sdk (I don't think I touched any public API in this PR) but if it does change the public API it probably needs to be a major version of the RN sdk too

lucas-zimerman:
It's safe to say it's not a break change since we don' document the segment field and the other SDKs are not using it

lucas-zimerman:
We could do a follow-up PR to remove the `segment` reference from the file `getsentry/sentry-react-native/packages/core/src/js/NativeRNSentry.ts` and `getsentry/sentry-react-native/packages/core/src/js/wrapper.ts`

lucas-zimerman (approved):
LGTM!

PR #5310, opened 2025-10-27
Remove use of integrations in tests

noahsmartin:
I previously removed some of these but forgot these two tests: https://github.com/getsentry/sentry-react-native/pull/5304

#skip-changelog


lucas-zimerman (approved):
LGTM!

Issue #4982, opened 2025-07-09
Orientation change doesn't work in replays on Android


romtsn:
### Description

Tested both in old implementation (v7) and the new one (v8) and we fail to properly continue recording after orientation change. For v7 it just stops ([replay](https://sentry-sdks.sentry.io/explore/replays/25de853b5f664db48a8c56c95e6477b2)), for v8 we continue recording but with a wrong aspect ratio so the screen appears stretched ([replay](https://sentry-sdks.sentry.io/explore/replays/7de9a3e57b3440259ecfaf7704e6b662/?project=5428561&query=&referrer=%2Fexplore%2Freplays%2F%3AreplaySlug%2F&statsPeriod=2h&yAxis=count%28%29&t=71)). Probably have to investigate how RN handles orientation change, because we're not getting the callbacks that work for native Android.

antonis:
I've spent some time investigating this today but I don't have a solution yet.
> for v8 we continue recording but with a wrong aspect ratio so the screen appears stretched ([replay](https://sentry-sdks.sentry.io/explore/replays/7de9a3e57b3440259ecfaf7704e6b662/?project=5428561&query=&referrer=%2Fexplore%2Freplays%2F%3AreplaySlug%2F&statsPeriod=2h&yAxis=count%28%29&t=71))

Confirmed that the behaviour persist on React Native SDK 7.0.1 with Android SDK from v8.21.1.

PR #5321, opened 2025-10-30
fix(session-replay): Fixes orientation change misalignment for session replay on Android

antonis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Fixes orientation change misalignment for session replay on Android by manually triggering the [onWindowSizeChanged](https://github.com/getsentry/sentry-java/blob/fcec2f233b744694bffee41e58e5a706e032ef51/sentry-android-replay/src/main/java/io/sentry/android/replay/ReplayIntegration.kt#L474) when the screen size changes. This should cover orientation changes or other screen size changes (e.g. foldables).

⚠️ This also fixes a PII issue since the orientation misalignment results in masking failures. 

Fixes https://github.com/getsentry/sentry-react-native/issues/4982

Tested with the [latest SDK](https://github.com/getsentry/sentry-java/releases/tag/8.25.0) and the `screenshotStrategy` makes a difference on how the bug is manifested

||canvas|pixelCopy|
|---|---|---|
| Before |||
| After |||
| After | [example replay](https://sentry-sdks.sentry.io/explore/replays/98b38199054b4433978160acfbf38466/?project=5428561&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&yAxis=count%28%29) | [example replay](https://sentry-sdks.sentry.io/explore/replays/98921bb21f82421f905c09558e8df935/?project=5428561&query=&referrer=%2Fexplore%2Freplays%2F&statsPeriod=1h&yAxis=count%28%29) |

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/android/src/main/java/io/sentry/react/replay/RNSentryReplayFragmentLifecycleTracer.java
+++ b/packages/core/android/src/main/java/io/sentry/react/replay/RNSentryReplayFragmentLifecycleTracer.java
@@ -25,7 +25,7 @@ public class RNSentryReplayFragmentLifecycleTracer extends FragmentLifecycleCall
   private int lastWidth = -1;
   private int lastHeight = -1;
 
-  private @Nullable View currentView;
+  private @Nullable WeakReference<View> currentViewRef;
   private @Nullable ViewTreeObserver.OnGlobalLayoutListener currentListener;
 
   public RNSentryReplayFragmentLifecycleTracer(@NotNull ILogger logger) {
@@ -62,16 +62,17 @@ public class RNSentryReplayFragmentLifecycleTracer extends FragmentLifecycleCall
           }
         };
 
-    currentView = view;
+    currentViewRef = new WeakReference<>(view);
     currentListener = listener;
 
     view.getViewTreeObserver().addOnGlobalLayoutListener(listener);
   }
 
   private void detachLayoutChangeListener() {
-    if (currentView != null && currentListener != null) {
+    final View view = currentViewRef != null ? currentViewRef.get() : null;
+    if (view != null && currentListener != null) {
       try {
-        ViewTreeObserver observer = currentView.getViewTreeObserver();
+        ViewTreeObserver observer = view.getViewTreeObserver();
         if (observer != null) {
           observer.removeOnGlobalLayoutListener(currentListener);
         }
@@ -80,7 +81,7 @@ public class RNSentryReplayFragmentLifecycleTracer extends FragmentLifecycleCall
       }
     }
 
-    currentView = null;
+    currentViewRef = null;
     currentListener = null;
   }
 


antonis:
@romtsn I took the liberty of adding you as a reviewer too since you have more context on the internals of SR on Android 🙇

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
```diff
@@ -95,6 +104,71 @@ public void onEventDispatch(Event event) {
...
+
+  private void attachLayoutChangeListener(final View view) {
+    view.getViewTreeObserver()
+        .addOnGlobalLayoutListener(
+            new ViewTreeObserver.OnGlobalLayoutListener() {
+              @Override
+              public void onGlobalLayout() {
+                checkAndNotifyWindowSizeChange(view);
+              }
+            });
```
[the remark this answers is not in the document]
antonis: Updated with c8266c346e5dca6dbff37cd8f9e5c8464e081997

lucas-zimerman:
important fix for pixel copy since it avoid leaks!

lucas-zimerman:
@sentry review

antonis:
@sentry review

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
```diff
@@ -95,6 +107,93 @@ public void onEventDispatch(Event event) {
...
+    try {
+      if (replayIntegration == null) {
+        replayIntegration = getReplayIntegration();
+      }
+
+      if (replayIntegration == null) {
+        return;
+      }
+
+      replayIntegration.onWindowSizeChanged(width, height);
```
lucas-zimerman: We can limit the try catch block to this area since the retrieval of replay is already covered.
```suggestion
   try {
      replayIntegration.onWindowSizeChanged(width, height);
```

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
```diff
@@ -95,6 +107,93 @@ public void onEventDispatch(Event event) {
...
+    } catch (Exception e) {
+      logger.log(SentryLevel.DEBUG, "Failed to check window size", e);
+    }
+  }
+
+  private void notifyReplayIntegrationOfSizeChange(int width, int height) {
+    try {
+      if (replayIntegration == null) {
+        replayIntegration = getReplayIntegration();
+      }
```
lucas-zimerman: This is a nit so feel free to not opt it out
```suggestion
      replayIntegration = (replayIntegration != null)
       ? replayIntegration
       : getReplayIntegration();
```

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
```diff
@@ -95,6 +107,93 @@ public void onEventDispatch(Event event) {
...
+
+        notifyReplayIntegrationOfSizeChange(currentWidth, currentHeight);
+      }
+    } catch (Exception e) {
+      logger.log(SentryLevel.DEBUG, "Failed to check window size", e);
+    }
+  }
+
+  private void notifyReplayIntegrationOfSizeChange(int width, int height) {
+    try {
```
lucas-zimerman: If you accept https://github.com/getsentry/sentry-react-native/pull/5321/files#r2481124409 we will need to remove this code and format the below code

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
antonis: Good catch Lucas 👍 Updated with 919dff2013c6fcfaf4b07d8124d81aff5b2af38f

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
```diff
@@ -95,6 +107,93 @@ public void onEventDispatch(Event event) {
...
+
+  private void checkAndNotifyWindowSizeChange(View view) {
+    try {
+      android.util.DisplayMetrics metrics = view.getContext().getResources().getDisplayMetrics();
+      int currentWidth = metrics.widthPixels;
+      int currentHeight = metrics.heightPixels;
+
+      if (lastWidth != currentWidth || lastHeight != currentHeight) {
+        lastWidth = currentWidth;
+        lastHeight = currentHeight;
```
lucas-zimerman: do we need to run notifyReplayIntegrationOfSizeChange when the size change didnt change? if not we should skip it early
```suggestion
      if (lastWidth == currentWidth && lastHeight == currentHeight) {
        return;
      }
      
      lastWidth = currentWidth;
      lastHeight = currentHeight;
```

lucas-zimerman:
Overall looks good! lets get the approval of @romtsn  before merging.

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
```diff
@@ -95,6 +107,94 @@ public void onEventDispatch(Event event) {
...
+    if (currentView != null && currentListener != null) {
+      try {
+        currentView.getViewTreeObserver().removeOnGlobalLayoutListener(currentListener);
+      } catch (Exception e) {
+        logger.log(SentryLevel.DEBUG, "Failed to remove layout change listener", e);
+      }
+    }
+
+    currentView = null;
+    currentListener = null;
```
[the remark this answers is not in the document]
antonis: Added with 5c6a512968ec3e0bf6d5408d16eced2a2e669ffe

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
antonis: I think the `notifyReplayIntegrationOfSizeChange` would run only if there is a size change and the flow enters the if block. Maybe the early return is cleaner but I'll consider applying it separately to also remove the closing brace

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
antonis: Yep, applied with 919dff2013c6fcfaf4b07d8124d81aff5b2af38f

alwx (approved):

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
```diff
@@ -95,6 +111,103 @@ public void onEventDispatch(Event event) {
...
+    }
+
+    try {
+      replayIntegration.onWindowSizeChanged(width, height);
+    } catch (Exception e) {
+      logger.log(SentryLevel.DEBUG, "Failed to notify replay integration of size change", e);
+    }
+  }
+
+  private @Nullable ReplayIntegration getReplayIntegration() {
```
romtsn: nit: I think you could simplify this method by calling `ScopesAdapter.getInstance().getOptions().getReplayController()` and then checking if it's `instanceof ReplayIntegration`. This would spare a few cycles.

romtsn: Alternatively, we could even expose the `onWindowSizeChanged` method in the [ReplayController](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/ReplayController.java) interface, but would require a new Android SDK release, so I'll leave it up to you to decide :)

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
```diff
@@ -95,6 +111,103 @@ public void onEventDispatch(Event event) {
             }
           }
         });
+
+    // Add layout listener to detect configuration changes after detaching any previous one
```
romtsn: so, as I understood this logic only applies when ReactNavigationIntegration is used and ttid-tracking is [enabled](https://github.com/getsentry/sentry-react-native/blob/136effd2f03627a2043cf21f2316f368524de585/packages/core/src/js/tracing/reactnavigation.ts#L113-L117). I was wondering if we should actually make it independent from the both above, and just introduce a replay-specific FragmentLifecycleCallbacks? Also, is this only a problem when fragments are used (i.e. reactNavigation), or also applies to activity-based apps?

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
antonis: Great idea @romtsn 🙇 
Updated with dbee1d40fe3d01701dca81f55ed4dbd27ceb7808 and verified that everything works as before.

> Alternatively, we could even expose the onWindowSizeChanged method in the ReplayController interface, but would require a new Android SDK release, so I'll leave it up to you to decide :)

I'll add a note to iterate on this as an improvement. It would be nice to ship the fix soon since this is also a PII.

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
antonis: Good catch @romtsn 🙇 

> make it independent from the both above, and just introduce a replay-specific FragmentLifecycleCallbacks? 

Makes sense 👍 Updated with bbddc271a3089ddb8e718df8e356969b09804d5e

> Also, is this only a problem when fragments are used (i.e. reactNavigation), or also applies to activity-based apps?

I was able to still [reproduce the issue](https://sentry-sdks.sentry.io/explore/replays/8dee59b11eba4046ba24b8c186f6f6cd/) on a bare app that seems to have only Activities. I think for most real world scenarios the fragment implementation would be used and I would opt for handling this case separately.

<details><summary>Example Activity only app</summary>
<p>

```
import React from 'react';
import { Button, View, Text, StyleSheet } from 'react-native';
import * as Sentry from '@sentry/react-native';

Sentry.init({
  dsn: 'DSN',

  // Adds more context data to events (IP address, cookies, user, etc.)
  // For more information, visit: https://docs.sentry.io/platforms/react-native/data-management/data-collected/
  sendDefaultPii: true,

  // Enable Logs
  enableLogs: true,

  // Configure Session Replay
  replaysSessionSampleRate: 0.1,
  replaysOnErrorSampleRate: 1,
  integrations: [Sentry.mobileReplayIntegration(), Sentry.feedbackIntegration()],

  // uncomment the line below to enable Spotlight (https://spotlightjs.com)
  // spotlight: __DEV__,
});

export default function App() {
  return (
    <View style={styles.container}>
      <Text style={styles.text}>Hello, Minimal App!</Text>
      <Button title='Try!' onPress={ () => { Sentry.captureException(new Error('First error')) }}/>
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#fff',
  },
  text: {
    fontSize: 20,
  },
});

export default Sentry.wrap(App);
```
</p>
</details>

romtsn: I agree, let's not care about it for now. If we get reports we can do that later 👍 Let me review the updated logic

antonis:
There seems to be a CI issue at this point irrelevant from this PR that causes failures like:

```
Error: Server answered with HTTP 500 when performing the request to https://repo.yarnpkg.com/3.6.4/packages/yarnpkg-cli/bin/yarn.js; for troubleshooting help, see https://github.com/nodejs/corepack#troubleshooting
```

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -183,6 +184,20 @@ private void initFragmentInitialFrameTracking() {
     }
   }
 
+  private void initFragmentReplayTracking() {
+    final RNSentryReplayFragmentLifecycleTracer fragmentLifecycleTracer =
+        new RNSentryReplayFragmentLifecycleTracer(logger);
+
+    final @Nullable FragmentActivity fragmentActivity = (FragmentActivity) getCurrentActivity();
```
romtsn: I assume it should be always a `FragmentActivity` but maybe worth to have an `instanceof` check just in case?

packages/core/android/src/main/java/io/sentry/react/replay/RNSentryReplayFragmentLifecycleTracer.java
```diff
@@ -0,0 +1,137 @@
...
+import io.sentry.android.replay.ReplayIntegration;
+import java.lang.ref.WeakReference;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class RNSentryReplayFragmentLifecycleTracer extends FragmentLifecycleCallbacks {
+  private @NotNull final ILogger logger;
+
+  @SuppressWarnings("PMD.AvoidUsingVolatile")
+  private @Nullable volatile ReplayIntegration replayIntegration;
```
romtsn: I assume `volatile` is not really needed here since everything is being accessed from the main thread (OnGlobalLayoutListener is likely called on the main thread too). But it's also fine as it is

packages/core/android/src/main/java/io/sentry/react/replay/RNSentryReplayFragmentLifecycleTracer.java
```diff
@@ -0,0 +1,137 @@
...
+
+  private int lastWidth = -1;
+  private int lastHeight = -1;
+
+  private @Nullable View currentView;
```
romtsn: theoretically this shouldn't leak since we nullify it in `onFragmentViewDestroyed` but maybe worth to have it as a WeakRef just in case?

antonis: Good idea 👍 Updated with f56e75cc97e8bad611a04ff032ecf9279fbe7316

packages/core/android/src/main/java/io/sentry/react/replay/RNSentryReplayFragmentLifecycleTracer.java
antonis: Updated with f96359044bb65155eee407a0185b337775913da2

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
antonis: I used the pattern from above but I think it's a good idea to add the check 👍 Updated with a8ad16396cf1d5d854924f48bb27fc6172f79bb4

PR #5354, opened 2025-11-10
feat: logsOrigin implementation

lucas-zimerman:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Before this change, users couldn't opt out for which layers they don't want logs, JavaScript and Native both produces logs that may or may not be interesting to their use.

The changes allows users to choose between `all`, `js` and `native. As the name implies you can choose if only JavaScript, Native or both will be allowed to create logs.

Simulator, Tests.

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,16 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
-## 7.6.0
+## Unreleased
 
 ### Features
 
 - Added `logsOrigin` to Sentry Options ([#5354](https://github.com/getsentry/sentry-react-native/pull/5354))
   - You can now choose which logs are captured: 'native' for logs from native code only, 'js' for logs from the JavaScript layer only, or 'all' for both layers.
   - Takes effect only if `enableLogs` is `true` and defaults to 'all', preserving previous behavior.
+
+## 7.6.0
+
 
 ### Fixes
 


packages/core/src/js/wrapper.ts
```diff
@@ -273,7 +277,7 @@ export const NATIVE: SentryNativeWrapper = {
 
     // filter out all the options that would crash native.
     /* eslint-disable @typescript-eslint/unbound-method,@typescript-eslint/no-unused-vars */
-    const { beforeSend, beforeBreadcrumb, beforeSendTransaction, integrations, ignoreErrors, ...filteredOptions } =
+    const { beforeSend, beforeBreadcrumb, beforeSendTransaction, integrations, ignoreErrors, loggerOrigin, ...filteredOptions } =
       options;
     /* eslint-enable @typescript-eslint/unbound-method,@typescript-eslint/no-unused-vars */
     const nativeIsReady = await RNSentry.initNativeSdk(filteredOptions);
```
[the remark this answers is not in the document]
lucas-zimerman: logEnricherIntegration() and consoleLoggingIntegration(), are integrations that only interact with JavaScript logs, they don't touch any Native logs.

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@
 
 ## Unreleased
 
+### Features
+
+- added `logsOrigin` to Sentry Options ([#5354](https://github.com/getsentry/sentry-react-native/pull/5354))
+  Now you can choose which logs will be captured: 'native' for only Logs that comes from the Native code, 'js' for logs that only comes from the JavaScript layer, or 'all' for both layers. This will only have any impact if `enableLogs` is set to `true`.
+  By default it is set to `all` so it behaves the same before this change.
+
```
antonis: nit: Formatting and rephrasing suggestion
```suggestion
- Added `logsOrigin` to Sentry Options ([#5354](https://github.com/getsentry/sentry-react-native/pull/5354))
  - You can now choose which logs are captured: 'native' for logs from native code only, 'js' for logs from the JavaScript layer only, or 'all' for both layers.
  - Takes effect only if `enableLogs` is `true` and defaults to 'all', preserving previous behavior.
```

packages/core/android/libs/replay-stubs.jar
antonis, line 1: nit: we can skip the replay-stubs change for this PR

packages/core/src/js/client.ts
antonis, line 84: q: Is the temporary `enableLogs` value checked only here? If yes maybe we can add the `options.logsOrigin !== 'native'` check to avoid the need for `originalEnableLogs`.

lucas-zimerman: The temporary enable log is set due to `super(options);` we need to disable `enableLogs` before calling it. Instead of copying a clean copy of options, I opted to only patch the boolean flag. I will add a comment about it

antonis: Sounds good 👍 
Thank you for the explanation 🙇

PR #5440, opened 2025-12-04
feat(metrics): Updates Metrics code to correctly process envelops with`application/vnd.sentry.items.trace-metric+json` content type

alwx:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Updates metrics to be processed correctly.
In addition to that, updates some of the tests that were incorrectly using content type for logs.

Tests, tests, tests. And just running the app and seing myself.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Features
 
-- Adds Metrics Beta ([#5402](https://github.com/getsentry/sentry-react-native/pull/5402))
+- Adds Metrics Beta for iOS ([#5402](https://github.com/getsentry/sentry-react-native/pull/5402))
 - Improves Expo Router integration to optionally include full paths to components instead of just component names ([#5414](https://github.com/getsentry/sentry-react-native/pull/5414))
 - Report slow and frozen frames as TTID/TTFD span data ([#5419](https://github.com/getsentry/sentry-react-native/pull/5419))
 - Report slow and frozen frames on spans created through the API ([#5420](https://github.com/getsentry/sentry-react-native/issues/5420))


packages/core/test/wrapper.test.ts
```diff
@@ -391,7 +399,7 @@ describe('Tests Native Wrapper', () => {
         base64StringFromByteArray(
           utf8ToBytes(
             '{"event_id":"event0","sent_at":"123"}\n' +
-              '{"type":"event","content_type":"application/vnd.sentry.items.log+json","length":87}\n' +
+              '{"type":"event","content_type":"application/octet-stream","length":87}\n' +
```
lucas-zimerman: Could you double check that the JavaScript SDK sends octet-stream for events?
I ask that because octet-stream is usually only used for Attachments.

lucas-zimerman: for reference, I think it should be application/json https://develop.sentry.dev/sdk/data-model/envelopes/#full-examples
but `application/vnd.sentry.items.log+json` worked with it so this field might be ignored when type is event

packages/core/test/wrapper.test.ts
```diff
@@ -1,4 +1,12 @@
-import type { Event, EventEnvelope, EventItem, SeverityLevel } from '@sentry/core';
+import type {
+  EnvelopeItem,
```
antonis: I think we can remove this since it is unused

```suggestion
```

packages/core/test/wrapper.test.ts
alwx: @lucas-zimerman I would say it doesn't matter much since it's simply a callback — the proper value is specified in the header itself.
but it makes sense to change it to `application/json` for clarity — will do that after having a quick lunch

alwx: done

lucas-zimerman:
@sentry review

alwx:
@lucas-zimerman that's done now

alwx:
@lucas-zimerman you seem to be correct — the enabled native integration kills it. But as I understood, it seems to be related to unknown spans on Android, which means it requires no changes to RN code and this PR in particular, right?

antonis:
> @lucas-zimerman you seem to be correct — the enabled native integration kills it. But as I understood, it seems to be related to unknown spans on Android, which means it requires no changes to RN code and this PR in particular, right?

@alwx I think this is the case 👍  Maybe we can [update the changelog](https://github.com/getsentry/sentry-react-native/blob/main/CHANGELOG.md#features) in this (or a separate PR) clarifying that the metrics are released in beta only for iOS and merge/release this fix.
We can then add a changelog for the Android beta release of metrics when we merge the Android SDK bump that includes the fix. Wdyt?

alwx:
@antonis makes sense, I'd do that. @lucas-zimerman what do you think?

lucas-zimerman:
I also agree.
A PR to fix this is on the way, but I have no idea when it will be released: https://github.com/getsentry/sentry-java/pull/4956

EDIT: There is no additional changes required on this PR, should we wait for the Android update so both iOS/Android have support for Metrics, or release this PR mentioning that it's only supported on iOS?

antonis:
> release this PR mentioning that it's only supported on iOS?

Since it is a beta release I think it would be ok to release only on iOS. I'll leave that up to @alwx 🙇

antonis (approved):
LGTM 🚀

alwx:
@lucas-zimerman @antonis I've updated the changelog entry with information that it only works on iOS.

Issue #5376, opened 2025-11-14
Replay: Refactor the retrieval of ReplayID


lucas-zimerman:
### Problem Statement

https://github.com/getsentry/sentry-react-native/blob/493c1a1f0b5ee2f015d3d3a440193a5c4234509c/packages/core/src/js/replay/mobilereplay.ts#L194-L199
Parts of the code uses this data, such as Logs, if we get an excessive amount of requests to this endpoint, we will generate a lot of overhead to the app.

### Solution Brainstorm

We should emit when replay IDs are created on the native layer and cache it on the JavaScript layer to avoid too much bridge interaction

### Are you willing to submit a PR?

None

PR #5449, opened 2025-12-09
Add caching to `getReplayId`

alwx:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Improves performance by introducing caching to `getReplayId` in `mobileReplayIntegration`

Fixes https://github.com/getsentry/sentry-react-native/issues/5376

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/.github/workflows/e2e-v2.yml
+++ b/.github/workflows/e2e-v2.yml
@@ -179,7 +179,7 @@ jobs:
     strategy:
       fail-fast: false # keeps matrix running if one fails
       matrix:
-        rn-version: ['0.71.11', '0.81.0']
+        rn-version: ['0.71.19', '0.82.1']
         rn-architecture: ['legacy', 'new']
         platform: ['android', 'ios']
         build-type: ['production']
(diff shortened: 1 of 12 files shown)


packages/core/src/js/replay/mobilereplay.ts
```diff
@@ -192,19 +211,23 @@ export const mobileReplayIntegration = (initOptions: MobileReplayOptions = defau
...
         );
       } else {
+        updateCachedReplayId(null);
         debug.log(`[Sentry] ${MOBILE_REPLAY_INTEGRATION_NAME} not sampled for event ${event.event_id}.`);
       }
-    } else {
-      debug.log(
-        `[Sentry] ${MOBILE_REPLAY_INTEGRATION_NAME} Captured recording replay ${replayId} for event ${event.event_id}.`,
-      );
     }
```
alwx, line 231: @lucas-zimerman I wonder if it's sufficient. In this case, `updateCachedReplayId` is being called only when `processEvent` is called, and I think it makes sense — we call `NATIVE.captureReplay` there, get the ID, and update it in our "cache".
BUT is it possible for replay ID to be updated outside of this scenario?

antonis: I think the replay id would change on the native side since we get the replay id from the current native scope ([Android](https://github.com/getsentry/sentry-react-native/blob/f6eddd20c3fa2ec57a1198da3df63541d65f7aee/packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java#L619), [iOS](https://github.com/getsentry/sentry-cocoa/blob/0af71b2e6cb990eeeb1f61a59b684695ba6609df/Sources/Sentry/PrivateSentrySDKOnly.m#L347))
I think there is no reliable way to cache the value on RN without a callback from the native code.

antonis (approved):
LGTM and worked as expected in my tests 🎸 
Thank you for this improvement @alwx 🙇 

We can iterate later with [better caching](https://github.com/getsentry/sentry-react-native/pull/5449/files#r2602412212) if this feasible from the native side.

PR #5493, opened 2025-12-19
fix(appstart): Skip span ID check when standalone mode is enabled

antonis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

When standalone mode is enabled, the app start integration creates its own transaction. However, the span ID validation check was still comparing against the first navigation transaction's span ID, causing app start data to be dropped when React Navigation auto-instrumentation starts a navigation transaction before the standalone app start transaction is created.

This fix skips the span ID check when `standalone: true`, allowing app start to be attached to the standalone transaction regardless of whether a navigation transaction started first.

Fixes https://github.com/getsentry/sentry-react-native/issues/5494

Manual, CI

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/.github/workflows/sample-application.yml
+++ b/.github/workflows/sample-application.yml
@@ -15,6 +15,7 @@ concurrency:
 env:
   SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
   MAESTRO_VERSION: '2.0.10'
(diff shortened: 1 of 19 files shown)


alwx (approved):
Looks good!

Issue #5499, opened 2025-12-22
Add AGENTS.md with instructions for AI agents


antonis:
[AGENTS.md](https://agents.md/) is the upcoming new standard for documenting instructions for AI agents supported by collaborative efforts across the AI software development ecosystem, including [OpenAI Codex](https://openai.com/codex/), [Amp](https://ampcode.com/), [Jules from Google](https://jules.google/), [Cursor](https://cursor.com/), and [Factory](https://factory.ai/).

See similar Cocoa file [https://github.com/getsentry/sentry-cocoa/blob/main/AGENTS.md](https://github.com/getsentry/sentry-cocoa/blob/main/AGENTS.md)

PR #5510, opened 2026-01-05
AGENTS.md file added

alwx:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Adds AGENTS.md, fixes https://github.com/getsentry/sentry-react-native/issues/5499
The added file is partially based on https://github.com/getsentry/sentry-cocoa/blob/main/AGENTS.md

#skip-changelog

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,631 @@
+# AGENTS.md
+
+This file provides comprehensive guidance for AI coding agents working with the Sentry React Native SDK repository.
+
+## Agent Responsibilities
+
+- **Continuous Learning**: Whenever an agent performs a task and discovers new patterns, conventions, or best practices that aren't documented here, it should add these learnings to AGENTS.md. This ensures the documentation stays current and helps future agents work more effectively.
+- **Context Management**: When using compaction (which reduces context by summarizing older messages), the agent must re-read AGENTS.md afterwards to ensure it's always fully available in context. This guarantees that all guidelines, conventions, and best practices remain accessible throughout the entire session.
+
+## Best Practices
+
+### Compilation & Testing
+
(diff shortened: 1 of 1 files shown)


lucas-zimerman:
is this only internal or clients will also use it?
Should we mark it with #skip-changelog and with the label ready-to-merge?

antonis:
> is this only internal or clients will also use it?

My understanding is that this would be used both internally and externally since it is open.

lucas-zimerman:
> > 
> 
> My understanding is that this would be used both internally and externally since it is open.

Does it make sense to add it to the changelog?

lucas-zimerman:
@sentry review

antonis:
>> My understanding is that this would be used both internally and externally since it is open.
>
> Does it make sense to add it to the changelog?

I do not have a strong opinion but I think not since it is targeting agents and not humans. Other platforms skipped the [changelog too](https://github.com/getsentry/sentry-dart/pull/3426)

AGENTS.md
```diff
@@ -0,0 +1,631 @@
...
+### Cross-Platform Dependencies
+
+When making changes, consider impact on:
+- **React Native**: Core platform
+- **Expo**: Managed workflow support
+- **Sentry Cocoa**: iOS native SDK
+- **Sentry Java/Android**: Android native SDK
+- **Flutter**: Flutter SDK depends on native SDKs
+- **.NET**: .NET MAUI depends on native SDKs
+- **Unity**: Unity SDK depends on native SDKs
```
lucas-zimerman, line 440: Q: shouldn't we only consider the impact on React Native and maybe cocoa/java?

antonis: +1 on this. We may also need to add the Javascript SDK here

PR #5519, opened 2026-01-09
ci(release): Switch from action-prepare-release to Craft

BYK:
## Summary

This PR migrates from the deprecated `action-prepare-release` to the new Craft GitHub Actions.

## Changes

- Migrated `.github/workflows/release.yml` to Craft reusable workflow

## Documentation

See https://getsentry.github.io/craft/github-actions/ for more information.

### DIFF
--- a/.github/workflows/changelog-preview.yml
+++ b/.github/workflows/changelog-preview.yml
@@ -13,5 +13,5 @@ permissions:
 
 jobs:
   changelog-preview:
-    uses: getsentry/craft/.github/workflows/changelog-preview.yml@v2
+    uses: getsentry/craft/.github/workflows/changelog-preview.yml@39ee616a6a58dc64797feecb145d66770492b66c # V2
     secrets: inherit


antonis:
Added [the `ready-to-merge` tag](https://github.com/getsentry/sentry-react-native/blob/main/CI.md) so that all CI checks would run

.github/workflows/changelog-preview.yml
```diff
@@ -0,0 +1,17 @@
...
+    - reopened
+    - edited
+    - labeled
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  changelog-preview:
+    uses: getsentry/craft/.github/workflows/changelog-preview.yml@v2
```
lucas-zimerman: Q: Shouldn't we use a fixed commit hash here?
```suggestion
    uses: getsentry/craft/.github/workflows/changelog-preview.yml@39ee616a6a58dc64797feecb145d66770492b66c # V2
```

BYK: I prefer floating tags for Craft since we control it and it's much easier to update that way but the consensus is to using SHAs for increased security. Will change.

Issue #5529, opened 2026-01-13
iOS Duplicate JS unhandled issues in new architecture


gkartalis:
### What React Native libraries do you use?

RN New Architecture, React Navigation, Hermes, Expo (mobile only)

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### @sentry/react-native SDK Version

6.21.0

### How does your development environment look like?

https://github.com/artsy/eigen

````
System:
  OS: macOS 26.1
  CPU: (12) arm64 Apple M2 Pro
  Memory: 411.53 MB / 32.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 24.6.0
    path: ~/.local/share/mise/installs/node/24.6.0/bin/node
  Yarn:
    version: 4.10.3
    path: ~/.local/share/mise/installs/yarn/4.10.3/bin/yarn
  npm:
    version: 11.5.1
    path: ~/.local/share/mise/installs/node/24.6.0/bin/npm
  Watchman:
    version: 2025.11.10.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.16.2
    path: /Users/georgioskartalis/.local/share/mise/installs/ruby/3.1.7/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.0
      - iOS 26.0
      - macOS 26.0
      - tvOS 26.0
      - visionOS 26.0
      - watchOS 26.0
  Android SDK:
    API Levels:
      - "30"
      - "34"
      - "35"
      - "36"
    Build Tools:
      - 33.0.1
      - 34.0.0
      - 35.0.0
      - 36.0.0
      - 36.1.0
    System Images:
      - android-30 | Google APIs ARM 64 v8a
      - android-34 | Google Play ARM 64 v8a
      - android-35 | Google Play ARM 64 v8a
      - android-35 | Pre-Release 16 KB Page Size Google Play ARM 64 v8a
    Android NDK: Not Found
IDEs:
  Android Studio: 2025.2 AI-252.25557.131.2521.14432022
  Xcode:
    version: 26.0.1/17A400
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.11
    path: /Users/georgioskartalis/.local/share/mise/installs/java/zulu-17.50.19.0/bin/javac
  Ruby:
    version: 3.1.7
[...]

PR #5532, opened 2026-01-14
fix(ios): Fix duplicate JS error reporting on iOS with New Architecture

antonis:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Fix duplicate JS error reporting on iOS with New Architecture.

With New Architecture, React Native wraps JS errors in C++ exceptions. These exceptions are caught by the native crash handler and should be filtered out since the JS error is already reported by the JS error handler. The key indicator is "ExceptionsManager.reportException" in the exception value, which is React Native's mechanism for reporting JS errors to the native layer.

Fixes https://github.com/getsentry/sentry-react-native/issues/5529

CI

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


antonis:
@sentry review

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### Fixes
 
+- Fix duplicate error reporting on iOS with New Architecture ([#5532](https://github.com/getsentry/sentry-react-native/pull/5532))
```
lucas-zimerman, line 13: Shouldn't we specify  if this was fixed on a specific version of React Native onwards?

lucas-zimerman (approved):
Other than the changelog question, LGTM!

CHANGELOG.md
antonis: I think this behavior is part of New Architecture's error handling, so the fix applies to all RN versions with New Architecture enabled (0.69+), not just a specific version

PR #5534, opened 2026-01-14
ci: Add changelog preview flow

BYK:
This PR adds the `unlabeled` trigger to the changelog preview workflow to ensure the preview is updated when a label is removed.


.github/workflows/changelog-preview.yml
BYK, line 9:
```suggestion
    - labeled
    - unlabled
```

PR #5542, opened 2026-01-15
chore(android): Increase Android diffMax to 5.5 MB

antonis:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

⚠️ Based on https://github.com/getsentry/sentry-react-native/pull/5539

## :scroll: Description

Updated the threshold to 5.5 MiB. This will pass since the actual difference is 5.44 MiB, and it provides a tighter limit with a small buffer (~62 KB) for minor fluctuations.

Opened a separate PR to keep track of the change since [the binary size check failed](https://github.com/getsentry/sentry-react-native/actions/runs/21030357158/job/60472951104?pr=5539) in the [8.30.0 Android bump](https://github.com/getsentry/sentry-react-native/pull/5539)

```
BinarySizeTest > app size() STANDARD_ERROR
    Jan 15, 2026 1:34:36 PM BinarySizeTest app size
    INFO: App com.testappplain  size is 43.94 MiB
    Jan 15, 2026 1:34:36 PM BinarySizeTest app size
    INFO: App com.testappsentry size is 49.38 MiB
    Jan 15, 2026 1:34:36 PM BinarySizeTest app size
    INFO: App com.testappsentry is 5.44 MiB larger than app com.testappplain

BinarySizeTest > app size() FAILED
    java.lang.AssertionError: 5705106 should be < 5242880
        at BinarySizeTest.app size(BinarySizeTest.kt:37)
```

CI

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/performance-tests/metrics-android.yml
+++ b/performance-tests/metrics-android.yml
@@ -13,4 +13,4 @@ startupTimeTest:
 
 binarySizeTest:
   diffMin: 1 MiB
-  diffMax: 5.0 MiB
+  diffMax: 5.5 MiB


alwx (approved):
I am wondering if we could increase it even more because otherwise it will likely break again once the new native SDKs are released. But that also works for now.

antonis:
> I am wondering if we could increase it even more because otherwise it will likely break again once the new native SDKs are released. But that also works for now.

Good point. I was thinking of bumping it to 6 directly but a 0.5 increase seems big already. We can bump it again when the test fails and keep track of the change.

PR #5571, opened 2026-01-23
ci(release): Fix changelog-preview permissions

BYK:
The changelog-preview reusable workflow now requires `statuses: write`
permission to function correctly with GitHub App installations that
declare permissions statically.

### DIFF
--- a/.github/workflows/changelog-preview.yml
+++ b/.github/workflows/changelog-preview.yml
@@ -10,6 +10,7 @@ on:
 permissions:
   contents: write
   pull-requests: write
+  statuses: write
 
 jobs:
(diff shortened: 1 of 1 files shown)


lucas-zimerman:
@BYK  @antonis The PR uses a fixed hash, so I assume we also need to bump it in order for this PR to make sense `getsentry/craft/.github/workflows/changelog-preview.yml@c6e2f04939b6ee67030588afbb5af76b127d8203`

antonis:
@lucas-zimerman I think this bump should fix it https://github.com/getsentry/sentry-react-native/pull/5574

Issue #5572, opened 2026-01-23
Investigate and fix Native Tests / ios flakiness


antonis:
The Native Tests / ios ([example](<https://github.com/getsentry/sentry-react-native/actions/runs/21285494916/job/61266250165?pr=5470>)) is flaky recently probably due to the unavailability of the requested simulator 

```
Run env NSUnbufferedIO=YES \
2026-01-23 12:18:28.286 xcodebuild[7038:22754] Writing error result bundle to /var/folders/kg/7q73ww8s3llgyl61c9z_j5g40000gn/T/ResultBundle_2026-23-01_12-18-0028.xcresult
xcodebuild: error: Unable to find a device matching the provided destination specifier:
		{ platform:iOS Simulator, OS:latest, name:iPhone 16 }

	The requested device could not be found because no available devices matched the request.

	Available destinations for the "RNSentryCocoaTester" scheme:
		{ platform:iOS, id:dvtdevice-DVTiPhonePlaceholder-iphoneos:placeholder, name:Any iOS Device }
		{ platform:iOS Simulator, id:dvtdevice-DVTiOSDeviceSimulatorPlaceholder-iphonesimulator:placeholder, name:Any iOS Simulator Device }
Error: Process completed with exit code 70.
```

PR #5577, opened 2026-01-26
chore(ci): Dynamically select available iPhone simulator for tests

antonis:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Instead of hardcoding device name and OS version, dynamically find the first
available iPhone simulator using xcrun simctl. This prevents failures when
specific device/OS combinations aren't available on CI runners.

Fixes https://github.com/getsentry/sentry-react-native/issues/5572

CI, I've [run the native ios test twice and was](https://github.com/getsentry/sentry-react-native/actions/runs/21354957156/job/61465492202?pr=5577) 🟢 

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/.github/workflows/native-tests.yml
+++ b/.github/workflows/native-tests.yml
@@ -58,8 +58,17 @@ jobs:
           SCHEME: RNSentryCocoaTester
           CONFIGURATION: Release
         run: |
-          # Find first available iPhone simulator
-          DEVICE_ID=$(xcrun simctl list devices available iPhone -j | jq -r '.devices | to_entries[].value[] | select(.isAvailable == true) | .udid' | head -1)
+          # Find first available iPhone simulator from latest iOS runtime
+          DEVICE_ID=$(xcrun simctl list devices available iPhone -j | jq -r '
+            .devices |
+            to_entries |
+            map(select(.key | startswith("com.apple.CoreSimulator.SimRuntime.iOS-"))) |
+            sort_by(.key) |
+            reverse |
+            .[0].value[] |
+            select(.isAvailable == true) |
+            .udid
+          ' | head -1)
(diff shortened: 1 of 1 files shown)


.github/workflows/native-tests.yml
```diff
@@ -49,17 +49,27 @@ jobs:
...
+
       - name: Run iOS Tests
         working-directory: packages/core/RNSentryCocoaTester
         env:
           SCHEME: RNSentryCocoaTester
           CONFIGURATION: Release
-          DESTINATION: 'platform=iOS Simulator,OS=latest,name=iPhone 16'
         run: |
+          # Find first available iPhone simulator
+          DEVICE_ID=$(xcrun simctl list devices available iPhone -j | jq -r '.devices | to_entries[].value[] | select(.isAvailable == true) | .udid' | head -1)
```
alwx: One quick question: I noticed that this command does return "iPhone 16" with iOS 18.6 when I run it so it seems like there is no sorting applied there, at least from what I see when running this locally. Could it be that the destination will become some older simulator with some older version of iOS or is it practically impossible?

antonis: Great catch @alwx 👍 Updated with a253400928cc3d242e489e44f14c9b44a6cd048a
Let's see if everything works on CI now 🤞

antonis: [Rerun the test a couple of times](https://github.com/getsentry/sentry-react-native/actions/runs/21359617492/job/61482192339?pr=5577) with the latest changes and stayed 🟢 
@alwx Ready for another pass 🙇

PR #5501, opened 2025-12-23
[React-Native]: v8

antonis:
## Description
- Bump Cocoa to v9
- Release Capture Startup Crashes

### 8.0.0-alpha.0
- CLI v3
- https://github.com/getsentry/sentry-react-native/pull/5356
- https://github.com/getsentry/sentry-react-native/pull/5582
- https://github.com/getsentry/sentry-react-native/pull/5578

### 8.0.0-beta.0
- https://github.com/getsentry/sentry-react-native/pull/5515

### GA
- https://github.com/getsentry/sentry-react-native/pull/5528
- https://github.com/getsentry/sentry-react-native/pull/5619
- https://github.com/getsentry/sentry-react-native/pull/5656

#### Minimum Version Requirements

- **iOS/macOS/tvOS**: ([#5356](https://github.com/getsentry/sentry-react-native/pull/5356))
  - iOS **15.0+** (previously 11.0+)
  - macOS **10.14+** (previously 10.13+)
  - tvOS **15.0+** (previously 11.0+)

- **Android**: ([#5578](https://github.com/getsentry/sentry-react-native/pull/5578))
  - Sentry Android Gradle Plugin **6.0.0** (previously 5.x)
  - Android Gradle Plugin **7.4.0+** (previously 7.3.0+)
  - Kotlin **1.8+**

- **Sentry Self-Hosted**: ([#5523](https://github.com/getsentry/sentry-react-native/pull/5523))
  - Sentry CLI v3 requires self-hosted **25.11.1+** (previously 25.2.0)

#skip-changelog

PR #5579, opened 2026-01-26
Support for applying scope attributes to logs

### LINKED ISSUE #5488: Apply Scope Attributes to Logs
**The **[**JS SDK 10.32.0**](https://github.com/getsentry/sentry-react-native/pull/5487)** adds scope attributes to logs (see below) and we need to check if this **[**should be synced to native**](https://github.com/getsentry/sentry-react-native/pull/5487#discussion_r2631119494)**.**<br>****<br>**feat(core): Apply scope attributes to logs (**[**#18184**](https://github-redirect.dependabot.com/getsentry/sentry-javascript/pull/18184)**)**

You can now set attributes on the SDK's scopes which will be applied to all logs as long as the respective scopes are active. For the time being, only `string`, `number` and `boolean` attribute values are supported.

```ts
Sentry.geGlobalScope().setAttributes({ is_admin: true, auth_provider: 'google' });
Sentry.withScope(scope => {
  scope.setAttribute('step', 'authentication');
  // scope attributes `is_admin`, `auth_provider` and `step` are added
  Sentry.logger.info(`user ${user.id} logged in`, { activeSince: 100 });
  Sentry.logger.info(`updated ${user.id} last activity`);
});
// scope attributes `is_admin` and `auth_provider` are added
Sentry.logger.warn('stale website version, reloading page');
```

alwx:
- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Implements support for applying scope attributes to logs in the Sentry React Native SDK, aligning with the JavaScript SDK 10.32.0 (https://github.com/getsentry/sentry-react-native/pull/5487).
Fixes #5488

Most of the changes actually come from JS SDK, this PR just exposes the required functions and does merging of scope attributes. I've also added tests for the changes to make sure the merging is being done correctly and that the scope attributes are being applied.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Features
 
-- Add support for applying scope attributes to logs ([#5579](https://github.com/getsentry/sentry-react-native/pull/5579))
+- Add support for applying scope attributes to JS logs ([#5579](https://github.com/getsentry/sentry-react-native/pull/5579))
 - Add experimental `sentry-span-attributes` prop to attach custom attributes to user interaction spans ([#5569](https://github.com/getsentry/sentry-react-native/pull/5569))
   ```tsx
   <Pressable


antonis:
@sentry review

antonis:
The flaky Native Tests / ios should be fixed with https://github.com/getsentry/sentry-react-native/pull/5577 🤞

lucas-zimerman:
@sentry review

lucas-zimerman:
I am reviewing the code but i dont think the bot is working

packages/core/src/js/integrations/logEnricherIntegration.ts
```diff
@@ -79,6 +79,13 @@ function processLog(log: Log, client: ReactNativeClient): void {
   // Save log.attributes to a new variable
   const logAttributes = log.attributes ?? {};
 
+  // Apply scope attributes from all active scopes (global, isolation, and current)
+  // These are applied first so they can be overridden by more specific attributes
+  const scopeAttributes = collectScopeAttributes();
```
lucas-zimerman, line 84: could we try using 
```
  const {
    user: { id, email, username },
    attributes: scopeAttributes = {},
  } = getCombinedScopeData(getIsolationScope(), currentScope);
```

I am not sure why Sentry JS Opted to only load data from isolationScope and currentScope, might be worth checking why.

alwx: That's a good suggestion and my first try as well but the thing is `getCombinedScopeData` is an internal function and not getting exported by the JS SDK. That's why I do the merging and use `getCombinedScopeAttributes` here.

alwx: Regarding the global scope: they do use it, they just call `getGlobalScope()` directly here: https://github.com/getsentry/sentry-javascript/blob/develop/packages/core/src/utils/scopeData.ts#L125

alwx:
@antonis I've added all the missing scope attributes sync methods for both iOS and Android in the latest two commits. Please check!

packages/core/src/js/integrations/logEnricherIntegration.ts
lucas-zimerman: Oh True, I did a quick test on capacitor and I was able to import it from `@sentry/core` but not on React Native

lucas-zimerman: EDIT: it should be be exported on @sentry/core

CHANGELOG.md
```diff
@@ -8,8 +8,7 @@
 
 ## Unreleased
 
-### Features
-
```
lucas-zimerman, line 12: Q: why remove the features header?

packages/core/ios/RNSentry.mm
```diff
@@ -723,6 +723,20 @@ + (SentryUser *_Nullable)userFrom:(NSDictionary *)userKeys
         configureScope:^(SentryScope *_Nonnull scope) { [scope setTagValue:value forKey:key]; }];
 }
 
+RCT_EXPORT_METHOD(setAttribute : (NSString *)key value : (NSString *)value)
+{
+    [SentrySDKWrapper
+        configureScope:^(SentryScope *_Nonnull scope) { [scope setAttribute:value forKey:key]; }];
```
lucas-zimerman: Does `setAttribute:forKey` exist here? CI is complaining about it.
Same for line  376

packages/core/src/js/integrations/logEnricherIntegration.ts
```diff
@@ -93,3 +100,44 @@ function processLog(log: Log, client: ReactNativeClient): void {
...
+  if (scope && typeof scope.getScopeData === 'function') {
+    const scopeData = scope.getScopeData();
+    const scopeAttrs = scopeData.attributes || {};
+    Object.keys(scopeAttrs).forEach((key: string) => {
+      const value = scopeAttrs[key];
+      if (typeof value === 'string' || typeof value === 'number' || typeof value === 'boolean') {
+        target[key] = value;
+      }
+    });
+  }
```
lucas-zimerman, line 123: nit: The code works great! I would suggest the following code so it's slightly shorter:
```suggestion
  const scopeAttrs = scope?.getScopeData?.().attributes || {};
  for (const [key, value] of Object.entries(scopeAttrs)) {
      if (typeof value === 'string' || typeof value === 'number' || typeof value === 'boolean') {
        target[key] = value;
      }
    }
```

packages/core/ios/RNSentry.mm
antonis: @alwx @lucas-zimerman I think [this was added recently in 9.x](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#910). Probably will need to [target v8](https://github.com/getsentry/sentry-react-native/pull/5501) for this changes. Would it make sense to ship just the JS side and do the native changes separately?

alwx: We could just leave the placeholders and TODOs in code so that when the native changes are done we're ready to quickly add them.
I can also create an issue to revisit that and move it to "Blocked".

Wdyt @lucas-zimerman @antonis ?

alwx: Something like this: https://github.com/getsentry/sentry-react-native/commit/bbdb7a5d3d6033b1df42d57ba4d8f17d0c9ec40a#diff-c68ee9159d63acd4d40c1527f895d595c9e743aa83eb69e1d8f33d31322097a1R735

antonis: Sounds good 👍 As you suggested we could open an issue to follow up

packages/core/src/js/integrations/logEnricherIntegration.ts
alwx: good one, thanks!

alwx:
@antonis could you check this one again? Should be ready to go now.

.github/workflows/e2e-v2.yml
```diff
@@ -17,7 +17,7 @@ env:
   SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
   MAESTRO_VERSION: '2.1.0'
   IOS_DEVICE: 'iPhone 17'
-  IOS_VERSION: '>=18.0'
+  IOS_VERSION: '26.2'
```
antonis: q: Why is this version bump needed?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### Features
 
+- Add support for applying scope attributes to logs ([#5579](https://github.com/getsentry/sentry-react-native/pull/5579))
```
antonis: nit: Should we mention that this is only for JS logs?

.github/workflows/e2e-v2.yml
alwx: reverted it (messed it up when experimenting with cirrus labs)

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -933,6 +933,26 @@ public void setTag(String key, String value) {
...
 
+  public void setAttribute(String key, String value) {
+    // TODO(alwx): This is not implemented in sentry-android yet
+    /*
+     * Sentry.configureScope(
+     * scope -> {
+     * scope.setAttribute(key, value);
+     * });
+     */
+  }
```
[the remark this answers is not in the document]
alwx: that's great, exactly what we wanted :)

antonis (approved):

PR #5586, opened 2026-01-27
SDK handling HTTP 413

### LINKED ISSUE #5566: SDK Handling HTTP 413 [React Native]
Relay previously returned HTTP 400 Bad Request for oversized envelopes. This was changed in Relay [#5474](<https://github.com/getsentry/relay/pull/5474>) to HTTP 413 (Content Too Large) to allow SDKs to distinguish size-related rejections from other errors. For more context on this decision, see the [DACI: Relay Returns 413 for too-large envelopes.](<https://www.notion.so/sentry/DACI-Relay-returns-413-for-too-large-envelopes-2c68b10e4b5d8034be83e524c855e73d>)

The main benefit is for our users. With the HTTP 413 response, SDKs can now log specific, actionable error messages when envelopes are rejected due to size limits. Previously, users had no clear indication of why their data was being dropped when envelopes exceeded size limits. Now, they'll understand the cause and can take steps to resolve it.

### Requirements

See the [413 section in Dealing With Network Failures in the SDK develop docs](<https://develop.sentry.dev/sdk/expected-features/#dealing-with-network-failures>) for the full specification.

alwx:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Adds handling for 413 error code.
- SDK now logs actionable error messages when envelopes are rejected due to size limits
- Prepared infrastructure to capture HTTP status codes from native transport layer for future enhancements
  
Fixes #5566 
  
## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 
 ### Features
 
+- Add handling for HTTP 413 (Content Too Large) responses from Relay ([#5566](https://github.com/getsentry/sentry-react-native/issues/5566))
+  - SDK now logs actionable error messages when envelopes are rejected due to size limits
+  - Prepared infrastructure to capture HTTP status codes from native transport layer for future enhancements
```
antonis: nit: I think we could skip this from the changelog since it is not a user facing change
```suggestion
```

packages/core/ios/RNSentry.mm
```diff
@@ -555,7 +555,9 @@ - (void)stopObserving
         [PrivateSentrySDKOnly captureEnvelope:envelope];
     }
 #endif
-    resolve(@YES);
+
+    // TODO(alwx): Capture transport response from sentry-cocoa when available
```
antonis, line 559: [The iOS implementation](https://github.com/getsentry/sentry-cocoa/pull/7261/changes#diff-bc1c8b281ca8b985ea409d4f715120884eee950d8574f24bb2d72e2cda22685dR26) may arrive soon in a next Cocoa 9.x.

antonis:
@sentry review

packages/core/ios/RNSentry.mm
alwx: could add it once it's there, shouldn't be too urgent anyway

alwx:
Removed that changelog entry in the last commit

packages/core/ios/RNSentry.mm
philippsuess: We released this yesterday with 9.3.0. But the transport doesn't expose the response code. It simply logs an error. Don't these Cocoa SDK logs don't show up in your RN logs? If yes, then all good nothing to do. If not, we need to align on how we can communicate this. This won't be straight forward, because we first store envelopes to disk and then send these async.

packages/core/src/js/wrapper.ts
```diff
@@ -218,7 +218,26 @@ export const NATIVE: SentryNativeWrapper = {
       envelopeBytes = newBytes;
     }
 
-    await RNSentry.captureEnvelope(base64StringFromByteArray(envelopeBytes), { hardCrashed });
+    const response = await RNSentry.captureEnvelope(base64StringFromByteArray(envelopeBytes), { hardCrashed });
+
+    if (response?.status) {
+      const statusCode = parseInt(response.status, 10);
```
lucas-zimerman, line 224: Thank you for the PR!, the solution is nice, but I fear just parsing an integer from a string may be flaky when other types of error happen. Would it be possible to return the `statusCode` as integer and the `errormessage` on separated parameters on the native side?

lucas-zimerman:
It would be nice to have on the sample app a button to trigger this case, would you be able to add it?

lucas-zimerman:
> We can mark the [issue](https://github.com/getsentry/sentry-react-native/issues/5566) as done once we ship an RN version with that includes both the native implementations. Wdyt

Ideally it would be nice for the native side todo that task, otherwise we may have the risk of having duplicated logs.

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -694,11 +694,16 @@ public void captureEnvelope(String rawBytes, ReadableMap options, Promise promis
     try {
       InternalSentrySdk.captureEnvelope(
           bytes, !options.hasKey("hardCrashed") || !options.getBoolean("hardCrashed"));
+
+      // TODO(alwx): Capture transport response from sentry-android when available
```
lucas-zimerman, line 698: I could be wrong, but I think the flush/transport is async so the http error will not be triggered here.

alwx:
After an investigation we figured that because `captureEnvelope` is async, we cannot use this direct approach and have to just somehow expose transport logs from the native side to JS. Will do a quick investigation to see how it could be achieved.

cc @philipphofmann

Issue #5288, opened 2025-10-21
iOS debug files (`.dSYM`) for primary application package not automatically uploading using Expo plugin


estrattonbailey:
### What React Native libraries do you use?

React Native Web, Expo Application Services (EAS), Expo (mobile only)

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### @sentry/react-native SDK Version

6.20.0

### How does your development environment look like?

````
System:
  OS: macOS 15.6
  CPU: (10) arm64 Apple M1 Max
  Memory: 99.55 MB / 64.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 20.18.3
    path: /tmp/yarn--1761081184117-0.9973840234076763/node
  Yarn:
    version: 1.22.22
    path: /tmp/yarn--1761081184117-0.9973840234076763/yarn
  npm:
    version: 10.8.2
    path: /Users/estrattonbailey/.nvm/versions/node/v20.18.3/bin/npm
  Watchman:
    version: 2025.04.14.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.16.2
    path: /opt/homebrew/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.0
      - iOS 26.0
      - macOS 26.0
      - tvOS 26.0
      - visionOS 26.0
      - watchOS 26.0
  Android SDK:
    API Levels:
      - "28"
      - "29"
      - "31"
      - "33"
      - "34"
      - "35"
    Build Tools:
      - 30.0.2
      - 30.0.3
      - 31.0.0
      - 33.0.0
      - 33.0.1
      - 34.0.0
      - 35.0.0
    System Images:
      - android-25 | Google APIs ARM 64 v8a
      - android-25 | Google APIs Intel x86 Atom
      - android-28 | Google ARM64-V8a Play ARM 64 v8a
      - android-30 | Google Play Intel x86 Atom
      - android-31 | Google APIs ARM 64 v8a
      - android-33 | Google APIs ARM 64 v8a
      - android-33 | Google Play ARM 64 v8a
      - android-35 | Google Play ARM 64 v8a
    Android NDK: Not Found
IDEs:
  Android Studio: 2023.1 AI-231.9392.1.2311.11076708
  Xcode:
    version: 26.0.1/17A400
    path: /usr/bin/xcodebuild
Languages:
  Java:
[...]

PR #5617, opened 2026-02-04
fix(iOS): Fix dSYM upload for main app in Expo EAS builds by adding inputPaths to build phase

antonis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

  1. Tests for existing functionality (test/expo-plugin/modifyXcodeProject.test.ts)
  - Added test suite "Upload Debug Symbols to Sentry build phase"
  - Tests that validate current behavior:
    - Creates build phase with correct shell script
    - Does not include inputPaths in current implementation
    - Skips creating build phase if it already exists

  2. Implementation of the fix (plugin/src/withSentryIOS.ts)
  - Added inputPaths array to the addBuildPhase options:
    - $(DWARF_DSYM_FOLDER_PATH)/$(DWARF_DSYM_FILE_NAME)/Contents/Resources/DWARF/$(PRODUCT_NAME)
    - $(DWARF_DSYM_FOLDER_PATH)/$(DWARF_DSYM_FILE_NAME)
  - These input files tell Xcode that the script depends on dSYM files being generated first, ensuring proper build ordering

  3. Tests for the fix (test/expo-plugin/modifyXcodeProject.test.ts)
  - Tests that validate the fix works correctly:
    - Includes inputPaths for dSYM files to establish build dependencies
    - Includes DWARF file path in inputPaths
    - Includes dSYM folder path in inputPaths

Fixes https://github.com/getsentry/sentry-react-native/issues/5288

CI, Manual

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/test/expo-plugin/modifyXcodeProject.test.ts
+++ b/packages/core/test/expo-plugin/modifyXcodeProject.test.ts
@@ -84,6 +84,11 @@ describe('Upload Debug Symbols to Sentry build phase', () => {
   const expectedShellScript =
     "/bin/sh `${NODE_BINARY:-node} --print \"require('path').dirname(require.resolve('@sentry/react-native/package.json')) + '/scripts/sentry-xcode-debug-files.sh'\"`";
 
+  const getOptions = () => {
+    const callArgs = addBuildPhaseSpy.mock.calls[0];
+    return callArgs[4];
+  };
+
   beforeEach(() => {
     addBuildPhaseSpy = jest.fn();
     mockXcodeProject = {
@@ -120,8 +125,7 @@ describe('Upload Debug Symbols to Sentry build phase', () => {
       shellScript: expectedShellScript,
     });
 
-    const callArgs = addBuildPhaseSpy.mock.calls[0];
-    const options = callArgs[4];
+    const options = getOptions();
 
     expect(options.inputPaths).toBeUndefined();
   });
@@ -144,8 +148,7 @@ describe('Upload Debug Symbols to Sentry build phase', () => {
       ],
     });
 
-    const callArgs = addBuildPhaseSpy.mock.calls[0];
-    const options = callArgs[4];
+    const options = getOptions();
 
     expect(options.inputPaths).toBeDefined();
     expect(options.inputPaths).toHaveLength(2);
@@ -161,8 +164,7 @@ describe('Upload Debug Symbols to Sentry build phase', () => {
       ],
     });
 
-    const callArgs = addBuildPhaseSpy.mock.calls[0];
-    const options = callArgs[4];
+    const options = getOptions();
 
     // Verify paths are wrapped in quotes to prevent pbxproj corruption
     expect(options.inputPaths[0]).toMatch(/^".*"$/);


antonis:
@sentry review

packages/core/test/expo-plugin/modifyXcodeProject.test.ts
```diff
@@ -77,3 +77,109 @@ describe('Configures iOS native project correctly', () => {
...
+
+  it('skips creating build phase if it already exists', () => {
+    mockXcodeProject.pbxItemByComment = jest.fn().mockReturnValue({
+      shellScript: 'existing',
+    });
+
+    expect(addBuildPhaseSpy).not.toHaveBeenCalled();
+  });
+
+  it('includes inputPaths with escaped quotes to avoid cordova-node-xcode serialization bug', () => {
```
lucas-zimerman: Q: Why cordova-node-xcode and what bug is it being avoided?

packages/core/plugin/src/withSentryIOS.ts
```diff
@@ -23,6 +23,10 @@ export const withSentryIOS: ConfigPlugin<string> = (config, sentryProperties: st
       xcodeProject.addBuildPhase([], 'PBXShellScriptBuildPhase', 'Upload Debug Symbols to Sentry', null, {
         shellPath: '/bin/sh',
         shellScript: `/bin/sh ${SENTRY_REACT_NATIVE_XCODE_DEBUG_FILES_PATH}`,
+        inputPaths: [
+          '"$(DWARF_DSYM_FOLDER_PATH)/$(DWARF_DSYM_FILE_NAME)/Contents/Resources/DWARF/$(PRODUCT_NAME)"',
+          '"$(DWARF_DSYM_FOLDER_PATH)/$(DWARF_DSYM_FILE_NAME)"',
+        ],
```
lucas-zimerman, line 29: magical 2 lines fix :D

packages/core/test/expo-plugin/modifyXcodeProject.test.ts
```diff
@@ -77,3 +77,109 @@ describe('Configures iOS native project correctly', () => {
...
+    };
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('creates Upload Debug Symbols build phase with correct shell script', () => {
+    const expectedShellScript =
+      "/bin/sh `${NODE_BINARY:-node} --print \"require('path').dirname(require.resolve('@sentry/react-native/package.json')) + '/scripts/sentry-xcode-debug-files.sh'\"`";
```
lucas-zimerman: This const parameter seems to be used on all tests from `Upload Debug Symbols to Sentry build phase`.
Would it be possible to move them outside `it` test but within the `describe` block?

packages/core/test/expo-plugin/modifyXcodeProject.test.ts
```diff
@@ -77,3 +77,109 @@ describe('Configures iOS native project correctly', () => {
...
+
+    mockXcodeProject.addBuildPhase([], 'PBXShellScriptBuildPhase', 'Upload Debug Symbols to Sentry', null, {
+      shellPath: '/bin/sh',
+      shellScript: expectedShellScript,
+    });
+
+    const callArgs = addBuildPhaseSpy.mock.calls[0];
+    const options = callArgs[4];
```
lucas-zimerman: We don't really need callArgs, all added tests seems to be using the same code:
```
    const callArgs = addBuildPhaseSpy.mock.calls[0];
    const options = callArgs[4];
```

We could transform this block on a function that receives the spy and return the options. What do you think?

EDIT: We don't need to pass `addBuildPhaseSpy` since it's inside the describe block, we can simply get it from there.

packages/core/test/expo-plugin/modifyXcodeProject.test.ts
antonis: Good catch @lucas-zimerman 👍

We are actually not using [`cordova-node-xcode`](https://github.com/apache/cordova-node-xcode) directly but we use `@expo/config-plugins` [that depends on it](https://github.com/expo/expo/blob/main/packages/@expo/config-plugins/package.json#L69).

The issue was [breaking the build](https://github.com/getsentry/sentry-react-native/actions/runs/21671405610/job/62483461094) and the fix was to [add escape quotes](https://github.com/getsentry/sentry-react-native/pull/5617/changes/e8453876d8730771f566880710327dd564efd554).

I've updated the test description with 987f21bbd23aeebd289d041da8c4942405f39f7f to make this less confusing.

packages/core/test/expo-plugin/modifyXcodeProject.test.ts
antonis: Good catch 👍 Updated with 1770e4af14b0b28a6c7586684c0e3f2b1957b876

packages/core/test/expo-plugin/modifyXcodeProject.test.ts
antonis: Makes sense 👍 Updated with 20bc4f0a0d8be8ddc4bdba72ce3bcd996c7efc64

Issue #5242, opened 2025-10-06
Show Native Logs


gajrajgchouhan:
### Problem Statement

Hi, 

Can we please add the native logs to Console tab as well ? Currently it is not possible to debug using only the javascript logs. Adding logs for kotlin/swift would help a lot.

### Are you willing to submit a PR?

None

antonis:
Thank you for your suggestion @gajrajgchouhan 🙇 We will consider this and iterate back.

alwx:
Hey @gajrajgchouhan!
Could you let us know what logs do you mean? Are you talking about Session Replays and logs shown there or is there something else you meant?
Thanks a lot.

antonis:
Closing due to inactivity. Feel free to comment and we will be glad to resume the investigation.

gajrajgchouhan:
Hi, sorry for the late reply. I meant the logs shown in session replay. Currently we only see from javascript side

antonis:
Thank you for the clarification @gajrajgchouhan 🙇 
We will consider this and iterate back.

alwx:
Hello @gajrajgchouhan and thanks again for your suggestions. 

The Console tab is designed to show log messages coming from your JavaScript console. If you want to combine logs from both the native side and the JavaScript side in one place, you can use the [Logs](https://docs.sentry.io/platforms/react-native/logs/) feature.

This feature includes Session Replay support — all logs you send will have a `session_id `specified, allowing you to search and filter by `session_id` and even view logs in a dedicated tab for each particular session replay.

This is how it looks in practice:

<img width="1228" height="743" alt="Image" src="https://github.com/user-attachments/assets/e08bf917-8b5b-4189-8887-0a3810a316e2" />

In this case, all but one log message were sent by JS code, while the remaining message was sent by native Android code. As you can see, everything is displayed in one place.

Please let me know if it fits your needs.

gajrajgchouhan:
Thanks @alwx , do you have the code of this screenshot? It would really help me here

alwx:
@gajrajgchouhan sure! You can find my sample app here: https://github.com/alwx-sentry/5242-logs-and-native-logs

Here is the brief description of it:
1) I've added native log statements only for Android (see `android/app/src/main/java/com/testappsentry/MainActivity.kt`): after you launch the app you have to wait 5 seconds, after that a warn message will be printed from the native side.
[...]

PR #5622, opened 2026-02-05
[iOS + Android] Add the ability to intercept errors from native side and forward them to JS console

alwx:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Was born as an attempt to fix https://github.com/getsentry/sentry-react-native/issues/5566 and turned out to be a good idea in general.
It also fixes https://github.com/getsentry/sentry-react-native/issues/5242

This PR makes it possible to intercept native logs (currently only works with iOS and only with a patch, see notes below) and forward them to JS console:
```
Sentry.init({
  onNativeLog: ({ level, component, message }) => {
    // Use logWithoutTracing to avoid feedback loops with Sentry's console integration
    logWithoutTracing(`[Sentry Native] [${level.toUpperCase()}] [${component}] ${message}`);
  },
  debug: true,
}
```
(that only works when `debug` is set to `true`)

This PR makes it so that instead of handling HTTP code 413 somehow separately we can allow users to just get native logs in their JS console and see code 413 or any other error codes themselves. Filtering per level or component is also possible.

## Notes

⚠️ Currently is works on Android without any further changes but on iOS it only works with a minor change to `SentrySDKLog.swift`. The change is done in this PR: https://github.com/getsentry/sentry-cocoa/pull/7444

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,17 +13,19 @@
 - Add `onNativeLog` callback to intercept and forward native SDK logs to JavaScript console ([#5622](https://github.com/getsentry/sentry-react-native/pull/5622))
   - The callback receives native log events with `level`, `component`, and `message` properties
   - Only works when `debug: true` is enabled in `Sentry.init`
-  - Use `logWithoutTracing` inside the callback to prevent feedback loops with Sentry's console integration
+  - Use `consoleSandbox` inside the callback to prevent feedback loops with Sentry's console integration
     ```js
     import * as Sentry from '@sentry/react-native';
 
     Sentry.init({
       debug: true,
       onNativeLog: ({ level, component, message }) => {
-        // Use logWithoutTracing to avoid feedback loops
-        Sentry.logWithoutTracing(
-          `[Sentry Native] [${level.toUpperCase()}] [${component}] ${message}`
-        );
+        // Use consoleSandbox to avoid feedback loops
+        Sentry.consoleSandbox(() => {
+          console.log(
+            `[Sentry Native] [${level.toUpperCase()}] [${component}] ${message}`
+          );
+        });
       }
     });
     ```
--- a/packages/core/src/js/index.ts
+++ b/packages/core/src/js/index.ts
@@ -47,6 +47,7 @@ export {
   setCurrentClient,
   addEventProcessor,
   lastEventId,
+  consoleSandbox,
 } from '@sentry/core';
 
 export {
--- a/packages/core/test/NativeLogListener.test.ts
+++ b/packages/core/test/NativeLogListener.test.ts
@@ -2,6 +2,10 @@ import { debug } from '@sentry/core';
 import { defaultNativeLogHandler, setupNativeLogListener } from '../src/js/NativeLogListener';
 import type { NativeLogEntry } from '../src/js/options';
 
+jest.mock('../src/js/utils/environment', () => ({
+  isExpoGo: jest.fn().mockReturnValue(false),
+}));
+
 jest.mock('react-native', () => ({
   NativeModules: {
     RNSentry: {},


packages/core/src/js/NativeLogListener.ts
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Sets up the native log listener that forwards logs from the native SDK to JS.
+ * This only works when `debug: true` is set in Sentry options.
+ *
+ * @param callback - The callback to invoke when a native log is received.
+ * @returns A function to remove the listener, or undefined if setup failed.
+ */
+export function setupNativeLogListener(callback: (log: NativeLogEntry) => void): (() => void) | undefined {
+  if (Platform.OS !== 'ios' && Platform.OS !== 'android') {
+    debug.log('Native log listener is only supported on iOS and Android.');
```
lucas-zimerman, line 17: This will work on the future once we have support for Android and iOS. For the time being, this PR is only adding support for iOS
```suggestion
  if (Platform.OS !== 'ios') {
    debug.log('Native log listener is only supported on iOS.'');
```

packages/core/src/js/NativeLogListener.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/**
+ * Default handler for native logs that logs to the JS console.
+ */
+export function defaultNativeLogHandler(log: NativeLogEntry): void {
+  const prefix = `[Sentry] [${log.level.toUpperCase()}] [${log.component}]`;
+  const message = `${prefix} ${log.message}`;
+
+  switch (log.level.toLowerCase()) {
```
lucas-zimerman, line 69: It would be nice if `defaultNativeLogHandler`  doesn't create any events when `captureConsoleIntegration` is added as an integration.
Q: Shouldn't we use debug instead of console here?

packages/core/src/js/NativeLogListener.ts
alwx: Now it supports Android as well :)

alwx:
I've updated this PR to support Android as well (that was easy on Android because all the required methods are already there) 🎉

samples/react-native/ios/sentryreactnativesample.xcodeproj/project.pbxproj
antonis, line 1: nit: Let's skip the changes on this file since it's mainly formatting

alwx:
@antonis rebased this one to the current `main` 🎉

lucas-zimerman:
lets add the suggested changelog and LGTM! Thank you for the PR!

CHANGELOG.md
```diff
@@ -65,6 +65,23 @@
     ],
   });
   ```
+- Add `onNativeLog` callback to intercept and forward native SDK logs to JavaScript console
```
antonis: Let's move the changelog entry in the unreleased section. Other than that LGTM

alwx: sorry, a stupid mistake :(

CHANGELOG.md
```diff
@@ -8,6 +8,26 @@
 
 ## Unreleased
 
+### Features
+
+- Add `onNativeLog` callback to intercept and forward native SDK logs to JavaScript console
```
lucas-zimerman: Nit to make danger happy

```suggestion
- Add `onNativeLog` callback to intercept and forward native SDK logs to JavaScript console ([#5622](https://github.com/getsentry/sentry-react-native/pull/5622))
```

packages/core/ios/RNSentryExperimentalOptions.m
```diff
@@ -34,7 +34,7 @@ + (void)setEnableSessionReplayInUnreliableEnvironment:(BOOL)enabled
     if (sentryOptions == nil) {
         return;
     }
-    sentryOptions.experimental.enableSessionReplayInUnreliableEnvironment = enabled;
+    // sentryOptions.experimental.enableSessionReplayInUnreliableEnvironment = enabled;
```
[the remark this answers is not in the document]
antonis: Q: Why did we comment this?

packages/core/ios/RNSentryNativeLogsForwarder.m
```diff
@@ -0,0 +1,144 @@
...
+            @"[Sentry] [info] [0] [RNSentryNativeLogsForwarder] Native log forwarding "
+            @"configured successfully"];
+}
+
+- (void)stopForwarding
+{
+    self.eventEmitter = nil;
+
+    // Reset to default print behavior
+    [SentrySDKLog setOutput:^(NSString *_Nonnull message) { NSLog(@"%@", message); }];
```
antonis, line 53: Q: should we save the previous block before overwriting it in `configureWithEventEmitter`, and restore it here?

samples/react-native/src/App.tsx
```diff
@@ -45,6 +45,8 @@ const StackNavigator: TypedNavigator<any, any> = isMobileOs
 const BottomTabNavigator = createBottomTabNavigator();
 
 Sentry.init({
+  debug: true,
+  environment: 'dev',
```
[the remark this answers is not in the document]
antonis: We should probably revert the debug changes before merging.

lucas-zimerman:
it's weird the ios test is failing with any error

packages/core/ios/RNSentryNativeLogsForwarder.m
alwx: Yes but that would require making getOutput() public in sentry-cocoa which would take quite some time.
I think it's a correct and valid approach and I will do that but we can do it later — setOutput is an API only used by hybrid SDKs thus in practice it won't change much.

packages/core/ios/RNSentryNativeLogsForwarder.m
```diff
@@ -0,0 +1,146 @@
...
+        RNSentryNativeLogsForwarder *strongSelf = weakSelf;
+        if (strongSelf) {
+            [strongSelf forwardLogMessage:message];
+        }
+    }];
+
+    // Send a test log to verify the forwarding works
+    [self forwardLogMessage:
+            @"[Sentry] [info] [0] [RNSentryNativeLogsForwarder] Native log forwarding "
+            @"configured successfully"];
```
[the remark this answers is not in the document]
alwx: I don't think it's the case here — it's actually to notify a user that native log forwarding works. Will update the wording.

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -170,8 +170,17 @@ public void initNativeReactNavigationNewFrameTracking(Promise promise) {
...
+
     RNSentryStart.startWithOptions(
-        getApplicationContext(), rnOptions, getCurrentActivity(), logger);
+        getApplicationContext(),
+        rnOptions,
+        options -> {
+          // Use our custom logger that forwards to JS
+          options.setLogger(rnLogger);
+        },
+        logger);
```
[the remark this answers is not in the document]
antonis: I think this might still be valid. The `startWithOptions` called here [is also used in the RNSentrySDK](https://github.com/getsentry/sentry-react-native/blob/8c27a92935986915e01f7b6323c0b53e2831ddfe/packages/core/android/src/main/java/io/sentry/react/RNSentrySDK.java#L40) but `RNSentrySDK` is a standalone utility class for initializing the SDK outside of a React Native module context. There's no `getCurrentActivity()` available at that call site.                                                                                             
The `RNSentryModuleImpl` is different: it runs inside the RN bridge and does have access to `getCurrentActivity()`. The issue is specifically that the refactored call switched from the activity-aware overload to the no-activity overload, dropping the activity that was previously available and passed.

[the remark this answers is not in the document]
antonis: @alwx Maybe we can create a new overload [in RNSentryStart](https://github.com/getsentry/sentry-react-native/blob/8c27a92935986915e01f7b6323c0b53e2831ddfe/packages/core/android/src/main/java/io/sentry/react/RNSentryStart.java#L47) that accepts both the activity and the options and then use it here. Something like:
```
  static void startWithOptions(                                                                                                                            
      @NotNull final Context context,                                                                                                                      
      @NotNull final ReadableMap rnOptions,                                                                                                                
      @Nullable Activity currentActivity,
      @NotNull Sentry.OptionsConfiguration<SentryAndroidOptions> configuration,
      @NotNull ILogger logger) {
    Sentry.OptionsConfiguration<SentryAndroidOptions> defaults =
        options -> updateWithReactDefaults(options, currentActivity);
    Sentry.OptionsConfiguration<SentryAndroidOptions> rnConfigurationOptions =
        options -> getSentryAndroidOptions(options, rnOptions, logger);
    RNSentryCompositeOptionsConfiguration compositeConfiguration =
        new RNSentryCompositeOptionsConfiguration(
            rnConfigurationOptions, defaults, configuration, RNSentryStart::updateWithReactFinals);
    SentryAndroid.init(context, compositeConfiguration);
  }

```

CHANGELOG.md
```diff
@@ -10,6 +10,23 @@
...
+  - Only works when `debug: true` is enabled in `Sentry.init`
+  - Use `logWithoutTracing` inside the callback to prevent feedback loops with Sentry's console integration
+    ```js
+    import * as Sentry from '@sentry/react-native';
+
+    Sentry.init({
+      debug: true,
+      onNativeLog: ({ level, component, message }) => {
+        // Use logWithoutTracing to avoid feedback loops
+        Sentry.logWithoutTracing(
```
lucas-zimerman: Could you double-check this? I am quite sure only the sample app implements `logWithoutTracing` and not the SDK

packages/core/src/js/NativeLogListener.ts
```diff
@@ -0,0 +1,78 @@
...
+ * @param callback - The callback to invoke when a native log is received.
+ * @returns A function to remove the listener, or undefined if setup failed.
+ */
+export function setupNativeLogListener(callback: (log: NativeLogEntry) => void): (() => void) | undefined {
+  if (Platform.OS !== 'ios' && Platform.OS !== 'android') {
+    debug.log('Native log listener is only supported on iOS and Android.');
+    return undefined;
+  }
+
+  if (!NativeModules.RNSentry) {
```
lucas-zimerman, line 25: Let's quietly ignore expo go too.

```suggestion
  if (isExpoGo()) {
    return undefined;
  }
  
  if (!NativeModules.RNSentry) {
```

packages/core/android/src/main/java/io/sentry/react/RNSentryLogger.java
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Custom ILogger implementation that wraps AndroidLogger and forwards log messages to React Native.
+ * This allows native SDK logs to appear in the Metro console when debug mode is enabled.
+ */
+public class RNSentryLogger implements ILogger {
+  private static final String TAG = "RNSentry";
+  private static final String EVENT_NAME = "SentryNativeLog";
+
+  private final AndroidLogger androidLogger;
+  private WeakReference<ReactApplicationContext> reactContextRef;
```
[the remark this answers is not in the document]
antonis: I think we can ignore this

packages/core/src/js/NativeLogListener.ts
```diff
@@ -17,6 +18,10 @@ export function setupNativeLogListener(callback: (log: NativeLogEntry) => void):
     return undefined;
   }
 
+  if (isExpoGo()) {
+    return undefined;
```
antonis, line 22: We should update the tests after this change
```
       FAIL  test/NativeLogListener.test.ts
        ● NativeLogListener › setupNativeLogListener › returns a cleanup function on success
      
          TypeError: Cannot read properties of undefined (reading 'expo')
      
            27 |  */
            28 | export function getExpoGo(): ExpoGo | undefined {
          > 29 |   return RN_GLOBAL_OBJ.expo?.modules?.ExpoGo ?? undefined;
               |                        ^
            30 | }
            31 |
      
            at expo (src/js/utils/expomodules.ts:29:24)
            at isExpoGo (src/js/utils/environment.ts:38:27)
            at setupNativeLogListener (src/js/NativeLogListener.ts:21:15)
            at Object.<anonymous> (test/NativeLogListener.test.ts:35:45)
      
        ● NativeLogListener › setupNativeLogListener › cleanup removes only its own listener, not a later one
      
          TypeError: Cannot read properties of undefined (reading 'expo')
      
            27 |  */
            28 | export function getExpoGo(): ExpoGo | undefined {
          > 29 |   return RN_GLOBAL_OBJ.expo?.modules?.ExpoGo ?? undefined;
               |                        ^
            30 | }
            31 |
      
            at expo (src/js/utils/expomodules.ts:29:24)
            at isExpoGo (src/js/utils/environment.ts:38:27)
            at setupNativeLogListener (src/js/NativeLogListener.ts:21:15)
            at Object.<anonymous> (test/NativeLogListener.test.ts:50:46)
      
        ● NativeLogListener › setupNativeLogListener › cleanup is idempotent and does not remove listener twice
      
          TypeError: Cannot read properties of undefined (reading 'expo')
      
            27 |  */
            28 | export function getExpoGo(): ExpoGo | undefined {
          > 29 |   return RN_GLOBAL_OBJ.expo?.modules?.ExpoGo ?? undefined;
               |                        ^
            30 | }
            31 |
      
            at expo (src/js/utils/expomodules.ts:29:24)
            at isExpoGo (src/js/utils/environment.ts:38:27)
            at setupNativeLogListener (src/js/NativeLogListener.ts:21:15)
            at Object.<anonymous> (test/NativeLogListener.test.ts:75:45)
      
      
      Test Suites: 1 failed, 1 skipped, 96 passed, 97 of 98 total
      Tests:       3 failed, 1 skipped, 1062 passed, 1066 total
```

CHANGELOG.md
alwx: Good point. Exposed another similar method from the JS SDK.

antonis (approved):
LGTM! Thank you for your work on this Alex 🙇

PR #5627, opened 2026-02-06
feat(feedback): Add autoCorrect and spellCheck config to FeedbackWidget

### LINKED ISSUE #5626: FeedbackWidget: no way to control autoCorrect/spellCheck on text inputs
The `FeedbackWidget` TextInputs don't expose `autoCorrect` or `spellCheck` props. There's also no `autoCapitalize="none"` on the email input, so email addresses get auto-capitalized on iOS.

**Expected:** Consumers can opt out of auto-correct/spell-check (useful for bug report forms where technical terms trigger corrections), and the email field doesn't auto-capitalize.

SDK version: 7.12.0

sbs44:
- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Add optional `autoCorrect` and `spellCheck` props to `FeedbackGeneralConfiguration`, forwarded to all `TextInput` elements in the feedback form. Both default to `true` (matching React Native defaults) so this is non-breaking.

Also adds `autoCapitalize="none"` to the email input — email addresses shouldn't be auto-capitalized.

Fixes #5626 — feedback forms used for bug reports often contain technical terms, stack traces, and identifiers that trigger aggressive auto-correct behavior. Consumers should be able to opt out.

- Test: `autoCorrect={false}` and `spellCheck={false}` forwarded to all three TextInputs
- Test: defaults to `true` when not specified
- Test: email input has `autoCapitalize="none"`
- All existing tests pass, snapshots updated

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

None.

### DIFF
--- a/packages/core/src/js/feedback/FeedbackWidget.tsx
+++ b/packages/core/src/js/feedback/FeedbackWidget.tsx
@@ -260,7 +260,6 @@ export class FeedbackWidget extends React.Component<FeedbackWidgetProps, Feedbac
     const config: FeedbackGeneralConfiguration = this.props;
     const imagePickerConfiguration: ImagePickerConfiguration = this.props;
     const text: FeedbackTextConfiguration = this.props;
-    const styles: FeedbackWidgetStyles = { ...defaultStyles(theme), ...this.props.styles };
     const _defaultStyles = defaultStyles(theme);
     const _propStyles = this.props.styles || {};
     const autoCorrect = config.autoCorrect !== false;


packages/core/src/js/feedback/FeedbackWidget.types.ts
```diff
@@ -60,6 +60,20 @@ export interface FeedbackGeneralConfiguration {
...
+   * @default true
+   */
+  autoCorrect?: boolean;
+
+  /**
+   * Enable spell check for text inputs.
+   *
+   * @default true
+   */
+  spellCheck?: boolean;
```
lucas-zimerman, line 75: The default values matches the default values imposed by React Native.

packages/core/src/js/feedback/FeedbackWidget.tsx
```diff
@@ -325,8 +329,11 @@ export class FeedbackWidget extends React.Component<FeedbackWidgetProps, Feedbac
                 testID="sentry-feedback-email-input"
                 placeholder={text.emailPlaceholder}
                 keyboardType={'email-address' as KeyboardTypeOptions}
+                autoCapitalize="none"
                 value={email}
                 onChangeText={value => this.setState({ email: value })}
+                autoCorrect={autoCorrect}
+                spellCheck={spellCheck}
```
lucas-zimerman: I would suggest to not use spellchecks on email, since the email itself could be on foreign language or even contains a spell and be the users email.

```suggestion
```

packages/core/src/js/feedback/FeedbackWidget.tsx
```diff
@@ -325,8 +329,11 @@ export class FeedbackWidget extends React.Component<FeedbackWidgetProps, Feedbac
                 testID="sentry-feedback-email-input"
                 placeholder={text.emailPlaceholder}
                 keyboardType={'email-address' as KeyboardTypeOptions}
+                autoCapitalize="none"
                 value={email}
                 onChangeText={value => this.setState({ email: value })}
+                autoCorrect={autoCorrect}
```
lucas-zimerman: same as below
```suggestion
```

packages/core/src/js/feedback/FeedbackWidget.tsx
```diff
@@ -310,6 +312,8 @@ export class FeedbackWidget extends React.Component<FeedbackWidgetProps, Feedbac
                 placeholder={text.namePlaceholder}
                 value={name}
                 onChangeText={value => this.setState({ name: value })}
+                autoCorrect={autoCorrect}
+                spellCheck={spellCheck}
```
lucas-zimerman: I would suggest to only leave the autocorrect and spellcheck on the message only and not on the name/email.
```suggestion
```

lucas-zimerman:
Thank you for the contribution! I have  a few suggestions on the email/username input, other than that, it's looking good!

lucas-zimerman:
@sentry review

packages/core/src/js/feedback/FeedbackWidget.tsx
sbs44: Good point on scoping these to message only. Quick clarification — if we remove `autoCorrect`/`spellCheck` from name and email entirely, React Native defaults both to `true`, so those fields would still get spell-checked and auto-corrected.

Should we:
1. Hardcode `autoCorrect={false} spellCheck={false}` on name and email (always off), or
2. Leave them off and rely on platform behavior (iOS suppresses autocorrect for `keyboardType="email-address"`, but spellCheck still applies)

lucas-zimerman: That's also a good point, let's hardcode it to false on email/username and also use keyboardType as `email-address` for the email field.

sbs44: Done — hardcoded `autoCorrect={false} spellCheck={false}` on both name and email inputs. Message input keeps the configurable props. Email already had `keyboardType="email-address"`.

packages/core/src/js/feedback/FeedbackWidget.tsx
```diff
@@ -260,8 +260,11 @@ export class FeedbackWidget extends React.Component<FeedbackWidgetProps, Feedbac
     const config: FeedbackGeneralConfiguration = this.props;
     const imagePickerConfiguration: ImagePickerConfiguration = this.props;
     const text: FeedbackTextConfiguration = this.props;
+    const styles: FeedbackWidgetStyles = { ...defaultStyles(theme), ...this.props.styles };
```
antonis: I think this is more of a merge issue

```suggestion
```

packages/core/test/feedback/FeedbackWidget.test.tsx
```diff
@@ -149,6 +149,35 @@ describe('FeedbackWidget', () => {
...
+    expect(getByTestId('sentry-feedback-name-input').props.autoCorrect).toBe(false);
+    expect(getByTestId('sentry-feedback-name-input').props.spellCheck).toBe(false);
+    expect(getByTestId('sentry-feedback-email-input').props.autoCorrect).toBe(false);
+    expect(getByTestId('sentry-feedback-email-input').props.spellCheck).toBe(false);
+  });
+
+  it('sets autoCapitalize to none on email input', () => {
+    const { getByTestId } = render(<FeedbackWidget {...defaultProps} />);
+
+    expect(getByTestId('sentry-feedback-email-input').props.autoCapitalize).toBe('none');
```
antonis, line 180:
```suggestion
    expect(getByTestId('sentry-feedback-email-input').props.autoCapitalize).toBe('none');
   });

```

packages/core/src/js/feedback/FeedbackWidget.tsx
antonis: I'll proceed with this change since it's a merge conflict and not an implementation feedback 🙇

packages/core/test/feedback/FeedbackWidget.test.tsx
antonis: As above

antonis:
This is now released with [7.13.0](https://github.com/getsentry/sentry-react-native/releases/tag/7.13.0) 🎉 
Thank you for your contribution @sbs44 🙇

PR #5667, opened 2026-02-13
fix(spotlight): Add Sentry Spotlight dependency for debug builds

romtsn:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Since 8.32.0 spotlight was separated into a different module, we now have to explicitly include it into the consuming project (when enabling spotlight in the Android SDK). The recommended way is to only do that for debug builds and not ship the artifact to production. See Flutter: https://github.com/getsentry/sentry-dart/blob/main/packages/flutter/android/build.gradle#L66

I haven't tested it, but afaik RN works in similar ways as Flutter, so the android project (from sentry-react-native) will be added as includedBuild locally to our customers' projects so this should suffice. But would be great if you can verify it on the sample app

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/packages/core/android/build.gradle
+++ b/packages/core/android/build.gradle
@@ -56,4 +56,5 @@ dependencies {
     compileOnly files('libs/replay-stubs.jar')
     implementation 'com.facebook.react:react-native:+'
     api 'io.sentry:sentry-android:8.32.0'
+    debugImplementation 'io.sentry:sentry-spotlight:8.32.0'
 }


buenaflor:
I think the update dep release script needs to be updated as well 

otherwise it's gonna stay at 8.32.0

romtsn:
> I think the update dep release script needs to be updated as well
> 
> otherwise it's gonna stay at 8.32.0

good point, will do on Monday!

lucas-zimerman:
I tested it locally and spotlight works just fine on debug, but also release? I thought it wasn't going to run in release
`./gradlew assembleRelease`

buenaflor:
@lucas-zimerman is it possible that the JS spotlight config is enabled as well?

buenaflor:
afaict envelopes go through both spotlight paths: [js spotlight integration](https://github.com/getsentry/sentry-react-native/blob/main/packages/core/src/js/integrations/spotlight.ts#L48) and native - prolly that's why you still see events coming through spotlight even though android is disabled 

could be an oversight in the impl

antonis:
I've opened a PR to update this along with the Android version https://github.com/getsentry/sentry-react-native/pull/5672

antonis:
> I tested it locally and spotlight works just fine on debug, but also release? I thought it wasn't going to run in release
./gradlew assembleRelease

I've opened https://github.com/getsentry/sentry-react-native/pull/5673 to avoid accidentally enabling js spotlight in release

PR #5690, opened 2026-02-20
Move task operations out of onVariants to fix AGP Artifacts API confl…

RyanCommits:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Fixes issue where calling tasks.matching{}.each{} inside androidComponents.onVariants()
forces task realization during AGP's variant configuration phase, disrupting the
Artifacts API transform chain used by other plugins (e.g. Fullstory).

This caused APK transforms from other plugins to output to build/intermediates/
instead of build/outputs/, breaking downstream tooling that expects APKs in the
standard location.

The fix:
- Collect only variant metadata (name, outputs) in onVariants callback
- Move all task-level operations (tasks.matching, tasks.register, etc.) into
  project.afterEvaluate within the plugins.withId block
- Update extractCurrentVariants to accept pre-collected data instead of live variant

This ensures the AGP Artifacts API transform chain is fully established before
Sentry accesses any tasks, preventing interference with other plugins.

This is currently conflicting with the Fullstory plugin

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/packages/core/sentry.gradle
+++ b/packages/core/sentry.gradle
@@ -119,7 +119,7 @@ plugins.withId('com.android.application') {
                 //     .join('\n')
 
                 def currentVariants = extractCurrentVariants(bundleTask, v)
-                if (currentVariants == null) return
+                if (currentVariants == null || currentVariants.isEmpty()) return
 
                 def previousCliTask = null
                 def applicationVariant = null
@@ -328,9 +328,10 @@ plugins.withId('com.android.application') {
                 previousCliTask.configure { finalizedBy cliCleanUpTask }
 
                 def packageTasks = tasks.matching {
-                    task -> ("package${applicationVariant}".equalsIgnoreCase(task.name) || "package${applicationVariant}Bundle".equalsIgnoreCase(task.name)) && task.enabled
+                    task -> ("package${applicationVariant}".equalsIgnoreCase(task.name) || "package${applicationVariant}Bundle".equalsIgnoreCase(task.name))
                 }
                 packageTasks.configureEach { packageTask ->
+                    if (!packageTask.enabled) return
                     packageTask.dependsOn modulesTask
                     packageTask.finalizedBy modulesCleanUpTask
                 }


antonis:
@sentry review

packages/core/sentry.gradle
```diff
@@ -65,13 +65,43 @@ tasks.register("cleanupTemporarySentryJsonConfiguration") {
...
             // sourcemap generation parameters.  In case for whatever reason no release
             // was found for the asset folder we just bail.
-            def bundleTasks = tasks.findAll { task -> (task.name.startsWith("createBundle") || task.name.startsWith("bundle")) && task.name.endsWith("JsAndAssets") && !task.name.contains("Debug") && task.enabled }
-            bundleTasks.each { bundleTask ->
+            tasks.matching { task ->
+                (task.name.startsWith("createBundle") || task.name.startsWith("bundle")) &&
+                task.name.endsWith("JsAndAssets") &&
+                !task.name.contains("Debug") &&
+                task.enabled
+            }.each { bundleTask ->
```
antonis: The change from `tasks.findAll` to `tasks.matching` is an improvement, but the task.enabled check inside the matching predicate still forces task realization for every candidate. Wdyt of moving it into the `.each` block instead:
```suggestion
            }.each { bundleTask ->
                if (!bundleTask.enabled) return
```

packages/core/sentry.gradle
```diff
@@ -65,13 +65,43 @@ tasks.register("cleanupTemporarySentryJsonConfiguration") {
...
+            }
+            releaseVariants << [name: v.name, outputs: outputs]
+        }
+    }
+
+    // All task-level operations must happen in afterEvaluate, not inside onVariants.
+    // By the time afterEvaluate runs, all plugins have registered their onVariants
+    // callbacks and the AGP Artifacts API transform chain is fully established.
+    project.afterEvaluate {
+        releaseVariants.each { variantData ->
```
antonis: This assumes onVariants populates `releaseVariants` before `afterEvaluate` runs. Since this might not be [guaranteed](https://developer.android.com/build/extend-agp#build-flow-extension-points) wdyt of adding a warning for this?
```suggestion
    // By the time afterEvaluate runs, all plugins have registered their onVariants
    // callbacks and the AGP Artifacts API transform chain is fully established.
    project.afterEvaluate {
        if (releaseVariants.isEmpty()) {
            project.logger.warn("[sentry] No release variants collected, onVariants may have run after afterEvaluate. Sourcemap upload tasks will not be registered.")
        }
        releaseVariants.each { variantData ->
```

antonis (commented):
Thank you for your contribution @RyanCommits! We really appreciate this🙇 
I understand that the PR is still draft but since it came to our attention I've added some early feedback. We would also appreciate a small repro to help us test the fix.

RyanCommits:
Hi @antonis, thank you for the feedback. I've updated the branch with your feedback, as well as simplified what I was doing.

I have a repro of it issue here: 
https://github.com/RyanCommits/sentry-fullstory-issue-repro/tree/master
You'll need to configure your own sentry account and EAS build settings in your environment, but the issue should show with build command `eas build --local --platform android`

This other branch has the fix implemented:
https://github.com/RyanCommits/sentry-fullstory-issue-repro/tree/issue-fix

packages/core/sentry.gradle
```diff
@@ -65,13 +65,37 @@ tasks.register("cleanupTemporarySentryJsonConfiguration") {
...
+            releaseVariants << v
+        }
+    }
+
+    // All task-level operations must happen in afterEvaluate, not inside onVariants.
+    // By the time afterEvaluate runs, all plugins have registered their onVariants
+    // callbacks and the AGP Artifacts API transform chain is fully established.
+    project.afterEvaluate {
+        if (releaseVariants.isEmpty()) {
+            project.logger.warn("[sentry] No release variants collected, onVariants may have run after afterEvaluate. Sourcemap upload tasks will not be registered.")
```
ReeceLaF, line 87: This warning is inaccurate. It should not be possible for onVariants to have been invoked `afterEvaluate` as it is called when the plugin is applied, which should not be in or after `afterEvaluate` in this case. Though it could indicate no release variants configured, as the AI review stated.

lucas-zimerman:
The PR looks good! Would you be able to edit the changelog and add the following snippet?

```

## Unreleased

### Fixes

- Android: defer Sentry sourcemap upload task wiring to afterEvaluate to avoid configuring Gradle tasks inside onVariants and improve AGP compatibility ([#5690](https://github.com/getsentry/sentry-react-native/pull/5690))

packages/core/sentry.gradle
antonis: wdyt of simplifying the message to `[sentry] No release variants found. Source map upload tasks will not be registered.`?

ReeceLaF: Sounds great. The one you've got in #5714 looks good too!

antonis:
> Heads up that we've proceeded with the alternative implementation in https://github.com/getsentry/sentry-react-native/pull/5714 due to the regression mentioned https://github.com/getsentry/sentry-react-native/pull/5690#pullrequestreview-3848137749. The fix is now shipped in [8.2.0](https://github.com/getsentry/sentry-react-native/releases/tag/8.2.0) 🚢
>
> Thank you again for taking the effort to prepare this PR and pushing us forward to fix the issue 🙇

I'll proceed with closing this PR now to avoid any confusion.

PR #5696, opened 2026-02-23
Cirrus Labs runners for other important workflows (where it makes sense to do so) + Ubuntu update (22.04 -> 24.04)

alwx:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Hopefully that last bit of Cirrus Labs runners project — this time it's about 
1) moving to CL Ubuntu runners where it makes sense
2) moving sample-application-expo.yml and sample-application.yml back to GH Ubuntu runners (for the sake of consuming less resources)
2) replacing the old Sequoia to Tahoe where it's possible
3) moving more workflows to Cirrus Labs
4) sample-applicaty

Speed!

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/.github/workflows/buildandtest.yml
+++ b/.github/workflows/buildandtest.yml
@@ -123,7 +123,7 @@ jobs:
         run: yarn install
       - name: Check lock file integrity
         run: |
-          ["$(diff yarn.lock.initial yarn.lock)" -eq ""]
+          [ "$(diff yarn.lock.initial yarn.lock)" = "" ]
 
   job_build:
     name: Build
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -56,7 +56,7 @@ only_rules:
   - private_unit_test
   - prohibited_super_call
   - redundant_nil_coalescing
-  - implicit_optional_initialization
+  - redundant_optional_initialization
   - redundant_string_enum_value
   - redundant_void_return
   - return_arrow_whitespace


.github/workflows/testflight.yml
```diff
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - main
-      - alwx/ci/macos-tahoe-cl-runners
```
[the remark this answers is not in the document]
antonis: Thank you for removing this 🙇  I think this feedback is invalid.

.github/workflows/add-platform-label.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   add_label:
-    runs-on: ubuntu-latest
+    runs-on: ["ghcr.io/cirruslabs/ubuntu-runner-amd64:24.04", "runner_group_id:12"]
```
lucas-zimerman: q: why 12?

lucas-zimerman: Is it possible to centralize `runs-on: ["ghcr.io/cirruslabs/ubuntu-runner-amd64:24.04", "runner_group_id:12"]` on a single job so others can share it?
That would make it easier in case we need to update it, avoiding updating the same reference on all jobs. It might be possible with reusable workflows.

lucas-zimerman:
The PR is looking good! I left a few topics in regard to the number 12 and also the possibility on reusing.

.github/workflows/add-platform-label.yml
alwx: because it's the ID of the runner group assigned to React Native SDK :)

alwx: unfortunately, GH doesn't support using env context in runs-on directly, neither it supports using any sort of constants to specify something that's gonna be shared between all the workflows. so we have to stick to manual specification for now

.github/workflows/testflight.yml
```diff
@@ -3,7 +3,7 @@ on:
...
+      - alwx/experiment/cirrus-labs-for-everythin
```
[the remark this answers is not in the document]
alwx: gonna be removed before merge!

.github/workflows/e2e-v2.yml
```diff
@@ -205,7 +205,7 @@ jobs:
             rn-version: '0.84.0'
             runs-on: ["ghcr.io/cirruslabs/macos-tahoe-xcode:26.2.0", "runner_group_id:10"]
           - platform: android
-            runs-on: ["ghcr.io/cirruslabs/ubuntu-runner-amd64:22.04", "runner_group_id:10"]
+            runs-on: ["ghcr.io/cirruslabs/ubuntu-runner-amd64:24.04", "runner_group_id:10"]
```
antonis, line 208: Maybe we can also revert the Android e2e, sample and expo-sample runners back to GH Ubuntu like https://github.com/getsentry/sentry-react-native/pull/5696/changes/8ddab36efb7d81a46b73096013685caa73ed51a1 and see how this goes. The Cirrus lab runners are way faster but we may end up being queued if more thing run on them.

alwx: I would say we can keep them for e2e since the benifit is pretty significant there but maybe use GH Ubuntu for sample and expo sample. what do you think, @antonis ?

antonis: Sounds good @alwx 👍

.github/workflows/buildandtest.yml
```diff
@@ -66,28 +66,46 @@ jobs:
...
+          # Install Swift 5.10 (compatible with SwiftLint 0.61.0)
+          swiftly install 5.10
+          swift --version
+
+          # Export Swift toolchain path for SwiftLint's SourceKit
+          # Dynamically determine the path from the swift binary location
+          SWIFT_BIN=$(which swift)
+          SWIFT_TOOLCHAIN=$(dirname $(dirname "$SWIFT_BIN"))
+          # Temp solution (todo(alwx): remove)
+          find "${SWIFT_TOOLCHAIN}" -name "libsourcekitdInProc.so" 2>/dev/null
```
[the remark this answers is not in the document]
lucas-zimerman: q: what is the solution solving?
is it only for debugging?

lucas-zimerman:
There is some issues on swiftlint, maybe updating it could help.


```diff
- LINUX_BIN="https://github.com/realm/SwiftLint/releases/download/0.61.0/swiftlint_linux_amd64.zip"
+ LINUX_BIN="https://github.com/realm/SwiftLint/releases/download/0.63.2/swiftlint_linux_amd64.zip"
- LINUX_SHA="sha256:02f4f580bbb27fb618dbfa24ce2f14c926461c85c26941289f58340151b63ae4"
+ LINUX_SHA="sha256:dd1017cfd20a1457f264590bcb5875a6ee06cd75b9a9d4f77cd43a552499143b"
- DARWIN_BIN="https://github.com/realm/SwiftLint/releases/download/0.61.0/portable_swiftlint.zip"
+ DARWIN_BIN="https://github.com/realm/SwiftLint/releases/download/0.63.2/portable_swiftlint.zip"
- DARWIN_SHA="sha256:2342f3784307a02117e18f745fcd350c6acc6cab0e521c0c0e01c32a53a3b274"
+ DARWIN_SHA="sha256:c59a405c85f95b92ced677a500804e081596a4cae4a6a485af76065557d6ed29"
```

.github/workflows/add-platform-label.yml
lucas-zimerman: Thank you for checking!

.github/workflows/release.yml
```diff
@@ -18,7 +18,7 @@ permissions:
 
 jobs:
   release:
-    runs-on: ubuntu-latest
+    runs-on: ["ghcr.io/cirruslabs/ubuntu-runner-amd64:24.04", "runner_group_id:10"]
```
antonis, line 21: Wdyt of leaving the release worflow on the GH runners? The releases are once a week and the gain from Cirruslabs might be small.
If we decide to include them let's run a test release and retract it (similar to what we do with Craft bumps https://github.com/getsentry/sentry-react-native/pull/5694#pullrequestreview-3846223409) to make sure that the flow won't break.

antonis:
> I've left a suggestion [related to the release workflow](https://github.com/getsentry/sentry-react-native/pull/5696/changes#r2863170059).
Let's [remove the test branch](https://github.com/getsentry/sentry-react-native/pull/5696/changes#r2847004964) and merge.

I realised that automerge was enabled after approving. I've opened a follow up PR with my suggestions https://github.com/getsentry/sentry-react-native/pull/5732

PR #5740, opened 2026-03-02
Fixes the issue with unit mismatch in `adjustTransactionDuration`

alwx:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
I've suddenly found out that `spanToJSON` from @sentry/core returns a structure that contains `start_timestamp` `timestamp` — but the "funny" thing is that both of this timestamps are in seconds, not in milliseconds (https://github.com/getsentry/sentry-javascript/blob/develop/packages/core/src/utils/spanUtils.ts#L170).
That means the way our `adjustTransactionDuration` works is wrong —  the `diff` is in seconds but it's compared against `maxDurationMs` which is in milliseconds. This means the safety net that should mark long-running        
 transactions as `deadline_exceeded` never triggers.
 
This PR fixes that.
 
In addition to that, there is a small formatting issue in `span.ts` — the default `finalTimeout` is specified as 60_0000 which should be milliseconds => needs to be 600_000 instead.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/src/js/tracing/onSpanEndUtils.ts
+++ b/packages/core/src/js/tracing/onSpanEndUtils.ts
@@ -1,8 +1,14 @@
 import type { Client, Span } from '@sentry/core';
 import { debug, getSpanDescendants, SPAN_STATUS_ERROR, spanToJSON } from '@sentry/core';
 import type { AppStateStatus } from 'react-native';
-import { AppState } from 'react-native';
+import { AppState, Platform } from 'react-native';
 import { isRootSpan, isSentrySpan } from '../utils/span';
+
+/**
+ * The time to wait after the app enters the `inactive` state on iOS before
+ * cancelling the span.
+ */
+const IOS_INACTIVE_CANCEL_DELAY_MS = 5_000;
 
 /**
  * Hooks on span end event to execute a callback when the span ends.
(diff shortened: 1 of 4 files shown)


packages/core/src/js/tracing/span.ts
```diff
@@ -36,12 +36,12 @@ export const defaultIdleOptions: {
    * The max. time an idle span may run.
    * If this time is exceeded, the idle span will finish no matter what.
    *
-   * @default 60_0000 (ms)
+   * @default 60_000 (ms)
```
antonis: nit: I think this should be 1000 like in the default value below
```suggestion
   * @default 1_000 (ms)
```

packages/core/src/js/tracing/onSpanEndUtils.ts
```diff
@@ -33,8 +33,8 @@ export const adjustTransactionDuration = (client: Client, span: Span, maxDuratio
       return;
     }
 
-    const diff = endTimestamp - startTimestamp;
-    const isOutdatedTransaction = endTimestamp && (diff > maxDurationMs || diff < 0);
+    const diff = endTimestamp - startTimestamp; // a diff in *seconds*
+    const isOutdatedTransaction = endTimestamp && (diff > maxDurationMs / 1000 || diff < 0);
```
antonis: nit: This is preexisting but we may simplify since `endTimestamp` is checked in the early return above
```suggestion
    const isOutdatedTransaction = diff > maxDurationMs / 1000 || diff < 0;
```

packages/core/src/js/tracing/span.ts
```diff
@@ -36,12 +36,12 @@ export const defaultIdleOptions: {
...
    */
   idleTimeout: number;
 } = {
   idleTimeout: 1_000,
-  finalTimeout: 60_0000,
+  finalTimeout: 60_000,
```
antonis: h: I think this would introduce a behavioral change that may not be desired. Changing `finalTimeout` from 600,000 ms → 60,000 ms reduces the hard cap on navigation span lifetime from 10 minutes to 1 minute. Any app with slow screens, heavy loading, or long user sessions that previously could create valid navigation spans up to 10 minutes would now have them force-terminated at 1 minute.
Looking at the history of this I think the problem is the `_` in the wrong position that was introduced [with this refactoring](https://github.com/getsentry/sentry-react-native/commit/63860627). We should probably keep this to 10m and just fix the formatting like:
```suggestion
  finalTimeout: 600_000,
```
Also looping in @lucas-zimerman who may have a better understanding of the history of this 🙏

packages/core/test/tracing/adjustTransactionDuration.test.ts
antonis, line 1: There is [a lint issue here](https://github.com/getsentry/sentry-react-native/actions/runs/22571626662/job/65380671331?pr=5740). A `yarn fix` should correct that.
```
  /home/admin/actions-runner/_work/sentry-react-native/sentry-react-native/packages/core/test/tracing/adjustTransactionDuration.test.ts
     39:28  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
     57:28  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
     75:28  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
     93:28  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
    112:28  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
    131:28  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
    153:33  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
    160:35  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
    178:28  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
```

antonis (approved):
LGTM once the CI is 🟢
Thank you for this fix @alwx 🙇

PR #5742, opened 2026-03-02
Handle `inactive` state for spans

alwx:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
On iOS, the app first goes to `inactive` state and only then to the `background` state. And during that `active` → `inactive` → `background` transition, there's a window where `AppState.currentState` is `inactive` and a navigation dispatch could create a new idle span.
That could explain the behavior when a navigation event starts when the app is in the background — it's likely starting during the `inactive` phase before full background.
This fix changes how we handle that — first of all, it introduces changes to `cancelInBackground` and makes sure we schedule the cancellation of a span when the app enters the `inactive` state. 
Secondly, `startIdleSpan` has changed to not start the span when the app is transitioning from foreground to background.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/src/js/tracing/span.ts
+++ b/packages/core/src/js/tracing/span.ts
@@ -12,7 +12,7 @@ import {
   spanToJSON,
   startIdleSpan as coreStartIdleSpan,
 } from '@sentry/core';
-import { AppState } from 'react-native';
+import { AppState, Platform } from 'react-native';
 import { isRootSpan } from '../utils/span';
 import { adjustTransactionDuration, cancelInBackground } from './onSpanEndUtils';
 import {
@@ -118,8 +118,10 @@ export const startIdleSpan = (
   }
 
   const currentAppState = AppState.currentState;
-  if (currentAppState === 'background' || currentAppState === 'inactive') {
-    debug.log(`[startIdleSpan] App is already in '${currentAppState}' state, not starting span for ${startSpanOption.name}`);
+  if (currentAppState === 'background' || (Platform.OS === 'ios' && currentAppState === 'inactive')) {
+    debug.log(
+      `[startIdleSpan] App is already in '${currentAppState}' state, not starting span for ${startSpanOption.name}`,
+    );
     return new SentryNonRecordingSpan();
   }
 


packages/core/src/js/tracing/span.ts
```diff
@@ -118,7 +118,7 @@ export const startIdleSpan = (
   }
 
   const currentAppState = AppState.currentState;
-  if (currentAppState === 'background') {
+  if (currentAppState === 'background' || currentAppState === 'inactive') {
```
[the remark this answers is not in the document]
antonis: > The guard here likely needs Platform.OS === 'ios' to match the approach in cancelInBackground.

I think this is correct

packages/core/src/js/tracing/span.ts
```diff
@@ -118,8 +118,8 @@ export const startIdleSpan = (
   }
 
   const currentAppState = AppState.currentState;
-  if (currentAppState === 'background') {
-    debug.log(`[startIdleSpan] App is already in background, not starting span for ${startSpanOption.name}`);
+  if (currentAppState === 'background' || currentAppState === 'inactive') {
+    debug.log(`[startIdleSpan] App is already either inactive or in background, not starting span for ${startSpanOption.name}`);
```
antonis: nit: Wdyt of adding the state here
```suggestion
    debug.log(`[startIdleSpan] App is already in '${currentAppState}' state, not starting span for ${startSpanOption.name}`);
```

packages/core/src/js/tracing/onSpanEndUtils.ts
```diff
@@ -1,9 +1,15 @@
...
 import type { AppStateStatus } from 'react-native';
-import { AppState } from 'react-native';
+import { AppState, Platform } from 'react-native';
 import { isRootSpan, isSentrySpan } from '../utils/span';
 
+/**
+ * The time to wait after the app enters the `inactive` state on iOS before
+ * cancelling the span.
+ */
+const IOS_INACTIVE_CANCEL_DELAY_MS = 5_000;
```
antonis, line 11: Q: Why did we choose 5s here?

alwx: It's somewhat arbitrary but the reasoning was the following:
- long enough for transient `inactive` states (like notification center or system dialogs), those should resolve back to `active` within a few seconds
- short enough that if the JS thread was suspended and then resumed, the span doesn't exist much longer than necessary

The transition from active to inactive to background should resolve in less than a second though, so the timeout mostly exists as a safety net for the case where JS resumes after being suspended and the 'background' event was lost/never delivered. A value like 2 or 3 sec would probably work just as well but I would suggest to reiterate in the future and see if situations described previously with events being sent after the app went in background are still the case.

PR #5752, opened 2026-03-03
fix(ci): Fix consistent iOS E2E flakiness on Cirrus Labs runners

antonis:
## Summary

Fixes consistent iOS E2E test failures that started after the Cirrus Labs Tart VM migration (`c1cade43`, Feb 27). Every single iOS E2E run on `main` has failed since that commit.

### Root Cause

The iOS E2E `react-native-test` job was moved from GitHub-hosted `macos-26` runners to Cirrus Labs Tart VMs (`ghcr.io/cirruslabs/macos-tahoe-xcode:26.2.0`). Tart VMs use nested virtualisation, which makes iOS simulator operations significantly slower and less stable:

- `xcrun simctl launch` takes ~23 seconds (vs <1s on bare metal)
- Maestro's XCTest driver takes ~10-20 seconds to connect to port 7001
- App launches intermittently fail — the app process exits before the JS bundle loads, causing Maestro to report "App crashed or stopped" despite the app briefly appearing on screen

This manifests as two distinct failure modes:

1. **XCTest driver instability** — After the driver connects, the first few app launches often fail. Screenshots show the app stuck on the default React Native welcome screen before crashing.

2. **crash.yml cascade** — After `nativeCrash()`, the bare `launchApp` caused Sentry to read the pending crash report and crash again within ~82ms, triggering iOS crash-loop protection that broke subsequent tests.

Note: PR #5735 (simulator-action v4→v5) is **not** the root cause — `erase_before_boot: true` was already the default in v4, and the failures predate that PR by 5 days.

### Fixes

1. **`wait_for_boot: true`** on simulator-action so the simulator is fully booted before proceeding
2. **`erase_before_boot: false`** — each Maestro flow already reinstalls the app via `clearState`, so erasing the entire simulator is redundant overhead
3. **Simulator warm-up step** — launches Settings.app so system services finish post-boot initialisation
4. **Retry the full Maestro suite up to 3 times** — the most reliable mitigation for intermittent app-launch failures on nested-virtualisation VMs
5. **`MAESTRO_DRIVER_STARTUP_TIMEOUT: 180000`** (3 min) to accommodate slower XCTest driver startup
6. **`clearState: true` in crash.yml** after the intentional `nativeCrash()` to prevent crash-loop cascade

## Test plan

- [x] Verify iOS E2E tests pass on this PR (`Test RN 0.84.0 legacy hermes ios production no` and `Test RN 0.84.0 new hermes ios production no`)
- [ ] If tests pass, merge to `main` and verify subsequent runs are stable

#skip-changelog


dev-packages/e2e-tests/cli.mjs
```diff
@@ -290,20 +290,51 @@
...
+        for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+          const label = `[${flowName}] Attempt ${attempt}/${maxAttempts}`;
+          console.log(`\n${'='.repeat(60)}\n${label}\n${'='.repeat(60)}`);
+          try {
+            execSync(
+              `maestro test "maestro/${flowFile}" \
+                --env=APP_ID="${appId}" \
+                --env=SENTRY_AUTH_TOKEN="${sentryAuthToken}" \
+                --debug-output maestro-logs \
+                --flatten-debug-output`,
```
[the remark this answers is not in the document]
antonis: Fixed with https://github.com/getsentry/sentry-react-native/pull/5752/commits/f7cb8905794818ce668a8f0d4cfa44eaa1523ea7

antonis (commented):
Note that I've opened a separate PR for the Sample e2e flakiness https://github.com/getsentry/sentry-react-native/pull/5755

.github/workflows/e2e-v2.yml
```diff
@@ -428,12 +428,28 @@ jobs:
...
+
+      - name: Warm up iOS simulator
+        if: ${{ matrix.platform == 'ios' }}
+        run: |
+          # Tart VMs are very slow right after boot.  Launch a stock app so
+          # that SpringBoard, backboardd, and other system services finish
+          # their post-boot initialisation before Maestro tries to connect.
+          xcrun simctl launch booted com.apple.Preferences
+          sleep 5
+          xcrun simctl terminate booted com.apple.Preferences
```
lucas-zimerman: The implementation on https://github.com/getsentry/sentry-react-native/pull/5755/changes#diff-c5263171d4b9aca7af53d70a3ee9f8423c7f4bebd6941e590c5610208f95c8bcR310 is different.
Sleep time is diffferent and we ignore when it fails.
Which pattern should we follow?

lucas-zimerman:
No issues so far, LGTM! once CI pass.

dev-packages/e2e-tests/cli.mjs
```diff
@@ -290,20 +290,50 @@ if (actions.includes('test')) {
   if (!sentryAuthToken) {
     console.log('Skipping maestro test due to unavailable or empty SENTRY_AUTH_TOKEN');
   } else {
+    const maxAttempts = 3;
```
alwx, line 293: I have mixed feelings about it because it basically is an attempt to fix the issue with flakiness by trying it three times to run the same test. I also wonder if it works considering that the flakiness is consistent + it doesn't seem like it guarantees to work.

.github/workflows/e2e-v2.yml
antonis: - PR #5752 (e2e-v2.yml): Added || true to both xcrun simctl commands so the warm-up is best-effort                                                                                
  - PR #5755 (sample-application.yml): Aligned comment and increased sleep 3 to sleep 5 to match e2e-v2.yml

dev-packages/e2e-tests/cli.mjs
antonis: That's a good point @alwx 👍 
I don't like this much either because it's a workaround. I don't have other ideas on this at this point though. It seems that the tests have started failing randomly (not always the same test fails) thus the retries attempt to overcome this by making individual automatic test retries rather than manually rerunning the whole test.

PR #5755, opened 2026-03-04
fix(ci): Fix Sample Application E2E test flakiness

antonis:
## Summary
- Increases `MAESTRO_DRIVER_STARTUP_TIMEOUT` from 90s to 180s for slow Cirrus Labs Tart VMs
- Adds `wait_for_boot: true` and `erase_before_boot: false` to `simulator-action` to ensure the simulator is fully ready before tests
- Adds a simulator warm-up step (launch/terminate Settings app) before running iOS tests
- Fixes `captureSpaceflightNewsScreenTransaction` test by sorting envelopes by timestamp instead of relying on arrival order (which is non-deterministic on slow VMs)
- Relaxes HTTP spans assertion from exactly 2 to at least 1, since not all HTTP tracing layers may complete within the transaction on slow VMs
- Fixes `captureErrorsScreenTransaction` Android test by searching all envelopes for the app start transaction instead of only the first matching envelope (on slow emulators, it may arrive separately)

#skip-changelog

## Test plan
- [x] Verify `Sample Application / Test ios production REV2` passes on CI
- [x] Verify `Sample Application / Test android production REV2` passes on CI

### DIFF
--- a/.github/workflows/sample-application.yml
+++ b/.github/workflows/sample-application.yml
@@ -305,9 +305,11 @@ jobs:
       - name: Warm up iOS Simulator
         if: ${{ matrix.platform == 'ios' }}
         run: |
-          # Launch and kill a dummy app to ensure the simulator is fully ready
+          # Tart VMs are very slow right after boot.  Launch a stock app so
+          # that SpringBoard, backboardd, and other system services finish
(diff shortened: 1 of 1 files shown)


lucas-zimerman:
I see no issues with the PR, LGTM! once CI pass.

PR #5756, opened 2026-03-04
fix(core): App start data lost when first transaction is unsampled

### LINKED ISSUE #5757: App Start data lost when first transaction is unsampled
### Environment

- `@sentry/react-native` (all versions with `appStartIntegration`)
- Any `tracesSampleRate` less than `1.0`

### Steps to Reproduce

1. Configure Sentry with `tracesSampleRate` less than `1.0` (e.g., `0.5`)
2. Launch the app so that a Cold/Warm Start is measured by the native layer
3. The first root span (navigation transaction) happens to be unsampled

### Expected Result

App start data (Cold/Warm Start spans and measurements) is attached to the next sampled transaction.

### Actual Result

App start data is permanently lost for that session. No sampled transaction ever receives it.

### Root Cause

In `appStartIntegration`, `recordFirstStartedActiveRootSpanId` locks `firstStartedActiveRootSpanId` to the first root span that starts, regardless of whether it is sampled. Since `processEvent` only runs for sampled events, the span ID match check at:

```typescript
if (firstStartedActiveRootSpanId !== event.contexts.trace.span_id) {
  return;
}
```

will always fail for every subsequent sampled transaction, because the stored ID belongs to the unsampled span that was never sent.

At lower sample rates this is statistically very likely. For example at `tracesSampleRate: 0.1`, there's a 90% chance the first transaction is unsampled, meaning 90% of sessions lose app start data entirely.

antonis:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

When `tracesSampleRate < 1.0`, the `appStartIntegration` locks `firstStartedActiveRootSpanId` to the first root span regardless of whether it is sampled. Since `processEvent` only runs for sampled events, if the first root span is unsampled, app start data is permanently lost for that session — subsequent sampled transactions fail the span ID match check.

This PR adds a `spanIsSampled()` guard in `recordFirstStartedActiveRootSpanId` so the ID only locks to the first **sampled** root span. The existing age/duration guards still protect against stale app starts.

Users with `tracesSampleRate < 1.0` were losing Cold/Warm Start data whenever the first transaction happened to be unsampled (which is statistically likely at low sample rates). This is a data loss bug — the app start is measured by the native layer but never attached to any event.

Fixes #5757

- Added unit test: "Does not lock firstStartedActiveRootSpanId to unsampled root span" — emits a `SentryNonRecordingSpan` (unsampled) followed by a real sampled span, verifies app start attaches to the sampled transaction
- All existing tests pass (`yarn test`, `yarn build`, `yarn lint`)

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

### DIFF
--- a/.github/workflows/skip-ci.yml
+++ b/.github/workflows/skip-ci.yml
@@ -45,7 +45,7 @@ jobs:
         if: steps.check-pr.outputs.is-pr == 'true'
         id: check_diff
         run: |
-          skipList=(".github/CODEOWNERS" ".prettierignore" ".github/workflows/update-deps.yml" ".github/workflows/skip-ci.yml" ".claude/settings.json")
+          skipList=(".github/CODEOWNERS" ".prettierignore" ".github/workflows/update-deps.yml" ".github/workflows/skip-ci.yml" ".claude/settings.json" "agents.toml")
           skipRegList=("(.*/)?\.gitignore" "\.github/ISSUE_TEMPLATE.*")
 
           for s in "${skipRegList[@]}"; do
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,322 +1,50 @@
 # AGENTS.md
 
-This file provides comprehensive guidance for AI coding agents working with the Sentry React Native SDK repository.
+Sentry React Native SDK — monorepo using yarn workspaces with a single package at `packages/core`.
 
 ## Agent Responsibilities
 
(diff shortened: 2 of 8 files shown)


lucas-zimerman:
q: what about users with sample rate near 0, aren't they going to lose the app start data anyway?

antonis:
> q: what about users with sample rate near 0, aren't they going to lose the app start data anyway?

Good point @lucas-zimerman 👍 
Users with with sample rate near 0 would lose the app data. This fix just ensures app start is included when they do get a sampled transaction improving the current functionality.
I'll revisit this and open a separate PR.

Issue #4493, opened 2025-01-30
[Expo Updates] Report errors causing emergency launches


krystofwoldrich:
### Description

If isEmergencyLaunch is set to true, this will contain a string error message describing what failed during initialization.

- https://docs.expo.dev/versions/latest/sdk/updates/#updatesemergencylaunchreason
- Human descriptions of the error codes https://docs.expo.dev/versions/latest/sdk/updates/#error-codes

PR #5794, opened 2026-03-10
Automatically capture a warning event when Expo Updates performs an emergency launch

alwx:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Another relatively low effort improvement to Expo Updates support — when `isEmergencyLaunch` is set to true, automatically capture a Sentry event (not just context). The current approach only adds `is_emergency_launch: true` as context on future errors. But if the app recovers gracefully after an emergency launch, the team never knows it happened. An emergency launch is significant enough to spawn its own event.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,13 @@
   ```
 - Adds tags with Expo Updates context variables to make them searchable and filterable ([#5788](https://github.com/getsentry/sentry-react-native/pull/5788))
 - Automatically capture a warning event when Expo Updates performs an emergency launch ([#5794](https://github.com/getsentry/sentry-react-native/pull/5794))
+- Adds environment configuration in the Expo config plugin. This can be set with the `SENTRY_ENVIRONMENT` env variable or in `sentry.options.json` ([#5796](https://github.com/getsentry/sentry-react-native/pull/5796))
+  ```json
+  ["@sentry/react-native/expo", {
+    "useNativeInit": true,
+    "environment": "staging"
+  }]
+  ```
 
 ### Fixes
 
--- a/packages/core/plugin/src/utils.ts
+++ b/packages/core/plugin/src/utils.ts
@@ -1,5 +1,6 @@
 import * as fs from 'fs';
 import * as path from 'path';
+import { warnOnce } from './logger';
 
 export function writeSentryPropertiesTo(filepath: string, sentryProperties: string): void {
   if (!fs.existsSync(filepath)) {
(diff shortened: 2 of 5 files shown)


lucas-zimerman:
Q: Do you have any events that were captured on an emergency update?
I am asking because When I was testing this feature I was unable to trigger it/ mark it as true, so I didn't open a PR for it.

alwx:
@lucas-zimerman technically that's considered as an "emergency" mechanism and shouldn't happen often — that's why I'm not surprised you didn't see it being marked as true.
I was able to reproduce it by simulating poor network conditions since `isEmergencyLaunch` along with `emergencyLaunchReason` gets populated when there is any exception being thrown during the update: https://github.com/expo/expo/blob/6878a766887031d762bd0220e5bbdd3bdce86d61/packages/expo-updates/android/src/main/java/expo/modules/updates/procedures/StartupProcedure.kt#L270C8-L270C35

alwx:
@lucas-zimerman https://github.com/getsentry/sentry-react-native/issues/4493 this is related, right?

lucas-zimerman:
> @lucas-zimerman technically that's considered as an "emergency" mechanism and shouldn't happen often — that's why I'm not surprised you didn't see it being marked as true. I was able to reproduce it by simulating poor network conditions since `isEmergencyLaunch` along with `emergencyLaunchReason` gets populated when there is any exception being thrown during the update: https://github.com/expo/expo/blob/6878a766887031d762bd0220e5bbdd3bdce86d61/packages/expo-updates/android/src/main/java/expo/modules/updates/procedures/StartupProcedure.kt#L270C8-L270C35

The way I was trying to simulate was by making an app update that throws an error when the app started. The expo app would download the update and once the app was restarted it would rollback to the previous version due to the error.

lucas-zimerman:
> @lucas-zimerman #4493 this is related, right?

Correct

lucas-zimerman:
After cloning expo code locally I got this output from claude 
```
  ┌────────────────────────────────────┬─────────────────────────────┐                                                                  
  │              Scenario              │ Triggers isEmergencyLaunch? │                                                                  
  ├────────────────────────────────────┼─────────────────────────────┤                                                                  
  │ Broken JavaScript code in update   │ ❌ No                       │                                                                  
  ├────────────────────────────────────┼─────────────────────────────┤                                                                  
  │ Failed to download update from EAS │ ❌ No                       │                                                                  
  ├────────────────────────────────────┼─────────────────────────────┤                                                                  
  │ Network timeout during download    │ ❌ No                       │                                                                  
  ├────────────────────────────────────┼─────────────────────────────┤                                                                  
  │ Update download corrupted          │ ❌ No                       │                                                                  
  ├────────────────────────────────────┼─────────────────────────────┤                                                                  
  │ SQLite database corrupt            │ ✅ Yes                      │                                                                  
  ├────────────────────────────────────┼─────────────────────────────┤                                                                  
  │ Updates directory inaccessible     │ ✅ Yes                      │                                                                  
  ├────────────────────────────────────┼─────────────────────────────┤                                                                  
  │ File system read failure           │ ✅ Yes                      │                                                                  
  └────────────────────────────────────┴─────────────────────────────┘                                                                  
```
Indeed the way I was testing would never trigger it (inserting errors on the newer build)

lucas-zimerman (approved):
Looking good! thank you for the PR!

PR #5795, opened 2026-03-10
Add `expoUpdatesListenerIntegration` that records breadcrumbs for Expo Updates lifecycle events

alwx:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Add `expoUpdatesListenerIntegration` that records breadcrumbs for Expo Updates lifecycle events
  - Tracks update checks, downloads, errors, rollbacks, and restarts as `expo.updates` breadcrumbs
  - Enabled by default in Expo apps (requires `expo-updates` to be installed)
  
Fixes https://github.com/getsentry/sentry-react-native/issues/5425

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


packages/core/src/js/integrations/expoupdateslistener.ts
```diff
@@ -0,0 +1,158 @@
...
+        return;
+      }
+
+      let previousContext: Partial<UpdatesNativeStateMachineContext> = {};
+
+      addListener((event: UpdatesNativeStateChangeEvent) => {
+        const ctx = event.context;
+        handleStateChange(previousContext, ctx);
+        previousContext = ctx;
+      });
```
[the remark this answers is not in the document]
antonis: This should be valid 👍

packages/core/src/js/integrations/default.ts
```diff
@@ -133,6 +134,7 @@ export function getDefaultIntegrations(options: ReactNativeClientOptions): Integ
 
   integrations.push(expoContextIntegration());
   integrations.push(expoConstantsIntegration());
+  integrations.push(expoUpdatesListenerIntegration());
```
antonis, line 137: m: I was wondering if we should add an option for this and not always add the integration unconditionally. Since this adds a listener it might add overhead and users might not always want to enable.
Wdyt of making it opt-in?

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@
   Sentry.wrapExpoAsset(Asset);
   ```
 - Adds tags with Expo Updates context variables to make them searchable and filterable ([#5788](https://github.com/getsentry/sentry-react-native/pull/5788))
+- Add `expoUpdatesListenerIntegration` that records breadcrumbs for Expo Updates lifecycle events ([#5795](https://github.com/getsentry/sentry-react-native/pull/5795))
+  - Tracks update checks, downloads, errors, rollbacks, and restarts as `expo.updates` breadcrumbs
+  - Enabled by default in Expo apps (requires `expo-updates` to be installed)
+
```
antonis: We should omit the blank line for consistency
```suggestion
```

packages/core/src/js/integrations/default.ts
lucas-zimerman: We could add it by default on the next major, what do you think?

alwx: the listener is not gonna be added if `expo-updates` is not installed so it comes with no additional overhead for anyone else — I guess it kinda makes sense, or do you still think it should be opt-in?

antonis: Thank you for the explanation @alwx 🙇 
Since it is guarded by the `expo-updates` let's not add extra config. We can still revisit the decision later if we have user feedback.

packages/core/src/js/integrations/expoupdateslistener.ts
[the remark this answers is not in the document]
antonis: This looks valid 👍

alwx:
@antonis both are fixed now!

PR #5301, opened 2025-10-23
feat(replay): Add screenshotStrategy option for Android

romtsn:
## :loudspeaker: Type of change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Adds new `screenshotStrategy` option which only applies to Android targets. Here's a docs [PR](https://github.com/getsentry/sentry-docs/pull/15270) for Android which explains both strategies better.

An example replay from the sample app: https://sentry-sdks.sentry.io/explore/replays/33ea588853ad4d0fa070424728b3f701/

## :bulb: Motivation and Context

Better masking

Closes #5300 

## :green_heart: How did you test it?
Sample app + we extensively tested it with BlueSky

## :pencil: Checklist

- [ ] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Add docs

PR #5808, opened 2026-03-12
Pass `screenshotStrategy` option from `mobileReplayIntegration` to the Android SDK

alwx:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Fixes the issue with `screenshotStrategy` not being passed correctly.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Fixes
 
-- Pass `screenshotStrategy` option from `mobileReplayIntegration` to the Android native SDK ([#XXXX](https://github.com/getsentry/sentry-react-native/pull/XXXX))
+- Pass `screenshotStrategy` option from `mobileReplayIntegration` to the Android native SDK ([#5808](https://github.com/getsentry/sentry-react-native/pull/5808))
 
 ### Features
 


packages/core/android/src/main/java/io/sentry/react/RNSentryStart.java
```diff
@@ -375,6 +376,15 @@ private static SentryReplayOptions getReplayOptions(@NotNull ReadableMap rnOptio
       androidReplayOptions.addMaskViewClass("com.horcrux.svg.SvgView"); // react-native-svg
     }
 
+    if (rnMobileReplayOptions.hasKey("screenshotStrategy")) {
+      final String strategy = rnMobileReplayOptions.getString("screenshotStrategy");
+      if ("canvas".equals(strategy)) {
```
antonis, line 381: We can probably do [a lowercase conversion similar to the original implementation](https://github.com/getsentry/sentry-react-native/pull/5301/changes#diff-15f06061acd36165e8cbd8ed2c25d6f4c3b2880a9d55a5f097ca1d9d2af0b219) but I think this is not needed since the passed values from JS are distinct

PR #5811, opened 2026-03-13
Ref: remove yarn from stub update

lucas-zimerman:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

There are issues with calling yarn so since it only calls a one line script to build the stubs I replaced it with that.

Fix the stub build command

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/scripts/update-android-stubs.sh
+++ b/scripts/update-android-stubs.sh
@@ -21,7 +21,6 @@ set-version)
     newValue="${BASH_REMATCH[1]}$2"
     echo "${content/${BASH_REMATCH[0]}/$newValue}" >$file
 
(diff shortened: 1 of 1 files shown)


scripts/update-android-stubs.sh
```diff
@@ -21,9 +21,11 @@ set-version)
     newValue="${BASH_REMATCH[1]}$2"
     echo "${content/${BASH_REMATCH[0]}/$newValue}" >$file
 
+cd android/replay-stubs && ./gradlew jar
```
lucas-zimerman:
```suggestion
```

Issue #4723, opened 2025-04-07
Report slow and frozen frames as span data too


markushi:
### Description

As of now slow and frozen frames are collected and attached to the root transaction. But for Mobile Vitals to work, the app start spans need to have the frame data set on the span `data` hash too.

Example span payload:
```
"spans": [
    {
      "timestamp": 1742508702.814,
      "start_timestamp": 1742508700.431,
      "op": "app.start.cold",
      "span_id": "42dbf44cc31249b4",
      "parent_span_id": "7168f0ee0fb14b00",
      "trace_id": "a8317f799fbb4d4e90103fc8dff13829",
      "status": "ok",
      "description": "Cold Start",
      "origin": "auto.ui.activity",
      "data": {
        "thread.name": "main",
        "thread.id": "13245",
        "frames.slow": 1, <<<
        "frames.frozen": 0, <<<
        "frames.total": 136, <<<
        "frames.delay": 0.131674152, <<<
      },
```

krystofwoldrich:
Let's tackle this in multiple steps, based on the type of span (app start, ttid/ttfd, started thru API).

antonis:
> Let's tackle this in multiple steps, based on the type of span (app start, ttid/ttfd, started thru API).

Opened sub issues for the above

kerenkhatiwada:
Related to Zendesk ticket #[166337](https://sentry.zendesk.com/agent/tickets/166337), is there an ETA for this fix?

antonis:
Thank you for bumping this @kerenkhatiwada 🙇
I'm planning to resume this work next week and ship it with a new version by the end of the month.

oferRounds:
Thank you @antonis, looking forward!

oferRounds:
Heyyyy
Do you might have an ETA for this one?

antonis:
Hey! Thanks for checking in 🙏 I don’t have a solid ETA just yet, we are working on this and I'll share a clearer timeline soon.

oferRounds:
Thank you @antonis! Sounds good :)

antonis:
Just as heads up that the implementation is currently in review. I'll update this issue when the fixes are shipped.

oferRounds:
Thank once again @antonis!

PR #5813, opened 2026-03-13
fix(tracing): Fix native frames measurements dropped for idle transactions

antonis:
- [x] Bugfix

## :scroll: Description

The NativeFrames integration was silently dropping transaction-level frame measurements (`frames_total`, `frames_slow`, `frames_frozen`) for idle transactions with the warning:
> `[NativeFrames] Frames were collected within larger than margin of error delay for spanId. Dropping the inaccurate values.`

This affected **all idle transactions** (navigation-based performance monitoring), causing slow/frozen frame metrics to be missing from the Performance pages.

Three issues fixed in the child span end frames fallback path of `processEvent`:

1. **Race condition**: Child span end frames were stored in the map only after two async `await`s completed (`_spanToNativeFramesAtStartMap.get()` and `fetchNativeFrames()`). If `processEvent` ran before those awaits finished, the fallback data was unavailable. Now stores a `Promise` immediately before any awaits, matching the pattern used by the root span end frames path.

2. **Timestamp mismatch**: Used `timestampInSeconds()` (wall-clock time) for the child span end timestamp, but compared it against `event.timestamp` which is the span's actual end time. For idle transactions where `event.timestamp` is trimmed to the last child span's end time, the wall-clock time at `spanEnd` event firing could differ. Now uses `spanToJSON(span).timestamp` — the same timestamp source the idle proxy trims to.

3. **Concurrent transaction interference**: The fallback was a single global variable (`_lastChildSpanEndFrames`) overwritten by any transaction's child spans. Now scoped per root span using `AsyncExpiringMap` keyed by root span ID, with a 60s TTL to survive idle timeouts + processing delays.

Bonus: child spans now reuse the same `fetchNativeFrames()` promise for both the fallback map and span attributes, eliminating a redundant native bridge call per child span.

Reported by a customer on SDK v6.15.1 seeing 100% of idle transactions missing slow/frozen frame measurements. The issue exists on both iOS and Android and affects all SDK versions (v6 and v7/v8) since the `processEvent` fallback logic has not changed.

Related: #4723

- Added test: `falls back to last child span end frames when root span end timestamp does not match event timestamp (idle transaction trim)` — simulates idle transaction trimming by shifting `event.timestamp` back to child span end time and advancing mock time for root span end, verifying the fallback path produces correct measurements.
- All existing NativeFrames tests pass (10/10).
- Build compiles successfully.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes


antonis:
@sentry review

antonis:
@claude review

packages/core/src/js/tracing/integrations/nativeFrames.ts
```diff
@@ -118,8 +120,31 @@ export const nativeFramesIntegration = (): Integration => {
...
+      const rootSpanId = getRootSpan(span).spanContext().spanId;
+      const spanTimestamp = spanToJSON(span).timestamp;
+      if (spanTimestamp) {
+        childEndFramesPromise = fetchNativeFrames();
+        _lastChildSpanEndFramesByRootSpan.set(
+          rootSpanId,
+          childEndFramesPromise
+            .then(frames => ({ timestamp: spanTimestamp, nativeFrames: frames }))
+            .then(undefined, error => {
+              debug.log(`[${INTEGRATION_NAME}] Error while fetching child span end frames.`, error);
```
lucas-zimerman: Won't this and line 195 generated duplicated error logs?
```
  1. Line 138: "Error while fetching child span end frames."                                                   
  2. Line 195: "Error while capturing end frames for span ${spanId}."                                          
```

lucas-zimerman:
Q: `Concurrent transaction interference` is there a test to validate this case?

packages/core/src/js/tracing/integrations/nativeFrames.ts
antonis: Good catch 👍 Removed with 88cae026653692d9a789624e2ca74fc871a97b2f

antonis:
> Q: Concurrent transaction interference is there a test to validate this case?

Good point. Added a test with 2d53119d0d9b5a2b0c175e0ca95d077430b95a42

PR #5828, opened 2026-03-17
feat(tracing): Add strict trace continuation support

antonis:
Add tests for strict trace continuation support (strictTraceContinuation and orgId options). The core logic is in @sentry/core v10.43.0.


antonis:
Closing in favor of #5829 which includes both JS and native bridge changes.

Issue #5665, opened 2026-02-13
Investigate reactNavigationIntegration performance on low end devices


antonis:
reactNavigationIntegration causes big performance overhead on low-end Android devices (Fire TV, Chromecast). We should investigate possible optimizations for these cases.

antonis:
We should also revisit [this older issue](https://github.com/getsentry/sentry-react-native/issues/3796)

antonis:
## Analysis: Performance overhead of `reactNavigationIntegration` on low-end devices

After a thorough code analysis cross-referenced with [this customer report](https://github.com/getsentry/sentry-react-native/issues/3796) and an internal support thread, the root cause is that registering the navigation container creates **idle root spans on every dispatch**, which activates a cascade of expensive work across multiple integrations simultaneously. Disabling individual features one by one doesn't resolve it because the overhead is distributed.

### What happens when `registerNavigationContainer` is called

The `__unsafe_action__` listener fires on **every** React Navigation dispatch, creating an idle navigation span. The existence of this root span activates:

| Integration | Overhead | Sampling check? |
|---|---|---|
| **Stall tracking** | 50ms `setTimeout` polling loop on JS thread | No |
| **Native frames** | `fetchNativeFrames()` native bridge call on start+end of every span (root + children) | No |
| **Profiling** | Starts Hermes CPU sampling | **Yes** (correct) |
| **HTTP instrumentation** | Every fetch/XHR during navigation becomes a child span, each triggering the above | N/A |
| **Per-span hooks** | 3-4 `client.on('spanEnd')` listeners per span, firing for all span ends system-wide | N/A |
| **AppState listener** | One per span via `cancelInBackground` | N/A |

### Why no single toggle fixes it

This matches what [#3796](https://github.com/getsentry/sentry-react-native/issues/3796) found — the reporter tried disabling stall tracking, native frames, and profiling individually, and the issue persisted each time. The current customer confirms: even with `enableNativeFramesTracking: false` + `enableTimeToInitialDisplay: false`, overhead remains ~2x.

### Additional issue: `useDispatchedActionData` defaults to `false`

[...]

PR #5840, opened 2026-03-18
fix(tracing): Unsubscribe spanEnd listeners after they fire to prevent accumulation

antonis:
- [x] Bugfix

## :scroll: Description

Each `client.on('spanEnd')` registered in `onSpanEndUtils` is a one-shot listener — it only cares about one specific span ending. Previously the unsubscribe function returned by `client.on()` was discarded, leaving listeners in the client's hook set indefinitely.

The fix captures and calls the unsubscribe function as the first action once the target span is identified, making every listener self-removing. All five functions in the file are updated: `onThisSpanEnd`, `adjustTransactionDuration`, `discardEmptyNavigationSpan` (shared by `ignoreEmptyBackNavigation` and `ignoreEmptyRouteChangeTransactions`), `onlySampleIfChildSpans`, and `cancelInBackground`.

Over the lifetime of an app with frequent navigation, these listeners accumulated: each navigation span left 3–4 zombie listeners that fired on every subsequent `spanEnd` event system-wide and returned early via an identity check. On low-end devices this contributes measurable overhead, especially when many child spans (HTTP requests) end during active navigation transactions.

Related issue: #5665

Added a dedicated test file `test/tracing/onSpanEndUtils.test.ts` with 11 tests covering:
- Correct callback behaviour for `onThisSpanEnd`
- Unsubscribe is called for every function after its target span ends
- AppState subscription is cleaned up in `cancelInBackground`
- Regression test: listener count does not grow across multiple spans — after 5 spans are created and ended, ending a new span does not trigger any of the old callbacks

All existing `reactnavigation` tests (93) continue to pass.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Part of a broader investigation into `reactNavigationIntegration` performance on low-end devices (#5665). Other improvements being tracked in that issue.

### DIFF
--- a/packages/core/src/js/tracing/integrations/nativeFrames.ts
+++ b/packages/core/src/js/tracing/integrations/nativeFrames.ts
@@ -1,5 +1,5 @@
 import type { Client, Event, Integration, Measurements, MeasurementUnit, Span } from '@sentry/core';
-import { debug, getRootSpan, spanIsSampled, spanToJSON, timestampInSeconds } from '@sentry/core';
+import { debug, getRootSpan, spanToJSON, timestampInSeconds } from '@sentry/core';
 import type { NativeFramesResponse } from '../../NativeRNSentry';
 import { AsyncExpiringMap } from '../../utils/AsyncExpiringMap';
 import { isRootSpan } from '../../utils/span';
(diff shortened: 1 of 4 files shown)


antonis:
@sentry review

antonis:
@cursor review

PR #5841, opened 2026-03-18
fix(android): Properly remove duplicated breadcrumbs

vovkasm:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
This should fix duplicated breadcrumbs on Android.
- Sentry syncs breadcrumbs from JS and native
- In the past JS breadcrumbs was filtered out before sent to JS side
- This filter worked on iterator of native Scope and broke when Native Android SDK replaces raw Scope with some composite scope class which plays role of "view" to several downstream Scope objects
- This patch change filter to work on materialized breadcrumbs array just after serialization

We had duplicated breadcrumbs on Android.

We applied this patch with patch-package. After that our applications did not duplicate breadcrumbs.

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
+++ b/packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
@@ -868,18 +868,19 @@ public class RNSentryModuleImpl {
     final @NotNull Map<String, Object> serialized =
         InternalSentrySdk.serializeScope(context, (SentryAndroidOptions) options, currentScope);
 
-    if (serialized.containsKey("breadcrumbs")) {
-      final @Nullable var breadcrumbs = (List<Map<String, Object>>) serialized.get("breadcrumbs");
-      if (breadcrumbs != null) {
-        var filtered = new ArrayList<Map<String, Object>>();
-        for (Map<String, Object> b : breadcrumbs) {
-          if ("react-native".equals(b.getOrDefault("origin", ""))) {
-            continue;
+    final @Nullable Object serializedBreadcrumbs = serialized.get("breadcrumbs");
+    if (serializedBreadcrumbs instanceof List) {
+      final List<?> breadcrumbs = (List<?>) serializedBreadcrumbs;
+      List<Map<?, ?>> filtered = new ArrayList<>();
+      for (Object o : breadcrumbs) {
+        if (o instanceof Map) {
+          final Map<?, ?> breadcrumb = (Map<?, ?>) o;
+          if (!"react-native".equals(breadcrumb.get("origin"))) {
+            filtered.add(breadcrumb);
           }
-          filtered.add(b);
         }
-        serialized.put("breadcrumbs", filtered);
       }
+      serialized.put("breadcrumbs", filtered);
     }
 
     final @Nullable Object deviceContext = RNSentryMapConverter.convertToWritable(serialized);


packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -814,19 +814,22 @@ protected void fetchNativeDeviceContexts(
...
+
+    if (serialized.containsKey("breadcrumbs")) {
+      final @Nullable var breadcrumbs = (List<Map<String, Object>>) serialized.get("breadcrumbs");
+      if (breadcrumbs != null) {
+        var filtered = new ArrayList<Map<String, Object>>();
+        for (Map<String, Object> b: breadcrumbs) {
+          if ("react-native".equals(b.getOrDefault("origin", ""))) continue;
+          filtered.add(b);
         }
+        serialized.put("breadcrumbs", filtered);
```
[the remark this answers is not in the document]
vovkasm: Instead it would be better to change JS <-> native bidirectional sync architecture. For now I'm let this decision to maintainers. I personally not support any changes which will slow down code. It is already pretty heavy.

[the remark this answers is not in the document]
antonis: Sounds good. Let's leave it as is for now 👍

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -814,19 +814,22 @@ protected void fetchNativeDeviceContexts(
...
-          breadcrumbsIterator.remove();
+
+    final @NotNull Map<String, Object> serialized =
+        InternalSentrySdk.serializeScope(context, (SentryAndroidOptions) options, currentScope);
+
+    if (serialized.containsKey("breadcrumbs")) {
+      final @Nullable var breadcrumbs = (List<Map<String, Object>>) serialized.get("breadcrumbs");
+      if (breadcrumbs != null) {
+        var filtered = new ArrayList<Map<String, Object>>();
+        for (Map<String, Object> b: breadcrumbs) {
```
antonis: nit: Minor formatting issue to make the linter happ
```suggestion
        for (Map<String, Object> b : breadcrumbs) {
```

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -814,19 +812,24 @@ protected void fetchNativeDeviceContexts(
...
-      while (breadcrumbsIterator.hasNext()) {
-        Breadcrumb breadcrumb = breadcrumbsIterator.next();
-        if ("react-native".equals(breadcrumb.getOrigin())) {
-          breadcrumbsIterator.remove();
+
+    final @NotNull Map<String, Object> serialized =
+        InternalSentrySdk.serializeScope(context, (SentryAndroidOptions) options, currentScope);
+
+    if (serialized.containsKey("breadcrumbs")) {
+      final @Nullable var breadcrumbs = (List<Map<String, Object>>) serialized.get("breadcrumbs");
```
antonis: I've notices some native tests are failing due to the Java version used. `var` is not supported and this needs explicit types:
```suggestion
      @SuppressWarnings("unchecked")                                                                                                                                                                                                                        
      final @Nullable List<Map<String, Object>> breadcrumbs = (List<Map<String, Object>>) serialized.get("breadcrumbs");
```

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -814,19 +812,24 @@ protected void fetchNativeDeviceContexts(
...
+      if (breadcrumbs != null) {
+        var filtered = new ArrayList<Map<String, Object>>();
```
antonis: Like above
```suggestion
        ArrayList<Map<String, Object>> filtered = new ArrayList<>();
```

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
vovkasm: Oh, will rewrite today! 😅
Thank you for review!

vovkasm:
After some thinking I'm rewrote filtering code to be more resilient to crashes and removed post java 8 language features. Tested on our live application, breadcrumbs still deduplicate as expected.

vovkasm:
Some tests failed again. But I suspect them failing in main branch also?  (https://github.com/getsentry/sentry-react-native/actions/runs/23300844792/job/67766824474)

antonis:
Awesome work @vovkasm 🚀 
Thank you for your contribution 🙇 
This will land in our next release 8.6.0 in about a week from now. I'll also add a changelog pointing to this PR with https://github.com/getsentry/sentry-react-native/pull/5851

Issue #5786, opened 2026-03-09
Split tests into iOS and Android


alwx:
For example, changes inside /ios probably don’t affect Android and thus, tests and sample apps shouldn’t run for those changes.<br>Using [https://github.com/getsentry/sentry-react-native/blob/main/.github/file-filters.yml](<https://github.com/getsentry/sentry-react-native/blob/main/.github/file-filters.yml>) it should be doable (in some cases).

Issue #5785, opened 2026-03-09
Changes to sample apps shouldn’t need to test the SDK


alwx:
Ideally PR should only run when related code/files were affected.

PR #5843, opened 2026-03-19
CI: detect-changes workflow to only check the affected components on the CI side

alwx:
Fixes https://github.com/getsentry/sentry-react-native/issues/5786
Fixes https://github.com/getsentry/sentry-react-native/issues/5785

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
`.github/workflows/detect-changes.yml` — New reusable workflow that uses `dorny/paths-filter`v3 with the file filters to produce `needs_ios` and `needs_android` boolean outputs.
That makes it all skip certain checks — for example, if the changes are in sample apps then there is no need to re-run tests for the whole SDK.

During the review I would love you to carefully look at `detect-changes.yml`, as well as the conditions specified in `platform-check` steps.

We want our CI to be fast and don't do redundant jobs.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


.github/workflows/sample-application.yml
```diff
@@ -67,51 +74,81 @@ jobs:
           - ios-use-frameworks: 'dynamic-frameworks'
             platform: 'macos'
     steps:
+      - name: Check if platform is needed
+        id: platform-check
```
antonis, line 78: nit: Wdyt of extracting the platform-check in a separate workflow and reuse it? (this can be part of a separate improvement PR)

.github/workflows/sample-application-expo.yml
```diff
@@ -52,56 +59,80 @@ jobs:
...
+            echo "Sample app changed — building ${{ matrix.platform }}."
+          elif [[ "${{ matrix.platform }}" == "ios" && "${{ needs.detect-changes.outputs.needs_ios }}" != "true" ]]; then
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+            echo "Skipping iOS — no relevant changes detected."
+          elif [[ "${{ matrix.platform }}" == "android" && "${{ needs.detect-changes.outputs.needs_android }}" != "true" ]]; then
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+            echo "Skipping Android — no relevant changes detected."
+          else
+            echo "skip=false" >> "$GITHUB_OUTPUT"
+          fi
```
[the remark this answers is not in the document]
antonis: This looks valid. Something like the following might work
```
elif [[ "${{ matrix.platform }}" == "web" && "${{ needs.detect-changes.outputs.js_source }}" != "true" ]]; then
  echo "skip=true" >> "$GITHUB_OUTPUT"                                                                                                                                                                                        
  echo "Skipping Web — no relevant JS changes detected."     
```

.github/workflows/native-tests.yml
```diff
@@ -98,8 +105,8 @@ jobs:
   test-android:
     name: android
     runs-on: ["ghcr.io/cirruslabs/ubuntu-runner-amd64:24.04", "runner_group_id:10"]
-    needs: [diff_check]
-    if: ${{ needs.diff_check.outputs.skip_ci != 'true' }}
+    needs: [diff_check, detect-changes]
+    if: ${{ needs.diff_check.outputs.skip_ci != 'true' && needs.detect-changes.outputs.needs_android == 'true' }}
```
antonis, line 109: Suggestion/idea for a future improvement:

It's nice that no runner is allocated here when the if is false here since each job is a single platform.

The other workflows use platform as a matrix dimension alongside others. We could apply the native-tests pattern on the other workflows too by splitting into separate build-ios / build-android / build-macos jobs, each with their own job-level if and a reduced matrix (without platform). That would avoid runner allocation entirely.

.github/file-filters.yml
```diff
@@ -16,4 +16,59 @@ high_risk_code: &high_risk_code
...
+  - 'packages/core/RNSentryAndroidTester/**'
+  - 'packages/core/sentry.gradle'
+
+# Changes to JS/TS source code (affects ALL platforms)
+js_source:
+  - 'packages/core/src/**'
+  - 'packages/core/plugin/**'
+  - 'packages/core/scripts/**'
+  - 'packages/core/package.json'
+  - 'packages/core/tsconfig.json'
```
antonis, line 43: No filter matches yarn.lock, root package.json, or .yarnrc.yml. A lockfile-only PR would have all detect-changes outputs as false, skipping all jobs.
```suggestion
  - 'packages/core/tsconfig.json'
  -  'yarn.lock'
  - 'package.json'
  - '.yarnrc.yml'
```

.github/workflows/sample-application.yml
alwx: Yes, let's better handle that separately

.github/workflows/native-tests.yml
alwx: I would also rather do it in another PR but that's a great suggestion. Will create an issue out of it!

.github/workflows/sample-application.yml
alwx: https://github.com/getsentry/sentry-react-native/issues/5865

.github/workflows/native-tests.yml
alwx: https://github.com/getsentry/sentry-react-native/issues/5864

PR #5849, opened 2026-03-19
fix(tracing): Guard getNewScreenTimeToDisplay behind enableTimeToInitialDisplay

antonis:
- [x] Bugfix

## :scroll: Description

`NATIVE.getNewScreenTimeToDisplay()` was called on every navigation state change regardless of the `enableTimeToInitialDisplay` option. This native bridge call posts to the main thread and registers a Choreographer callback (Android) or CADisplayLink (iOS) to capture the next frame timestamp.

The dispatch-time native calls (`setActiveSpanId`, `navigationProcessingSpan`) were already correctly guarded behind the `enableTimeToInitialDisplay` flag. The state-change-time call was the only one missing the guard.

A customer reported significant performance overhead on low-end devices (Fire TV, Chromecast) even after setting `enableTimeToInitialDisplay: false`. This unguarded native bridge call was still executing on every navigation, competing with the main thread on resource-constrained devices.

Related issue: #5665

Added a new test in `reactnavigation.ttid.test.tsx`:

- `should not call getNewScreenTimeToDisplay when ttid is disabled` — verifies the native bridge call is not made when `enableTimeToInitialDisplay` is `false`

All 94 reactnavigation tests pass (93 existing + 1 new).

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Part of a broader investigation into `reactNavigationIntegration` performance on low-end devices (#5665).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Fixes
 
 - Reduce `reactNavigationIntegration` performance overhead ([#5840](https://github.com/getsentry/sentry-react-native/pull/5840), [#5842](https://github.com/getsentry/sentry-react-native/pull/5842), [#5849](https://github.com/getsentry/sentry-react-native/pull/5849))
+- Fix duplicated breadcrumbs on Android ([#5841](https://github.com/getsentry/sentry-react-native/pull/5841))
 
 ## 8.5.0
 
@@ -28,7 +29,6 @@
 
 - Fix native frames measurements being dropped due to race condition ([#5813](https://github.com/getsentry/sentry-react-native/pull/5813))
 - Fix app start data lost when first navigation transaction is discarded ([#5833](https://github.com/getsentry/sentry-react-native/pull/5833))
-- Fix duplicated breadcrumbs on Android ([#5841](https://github.com/getsentry/sentry-react-native/pull/5841))
 
 ### Dependencies
 


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Fixes
+
+- Reduce `reactNavigationIntegration` performance overhead ([#5840](https://github.com/getsentry/sentry-react-native/pull/5840), [#5842](https://github.com/getsentry/sentry-react-native/pull/5842), [#5849](https://github.com/getsentry/sentry-react-native/pull/5849))
```
antonis, line 13: Mentioning here [all the three performance improvements](https://github.com/getsentry/sentry-react-native/issues/5665#issuecomment-4091082852)

packages/core/src/js/tracing/reactnavigation.ts
```diff
@@ -438,7 +438,9 @@ export const reactNavigationIntegration = ({
       return undefined;
     }
 
-    addTimeToInitialDisplayFallback(latestNavigationSpan.spanContext().spanId, NATIVE.getNewScreenTimeToDisplay());
+    if (enableTimeToInitialDisplay) {
```
lucas-zimerman, line 444: Q: should we mention the performance overhead when enabling this on changelog/docs?

antonis: I've added a note in the parameter documentation dd8106457b195c8fd67d9e50a15c6ce769c4d84e
I'll try to measure the actual performance overhead and update the doc accordingly

PR #5871, opened 2026-03-24
fix(android): Capture native exceptions consumed by Expo's bridgeless error handling

### LINKED ISSUE #5868: Support capturing native Android exceptions intercepted by Expo's error handler
### Problem Statement

On Expo SDK 53 Android apps, certain native exceptions (e.g., IllegalStateException from Fabric's SurfaceMountingManager) are caught by React Native's GuardedFrameCallback and routed through ReactHostImpl.handleHostException → handleInstanceException. Expo overrides handleInstanceException in its ExpoReactHostDelegate, which passes the error to registered hostHandlers (such as expo-updates) but does not rethrow it. Since hostHandlers is never empty in practice, the exception never reaches Java's UncaughtExceptionHandler, which is the mechanism sentry-java relies on to capture crashes.

The result is that reproducible native Android crashes go completely unreported in Sentry. The Sentry SDK itself works correctly, when the rethrow is forced, the event is captured with the expected UncaughtExceptionHandler mechanism. The problem is the exception never reaches the SDK.

This is complicated by the fact that expo-updates has a legitimate reason to intercept these errors (for OTA update rollback/error recovery), so simply rethrowing is not a viable solution, it would break expo-updates error recovery.

Related Expo PR: https://github.com/expo/expo/pull/37021

### Solution Brainstorm

Explore options for the SDK to capture these exceptions at a level before they are consumed by Expo's handler chain. One possibility is hooking into Expo's ReactNativeHostHandler interface, though that would introduce a dependency on Expo. There may be other approaches worth considering.

If the team feels this is best addressed on Expo's side rather than ours, please flag that so we can communicate it back to the customer, they are also in contact with the Expo team.

None

alwx:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Fixes #5868 

On Expo SDK 53+ Android apps, `ExpoReactHostDelegate.handleInstanceException` iterates registered `ReactNativeHostHandler`s but does **not rethrow** the exception. This causes native crashes (e.g., `IllegalStateException` from Fabric's `SurfaceMountingManager`) caught by React Native's `GuardedFrameCallback` to be silently swallowed — they never reach Java's `UncaughtExceptionHandler`, which `sentry-java` relies on for crash capture.

This PR registers a `ReactNativeHostHandler` via Expo's Package system that intercepts these exceptions and captures them directly through `Sentry.captureException()`.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### Fixes
 
+- Capture native exceptions consumed by Expo's bridgeless error handling ([#5871](https://github.com/getsentry/sentry-react-native/pull/5871))
```
antonis: nit: Let's also include that the scope of this fix is on Android

```suggestion
- Capture native exceptions consumed by Expo's bridgeless error handling on Android([#5871](https://github.com/getsentry/sentry-react-native/pull/5871))
```

packages/core/android/src/expo/java/io/sentry/react/expo/SentryReactNativeHostHandler.java
```diff
@@ -0,0 +1,44 @@
...
+
+    final Mechanism mechanism = new Mechanism();
+    mechanism.setType(MECHANISM_TYPE);
+    mechanism.setHandled(false);
+
+    final ExceptionMechanismException mechanismException =
+        new ExceptionMechanismException(mechanism, exception, Thread.currentThread());
+
+    Sentry.captureException(mechanismException);
+  }
```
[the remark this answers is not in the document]
antonis: I think that this is valid 👍

antonis (commented):
The fix LGTM. I think with the cursor feedback addressed and scoping the changelog would be ready to 🚢

packages/core/android/src/expo/java/io/sentry/react/expo/SentryReactNativeHostHandler.java
```diff
@@ -0,0 +1,44 @@
...
+    if (!Sentry.isEnabled()) {
+      return;
+    }
+
+    final Mechanism mechanism = new Mechanism();
+    mechanism.setType(MECHANISM_TYPE);
+    mechanism.setHandled(false);
+
+    final ExceptionMechanismException mechanismException =
+        new ExceptionMechanismException(mechanism, exception, Thread.currentThread());
```
lucas-zimerman: Q: Is the current thread consumed here the same as the one where the exception ocurred?

packages/core/android/src/expo/java/io/sentry/react/expo/SentryReactNativeHostHandler.java
```diff
@@ -0,0 +1,44 @@
+package io.sentry.react.expo;
```
lucas-zimerman, line 1: Would you be able to add some tests for this file?

packages/core/android/src/expo/java/io/sentry/react/expo/SentryReactNativeHostHandler.java
alwx: `Thread.currentThread()` captures the thread the handler runs on, which is the same thread where the exception propagated through `handleHostException` → `handleInstanceException`. Looking at `ReactHostImpl`, `handleHostException` is called from different contexts — the UI thread (via `GuardedFrameCallback`), the JS executor thread, or the BG executor — depending on where the crash originated. Since the handler is invoked synchronously (no thread hop between the crash and our callback), `Thread.currentThread()` correctly reflects the thread where the exception was being processed.

That said, it may not always be the _originating_ thread if React Native internally catches on one thread and re-dispatches to another before calling `handleHostException`. But this matches how sentry-java's own `UncaughtExceptionHandler` works — it also uses `Thread.currentThread()` at the point of handling.

packages/core/android/src/expo/java/io/sentry/react/expo/SentryReactNativeHostHandler.java
alwx: Added unit tests in the latest commit (9567e56f). Tests cover:
- Skip capture in dev support mode
- Skip capture when Sentry is not enabled
- Capture with correct unhandled mechanism (`type="expoReactHost"`, `handled=false`)
- Graceful handling when `Sentry.captureException` throws (the new try/catch)

Created an `expo-stubs.jar` with minimal `Package` and `ReactNativeHostHandler` interfaces for the test project (following the existing `replay-stubs.jar` pattern).

antonis: nit: Wdyt of adding a short readme with the context on what is the jar file and how to regenerate it if needed (similar to [replay-stubs](https://github.com/getsentry/sentry-react-native/blob/main/packages/core/android/replay-stubs/README.md))?

packages/core/RNSentryAndroidTester/app/src/test/java/io/sentry/react/expo/SentryReactNativeHostHandlerTest.kt
antonis, line 1: There are minor formatting errors here that can be autocorrected using 'ktlint --format'

packages/core/android/src/expo/java/io/sentry/react/expo/SentryReactNativeHostHandler.java
alwx: I didn't even know there is a document descrbing replay-stubs! :) Will check it out.

alwx:
Ok, I've replaced the conditional `findProject(':expo-modules-core')` approach with a `compileOnly` stubs jar (matching the existing `replay-stubs` pattern). Now it's more or less the same + there is also a README like it's done for `replay-stubs`

@antonis @lucas-zimerman please check :)

packages/core/android/expo-stubs/src/main/java/expo/modules/core/interfaces/ReactNativeHostHandler.java
```diff
@@ -0,0 +1,5 @@
+package expo.modules.core.interfaces;
+
+public interface ReactNativeHostHandler {
+  default void onReactInstanceException(boolean useDeveloperSupport, Exception exception) {}
```
antonis: This needs a comment to make the linter happy
```
ReactNativeHostHandler.java:4:	UncommentedEmptyMethodBody:	Document empty method body
```

PR #5873, opened 2026-03-24
refactor(ci): Extract platform-check into a reusable composite action

### LINKED ISSUE #5865: CI: Extract platform-check step into a reusable workflow
The platform-check step that decides whether to skip a build based on detect-changes outputs is duplicated across sample app workflows (sample-application.yml, sample-application-expo.yml). The logic is nearly identical — check if the sample itself changed, then check the platform-specific needs\_\* output.

We could extract this into a reusable workflow (or composite action) that takes the platform and the relevant detect-changes outputs as inputs and returns a skip output.

alwx:
- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

Extract the duplicated platform-check step into a composite action at `.github/actions/platform-check`.

Replaces three inline bash scripts across `sample-application.yml` and `sample-application-expo.yml` with a single reusable action.

Closes #5865

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

#skip-changelog

### DIFF
--- a/.github/workflows/sample-application-expo.yml
+++ b/.github/workflows/sample-application-expo.yml
@@ -59,6 +59,8 @@ jobs:
           - platform: 'android'
             ios-use-frameworks: 'dynamic-frameworks'
     steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
       - name: Check if platform is needed
         id: platform-check
         uses: ./.github/actions/platform-check
@@ -68,9 +70,6 @@ jobs:
           needs_ios: ${{ needs.detect-changes.outputs.needs_ios }}
           needs_android: ${{ needs.detect-changes.outputs.needs_android }}
           needs_web: ${{ needs.detect-changes.outputs.needs_web }}
-
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
-        if: ${{ steps.platform-check.outputs.skip != 'true' }}
 
       - name: Enable Corepack (NPM)
         if: ${{ steps.platform-check.outputs.skip != 'true' && matrix.platform != 'ios' }}
--- a/.github/workflows/sample-application.yml
+++ b/.github/workflows/sample-application.yml
@@ -74,6 +74,8 @@ jobs:
           - ios-use-frameworks: 'dynamic-frameworks'
             platform: 'macos'
     steps:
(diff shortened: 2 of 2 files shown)


.github/workflows/sample-application-expo.yml
```diff
@@ -61,22 +61,13 @@ jobs:
...
-            echo "skip=true" >> "$GITHUB_OUTPUT"
-            echo "Skipping iOS — no relevant changes detected."
-          elif [[ "${{ matrix.platform }}" == "android" && "${{ needs.detect-changes.outputs.needs_android }}" != "true" ]]; then
-            echo "skip=true" >> "$GITHUB_OUTPUT"
-            echo "Skipping Android — no relevant changes detected."
-          elif [[ "${{ matrix.platform }}" == "web" && "${{ needs.detect-changes.outputs.needs_web }}" != "true" ]]; then
-            echo "skip=true" >> "$GITHUB_OUTPUT"
-            echo "Skipping Web — no relevant changes detected."
-          fi
+        uses: ./.github/actions/platform-check
```
antonis, line 66: We may need an unconditional `actions/checkout` to make the `platform-check` accessible and prevent the Ci failures 🤔

lucas-zimerman:
Would you be able to make sub PRs targeting this PR for validating it?

Ex: PR1 contains Android only change, PR2 contains iOS only change, PR3 contains Android & iOS changes, PR4 contains JS changes, ...

.github/workflows/sample-application-expo.yml
alwx: done!

.github/workflows/sample-application-expo.yml
lucas-zimerman, line 31: This change is part of another PR that was merged, but shouldn't we also add `diff_check` on the needs list here?

Same question on `.github/workflows/sample-application.yml` for `detect-changes`

lucas-zimerman:
Thanks for the PR! Tests will improve a lot with the changes!

.github/workflows/sample-application-expo.yml
alwx: Making `detect-changes` depend on `diff_check` can save a CI run when `skip_ci=true` but I think it's more of an optimization thing, not a correctness issue. So I'd rather skip it for now, especially since it's out of context of this PR.

Issue #4908, opened 2025-06-11
Sentry Replays counting muted Errors or something off


varunraghu3112:
### What React Native libraries do you use?

React Navigation

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### @sentry/react-native SDK Version

6.5.0

### How does your development environment look like?

````
I have blacklisted a few errors in ignoreErrors/beforeSend being captured in Sentry, and they are not being reported as errors — which is expected.

However, it seems that these blacklisted errors are still being treated as recordings in Sentry Replays but errors are 0 in the replay. I’ve configured the following settings for a specific release:
	•	replaysOnErrorSampleRate: 0.2
	•	replaysSessionSampleRate: 0

Despite this, I’m still seeing replays without any captured errors, which is unnecessarily increasing our replay usage.



````

### Sentry.init()

````js
Sentry.init({
  dsn: 'xxxx',
  release: `x.x`,
  dist: `${versionCode}`,
  beforeSend(event, hint) {
    // since ignoreErrors doesnt seem to ignore 'Non-Error exception captured with keys: code, error, message, payload'
    // adding custom handling block to filter out Non-Error type exception

    const nonErrorPatterns = sentryNonErrorPatterns || [
      'Non-Error exception captured',
      'Object captured as exception with keys: code, error, message, payload',
      'Unable to resolve host'
    ];

    const isNonErrorException = nonErrorPatterns.some(
      (pattern) =>
        event.exception.values[0]?.value?.includes(pattern) || hint.originalException?.message?.includes(pattern)
    );

    if (isNonErrorException) {
      // ignore these kind of issues
      return null;
    }
    // check if the originalException is object and add the data as extras to the Sentry Event
    if (
      typeof hint.originalException === 'object' &&
      hint.originalException != null &&
      'key' in hint.originalException
    ) {
      event.extra = {
        ...event.extra,
        ...hint.originalException
      };
    }
    return event;
  },
  replaysSessionSampleRate: '0',
  replaysOnErrorSampleRate: '0.2',
  beforeSend(event, hint) {
  },
  enableAutoSessionTracking: true,
  environment: __DEV__ ? 'development' : logEnabled ? 'qa' : 'production',
  ignoreErrors: [
    'Non-Error exception captured with keys: code, error, message, payload',
    'Non-Error exception captured',
    'Non-Error exception captured with keys',
    'Non-Error exception captured with keys:'
  ],
  tracesSampleRate: performaceSample,
[...]

PR #5885, opened 2026-03-25
Fix(Replay): dont record ignored/dropped errors

lucas-zimerman:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Users wanted to only track errors, but the way `mobileReplay` was set, it triggered replay captured even to dropped errors, since it was doing this on the middle of the event processing workflow, which could lead to dropped events before sending it.

The fix is a hook into `beforeSend`, since it's the last stop before sending an event.
Ideally we shold use something like `client.on('beforeBeforeSend...` but the client hook was syncronous, which blocks the usage of Native APIs to fetch the replay data to fill into the event's context.

This way, when users set to only capture errors, Replay will not send recording data when the only captured events were dropped by IgnoredError/eventProcessor or beforeSend.

On a device, sample app.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

Close https://github.com/getsentry/sentry-react-native/issues/4908


packages/core/src/js/replay/mobilereplay.ts
```diff
@@ -303,6 +303,21 @@ export const mobileReplayIntegration = (initOptions: MobileReplayOptions = defau
...
+    const clientOptions = client.getOptions();
+    const originalBeforeSend = clientOptions.beforeSend;
+    clientOptions.beforeSend = async (event: ErrorEvent, hint: EventHint): Promise<ErrorEvent | null> => {
+      if (originalBeforeSend) {
+        const result = await originalBeforeSend(event, hint);
+        if (result === null) {
+          // Event was dropped by user's beforeSend, don't capture replay
+          return null;
+        }
+        return processEvent(result, hint);
```
lucas-zimerman: This is redundant.
```suggestion
```

packages/core/src/js/replay/mobilereplay.ts
```diff
@@ -303,6 +303,20 @@ export const mobileReplayIntegration = (initOptions: MobileReplayOptions = defau
...
+    const originalBeforeSend = clientOptions.beforeSend;
+    clientOptions.beforeSend = async (event: ErrorEvent, hint: EventHint): Promise<ErrorEvent | null> => {
+      if (originalBeforeSend) {
+        const result = await originalBeforeSend(event, hint);
+        if (result === null) {
+          // Event was dropped by user's beforeSend, don't capture replay
+          return null;
+        }
+      }
+      return processEvent(event, hint);
```
lucas-zimerman:
```suggestion
      return processEvent(result, hint);
```

packages/core/src/js/replay/mobilereplay.ts
```diff
@@ -303,6 +303,20 @@ export const mobileReplayIntegration = (initOptions: MobileReplayOptions = defau
     });
 
     client.on('beforeAddBreadcrumb', enrichXhrBreadcrumbsForMobileReplay);
+
+    // Wrap beforeSend to run processEvent after user's beforeSend
+    const clientOptions = client.getOptions();
+    const originalBeforeSend = clientOptions.beforeSend;
+    clientOptions.beforeSend = async (event: ErrorEvent, hint: EventHint): Promise<ErrorEvent | null> => {
+      if (originalBeforeSend) {
+        const result = await originalBeforeSend(event, hint);
```
lucas-zimerman:
```suggestion
      var result = event;
      if (originalBeforeSend) {
        result = await originalBeforeSend(event, hint);
```

packages/core/src/js/replay/mobilereplay.ts
```diff
@@ -303,6 +303,21 @@ export const mobileReplayIntegration = (initOptions: MobileReplayOptions = defau
...
+        if (result === null) {
+          // Event was dropped by user's beforeSend, don't capture replay
+          return null;
+        }
+        return processEvent(result, hint);
+      }
+      return processEvent(event, hint);
+    };
```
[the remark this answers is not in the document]
antonis: I think this is valid 👍

lucas-zimerman:
> LGTM 🚀 It would be nice to add [the try/catch cursor suggests](https://github.com/getsentry/sentry-react-native/pull/5885/changes#r2993408741) and run the full test suite.

Done on https://github.com/getsentry/sentry-react-native/pull/5885/commits/aac59684505678cbb1f58b242be4b85704beb1bb

PR #5940, opened 2026-03-31
feat(core): Add Sentry.appLoaded() API to signal app start end

alwx:
## 📢 Type of change

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## 📜 Description

Fixes [RN-565](https://linear.app/getsentry/issue/RN-565/a-method-to-notify-sentry-that-the-app-is-loaded)

Adds a public `Sentry.appLoaded()` function that users can call to explicitly signal when their app is fully ready for user interaction. This provides a more accurate app start end timestamp for apps that do significant async work after the root component mounts.

When `appLoaded()` is called it:

* Records the current timestamp as the manual app start end
* Fetches native frames for frame data attachment
* Triggers standalone app start capture when in standalone mode
* Marks the span origin as `manual.app.start`

**Existing behaviour is unchanged when** `appLoaded()` is not called.

### Usage

```ts
await loadRemoteConfig();
await restoreSession();
SplashScreen.hide();
Sentry.appLoaded();
```

## 💚 How did you test it?

Tests were added :)

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/packages/core/src/js/tracing/integrations/appStart.ts
+++ b/packages/core/src/js/tracing/integrations/appStart.ts
@@ -38,6 +38,8 @@ const INTEGRATION_NAME = 'AppStart';
 export type AppStartIntegration = Integration & {
   captureStandaloneAppStart: () => Promise<void>;
   resetAppStartDataFlushed: () => void;
+  cancelDeferredStandaloneCapture: () => void;
+  scheduleDeferredStandaloneCapture: () => void;
 };
 
 /**
@@ -112,6 +114,9 @@ export async function _appLoaded(): Promise<void> {
 
   const integration = client.getIntegrationByName<AppStartIntegration>(INTEGRATION_NAME);
   if (integration) {
+    // Cancel any deferred standalone send from auto-capture — we'll send our own
+    // with the correct manual timestamp instead of sending two transactions.
+    integration.cancelDeferredStandaloneCapture();
     // In standalone mode, auto-capture may have already flushed the transaction.
     // Reset the flag so captureStandaloneAppStart can re-send with the manual timestamp.
     integration.resetAppStartDataFlushed();
@@ -150,7 +155,19 @@ export async function _captureAppStart({ isManual }: { isManual: boolean }): Pro
   });
 
   await fetchAndUpdateEndFrames();
-  await client.getIntegrationByName<AppStartIntegration>(INTEGRATION_NAME)?.captureStandaloneAppStart();
+
+  const integration = client.getIntegrationByName<AppStartIntegration>(INTEGRATION_NAME);
+  if (integration) {
+    if (!isManual) {
+      // For auto-capture, defer the standalone send to give appLoaded() a chance
+      // to override the end timestamp before the transaction is sent.
+      // If appLoaded() is called, it cancels this deferred send and sends its own.
+      // In non-standalone mode, scheduleDeferredStandaloneCapture is a no-op.
+      integration.scheduleDeferredStandaloneCapture();
+    } else {
+      await integration.captureStandaloneAppStart();
+    }
+  }
 }
 
 /**
@@ -276,6 +293,7 @@ export const appStartIntegration = ({
   let firstStartedActiveRootSpanId: string | undefined = undefined;
   let firstStartedActiveRootSpan: Span | undefined = undefined;
   let cachedNativeAppStart: NativeAppStartResponse | null | undefined = undefined;
+  let deferredStandaloneTimeout: ReturnType<typeof setTimeout> | undefined = undefined;
 
   const setup = (client: Client): void => {
     _client = client;
@@ -305,6 +323,10 @@ export const appStartIntegration = ({
         firstStartedActiveRootSpan = undefined;
         isAppLoadedManuallyInvoked = false;
         cachedNativeAppStart = undefined;
+        if (deferredStandaloneTimeout !== undefined) {
+          clearTimeout(deferredStandaloneTimeout);
+          deferredStandaloneTimeout = undefined;
+        }
       } else {
         debug.log(
           '[AppStartIntegration] Waiting for initial app start was flush, before updating based on runApplication call.',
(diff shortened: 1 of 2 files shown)


packages/core/src/js/tracing/integrations/appStart.ts
antonis, line 76: Q: would it make sense to deprecate this method?

packages/core/test/tracing/integrations/appStart.test.ts
```diff
@@ -1024,6 +1027,134 @@ describe('App Start Integration', () => {
...
+  });
+
+  afterEach(() => {
+    clearReactNativeBundleExecutionStartTimestamp();
+    _clearAppStartEndData();
+    _clearRootComponentCreationTimestampMs();
+  });
+
+  function makeIntegration(): AppStartIntegrationTest {
+    const integration = appStartIntegration({ standalone: false }) as AppStartIntegrationTest;
```
antonis, line 1055: Q: Should we also test with `standalone: true`?

packages/core/src/js/sdk.tsx
```diff
@@ -220,6 +221,29 @@ export function nativeCrash(): void {
...
+ *
+ * @example
+ * ```ts
+ * await loadRemoteConfig();
+ * await restoreSession();
+ * SplashScreen.hide();
+ * Sentry.appLoaded();
+ * ```
+ */
+export function appLoaded(): void {
```
antonis, line 244: I have no strong opinion on this but wdyt of marking as experimental since it's a new api?

packages/core/src/js/tracing/integrations/appStart.ts
```diff
@@ -71,12 +72,64 @@ export function captureAppStart(): Promise<void> {
...
+    try {
+      const endFrames = await NATIVE.fetchNativeFrames();
+      debug.log('[AppStart] Captured end frames for app start.', endFrames);
+      _updateAppStartEndFrames(endFrames);
+    } catch (error) {
+      debug.log('[AppStart] Failed to capture end frames for app start.', error);
+    }
+  }
+
+  await client.getIntegrationByName<AppStartIntegration>(INTEGRATION_NAME)?.captureStandaloneAppStart();
```
antonis: nit: we could extract this part from [_captureAppStart](https://github.com/getsentry/sentry-react-native/blob/a5000dcb195873ebc7eaa6a7b904b566c1957e86/packages/core/src/js/tracing/integrations/appStart.ts#L97) and reuse it

packages/core/src/js/sdk.tsx
alwx: done!

packages/core/src/js/tracing/integrations/appStart.ts
```diff
@@ -59,24 +59,71 @@ interface AppStartEndData {
...
+    appStartEndData.timestampMs = timestampMs;
+    appStartEndData.endFrames = null;
+  } else {
+    _setAppStartEndData({ timestampMs, endFrames: null });
+  }
+  isRecordedAppStartEndTimestampMsManual = true;
+
+  await fetchAndUpdateEndFrames();
+  await client.getIntegrationByName<AppStartIntegration>(INTEGRATION_NAME)?.captureStandaloneAppStart();
+}
```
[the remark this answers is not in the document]
antonis: This looks valid 👍

packages/core/src/js/tracing/integrations/appStart.ts
```diff
@@ -59,24 +59,71 @@ interface AppStartEndData {
 
 let appStartEndData: AppStartEndData | undefined = undefined;
 let isRecordedAppStartEndTimestampMsManual = false;
+let isAppLoadedManuallyInvoked = false;
```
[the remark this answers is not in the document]
antonis: This looks valid 👍

packages/core/test/tracing/integrations/appStart.test.ts
```diff
@@ -1243,6 +1242,96 @@ describe('appLoaded() standalone mode', () => {
...
+    expect(autoSpan?.timestamp).toBeCloseTo(autoTimeSeconds, 1);
+
+    // Now call appLoaded() with a later timestamp
+    const manualTimeSeconds = autoTimeSeconds + 2;
+    mockFunction(timestampInSeconds).mockReturnValue(manualTimeSeconds);
+    await _appLoaded();
+    await new Promise(resolve => setTimeout(resolve, 0));
+
+    // appLoaded() should have sent a second standalone transaction with the manual timestamp
+    expect(standaloneClient.eventQueue.length).toBe(2);
```
antonis: Q: With the latest fix two standalone app start transactions are sent to Sentry for the same launch. Is this intentional and how does it appear on the dashboard?

alwx: yes, that's certainly not how it needs to be. let me check that again

alwx: ok, so here is how it changed now: when `appLoaded()` is called (even synchronously after mount), it cancels the deferred send and sends only one transaction with the correct manual timestamp
(we can reiterate over this approach later after the potential discussion here: https://github.com/getsentry/sentry-cocoa/issues/6886)

antonis: thank you 🙇

packages/core/src/js/tracing/integrations/appStart.ts
```diff
@@ -540,12 +647,38 @@ export const appStartIntegration = ({
...
+  const scheduleDeferredStandaloneCapture = (): void => {
+    if (!standalone) {
+      return;
+    }
+    deferredStandaloneTimeout = setTimeout(() => {
+      deferredStandaloneTimeout = undefined;
+      // oxlint-disable-next-line typescript-eslint(no-floating-promises)
+      captureStandaloneAppStart();
+    }, 0);
+  };
```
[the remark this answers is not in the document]
antonis: This looks valid but it might be an edge case not affecting the reported use case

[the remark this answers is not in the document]
alwx: Yes, I think the combination of standalone + Sentry.wrap() + appLoaded() after significant async work is a bit of a rare case 😅 In the primary use case (non-standalone), `processEvent` naturally defers attachment to the first navigation transaction, and `appLoaded()` works correctly.

PR #5964, opened 2026-04-07
fix(android): Recover missing JS stack traces from native JavascriptException

### LINKED ISSUE #5071: Missing stacktrace in Value is undefined, expected an Object Error
There was a report of an error with a missing stacktrace and `Error: Value is undefined, expected an Object`<br><br>The same error seems to have been reported with a stacktrace by Crashlytics as

`Fatal Exception: com.facebook.react.common.JavascriptException TypeError: Cannot read property 'content' of undefined`<br><br>The problem occurred with SDK `6.15.1` and RN `0.79.4` with `Hermes`

antonis:
- [x] Bugfix

## :scroll: Description

On Android with Hermes, React render errors can arrive at the JS-side `ErrorUtils.setGlobalHandler` without a `.stack` property. The same error is captured by React Native's native layer as a `JavascriptException` which contains the full JS stack trace string (with component names and bundle offsets).

Previously, `JavascriptException` was fully ignored via `addIgnoredExceptionForType`, discarding the stack information entirely. This PR:

1. **Intercepts** `JavascriptException` in `beforeSend` instead of ignoring it — caches the stack trace string, then drops the event (same dedup behavior)
2. **Exposes** the cached stack to JS via a sync bridge method (`fetchCachedJavascriptExceptionStack`)
3. **Recovers** the stack in a new `NativeStackRecovery` integration that fetches and parses the cached stack when an event has no frames

The cache uses `AtomicReference` with a 5-second TTL and single-consumption (`getAndClear`) semantics.

Fixes https://github.com/getsentry/sentry-react-native/issues/5071

A user reported a fatal error `"Value is undefined, expected an Object"` arriving in Sentry with zero stacktrace frames, while Crashlytics captured the same error with a full stack as a `JavascriptException`. The root cause is that React's error handling transforms render errors before they reach the JS global error handler, stripping the `.stack` property. The native `JavascriptException` retains the original stack but was being discarded.

See also: https://x.com/TheAnirudh/status/1954761591515476463

- 9 new JS tests for `NativeStackRecovery` integration (all branches covered)
- 3 new/updated Android unit tests for `beforeSend` JavascriptException handling
- Integration ordering test verifying `NativeStackRecovery` runs before `RewriteFrames`
- Verified `defaultStackParser` correctly parses `JavascriptException.getMessage()` format including `address at` prefixes
- All 1219 tests pass, 0 lint errors, no circular dependencies

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- Verify with a real Android/Hermes app that the recovered stack frames are correctly symbolicated by Sentry's server
- Consider if a similar issue exists on iOS (currently iOS does not exhibit this problem)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Fixes
 
+- Recover missing JS stack traces from native `JavascriptException` on Android ([#5964](https://github.com/getsentry/sentry-react-native/pull/5964))
 - Lazy-load Metro internal modules to prevent Expo 55 import errors ([#5958](https://github.com/getsentry/sentry-react-native/pull/5958))
 
 ### Dependencies
--- a/packages/core/RNSentryAndroidTester/app/src/androidTest/java/io/sentry/react/RNSentrySDKTest.kt
+++ b/packages/core/RNSentryAndroidTester/app/src/androidTest/java/io/sentry/react/RNSentrySDKTest.kt
@@ -176,7 +176,7 @@ class RNSentrySDKTest {
     }
 
     private fun verifyDefaults(actualOptions: SentryAndroidOptions) {
-        assertTrue(actualOptions.ignoredExceptionsForType.contains(JavascriptException::class.java))
+        assertFalse(actualOptions.ignoredExceptionsForType.contains(JavascriptException::class.java))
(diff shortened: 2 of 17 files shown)


antonis:
@cursor review

antonis:
Superseded by a simpler approach — using React's componentStack as a fallback instead of bridging the native JavascriptException stack. See new PR.

PR #5965, opened 2026-04-07
fix(android): Use componentStack as fallback for missing error stack traces

### LINKED ISSUE #5071: Missing stacktrace in Value is undefined, expected an Object Error

antonis:
- [x] Bugfix

## :scroll: Description

When React render errors occur on Android with Hermes, the error may arrive at `ErrorUtils.setGlobalHandler` without a usable `.stack` property (empty or message-only). However, React's `ReactFiberErrorDialog` attaches a `.componentStack` to the error object before it reaches the handler. The `componentStack` contains component locations with bundle offsets that can be symbolicated by Sentry.

This PR uses `componentStack` as a fallback for `error.stack` when the stack has no frames, so that `eventFromException` can extract frames with source locations.

The fix handles both cases:
- `error.stack` is `undefined`/empty
- `error.stack` is a message-only string with no frame lines (e.g. `"Error: Value is undefined, expected an Object"`)

**Note:** The `componentStack` is a React component tree (not a JS execution stack). After symbolication, it shows which component caused the error and its source location, but not the exact call chain within the component. This is the best information available on the JS side for this class of errors.

Fixes https://github.com/getsentry/sentry-react-native/issues/5071

A user reported a fatal error `"Value is undefined, expected an Object"` arriving in Sentry with no stacktrace, while Crashlytics captured the same error with a full component stack as a `JavascriptException`. The root cause is that some Hermes/JSI errors have no `.stack` property, but React attaches `.componentStack` which contains the same location information.

See also: https://x.com/TheAnirudh/status/1954761591515476463

- Added test verifying `componentStack` is used as fallback when `.stack` is missing
- Added test verifying `componentStack` is used when `.stack` is message-only (no frames)
- Added test verifying existing `.stack` with frames is not overridden
- All 1216 tests pass, 0 lint errors
- Verified `defaultStackParser` correctly parses the `componentStack` format including `address at` prefixes

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- Verify with a real Android/Hermes app that the recovered component stack frames are correctly symbolicated


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### Fixes
 
+- Use React `componentStack` as fallback when error has no stack trace on Android ([#5071](https://github.com/getsentry/sentry-react-native/issues/5071))
```
```suggestion
- Use React `componentStack` as fallback when error has no stack trace on Android ([#5965](https://github.com/getsentry/sentry-react-native/pull/5965)
```

antonis:
@sentry review

antonis:
@cursor review

packages/core/src/js/integrations/reactnativeerrorhandlers.ts
```diff
@@ -170,6 +170,16 @@ function setupErrorUtilsGlobalHandler(): void {
...
     }
 
+    // React render errors may arrive without useful frames in .stack but with a
+    // .componentStack (set by ReactFiberErrorDialog) that contains component
+    // locations with bundle offsets. Use componentStack as a fallback so
+    // eventFromException can extract frames with source locations.
+    // oxlint-disable-next-line typescript-eslint(no-unsafe-member-access)
+    if (error?.componentStack && (!error.stack || !hasStackFrames(error.stack))) {
+      // oxlint-disable-next-line typescript-eslint(no-unsafe-member-access)
+      error.stack = `${error.message || 'Error'}${error.componentStack}`;
```
lucas-zimerman, line 180: shouldn't we add an extra space between the parameters, so Error is not clued to the componentStack message?

lucas-zimerman (approved):
nit comment, other than that, LGTM!

PR #5981, opened 2026-04-13
fix(core): Retry native module resolution to prevent silent event drops

antonis:
## 📢 Type of change

- [X] Bugfix

## 📜 Description

The `RNSentry` TurboModule reference is resolved once at module load time (`wrapper.ts:48`). In production Hermes bytecode builds, JS execution can start before the TurboModule registry is fully populated, causing `getRNSentryModule()` to return `undefined` permanently. This silently disables native transport and drops all events.

The fix re-resolves the native module in `isNativeAvailable()` if the initial resolution returned `undefined`. By the time `Sentry.init()` calls this method (milliseconds after module load), TurboModules are registered and the module resolves correctly.

Changes:

* `const RNSentry` → `let RNSentry` to allow re-assignment
* `isNativeAvailable()` retries `getRNSentryModule()` if `RNSentry` is `undefined`

## 💡 Motivation and Context

Fixes [RN-447](https://linear.app/getsentry/issue/RN-447/events-silently-dropped-when-sentryinit-is-first-statement-production)

Users report that events are silently dropped in production builds (Hermes + New Architecture) when `Sentry.init()` is the first statement after an early return guard. Adding any statement (e.g., `console.log('test')`) before `Sentry.init()` resolves the issue, pointing to a race condition in TurboModule registration timing.

## 💚 How did you test it?

* Added unit test simulating the race condition (module initially undefined, then available)
* All existing tests pass (78/78 wrapper tests, full suite green)
* On-device testing with Pixel 8 Pro, Android 16, production Hermes build

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## 🔮 Next steps

* Separately investigate the `__SENTRY_OPTIONS__` + `autoInitializeNativeSdk` code path which has a related but distinct silent event drop bug in 8.x when `useNativeInit: false` is combined with `options` in the Expo plugin config.


antonis:
@sentry review

antonis:
@cursor review

alwx (approved):
Looks good, thanks!

PR #5982, opened 2026-04-13
feat(core): Name navigation spans using dispatched action payload

### LINKED ISSUE #4429: [ReactNavigation] Name navigation spans using dispatched action data
### Description

The ReactNavigationIntegration doesn't read dispatch action data and starts generic `Route Changed` navigation span on every dispatch.

Depending on the action type the new route name can be specified in the data. For example navigation action has `name` of the new route in it's payload. This might not be the final route as that depends on the router processing the action.

Using the data from the dispatched action will reduce the number of `Route Changed` spans. We should still update the name once the router processed the action with the final route name.

---

[current integration implementation](https://github.com/getsentry/sentry-react-native/blob/f2d10b7bd12e10b57209f3a1651d1f37b33e08b1/packages/core/src/js/tracing/reactnavigation.ts#L160)
[react-navigation onDispatch implementation](https://github.com/react-navigation/react-navigation/blob/bddcc44ab0e0ad5630f7ee0feb69496412a00217/packages/core/src/BaseNavigationContainer.tsx#L236) 

---

Related Issue: not all actions are expected to trigger navigation

alwx:
- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

When `useDispatchedActionData` is enabled, the `ReactNavigationIntegration` now extracts the route name from the dispatched action's `payload.name` and uses it to name the navigation span immediately at dispatch time, instead of starting with the generic "Route Change" name.

Fixes #4429

- New unit tests are actually 90% of this PR :D

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/src/js/tracing/reactnavigation.ts
+++ b/packages/core/src/js/tracing/reactnavigation.ts
@@ -379,7 +379,7 @@ export const reactNavigationIntegration = ({
 
     // Extract route name from dispatch action payload when available
     const dispatchedRouteName = useDispatchedActionData ? getRouteNameFromAction(event) : undefined;
-    if (useDispatchedActionData && !dispatchedRouteName && !isAppRestart) {
+    if (useDispatchedActionData && event && !dispatchedRouteName && !isAppRestart) {
       debug.log(`${INTEGRATION_NAME} Navigation action has no route name in payload, not starting navigation span.`);
       return;
     }
@@ -409,12 +409,8 @@ export const reactNavigationIntegration = ({
     }
     // Always discard transactions that never receive route information
     const spanToCheck = latestNavigationSpan;
-    ignoreEmptyRouteChangeTransactions(
-      getClient(),
-      spanToCheck,
-      DEFAULT_NAVIGATION_SPAN_NAME,
-      () => latestNavigationSpan === spanToCheck,
-    );
+    const spanName = finalSpanOptions.name ?? DEFAULT_NAVIGATION_SPAN_NAME;
+    ignoreEmptyRouteChangeTransactions(getClient(), spanToCheck, spanName, () => latestNavigationSpan === spanToCheck);
 
     if (enableTimeToInitialDisplay && latestNavigationSpan) {
       NATIVE.setActiveSpanId(latestNavigationSpan.spanContext().spanId);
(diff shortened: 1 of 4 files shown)


packages/core/src/js/tracing/reactnavigation.ts
```diff
@@ -376,18 +377,31 @@ export const reactNavigationIntegration = ({
       return;
     }
 
+    // Extract route name from dispatch action payload when available
+    const dispatchedRouteName = useDispatchedActionData ? getRouteNameFromAction(event) : undefined;
+    if (useDispatchedActionData && !dispatchedRouteName && !isAppRestart) {
+      debug.log(`${INTEGRATION_NAME} Navigation action has no route name in payload, not starting navigation span.`);
+      return;
+    }
```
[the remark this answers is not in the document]
alwx: ok, that's reasonable

packages/core/test/tracing/reactnavigationutils.ts
```diff
@@ -95,6 +121,34 @@ export function createMockNavigationAndAttachTo(sut: ReturnType<typeof reactNavi
...
+    },
+    emitGoBackWithStateChange: () => {
+      mockedNavigationContained.listeners['__unsafe_action__'](goBackAction);
+      mockedNavigationContained.currentRoute = {
+        key: 'previous_screen',
+        name: 'Previous Screen',
+      };
+      mockedNavigationContained.listeners['state']({});
+    },
+    emitGoBackWithoutStateChange: () => {
```
antonis: nit: I think this is never used

antonis (approved):
LGTM 🚀 
Added a minor nit comment

packages/core/test/tracing/reactnavigationutils.ts
lucas-zimerman: Q: missing tests for this call? should we remove it since it's unused?

lucas-zimerman:
might be a nice idea to default it as true on the next major

antonis:
> might be a nice idea to default it as true on the next major

Good idea 👍 Let's open an issue for that if you agree Alex

alwx:
@lucas-zimerman @antonis https://github.com/getsentry/sentry-react-native/issues/5986

PR #5992, opened 2026-04-14
feat(core): Add rage tap detection with ui.frustration breadcrumbs

alwx:
- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Detects rage taps (rapid consecutive taps on the same UI element) and surfaces them as first-class frustration signals across the SDK.

### Design decisions

- **Component identity over coordinates** — taps are matched by label or component name + file rather than screen coordinates. More robust for shifting layouts and list items, and reuses data already captured by `TouchEventBoundary`.
- **Breadcrumb-first** — emits `ui.frustration` breadcrumbs.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes


alwx:
@cursor review

packages/core/android/src/main/java/io/sentry/react/RNSentryReplayBreadcrumbConverter.java
```diff
@@ -32,6 +32,9 @@ public final class RNSentryReplayBreadcrumbConverter extends DefaultReplayBreadc
     if ("touch".equals(breadcrumb.getCategory())) {
       return convertTouchBreadcrumb(breadcrumb);
     }
+    if ("ui.frustration".equals(breadcrumb.getCategory())) {
```
antonis: I think we should either align with the [ui.multiClick naming from JS](https://github.com/search?q=repo%3Agetsentry%2Fsentry-javascript%20ui.multiClick&type=code) (which probably doesn't make sense on mobile) or [update the backend to handle the new category name](https://github.com/search?q=repo%3Agetsentry%2Fsentry%20ui.multiClick&type=code).
As is the frustration will probably appear as a generic breadcrumb on the replay timeline but without the special rage click treatment (fire icon, "Rage Click" label, click count display).

Looping in @romtsn who has an overview of all the mobile replay implementation for more context 🙇 

Also linking the related docs I could find:
* Rage Click Issues (product docs): [[Rage Click Issues](https://docs.sentry.io/product/issues/issue-details/replay-issues/rage-clicks/)]
* Replay Issue Types (JavaScript SDK docs): [[Replay Issues](https://docs.sentry.io/platforms/javascript/session-replay/issue-types/)]
* Intro blog post on Rage & Dead Clicks: [[Rage & Dead Clicks](https://blog.sentry.io/introducing-rage-and-dead-click-detection-for-session-replay/)]

packages/core/src/js/ragetap.ts
```diff
@@ -0,0 +1,118 @@
...
+  identity: string;
+  timestamp: number;
+}
+
+export interface TouchedComponentInfo {
+  name?: string;
+  label?: string;
+  element?: string;
+  file?: string;
+}
```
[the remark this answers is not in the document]
antonis: Looks valid 👍

packages/core/android/src/main/java/io/sentry/react/RNSentryReplayBreadcrumbConverter.java
alwx: @romtsn do you have any thoughts on this?

romtsn: Yes, I think we should follow already existing conventions (that is, `ui.multiClick` and `ui.slowClickDetected`) as well as the timeout thresholds and the entire breadcrumb structure from [here](https://github.com/getsentry/sentry-javascript/blob/fa9fea2cf7babee0afad8ada19da80426616c662/packages/replay-internal/src/coreHandlers/handleClick.ts#L216-L270). We could then only change the frontend part to say "tap" instead of "click", for example here: https://github.com/getsentry/sentry/blob/8c43a8eb55c4b134d12376102d9338fab3c23166/static/app/utils/replays/getFrameDetails.tsx#L143-L155

Perhaps, we'll also have to change how selectors are [being stringified](https://github.com/getsentry/sentry/blob/8c43a8eb55c4b134d12376102d9338fab3c23166/static/app/utils/replays/getFrameDetails.tsx#L490-L504), but uncertain if it's necessary yet.

alwx: After checking closely:
The `stringifyNodeAttributes` in the Sentry frontend already handles our node shape correctly: It already special-cases `data-sentry-component` meaning it uses it as the display name and hides it from the attribute list. So mobile components will show as `SubmitButton[sentry-label="Submit"]` on the replay timeline.

The `tap` -> `click` change you proposed is reasonable but purely cosmetic — will draft a PR to the main repo's frontend later on. Here is what I did for now:
 - `ui.frustration`-> `ui.multiClick` across all files (JS, Android, iOS, tests, changelog, etc)
 - breadcrumb data reshaped to match web SDK: `clickCount`, `metric: true`, route, node with DOM-compatible shape

romtsn: Nice, I guess it's going to be a challenge to reshape the data from the native SDKs, but we could also adapt the frontend/backend part for it, if necessary

packages/core/src/js/ragetap.ts
```diff
@@ -0,0 +1,190 @@
...
+    const tapCount = this._detect(identity, now);
+
+    if (tapCount > 0) {
+      const message = buildTouchMessage(root, label);
+      const node = buildNodeFromTouchPath(root, label);
+
+      addBreadcrumb({
+        category: 'ui.multiClick',
+        type: 'default',
+        level: 'warning' as SeverityLevel,
```
antonis: Should we omit and default to `info` like [the web](https://github.com/getsentry/sentry-javascript/blob/develop/packages/replay-internal/src/coreHandlers/handleClick.ts#L254)?

packages/core/src/js/ragetap.ts
```diff
@@ -0,0 +1,190 @@
+import type { SeverityLevel } from '@sentry/core';
+import { addBreadcrumb, debug } from '@sentry/core';
+
+import { getCurrentReactNativeTracingIntegration } from './tracing/reactnativetracing';
+
+const DEFAULT_RAGE_TAP_THRESHOLD = 3;
+const DEFAULT_RAGE_TAP_TIME_WINDOW = 1000;
```
antonis: Q: Wdyt of increasing this to 7s [like the web](https://github.com/getsentry/sentry-javascript/blob/a87183e687797be2dcc30fd5f9d9af05be7d5815/packages/replay-internal/src/integration.ts#L125)?

packages/core/src/js/touchevents.tsx
```diff
@@ -203,6 +222,12 @@ class TouchEventBoundary extends React.Component<TouchEventBoundaryProps> {
     const label = touchPath.find(info => info.label)?.label;
     if (touchPath.length > 0) {
       this._logTouchEvent(touchPath, label);
+      this._rageTapDetector.updateOptions({
```
antonis: nit: We could move the `updateOptions` in initialization instead of calling on every tap

packages/core/src/js/touchevents.tsx
```diff
@@ -96,10 +110,15 @@ class TouchEventBoundary extends React.Component<TouchEventBoundaryProps> {
     breadcrumbType: DEFAULT_BREADCRUMB_TYPE,
     ignoreNames: [],
     maxComponentTreeSize: DEFAULT_MAX_COMPONENT_TREE_SIZE,
+    enableRageTapDetection: true,
+    rageTapThreshold: 3,
+    rageTapTimeWindow: 1000,
```
antonis: h: we should use the default from ragetap.ts and have a single source of truth

packages/core/android/src/main/java/io/sentry/react/RNSentryReplayBreadcrumbConverter.java
```diff
@@ -72,6 +75,22 @@ public final class RNSentryReplayBreadcrumbConverter extends DefaultReplayBreadc
     return rrWebBreadcrumb;
   }
 
+  @TestOnly
```
antonis: m: since it is called from `convert` we could remove the @TestOnly annotation

packages/core/src/js/ragetap.ts
```diff
@@ -0,0 +1,190 @@
...
+  public constructor(options?: Partial<RageTapDetectorOptions>) {
+    this._enabled = options?.enabled ?? true;
+    this._threshold = options?.threshold ?? DEFAULT_RAGE_TAP_THRESHOLD;
+    this._timeWindow = options?.timeWindow ?? DEFAULT_RAGE_TAP_TIME_WINDOW;
+  }
+
+  /**
+   * Update options at runtime (e.g. when React props change).
+   */
+  public updateOptions(options: Partial<RageTapDetectorOptions>): void {
```
antonis, line 49: l: it would be nice to test the `updateOptions` logic

packages/core/src/js/ragetap.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+  if (root.file) {
+    attributes['data-sentry-source-file'] = root.file;
+  }
+  if (label) {
+    attributes['sentry-label'] = label;
+  }
+
+  return {
+    id: 0,
```
antonis, line 177: q: why do we pass `0` here?

antonis (commented):
Thank you for your work on this @alwx and the changes 🙇 
Overall looks good :) I've left some comments and questions.

packages/core/test/ragetap.test.ts
antonis, line 1: I think the lint issues are just import ordering and should be fixable with `yarn fix`

packages/core/src/js/ragetap.ts
romtsn: i'm curious how this works for the case of app hanging (i.e. dead clicks) -- I'm guessing we won't be able to detect the following touches after the first one that triggered a hang -- the main thread would be occupied/congested, right?

lucas-zimerman:
No additional comments from my side than the one pointed by the other reviewers.

packages/core/src/js/ragetap.ts
alwx: Web SDK uses actual rrweb node IDs for DOM element mapping. On mobile we don't have rrweb node IDs so 0 is no more than just a placeholder (and `ReplayBaseDomFrameData` type requires `id: number`)

packages/core/src/js/ragetap.ts
alwx: @romtsn that's the correct assumption — rage taps (rapid taps that register) and dead taps (taps during a hang) are different signals. Dead tap detection would need native-side work.

alwx: @antonis the web SDK's 7s is more like "how long to wait before emitting the breadcrumb", not "how close clicks need to be". So the direct comparison of numbers doesn't seem applicable — in our case it's like a "rolling" time window.

packages/core/android/src/main/java/io/sentry/react/RNSentryReplayBreadcrumbConverter.java
```diff
@@ -72,6 +75,21 @@ public final class RNSentryReplayBreadcrumbConverter extends DefaultReplayBreadc
     return rrWebBreadcrumb;
   }
 
+  public @Nullable RRWebEvent convertMultiClickBreadcrumb(final @NotNull Breadcrumb breadcrumb) {
+    if (breadcrumb.getData("path") == null) {
```
antonis: l: Should we also [check the type similar to iOS](https://github.com/getsentry/sentry-react-native/pull/5992/changes#diff-93882eebfc985075269015925655aacd860b6cfd4c10be52dad015afdd20d8c6R89)?

packages/core/src/js/ragetap.ts
antonis: Sounds good 👍 We should double check that this get's through on the backend

packages/core/src/js/ragetap.ts
antonis: Sounds good 👍 Let's add a comment explaining the reasoning behind the 1s window if possible 🙇

antonis (approved):
LGTM!
Iterated on the comments and [added one more q](https://github.com/getsentry/sentry-react-native/pull/5992#discussion_r3146071568) but nothing blocking on my side.

Issue #5303, opened 2025-10-23
using withSentry prevent sourcemap upload for expo update to work


wcastand:
### What React Native libraries do you use?

Expo Application Services (EAS)

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### @sentry/react-native SDK Version

~7.2.0

### How does your development environment look like?

it's `npx expo-env-info`

````
 expo-env-info 2.0.7 environment info:
    System:
      OS: macOS 26.0.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.16.0 - ~/.nvm/versions/node/v22.16.0/bin/node
      Yarn: 1.22.22 - ~/.nvm/versions/node/v22.16.0/bin/yarn
      npm: 10.9.2 - ~/.nvm/versions/node/v22.16.0/bin/npm
      Watchman: 2025.09.15.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.0, iOS 26.0, macOS 26.0, tvOS 26.0, visionOS 26.0, watchOS 26.0
    IDEs:
      Android Studio: 2025.1 AI-251.26094.121.2513.14007798
      Xcode: 26.0.1/17A400 - /usr/bin/xcodebuild
    npmPackages:
      expo: 54.0.19 => 54.0.19
      expo-router: ~6.0.13 => 6.0.13
      expo-updates: ~29.0.12 => 29.0.12
      react: 19.1.0 => 19.1.0
      react-dom: 19.1.0 => 19.1.0
      react-native: 0.81.5 => 0.81.5
    npmGlobalPackages:
      eas-cli: 16.24.1
    Expo Workflow: bare


````

### Sentry.init()

````js
Sentry.init({

	attachViewHierarchy: false,
	defaultIntegrations: false,
	dsn: "my stuff",
	enableAppHangTracking: true,
	enableLogs: true,
	enableNativeFramesTracking: !isRunningInExpoGo(),
	environment: Env.IS_DEV ? "development" : "production",
	integrations: [
		Sentry.functionToStringIntegration,
		Sentry.nativeLinkedErrorsIntegration,
		Sentry.httpContextIntegration,
		Sentry.dedupeIntegration(),
		Sentry.reactNativeErrorHandlersIntegration(),
		Sentry.hermesProfilingIntegration(),
	],
[...]

PR #6006, opened 2026-04-16
fix(core): Fix sourcemap upload when withSentry is used programmatically

antonis:
- [x] Bugfix

## :scroll: Description

When `withSentry` is used programmatically in `app.config.ts`, the Expo sourcemap upload script (`sentry-expo-upload-sourcemaps`) fails because it discovers Sentry config by looking for `@sentry/react-native/expo` in the `plugins` array of `expo config --json`. Programmatically-applied plugins don't appear in this array.

This PR fixes the issue by:

1. **Stashing build-time properties in `config._internal`** — The `withSentry` plugin now stores `organization`, `project`, and `url` in `config._internal.sentryBuildProperties` during config resolution. This field is included in `expo config --json` output but is stripped from the public app manifest (`isPublicConfig: true` deletes `_internal`), so no metadata leaks into production builds.

2. **Reading `_internal.sentryBuildProperties` in the upload script** — When the plugin isn't found in the `plugins` array, the script now checks `config._internal.sentryBuildProperties` from the same `expo config --json` response.

3. **Fallback to `sentry.properties` files** — As additional defense-in-depth, the script also tries reading `android/sentry.properties` or `ios/sentry.properties` (created during `expo prebuild`).

4. **Improved error message** — When no config source is found, the error now mentions the `SENTRY_ORG`/`SENTRY_PROJECT`/`SENTRY_URL` environment variable workaround.

### Resolution chain (in order of priority)
1. Environment variables (`SENTRY_ORG`, `SENTRY_PROJECT`, `SENTRY_URL`)
2. Declarative plugin in `config.plugins` array
3. `config._internal.sentryBuildProperties` (new — fixes the reported issue)
4. `sentry.properties` files from prebuild (new — defense-in-depth)
5. Error with actionable guidance

Follow up to https://github.com/getsentry/sentry-react-native/issues/5303#issuecomment-4239823688

Users following the `withSentry` documentation for `app.config.ts` hit this error when uploading sourcemaps for EAS Update:
```
Could not fetch '@sentry/react-native' plugin properties from expo config.
```

The root cause is that `createRunOncePlugin` (used by `withSentry`) applies config modifications through Expo's mod system but doesn't add the plugin to `config.plugins`. The `expo config --json` output therefore lacks the plugin entry, and the upload script can't find it.

- Added 6 new tests covering `_internal.sentryBuildProperties` fallback, `sentry.properties` fallback, priority ordering, and error messaging
- End-to-end verified locally: created a minimal `app.config.js` using programmatic `withSentry`, confirmed `expo config --json` includes `_internal.sentryBuildProperties` while `plugins` array does not contain the sentry plugin
- Verified `_internal` is stripped from the public config (app manifest) by Expo's `getConfig({ isPublicConfig: true })`, so no metadata ships in the production app
- All 196 existing SDK tests pass
- Lint passes

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

### DIFF
--- a/packages/core/scripts/expo-upload-sourcemaps.js
+++ b/packages/core/scripts/expo-upload-sourcemaps.js
@@ -21,11 +21,7 @@ function getSentryPluginPropertiesFromExpoConfig() {
       throw result.error || new Error(`expo config exited with status ${result.status}`);
     }
     const config = JSON.parse(result.stdout);
-    const plugins = config.plugins;
-    if (!plugins) {
-      return null;
-    }
-
+    const plugins = config.plugins || [];
     const sentryPlugin = plugins.find(plugin => {
       if (!Array.isArray(plugin) || plugin.length < 2) {
         return false;
--- a/packages/core/scripts/expo-upload-sourcemaps.js
+++ b/packages/core/scripts/expo-upload-sourcemaps.js
@@ -34,15 +34,65 @@ function getSentryPluginPropertiesFromExpoConfig() {
       return pluginName === '@sentry/react-native/expo';
     });
 
-    if (!sentryPlugin) {
-      return null;
+    if (sentryPlugin) {
+      const [, pluginConfig] = sentryPlugin;
+      return pluginConfig;
     }
-    const [, pluginConfig] = sentryPlugin;
-    return pluginConfig;
+
+    // When withSentry is used programmatically in app.config.ts, the plugin
+    // doesn't appear in the plugins array. Check config._internal where the
+    // plugin stashes build-time properties as a fallback.
+    if (config._internal?.sentryBuildProperties) {
+      return config._internal.sentryBuildProperties;
+    }
+
+    return null;
   } catch (error) {
     console.error('Error fetching expo config:', error);
     return null;
   }
+}
+
+function getSentryPropertiesFromFile() {
+  const candidates = [
+    path.join(projectRoot, 'android', 'sentry.properties'),
+    path.join(projectRoot, 'ios', 'sentry.properties'),
+  ];
+  for (const candidate of candidates) {
+    if (!fs.existsSync(candidate)) {
+      continue;
+    }
+    try {
+      const content = fs.readFileSync(candidate, 'utf8');
+      const props = {};
+      for (const line of content.split('\n')) {
+        const trimmed = line.trim();
+        if (!trimmed || trimmed.startsWith('#')) {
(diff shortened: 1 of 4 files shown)


antonis:
@sentry review

antonis:
@cursor review

packages/core/scripts/expo-upload-sourcemaps.js
```diff
@@ -34,17 +34,67 @@ function getSentryPluginPropertiesFromExpoConfig() {
       return pluginName === '@sentry/react-native/expo';
     });
 
-    if (!sentryPlugin) {
-      return null;
+    if (sentryPlugin) {
```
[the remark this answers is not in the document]
antonis: Good catch! Fixed in https://github.com/getsentry/sentry-react-native/commit/5083219cba57ee39f377256dc01b208bb05da414 — changed if (!plugins) { return null; } to const plugins = config.plugins || [] so the _internal fallback is always reachable. Added a test for this exact case (no plugins array, only _internal.sentryBuildProperties).

packages/core/scripts/expo-upload-sourcemaps.js
```diff
@@ -34,17 +30,67 @@ function getSentryPluginPropertiesFromExpoConfig() {
...
-    return pluginConfig;
+
+    return null;
   } catch (error) {
     console.error('Error fetching expo config:', error);
     return null;
   }
 }
 
+function getSentryPropertiesFromFile() {
```
alwx, line 52: this seems to be a more general function to read .properties — I think it could go to some sort of utils, for example the code to write properties already exists in `packages/core/plugin/src/utils.ts` so maybe it should be there as well? 
but also feel free to skip it since it doesn't seem to be very related to the purpose of this PR

alwx (approved):
only one small comment regarding the code to read properties — it feels like we could create one file with everything related to reading/writing properties

but again, it's not really important to do that as part of this PR

packages/core/scripts/expo-upload-sourcemaps.js
antonis: Considered this but the upload script is a standalone .js file that runs directly with Node. I'll keep a note to recheck if we can extract a utils for the scripts under https://github.com/getsentry/sentry-react-native/tree/main/packages/core/scripts

PR #5611, opened 2026-02-03
iOS UI Profiling

alwx:
## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Added missing support for UI profiling on iOS.

## :bulb: Motivation and Context
Fixes https://github.com/getsentry/sentry-react-native/issues/5517

## :pencil: Checklist

- [x] I added tests to verify changes — **or rather updated the existing Android tests**
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Merge it and enjoy!

PR #5582, opened 2026-01-26
feat: v8: Capture app start errors before JS

antonis:
## :loudspeaker: Type of change
<!--- Put an `x` in the boxes that apply -->
- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
<!--- Describe your changes in detail -->
Follow up to https://github.com/getsentry/sentry-react-native/pull/4472 (v6) > https://github.com/getsentry/sentry-react-native/pull/5470 (v7) targeting v8 https://github.com/getsentry/sentry-react-native/pull/5501

- Capture App Start errors and crashes by initializing Sentry from `sentry.options.json` ([#4472](https://github.com/getsentry/sentry-react-native/pull/4472))

  Create `sentry.options.json` in the React Native project root and set options the same as you currently have in `Sentry.init` in JS.

  ```json
  {
      "dsn": "https://key@example.io/value",
  }
  ```

  Initialize Sentry on the native layers by newly provided native methods.

  ```kotlin
  import io.sentry.react.RNSentrySDK

  class MainApplication : Application(), ReactApplication {
      override fun onCreate() {
          super.onCreate()
          RNSentrySDK.init(this)
      }
  }
  ```

  ```obj-c
  #import <RNSentry/RNSentry.h>

  @implementation AppDelegate
  - (BOOL)application:(UIApplication *)application
      didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
  {
      [RNSentrySDK start];
      return [super application:application didFinishLaunchingWithOptions:launchOptions];
  }
  @end
  ```

- Add RNSentrySDK APIs support to @sentry/react-native/expo plugin ([#4633](https://github.com/getsentry/sentry-react-native/pull/4633))
  - Adds `useNativeInit` option to automatically initialize Sentry natively before JavaScript loads, enabling capture of app start errors
  ```json
  {
    "expo": {
      "plugins": [
        [
          "@sentry/react-native/expo",
          {
            "useNativeInit": true
          }
        ]
      ]
    }
  }
  ```

### Changes

- Load `optionsFile` into the JS bundle during Metro bundle process ([#4476](https://github.com/getsentry/sentry-react-native/pull/4476))
- Add experimental version of `startWithConfigureOptions` for Apple platforms ([#4444](https://github.com/getsentry/sentry-react-native/pull/4444))
[...]

PR #6012, opened 2026-04-17
fix(profiling): iOS UI profiling on v8

antonis:
- [x] Bugfix

## :scroll: Description

`RNSentryStart.createOptionsWithDictionary` — the live iOS init path reached from `Sentry.init()` via `initNativeSdk` — does not read `_experiments.profilingOptions`. The JS wrapper packs `profilingOptions` into `_experiments.profilingOptions` (`wrapper.ts:306-312`) and sends it across the bridge, but on iOS the dictionary is handed to Sentry Cocoa without the `configureProfiling` callback being installed. `profileSessionSampleRate`, `lifecycle`, and `startOnAppStart` are silently dropped.

The handling exists in `SentrySDKWrapper.createOptionsWithDictionary` (ios/SentrySDKWrapper.m:103-116), but that method hasn't been part of `initNativeSdk`'s call path since #5582 moved init to `RNSentryStart`. #5611 then added the profiling block to `SentrySDKWrapper` (the already-bypassed file) a week later, so the feature has never reached Cocoa through the JS init path in any v8 release.

This PR ports the `_experiments.profilingOptions` handling to `RNSentryStart.createOptionsWithDictionary` and adds XCTest coverage — including one test against `startWithOptions` so the full entry point used by `initNativeSdk` stays pinned. The positive tests invoke the installed callback on a probe `SentryProfileOptions` and assert `sessionSampleRate`, `lifecycle`, and `profileAppStarts` all land correctly, so a future incomplete callback regresses loudly.

Scope is intentionally narrow. The same dead `SentrySDKWrapper` options surface also contains `_experiments.enableUnhandledCPPExceptionsV2` handling that is unreached in the same way — that should be ported in a follow-up. (Top-level `enableLogs` in `SentrySDKWrapper` is redundant, not broken: Cocoa's `SentryOptionsInternal.m:96` already maps it from the raw dictionary.) Cleaning up the dead `SentrySDKWrapper.createOptionsWithDictionary` / `setupWithDictionary` (and their tests) is also a good follow-up — the reason this bug shipped is precisely that those dead-code tests kept passing.

iOS UI profiling is documented as available for React Native (https://docs.sentry.io/platforms/react-native/profiling/#ui-profiling-experimental), but no `profilingOptions` set through JS `Sentry.init({ _experiments: { profilingOptions: { ... } } })` actually takes effect on iOS in any 8.x release. Users cannot opt into UI profiling via the documented path.

- Added four XCTest cases under `#pragma mark - RNSentryStart Tests`:
  - `testStartWithDictionaryInstallsConfigureProfilingFromExperimentsProfilingOptions` — pins the full `startWithOptions` entry point used by `initNativeSdk`; invokes the installed callback on a probe and asserts all three fields land.
  - `testStartCreateOptionsWithDictionaryProfilingOptionsInstallsConfigureProfiling` — unit test for the ported block; invokes the callback on a probe and asserts all three fields.
  - `testStartCreateOptionsWithDictionaryProfilingOptionsMissingDoesNotInstallConfigureProfiling` + `…EmptyExperimentsDoesNotInstall…` — negative cases.
- All new tests are guarded by `#if SENTRY_TARGET_PROFILING_SUPPORTED` matching the implementation in `RNSentryExperimentalOptions.m`.
- Locally: `yarn build`, `yarn test` (210 passing), `yarn circularDepCheck`, `./scripts/clang-format.sh lint`, `yarn lint:android`, `yarn lint:kotlin` all clean. `yarn lint:oxlint` has 2 pre-existing errors in `sentryExpoNativeCheck.ts` (unrelated; present on main). Full Cocoa XCTest run relies on CI's simulator.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- Follow-up: port `_experiments.enableUnhandledCPPExceptionsV2` handling from `SentrySDKWrapper` to `RNSentryStart` (same dead-code issue). https://github.com/getsentry/sentry-react-native/pull/6014
- https://github.com/getsentry/sentry-react-native/issues/6016
- https://github.com/getsentry/sentry-react-native/issues/6015


antonis:
@sentry review

antonis (commented):
The linter issue should be fixed after https://github.com/getsentry/sentry-react-native/pull/6013

lucas-zimerman (approved):
LGTM!

Issue #2837, opened 2023-02-18
can't build my react native app with sentry


amineoutmal:
this is the error i got : 
```
* Where:
Script '/Users/administrateur/Desktop/application-reactnative/node_modules/@sentry/react-native/sentry.gradle' line: 158

* What went wrong:
Illegal null value provided in this collection: [--log-level, debug, --url, null, --auth-token, null, react-native, gradle, --bundle, /Users/administrateur/Desktop/application-reactnative/android/app/build/generated/assets/react/release/index.android.bundle, --sourcemap, /Users/administrateur/Desktop/application-reactnative/android/app/build/generated/sourcemaps/react/release/index.android.bundle.map, --release, com.heyme.ep@1.4.5+9, --dist, 9, --org, null, --project, null]

* Try:
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Exception is:
java.lang.IllegalArgumentException: Illegal null value provided in this collection: [--log-level, debug, --url, null, --auth-token, null, react-native, gradle, --bundle, /Users/administrateur/Desktop/application-reactnative/android/app/build/generated/assets/react/release/index.android.bundle, --sourcemap, /Users/administrateur/Desktop/application-reactnative/android/app/build/generated/sourcemaps/react/release/index.android.bundle.map, --release, com.heyme.ep@1.4.5+9, --dist, 9, --org, null, --project, null]
        at org.gradle.util.internal.GUtil.addToCollection(GUtil.java:173)
        at org.gradle.process.internal.ProcessArgumentsSpec.args(ProcessArgumentsSpec.java:88)
        at org.gradle.process.internal.ProcessArgumentsSpec.setArgs(ProcessArgumentsSpec.java:100)
        at org.gradle.process.internal.ProcessArgumentsSpec.commandLine(ProcessArgumentsSpec.java:61)
        at org.gradle.process.internal.ProcessArgumentsSpec.commandLine(ProcessArgumentsSpec.java:54)
        at org.gradle.process.internal.DefaultExecSpec.commandLine(DefaultExecSpec.java:71)
        at org.gradle.api.tasks.AbstractExecTask.commandLine(AbstractExecTask.java:81)
        at org.gradle.api.tasks.AbstractExecTask.commandLine(AbstractExecTask.java:46)
        at jdk.internal.reflect.GeneratedMethodAccessor357.invoke(Unknown Source)
[...]

PR #6031, opened 2026-04-22
fix(android): Report missing properties file or keys when flavorAware is enabled

### LINKED ISSUE #2839: Provide missing path or properties value when using flavors on Android
### Description

Most of the issues when using flavors on Android are regarding reading the correct properties file. This is difficult to debug as the error user gets looks like [this](https://github.com/getsentry/sentry-react-native/issues/2837).

We can improve this by reporting what file we expected or what property was missing in the file we used.

antonis:
- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

When `flavorAware` is set in `project.ext.sentryCli` and the per-variant properties file is missing, or required keys are absent, `sentry.gradle` previously swallowed the `FileNotFoundException` (logging at `info`) and passed `null` values to sentry-cli. Gradle then failed with an opaque:

```
Illegal null value provided in this collection: [..., --url, null, --auth-token, null, ..., --org, null, --project, null]
```

This PR makes the error actionable:

- Throws a `GradleException` naming the expected `sentry-${variant}.properties` path when the file is missing (flavor-aware only).
- Collects missing required keys (`auth.token` / `SENTRY_AUTH_TOKEN`, `defaults.org`, `defaults.project`) and throws a single error listing them all alongside the file path.
- Only appends `--url` when `defaults.url` is set, so sentry-cli falls back to its default instead of being handed `null`.
- Non-flavor-aware path (the `SENTRY_PROPERTIES` env flow) is unchanged.

Users hitting this flow today (see #2837 as an example of the cryptic surface error) have no indication of which file was expected or which key was missing. This change surfaces both directly.

Fixes #2839.

**Parse check** — `./gradlew :app:help` on `samples/react-native/android` with the updated `sentry.gradle` symlinked via the yarn workspace: builds the task graph without errors.

**Validation harness** — built a standalone Gradle project that copies the exact validation block (same message strings, same control flow) and drives 8 scenarios. All pass:

| # | Scenario | Expectation | Result |
|---|----------|-------------|--------|
| A | `flavorAware=true`, properties file missing | `GradleException` naming the expected path and `flavorAware` | PASS |
| B | `flavorAware=true`, file present, only `defaults.org` missing | `GradleException` reporting only that key, with file path | PASS |
| C | `flavorAware=true`, file fully populated | Args contain `--url`, `--auth-token`, `--org`, `--project` with no nulls | PASS |
| D | `flavorAware=true`, `auth.token` absent but `SENTRY_AUTH_TOKEN` set | Succeeds; `--auth-token` uses env value | PASS |
| E | `flavorAware=true`, all three required keys missing | Single `GradleException` listing all three keys | PASS |
| F | `flavorAware=true`, `defaults.url` empty string | `--url` flag omitted; required args still present | PASS |
| G | `flavorAware=false`, properties file missing | No exception (existing silent behavior preserved) | PASS |
| H | `flavorAware=false`, file present | No flavor-specific args added | PASS |

**Not run locally**: a full `:app:bundleRelease` against real Sentry infrastructure. The change is isolated to the `doLast` block inside the `_SentryUpload_` task, and the scenarios above cover every branch touched. No Gradle/Groovy test harness exists in this repo to check these in automatically.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 ### Fixes
 
 - Check `captureReplay` return value in iOS bridge to avoid linking error events to uncaptured replays ([#6008](https://github.com/getsentry/sentry-react-native/pull/6008))
+- Report the expected properties file path and any missing keys when using `flavorAware` on Android, instead of failing with an opaque `Illegal null value provided in this collection` error ([#2839](https://github.com/getsentry/sentry-react-native/issues/2839))
 
 ### Dependencies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 ### Fixes
 
 - Check `captureReplay` return value in iOS bridge to avoid linking error events to uncaptured replays ([#6008](https://github.com/getsentry/sentry-react-native/pull/6008))
-- Report the expected properties file path and any missing keys when using `flavorAware` on Android, instead of failing with an opaque `Illegal null value provided in this collection` error ([#2839](https://github.com/getsentry/sentry-react-native/issues/2839))
+- Report the expected properties file path and any missing keys when using `flavorAware` on Android, instead of failing with an opaque `Illegal null value provided in this collection` error ([#6031](https://github.com/getsentry/sentry-react-native/pull/6031))
 
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


antonis:
@sentry review

antonis:
@cursor review

alwx (approved):

Issue #6016, opened 2026-04-17
Add Android JUnit coverage for configureAndroidProfiling


antonis:
## Context

`packages/core/android/src/main/java/io/sentry/react/RNSentryStart.java:186` calls `configureAndroidProfiling(options, rnOptions, logger)`, which reads `_experiments.profilingOptions` and wires three fields onto `SentryAndroidOptions`:

- `profileSessionSampleRate` → `options.setProfileSessionSampleRate(...)` (line 241)
- `lifecycle` → `options.setProfileLifecycle(ProfileLifecycle.MANUAL|TRACE)` (line 258 / 261)
- `startOnAppStart` → `options.setStartProfilerOnAppStart(...)` (line 274)

This works correctly today. **It has no test coverage.** `packages/core/android/src/test/` doesn't exist — the Android side of the SDK ships no JUnit tests for the init path at all. The audit run while investigating the iOS profiling bug (#6012) confirmed this gap.

## Why this matters

We just finished fixing two iOS bugs (#6012, #6014) where option-parsing code was added to a file that was no longer on the live init path, and XCTest kept passing because it was exercising the dead surface. On Android today, the active init path *is* correct — but if someone refactors `configureAndroidProfiling` out of `RNSentryStart.startWithOptions`, or introduces a parallel init path (as happened on iOS), there is literally no test to catch it. We'd find out from a customer report.

Pinning the wiring with tests would have caught the analogous iOS bug.

## What to do

1. Set up the Android test harness under `packages/core/android/src/test/java/io/sentry/react/` with a Robolectric or plain JUnit + Mockito setup (whichever aligns with the rest of the Sentry Android SDK's test conventions — check `sentry-java` for the pattern).

2. Wire the test target into `packages/core/android/build.gradle` so `./gradlew test` runs it.

3. Add `RNSentryStartTest` (or similar) with at least these cases, mirroring the iOS tests added in #6012:
   - Passing `_experiments.profilingOptions = { profileSessionSampleRate: 1.0, lifecycle: "trace", startOnAppStart: true }` results in `options.getProfileSessionSampleRate() == 1.0`, `options.getProfileLifecycle() == ProfileLifecycle.TRACE`, `options.isStartProfilerOnAppStart() == true`.
   - `lifecycle: "manual"` → `ProfileLifecycle.MANUAL`.
   - No `_experiments` / no `profilingOptions` → defaults preserved, no calls made.
[...]

PR #4442, opened 2025-01-14
ref(ios): Extract Cocoa SDK init into standalone file

krystofwoldrich:
## :loudspeaker: Type of change

- [x] Refactoring

## :scroll: Description

This PR doesn't change any of the initialization logic, only moves the code from the main module file to a standalone class for future changes.

## :bulb: Motivation and Context

- this PR is required by https://github.com/getsentry/sentry-react-native/pull/4444

## :green_heart: How did you test it?
sample app

## :pencil: Checklist

- [ ] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

PR #6059, opened 2026-04-28
refactor(ios): remove dead SentrySDKWrapper init surface

### LINKED ISSUE #6015: Delete abandoned SentrySDKWrapper init surface and migrate tests to RNSentryStart
## Context

When the iOS init path was moved from `SentrySDKWrapper` to `RNSentryStart` in v8 (via #5582 / sub-PR #4442), only the call sites were repointed — the old option-parsing methods in `SentrySDKWrapper` were left in place "just in case," along with their XCTest suite. Nothing in production calls them anymore, but the tests kept passing, which is precisely how two bugs shipped in every 8.x release until we noticed:

- `_experiments.profilingOptions` silently dropped on iOS — fixed in #6012
- `_experiments.enableUnhandledCPPExceptionsV2` silently dropped on iOS — fixed in #6014

The dead code also parses `enableLogs` (harmless — Cocoa's dict parser handles it), `enableNativeCrashHandling`, `spotlight`, etc. — all duplicated in `RNSentryStart` today.

## Why this matters

As long as `SentrySDKWrapper.createOptionsWithDictionary` and its test coverage exist, any future author who wants to add a new option or `_experiments.*` field has two visually-equivalent files to update, no signal about which one is live, and XCTest runs that pass even when the live path is broken. That's exactly the setup that produced the two bugs above — the #5611 author added the profiling block to `SentrySDKWrapper` because it looked like the right file.

## What to do

1. In `packages/core/ios/SentrySDKWrapper.{h,m}`, delete:
   - `createOptionsWithDictionary:isSessionReplayEnabled:error:`
   - `setupWithDictionary:isSessionReplayEnabled:error:`
   - `startWithOptions:` (the internal one that the above methods call)

   Keep the methods that are still used from `RNSentry.mm`: `crash`, `close`, `crashedLastRun`, `configureScope:`, `debug`, `releaseName`, `enableAutoSessionTracking`, `enableWatchdogTerminationTracking`. (Grep `SentrySDKWrapper` in `RNSentry.mm` for the exact list; it's ~10 call sites.)

2. In `packages/core/RNSentryCocoaTester/RNSentryCocoaTesterTests/RNSentryTests.m`, delete the `[SentrySDKWrapper createOptionsWithDictionary:…]` tests (~lines 34–1065 in the current file). For any assertion in there that covers behavior still worth testing but *not* already covered by the `RNSentryStart Tests` block (e.g., the `ignoreErrors` filtering tests), migrate the test to use `RNSentryStart createOptionsWithDictionary:` / `startWithOptions:` instead.

3. Keep the remaining `RNSentryStart`-based tests as the single source of truth for init-path behavior.

## Acceptance

- `grep -r 'SentrySDKWrapper createOptionsWithDictionary\|SentrySDKWrapper setupWithDictionary' packages/core` returns nothing.
- `xcodebuild test` (Cocoa suite) stays green.
- No orphaned `#import "SentrySDKWrapper.h"` in the tests if nothing else from the header is used.

## Notes

- Do this **after** #6012 and #6014 land — the cleanup will have trivial merge conflicts with them otherwise.
- Not user-visible. Purely an anti-footgun cleanup.

antonis:
- [x] Refactoring

## :scroll: Description

`SentrySDKWrapper.createOptionsWithDictionary:isSessionReplayEnabled:error:` and friends became unreachable from production when v8 moved init to `RNSentryStart` in #5582 (sub-PR #4442), but were left in place along with their parallel XCTest suite. Those dead-code tests are the reason two bugs shipped in every 8.x release until we noticed them — #5611's author added profiling parsing to `SentrySDKWrapper` because the tests there made it look like the live surface, and the tests kept passing while the live `RNSentryStart` path silently dropped the option (#6012, #6014).

This PR removes the dead surface so the same mistake can't recur:

**`packages/core/ios/SentrySDKWrapper.{h,m}`** — delete:
- `+ createOptionsWithDictionary:isSessionReplayEnabled:error:`
- `+ setupWithDictionary:isSessionReplayEnabled:error:`
- `+ startWithOptions:` (only callsite was the deleted `setupWithDictionary:`)
- `+ enableAutoSessionTracking` (zero callers in `packages/core/`)
- `+ enableWatchdogTerminationTracking` (zero callers in `packages/core/`)
- the now-orphaned `getURLFromDSN:` helper and unused imports

**Kept** (still called 18 times from `RNSentry.mm`): `configureScope:`, `crash`, `close`, `crashedLastRun`, `debug`, `releaseName`.

**`packages/core/RNSentryCocoaTester/RNSentryCocoaTesterTests/RNSentryTests.m`** — 22 SentrySDKWrapper-based tests deleted (each had an existing `RNSentryStart`-based equivalent: spotlight, native crash handling, auto-perf tracing, breadcrumb filtering, CPP exceptions v2, `testRemovesPerformanceProperties`, etc.). 15 still-relevant tests migrated to `RNSentryStart`-based equivalents under the existing `#pragma mark - RNSentryStart Tests` block:
- `enableLogs` enabled / disabled (2)
- `beforeBreadcrumb` keeps-when-no-devServer edge case (1)
- `ignoreErrors` family (5)
- `beforeSend` JS-exception / CPP-exception filtering (2) — these now also call `[RNSentryStart updateWithReactFinals:]` so the full two-layer `beforeSend` chain is exercised, which is stronger coverage than the old tests (the wrapped outer filter wasn't in the old SentrySDKWrapper path).
- `enableSessionReplayInUnreliableEnvironment` Default / WithErrorSampleRate / WithSessionSampleRate / WithSessionSampleRates / Disabled (5) — drops the explicit `isSessionReplayEnabled:` parameter; `RNSentryStart.createOptionsWithDictionary:` computes it internally via `[RNSentryReplay updateOptions:]` on its mutable copy, driven by the same `replays*SampleRate` keys the fixtures already pass.

`#import "SentrySDKWrapper.h"` removed from the test file. Class `RNSentryInitNativeSdkTests` retained (could be renamed in a follow-up since it's no longer init-only, but that's bikeshed). Tests for `setEventOriginTag`, `fetchNativeStackFramesBy`, and user creation that don't touch `SentrySDKWrapper` are unmodified.

Net diff: −582 lines.

Closes #6015. Companion to #6012 and #6014, which fixed two real bugs caused by code being added to this dead surface. The cleanup removes the conditions that allowed those bugs to ship undetected — the tests on the dead path will no longer make the wrong file look "tested."

`SentrySDKWrapper.h` is **not** in `s.public_header_files` in `packages/core/RNSentry.podspec` (only `RNSentry.h`, `RNSentrySDK.h`, `RNSentryStart.h`, `RNSentryVersion.h`, `RNSentryBreadcrumb.h`, `RNSentryReplay.h`, `RNSentryReplayBreadcrumbConverter.h`, `Replay/RNSentryReplayMask.h`, `Replay/RNSentryReplayUnmask.h`, `RNSentryTimeToDisplay.h` are). The deleted methods are not exposed to downstream Swift/Obj-C consumers, so no API break.

- Searched the entire repo (samples, performance-tests, expo plugin, Swift test suites, sample apps' `AppDelegate.swift`) for any reference to the deleted methods or to `SentrySDKWrapper.setupWithDictionary` — zero hits.
- Grepped CHANGELOG / README / MIGRATION / CONTRIBUTING / CLAUDE.md for the deleted symbols — only matches are user-facing `SentryOptions.enableAutoSessionTracking` / `enableWatchdogTerminationTracking` documentation, which is unrelated to the deleted accessor wrappers.
- Verified the 22 deletions are duplicate-of-existing by spot-checking pairs (e.g., `testCreateOptionsWithDictionaryRemovesPerformanceProperties` ↔ `testStartWithDictionaryRemovesPerformanceProperties`): semantically equivalent, both pin the user-visible behavior.
- Verified the 15 migrations preserve coverage: replay tests still drive `RNSentryReplay.updateOptions:` via the same dict keys; `beforeSend` filter tests now exercise the full `prepareOptions` + `updateWithReactFinals` chain.
- Local: `yarn build`, `yarn test` (211 passing), `yarn circularDepCheck`, `./scripts/clang-format.sh lint`, `yarn lint` (full suite — oxlint, prettier, lerna, android, kotlin, clang, swift) all clean. Cocoa XCTest run is delegated to CI (local has a pre-existing fmt-pod / Xcode 22 compatibility issue unrelated to this change).

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- #6016


lucas-zimerman (approved):
LGTM!

Issue #6029, opened 2026-04-21
iOS 26: EXC_BAD_ACCESS crash in SentrySDKInternal startWithOptions when   _experiments is set


maaz-coros:
### What React Native libraries do you use?

Expo Application Services (EAS)

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### Are you using any other error monitoring solution alongside Sentry?

No

### @sentry/react-native SDK Version

8.8.0

### How does your development environment look like?

````
  react-native: 0.81.5
  expo: ~54.0.0                                                                 
  react: 19.1.0   
  OS: iPhone OS 26.3.1 (23D771330a)
  Device: iPhone 14 (iPhone14,5)   
  Distribution: TestFlight (EAS Build)


````

### Sentry.init()

````js
  Sentry.init({   
    dsn: process.env.EXPO_PUBLIC_SENTRY_DSN,                                    
    integrations: [], // mobileReplayIntegration excluded on iOS
    tracesSampleRate: 1.0,                                      
    _experiments: { profilesSampleRate: 1.0 },                                  
    enabled: true,                            
    debug: false,                                                               
  });
````

### Steps to Reproduce

 1. Build iOS using EAS Build with profile targeting iOS 26 device             
  2. Install via TestFlight on iPhone running iOS 26.3.1           
  3. Launch the app                                                             
  4. App crashes immediately before any screen renders  

### Expected Result

App launches successfully, Sentry initializes without crashing.

### Actual Result

  App crashes immediately on launch with EXC_BAD_ACCESS (SIGSEGV).
  KERN_INVALID_ADDRESS at 0x0000000000000010                                    
                                                                                
  Crash trace:
  0  +[SentrySDKInternal startWithOptions:] + 436 (SentrySDKInternal.m:236)     
  1  specialized static SentrySDK.start(options:) + 516 (SentrySDK.swift:105)
  2  +[RNSentryStart startWithOptions:] + 136 (RNSentryStart.m:29)              
  3  -[RNSentry initNativeSdk:resolve:rejecter:] + 112 (RNSentry.mm:148)        
                                                                                
  Session Replay is already disabled on iOS (mobileReplayIntegration not used). 
  Suspected trigger: _experiments: { profilesSampleRate: 1.0 }.                 
  Removing _experiments stops the crash.

antonis:
Thank you for reporting 🙇 We will investigate this and iterate back.

antonis:
[...]

PR #6081, opened 2026-05-04
fix(core): Trim DSN whitespace before native initialization

antonis:
## 📢 Type of change

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## 📜 Description

Trims leading/trailing whitespace from the DSN before passing it to the native SDK during initialization. If whitespace is detected, a warning is logged to inform the user.

## 💡 Motivation and Context

A user reported a crash (`EXC_BAD_ACCESS` in `SentrySDKInternal.startWithOptions`) caused by a stray space in their DSN string (getsentry/sentry-react-native#6029). The whitespace caused a null pointer dereference in the native Cocoa SDK during initialization.

See [https://github.com/getsentry/sentry-react-native/issues/6029#issuecomment-4291290711](<https://github.com/getsentry/sentry-react-native/issues/6029#issuecomment-4291290711>)

## 💚 How did you test it?

* Added test verifying whitespace is trimmed, warning is logged, and the trimmed DSN is passed to native init
* Updated existing test DSN values from `'test'` to a valid DSN format (`VALID_DSN` constant)
* All 197 wrapper tests pass

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## 🔮 Next steps

#skip-changelog is this is a minor change

### DIFF
--- a/packages/core/src/js/wrapper.ts
+++ b/packages/core/src/js/wrapper.ts
@@ -270,6 +270,12 @@ export const NATIVE: SentryNativeWrapper = {
       return false;
     }
 
+    const trimmedDsn = options.dsn.trim();
+    if (trimmedDsn !== options.dsn) {
+      debug.warn('The DSN contains whitespace which can cause native crashes. Whitespace has been trimmed.');
+      options.dsn = trimmedDsn;
+    }
+
     if (!this._isModuleLoaded(RNSentry)) {
       throw this._NativeClientError;
     }
--- a/packages/core/src/js/wrapper.ts
+++ b/packages/core/src/js/wrapper.ts
@@ -262,18 +262,17 @@ export const NATIVE: SentryNativeWrapper = {
       return false;
     }
 
+    if (options.dsn && options.dsn !== options.dsn.trim()) {
+      debug.warn('The DSN contains whitespace which can cause native crashes. Whitespace has been trimmed.');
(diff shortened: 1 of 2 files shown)


alwx (commented):
Should we actually fix that? I mean it seems more like a configuration issue made by customer, not an issue on our side.

antonis:
> Should we actually fix that? I mean it seems more like a configuration issue made by customer, not an issue on our side.

Good point @alwx 👍  The main issue is that the extra whitespace passes [the core validation](https://github.com/getsentry/sentry-javascript/blob/23ab0f1ad445cd25e8e435744d5b585a1ad251ed/packages/core/src/utils/dsn.ts#L83) (and works for the most part) but causes the native side to crash as with the specific report.
An alternative would be to just add a warning log to avoid this getting unnoticed. Wdyt?

antonis:
Regarding the bot feedback about whitespace-only DSNs (e.g. `"   "`):

After trimming, a whitespace-only DSN becomes `""` which is falsy in JS — so the existing `!options.dsn` check right below already catches it and disables the native SDK. No extra guard needed.

alwx:
Probably it's fine for now — was just thinking that the fix is a bit redundant. But doesn't hurt much, I guess.

alwx (approved):

PR #6106, opened 2026-05-07
feat(core): Extract text from children of touched components for breadcrumb labels

### LINKED ISSUE #6097: Extract text children from touchable components for breadcrumb labels
Walk `props.children` at runtime to extract text content from touched components. This is the only option that catches dynamic text like `<Text>{t('save')}</Text>`.

Rough approach:

* Walk `memoizedProps.children` up to depth 2
* If children are strings or `<Text>` with string children, use as label fallback (after sentry-label, accessibilityLabel, testID)
* Cap string length (e.g. 64 chars)

**Must be gated on replay masking settings.** If `maskAllText` is enabled, extracting text children into breadcrumbs would leak content the user asked us to mask. Options:

* Gate on masking setting (don't extract when masking is on)
* Require opt-in via a `TouchEventBoundary` prop (e.g. `extractTextChildren: true`, default off)
* Both

**Scope:** `touchevents.tsx` — child walk logic + masking gate + tests.

antonis:
- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Walks the React fiber tree downward (`child`/`sibling` pointers) to extract text content from children of touched components. This text is used as a **lowest-priority label fallback** when no explicit label (`sentry-label`, custom `labelKey`) is set.

For example, a button like:
```jsx
<TouchableOpacity onPress={save}>
  <Text>Save workout</Text>
</TouchableOpacity>
```
will now produce a breadcrumb with label `"Save workout"` instead of just `"TouchableOpacity"`.

### Masking behavior

- **When `maskAllText` is enabled (the default for replay users):** text extraction is automatically disabled to avoid leaking masked content via breadcrumbs. No behavior change for these users.
- **When `maskAllText` is explicitly `false`:** text extraction is enabled.
- **When MobileReplay integration is not installed:** text extraction is enabled. **This is a behavior change** — users without replay will now see text labels in their touch breadcrumbs where previously they only saw component names. This can surface new data in breadcrumbs (component text content). Users can opt out via `extractTextFromChildren: false` on `TouchEventBoundary`.
- **Per-view `Sentry.Mask` boundaries are respected:** text inside `<Sentry.Mask>` is never extracted.

### Safety limits

- Max text length: 64 characters (truncated with `...`)
- Max depth: 3 levels of child traversal
- Max siblings: 5 per level (prevents runaway traversal in long lists)

### New prop

`extractTextFromChildren` (default `true`) — set to `false` to opt out entirely.

### Label priority (lowest to highest)

1. Extracted text children ← **new**
2. Custom `labelKey` prop value
3. `sentry-label` prop

Touch breadcrumbs currently produce unhelpful labels like `"Touch event within element: View"` which makes Session Replay AI summaries and issue debugging less useful. This change automatically extracts meaningful labels from component text content.

Closes https://github.com/getsentry/sentry-react-native/issues/6097

- 11 new unit tests covering: basic extraction, nested children, siblings, truncation, masking gate (`maskAllText: true`, `false`, undefined), `Sentry.Mask` boundary, priority over extracted text, `extractTextFromChildren: false` opt-out, raw text fiber nodes, integration without options
- All 31 touchevents tests pass
- Type check passes (no new TS errors)
- Lint passes (0 warnings, 0 errors)

## :pencil: Checklist
- [x] I updated the docs if needed https://github.com/getsentry/sentry-docs/pull/17665
- [x] All tests passing
- [x] No breaking changes


antonis:
@sentry review

antonis:
@cursor review

packages/core/src/js/touchevents.tsx
```diff
@@ -391,6 +433,47 @@ function getSpanAttributes(currentInst: ElementInstance): Record<string, SpanAtt
...
+  depth: number,
+  siblingIndex: number = 0,
+): void {
+  if (!inst || depth > MAX_TEXT_EXTRACTION_DEPTH || siblingIndex >= MAX_SIBLINGS_TO_VISIT) {
+    return;
+  }
+
+  if (inst.elementType?.name === MASK_COMPONENT_NAME || inst.elementType?.displayName === MASK_COMPONENT_NAME) {
+    return;
+  }
```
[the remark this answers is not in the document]
antonis: Good catch! Fixed in 142085ae — we now recurse into `inst.sibling` before returning when hitting a Mask boundary. Added a test covering this case.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 ### Features
 
 - Use `accessibilityLabel`, `aria-label`, and `testID` as fallback labels for touch breadcrumbs when `sentry-label` is not set ([#6103](https://github.com/getsentry/sentry-react-native/pull/6103))
+- Extract text content from children of touched components as a label fallback for touch breadcrumbs ([#6106](https://github.com/getsentry/sentry-react-native/pull/6106))
```
antonis: Fixed with 92517ae3a5d38f98bdcc2bda7ba26776029fa2b8

packages/core/src/js/touchevents.tsx
```diff
@@ -410,6 +450,49 @@ function getSpanAttributes(currentInst: ElementInstance): Record<string, SpanAtt
...
+
+  const props = inst.memoizedProps;
+  if (typeof props === 'string') {
+    parts.push(props);
+  } else if (typeof props?.children === 'string') {
+    parts.push(props.children);
+  }
+
+  collectTextFromFiber(inst.child, parts, depth + 1, 0);
+  collectTextFromFiber(inst.sibling, parts, depth, siblingIndex + 1);
```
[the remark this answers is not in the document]
antonis: Valid catch. In production RN fiber trees, `<Text>Hello</Text>` has both `memoizedProps.children = 'Hello'` on the Text fiber and a child HostText fiber with `memoizedProps = 'Hello'`. Fixed in f14d1f93 — now we skip child recursion when `props.children` is already a string, and added a test modeling the real fiber structure.

packages/core/test/touchevents.test.tsx
```diff
@@ -903,4 +903,427 @@ describe('TouchEventBoundary._onTouchStart', () => {
...
+          },
+        },
+      };
+
+      // @ts-expect-error Calling private member
+      boundary._onTouchStart(event);
+
+      expect(addBreadcrumb).toHaveBeenCalledWith(
+        expect.objectContaining({
+          message: 'Touch event within element: Save workout',
```
lucas-zimerman: Test says `does not extract text when MobileReplay integration has no options property` but it seems like it extracted the text `save workout`?

packages/core/src/js/touchevents.tsx
```diff
@@ -280,6 +301,24 @@ class TouchEventBoundary extends React.Component<TouchEventBoundaryProps> {
...
+  private _shouldExtractText(): boolean {
+    if (!this.props.extractTextFromChildren) {
+      return false;
+    }
+    const client = getClient();
+    if (!client) {
+      return true;
+    }
+    const replayIntegration = client.getIntegrationByName(MOBILE_REPLAY_INTEGRATION_NAME);
+    if (replayIntegration && 'options' in replayIntegration) {
```
lucas-zimerman: What if options is not set here? does it mean it will always extract it?

packages/core/src/js/touchevents.tsx
```diff
@@ -70,6 +72,15 @@ export type TouchEventBoundaryProps = {
...
   rageTapTimeWindow?: number;
+  /**
+   * Extract text content from children of touched components as a label fallback.
+   * Automatically disabled when Session Replay's `maskAllText` is enabled (the default)
+   * to avoid leaking masked content via breadcrumbs. Set `maskAllText: false` in your
+   * `mobileReplayIntegration` config to enable text extraction.
+   * Per-view `Sentry.Mask` boundaries are also respected.
+   * Set to `false` to opt out of text extraction entirely.
+   */
+  extractTextFromChildren?: boolean;
```
lucas-zimerman, line 84: Might be verbose, but worth adding a `@default` mentioning to see the above description for the default value

lucas-zimerman:
Would it be possible to add a depth limit & sibling limit test?

packages/core/test/touchevents.test.tsx
antonis: Good catch Lucas 👍 Updated with 2c4692048b74874deb4536e695b37b4a9bb56754

packages/core/src/js/touchevents.tsx
antonis: Fixed with  2c4692048b74874deb4536e695b37b4a9bb56754

packages/core/src/js/touchevents.tsx
antonis: Added with 2c4692048b74874deb4536e695b37b4a9bb56754

antonis:
> Would it be possible to add a depth limit & sibling limit test?

Good idea 👍 Added with 6fdacc4b9f2bc8462ab09a21d08426f81eeab0cd

lucas-zimerman (approved):
LGTM!

PR #6130, opened 2026-05-11
fix(android): Handle boolean values in JSON options converter

### LINKED ISSUE #6122: Mobile Replay (Android, Fabric/New Architecture): maskAllText: false silently ignored — replays always fully masked
## Summary

Setting `maskAllText: false` (or any of `maskAllImages` / `maskAllVectors: false`) on `mobileReplayIntegration` has **no effect** when the React Native app uses the New Architecture (Fabric + TurboModules) on Android. Captured replays show all text / images / vectors fully masked regardless of the option value.

Confirmed on-device on `@sentry/react-native@7.2.0` (which pulls in `sentry-android-replay@8.21.1`).

## Hypothesis

The v7.5 changelog mentions the experimental Canvas Capture Strategy: *"If this strategy is used, all text and images will be masked, regardless of any masking configuration."* Our observation is consistent with the SDK selecting Canvas (not PixelCopy) when running on Fabric, which forces the all-mask behavior.

What's missing is either (a) an opt-in to PixelCopy on Fabric or (b) a Canvas implementation that honors `maskAllText: false`. Either would unblock the use case.

## Environment

- `@sentry/react-native`: 7.2.0
- `sentry-android-replay`: 8.21.1 (transitive)
- React Native: 0.81.5
- Fabric: **enabled** (`react_native_context.fabric: true` on captured events)
- TurboModules: enabled
- JS engine: Hermes
- Device: Pixel 7, Android 16
- App built with EAS, `expo-dev-client` runtime

## Reproduction

```ts
Sentry.init({
  dsn,
  enableNative: true,
  integrations: [
      maskAllText: false,
      maskAllImages: false,
      maskAllVectors: false,
    }),
  ],
});

// Then trigger an error:
Sentry.captureException(new Error("test"));
```

1. Build an APK with Fabric / New Architecture enabled.
2. Install on Android device, trigger the error.
3. Open the resulting replay in the Sentry dashboard.
4. **Expected**: text content visible in the captured frames.
5. **Actual**: all text masked.

## What we verified

1. The merged options reach the bundle correctly: `maskAllText: false` is present in the JS bundle Metro serves.
2. `RNSentryModuleImpl.java` (`packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java`) reads `mobileReplayOptions.maskAllText` and calls `androidReplayOptions.setMaskAllText(false)`.
3. `SentryReplayOptions.setMaskAllText(false)` (`io.sentry:sentry@8.21.1`) adds `android.widget.TextView` to `unmaskViewClasses` — verified against the published source.
4. `<Text>` in RN renders as `ReactTextView extends android.widget.TextView`, which would match the unmask rule via the `Class.superclass` walk in `ViewHierarchyNode.shouldMask` — but only if the view-hierarchy strategy is reading the actual native views (PixelCopy), not screen pixels (Canvas).

So the option flow is correct end-to-end on the SDK; the result mismatch points to the recording strategy.

## Asks

1. Is the Canvas strategy auto-selected on Fabric in 7.x / 8.x? Confirming intent helps us decide whether to upgrade or wait.
2. Is there a config knob to force PixelCopy on Fabric (Sentry RN init option, gradle property, or replay option)?
3. Is there a roadmap version where Canvas honors `maskAllText: false` or PixelCopy works on Fabric?

Happy to provide more diagnostics. Thanks!

antonis:
- [x] Bugfix

## :scroll: Description

`RNSentryJsonConverter.convertToWritable()` did not handle `Boolean` values when converting `sentry.options.json` to a React Native `ReadableMap`. Boolean values were silently dropped because the `instanceof` chain only checked for `Float`, `Double`, `Number`, `String`, `JSONObject`, `JSONArray`, and `null`.

This caused all boolean options to fall back to their defaults when using the `RNSentrySDK.init()` native-first initialization path.

Fixes #6122

When users set `maskAllText: false` (or other boolean replay options) in `sentry.options.json` and initialize via `RNSentrySDK.init()`, the `false` value was dropped. The SDK then treated the missing key as `true` (mask everything), making it impossible to disable text masking through the JSON config.

This affects **all boolean options** read from `sentry.options.json`, not just replay masking — including `debug`, `attachScreenshot`, `attachViewHierarchy`, `enableCaptureFailedRequests`, `spotlight`, etc. However, the impact is limited because most boolean options default to sensible values when missing, and the JS-side `Sentry.init()` options (which go through the React Native bridge, not the JSON converter) were never affected.

- Added boolean assertions to existing `RNSentryJsonConverterTest` for both map and array conversions
- Manually verified on Android emulator (API 36, Fabric/New Architecture enabled): replay screenshots showed unmasked text after the fix, confirming `maskAllText: false` is now properly applied

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 - Extract text content from children of touched components as a label fallback for touch breadcrumbs ([#6106](https://github.com/getsentry/sentry-react-native/pull/6106))
 
+### Fixes
+
+- Fix boolean options from `sentry.options.json` being ignored on Android when using `RNSentrySDK.init` ([#6130](https://github.com/getsentry/sentry-react-native/pull/6130))
+
 ### Dependencies
 
 - Bump JavaScript SDK from v10.51.0 to v10.52.0 ([#6108](https://github.com/getsentry/sentry-react-native/pull/6108))
--- a/packages/core/RNSentryAndroidTester/app/src/androidTest/java/io/sentry/react/RNSentryJsonConverterTest.kt
+++ b/packages/core/RNSentryAndroidTester/app/src/androidTest/java/io/sentry/react/RNSentryJsonConverterTest.kt
@@ -22,6 +22,8 @@ class RNSentryJsonConverterTest {
                 put("doubleKey", 12.3)
                 put("intKey", 123)
                 put("stringKey", "test")
+                put("boolTrueKey", true)
+                put("boolFalseKey", false)
                 put("nullKey", JSONObject.NULL)
             }
 
@@ -32,6 +34,8 @@ class RNSentryJsonConverterTest {
         assertEquals(12.3, result.getDouble("doubleKey"), 0.0)
         assertEquals(123, result.getInt("intKey"))
         assertEquals("test", result.getString("stringKey"))
+        assertEquals(true, result.getBoolean("boolTrueKey"))
+        assertEquals(false, result.getBoolean("boolFalseKey"))
         assertNull(result.getString("nullKey"))
     }
 
@@ -62,6 +66,8 @@ class RNSentryJsonConverterTest {
                 put(1)
                 put(2.5)
                 put("string")
+                put(true)
+                put(false)
                 put(JSONObject.NULL)
             }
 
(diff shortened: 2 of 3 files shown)


lucas-zimerman (approved):
LGTM!

alwx (approved):

PR #6132, opened 2026-05-11
fix(core): Deduplicate native HTTP breadcrumbs

### LINKED ISSUE #3045: Duplicate http / xhr breadcrumbs
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [x]  iOS
- [ ]  Android

**SDK:**

**SDK version:** 4.13.0

**`react-native` version:** 0.71.7

**Are you using Expo?**
- [x]  Yes
- [ ]  No

- [ ]  sentry.io (SaaS)
- [ ]  on-premise

[Link to issue]

**Configuration:**

(`@sentry/react-native`)
```
Sentry.init({
  enableInExpoDevelopment: true,
  tracesSampleRate: 1,
  normalizeDepth: 10,
  integrations: [
       new Sentry.Native.ReactNativeTracing({
          routingInstrumentation,
          tracePropagationTargets: [appEnv.apiUrl],
        }),
    ],
});
```

---
**I have following issue:**

When I have a natif crash, the breadcrumbs for the http request appears 3 times :
Capture d’écran 2023-05-05 à 10 16 10

After looking into it, I have the impression that the breadcrumbs of category "http" are coming from the native : if I look into the beforeBreadcrumb in the Sentry.init (in JS), I can't find those breadcrumbs. I only find the "xhr" breadcrumb.

**Steps to reproduce:**
- Init Sentry
- Make an api call
- Crash the app using `Sentry.Native.nativeCrash`

**Actual result:**

In the error, I have 3 times the same call API. If I filter my breadcrumbs with beforeBreadcrumbs, only the xhr will be affected.

Capture d’écran 2023-05-05 à 10 16 10

**Expected result:**

There should be only one log of the API call. It should be affected by the beforeBreadcrumbs in the Sentry.init.

antonis:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

When JS makes an HTTP request in React Native, breadcrumbs are created independently by both the JS SDK (`xhr` category) and the native SDK (`http` category), because RN's networking layer (fetch/XHR) is implemented via native APIs (`NSURLSession` on iOS, `OkHttp` on Android). This results in duplicate breadcrumbs for every HTTP request.

This PR adds deduplication logic in `deviceContextIntegration.processEvent` that filters out native `http` breadcrumbs when a matching JS `xhr`/`fetch` breadcrumb exists (same method, URL, status code, and timestamp within 2s tolerance).

Key design choices:
- **Keeps the JS breadcrumb, drops the native one** — the JS breadcrumb goes through `beforeBreadcrumb`, giving users filtering control
- **Consumed-match tracking** — each JS breadcrumb can only absorb one native duplicate, preventing false positives with consecutive identical requests
- **Only targets native HTTP breadcrumbs** — non-HTTP native breadcrumbs (ui.click, etc.) are untouched
- **Native-only HTTP requests preserved** — requests from native code (ObjC/Swift/Java) with no JS counterpart are kept
- **Handles string/number `status_code` mismatch** — uses `Number()` coercion for comparison

Also updates the FIXME comment in `breadcrumbs.ts` to reflect that deduplication now happens in `deviceContextIntegration`.

Closes #3045

Users see duplicate HTTP breadcrumbs (2x per request) which crowd out meaningful breadcrumbs and cannot be filtered via `beforeBreadcrumb` since native breadcrumbs bypass that callback.

The sentry-cocoa side of this (cocoa producing its own internal duplicates, #2971) was fixed in sentry-cocoa 8.8.0. The remaining duplication is the JS vs native layer, which this PR addresses.

- Added 8 unit tests covering: basic dedup, no-match preservation, non-HTTP breadcrumb preservation, timestamp tolerance boundary, multiple identical requests, string/number status_code coercion, empty JS breadcrumbs, and fetch category support
- All existing tests pass unmodified
- Build passes (`yarn build`)
- Lint passes (`yarn lint`)

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes


antonis:
@sentry review

packages/core/src/js/integrations/devicecontext.ts
```diff
@@ -86,11 +86,70 @@ async function processEvent(event: Event, _hint: EventHint, client: Client): Pro
...
+      if (consumedJsIndices.has(index)) {
+        return false;
+      }
+
+      const sameMethod = nativeBreadcrumb.data?.method === jsBreadcrumb.data?.method;
+      const sameUrl = nativeBreadcrumb.data?.url === jsBreadcrumb.data?.url;
+      const sameStatus =
+        nativeBreadcrumb.data?.status_code != null &&
+        jsBreadcrumb.data?.status_code != null &&
+        Number(nativeBreadcrumb.data.status_code) === Number(jsBreadcrumb.data.status_code);
```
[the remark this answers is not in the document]
antonis: Fixed in c1d65b6 — `sameStatus` now treats both sides being `null`/`undefined` as a match, and uses `Number()` coercion for string/number mismatches. Added tests for both cases.

packages/core/src/js/integrations/devicecontext.ts
```diff
@@ -86,11 +86,72 @@ async function processEvent(event: Event, _hint: EventHint, client: Client): Pro
...
+      const jsStatus = jsBreadcrumb.data?.status_code;
+      const sameStatus =
+        nativeStatus == null && jsStatus == null
+          ? true
+          : nativeStatus != null && jsStatus != null && Number(nativeStatus) === Number(jsStatus);
+      const withinTimeTolerance =
+        nativeBreadcrumb.timestamp != null &&
+        jsBreadcrumb.timestamp != null &&
+        Math.abs(nativeBreadcrumb.timestamp - jsBreadcrumb.timestamp) <=
+          HTTP_BREADCRUMB_DEDUP_TIMESTAMP_TOLERANCE_SECONDS;
```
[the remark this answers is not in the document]
antonis: Good catch — this was a pre-existing gap in `breadcrumbFromObject`. Fixed in b8adccf: numeric timestamps are now passed through directly. Added tests for numeric timestamp handling and NaN rejection.

lucas-zimerman:
Nice fix!

packages/core/src/js/integrations/devicecontext.ts
```diff
@@ -86,11 +86,72 @@ async function processEvent(event: Event, _hint: EventHint, client: Client): Pro
...
+      return true;
+    }
+
+    const matchIndex = jsHttpBreadcrumbs.findIndex((jsBreadcrumb, index) => {
+      if (consumedJsIndices.has(index)) {
+        return false;
+      }
+
+      const sameMethod = nativeBreadcrumb.data?.method === jsBreadcrumb.data?.method;
+      const sameUrl = nativeBreadcrumb.data?.url === jsBreadcrumb.data?.url;
```
lucas-zimerman, line 134: What do you think of using something like `endsWith` here? Usually the start of the backend url will always match, only the end/center will vary. That could skip checking the initial URL.

lucas-zimerman: Although if almost all the time this function is called is a match, then it's a bit pointless to check for it

lucas-zimerman:
Question: Can't we simply disable the auto capture for http request on the native side?
Also thinking if on this case, we cant automatically tag the breadcrumb as manually captured by the end user or not.

packages/core/src/js/integrations/devicecontext.ts
antonis: Yes, and having a looser check with `endsWith` might risk false positives

antonis:
> Question: Can't we simply disable the auto capture for http request on the native side? Also thinking if on this case, we cant automatically tag the breadcrumb as manually captured by the end user or not.

@lucas-zimerman Good question 👍 I considered it as an approach ([option 1 here](https://github.com/getsentry/sentry-react-native/issues/3045#issuecomment-1559722295)) and I agree that is is a cleaner architecturally. In reality disabling didn't work since we don't know if a request came from JS or from native code. The solution is flagging the JS requests (e.g. extra flag, [see option 3](https://github.com/getsentry/sentry-react-native/issues/3045#issuecomment-1559722295)) and filtering out on the native side resulting in a more complicated solution. [Approach 4](https://github.com/getsentry/sentry-react-native/issues/3045#issuecomment-1559722295) seemed simpler to implement but I'm open to discuss the alternatives.

lucas-zimerman (approved):
LGTM!
Lets also open an issue to see if we can refactor this change in the future by finding a simpler way of deduplicating the http requests

antonis (commented):
Opened https://github.com/getsentry/sentry-react-native/issues/6187 to follow up on this

PR #6025, opened 2026-04-20
feat(core): Add includeFeedback Metro option to exclude feedback widget from bundle

antonis:
## :loudspeaker: Type of change
- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Adds an `includeFeedback` option to `SentryMetroConfigOptions` (used by both `withSentryConfig` and `getSentryExpoConfig`). When set to `false`, a Metro resolver returns `{ type: 'empty' }` for any module matching `@sentry(?:-internal)?/feedback`, removing the web user-feedback widget from the bundle.

Example:

```ts
const config = getSentryExpoConfig(__dirname, {
  includeFeedback: false,
});
```

Also factored the shared resolver logic into `buildSentryPackageExcludeResolver`, so `withSentryResolver` (replay) and the new `withSentryFeedbackResolver` no longer duplicate the resolver body.

## :bulb: Motivation and Context
Users who do not use the feedback feature are still paying for `@sentry-internal/feedback` in their bundle (see the expo-atlas screenshot in the linked issue). The existing `includeWebReplay` option solves the same problem for replay; this PR mirrors that pattern for feedback.

Closes #5629

## :green_heart: How did you test it?
- Added unit tests in `packages/core/test/tools/metroconfig.test.ts` for the new `withSentryFeedbackResolver`, covering:
  - `includeFeedback` true/false/undefined across `web`/`android`/`ios`/`null` platforms
  - Both `@sentry-internal/feedback` and `@sentry/feedback` module names
  - Falls through to the original resolver for non-feedback modules
  - Old-Metro (<0.68) fallback path with `process.exit(-1)`
- `yarn test:tools` — 254 passed
- `yarn test` (full suite) — passing
- `yarn lint` — passing
- `yarn circularDepCheck` — passing
- `yarn build` — passing

## :pencil: Checklist
- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [x] I updated the docs if needed. https://github.com/getsentry/sentry-docs/pull/17423
- [x] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

PR #6150, opened 2026-05-15
fix(core): Fix includeWebFeedback crash by stubbing browser wrapper modules

### LINKED ISSUE #6149: Docs example for includeWebFeedback/includeWebReplay crashes at startup

React Native without Frameworks, RN New Architecture, Hermes, React Navigation

sentry.io (SaS)

No

_No response_

8.10.0

````
  react-native:
    installed: 0.81.6
````

### Sentry.init()

````js
Sentry.init({
  environment: '...',
  integrations: [sentryNavigationIntegration],
  tracesSampleRate: 0.1,
  profilesSampleRate: 0.1,
  sendDefaultPii: false,
  attachScreenshot: false,
  attachViewHierarchy: false,
  enableAutoPerformanceTracing: true,
});
````

### Steps to Reproduce

Build and run the app

### Expected Result

Doesn't crash

### Actual Result

Following the official "Reduce Bundle Size" docs for the metro config setup:

https://docs.sentry.io/platforms/react-native/manual-setup/metro/#reduce-bundle-size

```
  module.exports = withSentryConfig(config, {
    includeWebReplay: false,
    includeWebFeedback: false,
  });
```

**Crash**

```
[runtime not ready]: TypeError: ... buildFeedbackIntegration is not a function (it is undefined)
```

Thrown at module evaluation, before any app code runs.

### PR DESCRIPTION
**Pull request by antonis**
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

When `includeWebFeedback: false` is set in the Metro config, the app crashes at startup with:

```
TypeError: buildFeedbackIntegration is not a function (it is undefined)
```

**Root cause:** The feedback resolver stubs `@sentry-internal/feedback` to an empty module, but `@sentry/browser`'s barrel file imports `./feedbackSync.js` and `./feedbackAsync.js`, which **call** `buildFeedbackIntegration()` at module evaluation time. Since the feedback package is empty, the function is `undefined` and the call crashes before any app code runs.

This differs from replay (`includeWebReplay: false`), which works because `@sentry/browser` only re-exports symbols from `@sentry-internal/replay` without calling them at evaluation time.

**Fix:** The feedback resolver now also stubs `feedbackSync.js` and `feedbackAsync.js` when imported from `@sentry/browser`, preventing them from evaluating and calling into the empty module. The `buildSentryPackageExcludeResolver` helper is changed from a `RegExp` to a `shouldExcludeModule` function to support this origin-aware matching.

Fixes https://github.com/getsentry/sentry-react-native/issues/6149

Introduced in https://github.com/getsentry/sentry-react-native/pull/6025 — the `includeWebFeedback` option was modeled after `includeWebReplay`, but feedback has a different integration pattern in `@sentry/browser` (eager function call vs plain re-export).

- Added 16 unit tests for the wrapper module exclusion (`feedbackSync.js`, `feedbackAsync.js`)
- Added an integration-style test that simulates all `@sentry/browser` barrel imports and verifies only feedback-related modules are stubbed
- All 169 tests pass (142 existing + 27 new)
- Build and lint pass

## :pencil: Checklist
- [x] I updated the docs if needed. https://github.com/getsentry/sentry-docs/pull/17775
- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 
 - Fix duplicate JS error reporting on iOS New Architecture when the native SDK is initialized early via `sentry.options.json` ("Capture App Start Errors"). It's done by applying the `ExceptionsManager.reportException` C++ wrapper filter in both init paths ([#6145](https://github.com/getsentry/sentry-react-native/pull/6145))
 - Fix boolean options from `sentry.options.json` being ignored on Android when using `RNSentrySDK.init` ([#6130](https://github.com/getsentry/sentry-react-native/pull/6130))
+- Fix `includeWebFeedback: false` Metro config option causing crash at startup ([#6150](https://github.com/getsentry/sentry-react-native/pull/6150))
 
 ### Dependencies
 
--- a/packages/core/src/js/tools/metroconfig.ts
+++ b/packages/core/src/js/tools/metroconfig.ts
@@ -257,7 +257,7 @@ type CustomResolverBeforeMetro068 = (
 function buildSentryPackageExcludeResolver(
   config: MetroConfig,
   includePackage: boolean | undefined,
-  moduleRegex: RegExp,
+  shouldExcludeModule: (moduleName: string, context: CustomResolutionContext) => boolean,
   optionName: string,
 ): MetroConfig {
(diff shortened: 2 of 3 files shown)


lucas-zimerman (approved):
LGTM!

Issue #6170, opened 2026-05-18
Experiment: build RNSentry against sentry-cocoa pulled via SPM


alwx:
Use spm_dependency in our sample app's Podfile to source Sentry from SPM instead of CocoaPods. Keep our `s.dependency 'Sentry'` in the podspec or replace it with vendored xcframework. Verifies that all our private-header reaches work through SPM's distribution of sentry-cocoa. <br><br>This is the single most useful experiment because it tells us whether [https://github.com/getsentry/sentry-cocoa/issues/7881](<https://github.com/getsentry/sentry-cocoa/issues/7881>) is actually blocking us today.

PR #6181, opened 2026-05-18
fix(ios): use framework-style import for SentrySwizzle.h

alwx:
The quote-style `#import "SentrySwizzle.h"` only resolves when sentry-cocoa is consumed via CocoaPods source build, where the header sits on the HEADER_SEARCH_PATHS. When sentry-cocoa is consumed as a pre-built xcframework (via SPM `binaryTarget` or manual drop-in), the header lives in the framework's `PrivateHeaders/` directory and the quote-style import fails.

It's required for the sentry-cocoa SPM consumption: getsentry/sentry-react-native#6170.

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/core/ios/RNSentryRNSScreen.m
+++ b/packages/core/ios/RNSentryRNSScreen.m
@@ -4,7 +4,7 @@
 
 #    import "RNSentryDependencyContainer.h"
 #    import "RNSentryFramesTrackerListener.h"
-#    import "SentrySwizzle.h"
+#    import <Sentry/SentrySwizzle.h>
 @import Sentry;
 
 @implementation RNSentryRNSScreen


antonis (approved):
LGTM!
Let's leave the extra CI iOS checks to run just in case

antonis:
@alwx I think we'll need to temporarily revert the change since [it causes build failures with some configurations](https://github.com/getsentry/sentry-react-native/actions/runs/26042349623/job/76557219795)

```
[ReactAppDependencyProvider] Generating ReactAppDependencyProvider.framework.dSYM
Error: 'Sentry/SentrySwizzle.h' file not found (in target 'RNSentry' from project 'Pods')
** BUILD FAILED **
```

> It's required for the sentry-cocoa SPM consumption: https://github.com/getsentry/sentry-react-native/issues/6170.

We can make the change on a feature branch if needed to avoid blocking the release process later this week.

antonis (commented):
Probably we can use a `__has_include` [similar to this](https://github.com/getsentry/sentry-react-native/blob/1605536674d9a45cfa47f54f45a01dae191945b5/packages/core/ios/RNSentry.h#L4) like:

```
#if __has_include(<Sentry/SentrySwizzle.h>)
#    import <Sentry/SentrySwizzle.h>
#else
#    import "SentrySwizzle.h"
#endif
```

I'll test this out in https://github.com/getsentry/sentry-react-native/pull/6186 🤞

PR #6197, opened 2026-05-21
Correct route and dynamic param extraction for Expo Router (#6157)

### LINKED ISSUE #6157: Correct route and dynamic param extraction for Expo Router
`extractDynamicRouteParams()` works but:

* Transaction names are noisy/uninformative (`index`, `(tabs)`).
* Dynamic param extraction almost never finds matching keys, so spans rarely carry `route.params.*` attributes.
* `useFullPathsForNavigationRoutes` builds a path from nested navigators, but it joins React Navigation names, not Expo Router pathnames.<br>

Here is what we can do: When Expo Router is active, derive the route from Expo Router's own state rather than from `getCurrentRoute().name`:

* Use the Expo Router path (e.g. `/profile/[id]`) as the canonical span name and `route.name` attribute.
* Extract dynamic params from the matched route pattern + the actual `params` object, including catch-all `[...slug]`.
* Keep query-string params off the span by default; gate them behind `sendDefaultPii` (matching the existing PII model in `deeplink.ts` and the current `extractDynamicRouteParams` contract).
* Set `route.path` (templated) and, under `sendDefaultPii`, `route.url` (concrete).

alwx:
## 📢 Type of change

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## 📜 Description

Fixes getsentry/sentry-react-native#6157

The idea is simple: when Expo Router is active, we can get the route from Expo Router's own state rather than from getCurrentRoute().name. That allows for this information about route and params to be more precise and detailed compared to the filename-based approach of React Navigation.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## 🔮 Next steps


alwx:
@cursor review this please

packages/core/src/js/tracing/reactnavigation.ts
```diff
@@ -193,9 +212,15 @@ export const reactNavigationIntegration = ({
    * @param navigationContainerRef Ref to a `NavigationContainer`
    */
   registerNavigationContainer: (navigationContainerRef: unknown) => void;
+  /**
+   * Internal API: allow another integration (for example, Expo Router integration) to
```
antonis: I think an `@internal` annotation [should make the api checker happy](https://github.com/getsentry/sentry-react-native/actions/runs/26226852806/job/77178967932?pr=6197) 🤞 
```suggestion
   * @internal API: allow another integration (for example, Expo Router integration) to
```

antonis (commented):
Other than [the API change issue](https://github.com/getsentry/sentry-react-native/pull/6197/changes#r3282219348) LGTM!

packages/core/src/js/tracing/reactnavigation.ts
alwx: Good catch, thanks! Applied the `@internal` tag. The CI failure was actually a stale API report — api-extractor still emits `@internal` symbols in the report by default (since `dtsRollup` is disabled), so I also regenerated `etc/sentry-react-native.api.md` to include the new field with the correct tag.

antonis (approved):
LGTM 🚀

PR #6202, opened 2026-05-21
fix(ios): Return NO from requiresMainQueueSetup

### LINKED ISSUE #6201: Return NO from requiresMainQueueSetup to avoid main-queue dispatch_sync at bridge init

RN New Architecture, Expo Application Services (EAS), Expo (mobile only), React Navigation

sentry.io (SaS)

No

_No response_

6.20.0

````

(REDACTED)
iOS:
  hermesEnabled: true
  newArchEnabled: true


````

### Sentry.init()

````js
Sentry.init({
  // other options
});
````
N/A

### Steps to Reproduce

RNSentry's +requiresMainQueueSetup currently returns YES (packages/core/ios/RNSentry.mm). Its init only performs a trivial NSObject allocation — no UIKit or main-thread-only work — so it can safely return NO.

Returning YES forces a dispatch_sync onto the main queue during bridge setup, which has been observed as a rare "App Hang" crasher (notably reported by Shopify: https://arc.net/l/quote/kkxmwgdb). Concern grows under the New Architecture where bridge-init timing is tighter.

Proposed change: return NO. Happy to open a PR if it's welcome.

### Expected Result

No Crash

### Actual Result

Startup crashes

(This is more of a preventative measure as we are cleaning up things and moving to new arch)

antonis:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Return `NO` from `+requiresMainQueueSetup` in `RNSentry.mm`. The module's `init` only performs a trivial `NSObject` allocation (`RNSentryTimeToDisplay`), which has no main-thread requirement.

Note that `methodQueue` still returns `dispatch_get_main_queue()`, so all exported RCT methods continue to execute on the main thread as before.

When `requiresMainQueueSetup` returns `YES`, React Native uses `dispatch_sync(mainQueue, ...)` to initialize the module during bridge setup. This unnecessary synchronization point can contribute to startup hangs if the main thread is blocked.

This was originally set to `YES` in 2017 when `methodQueue` was changed to return the main queue, but `init` never actually required main-thread execution.

Fixes https://github.com/getsentry/sentry-react-native/issues/6201

This is a one-line change to a React Native module lifecycle constant. The sample app CI builds validate that the app starts correctly.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps


lucas-zimerman (approved):
LGTM!
But worth adding a changelog since it fixes an user issue.

PR #6221, opened 2026-05-27
feat(tracing): Wrap Expo Router `push`, `replace`, `navigate`, `back`, `dismiss` in addition to `prefetch`

### LINKED ISSUE #6158: Wrap `push`, `replace`, `navigate`, `back`, `dismiss` in addition to `prefetch`
`wrapExpoRouter()` in `tracing/expoRouter.ts` only wraps `prefetch`. The `ExpoRouter` interface already enumerates `push`, `replace`, `back`, `navigate`, but none of them are instrumented. As a result, programmatic navigations are invisible at the call-site level — we only see the resulting state change via the React Navigation listener, with no link back to the code that triggered it.<br><br>Wrap the other router methods so each navigation call:

* Adds a breadcrumb (`category: 'navigation'`, `type: 'navigation'`) carrying the method (`push` / `replace` / `navigate` / `back` / `dismiss`), the target href, and the parsed pathname/params.
* Associates the call with the subsequent idle navigation transaction so the span can be attributed to the initiating method (e.g. `navigation.method = "push"`).
* Mirrors the existing prefetch error/status handling (`SPAN_STATUS_OK` / `SPAN_STATUS_ERROR`).
* Is safe to call repeatedly — keep the `__sentryPrefetchWrapped` guard pattern and extend it to a single `__sentryWrapped` flag.

alwx:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Extends the Expo Router instrumentation beyond `prefetch` so that programmatic navigation calls produce breadcrumbs and spans, and the resulting idle navigation transaction can be attributed back to the call that triggered it.

What changed:

- `wrapExpoRouter` now wraps `push`, `replace`, `navigate`, `back`, and `dismiss` in addition to `prefetch`.
- Each wrapped call:
  - emits a `navigation` breadcrumb (`category: 'navigation'`, `data.method`, `data.pathname`),
  - opens a short-lived `navigation.<method>` span tagged with `sentry.origin = auto.navigation.expo_router`, `navigation.method`, and `route.name`,
  - stores a pending hand-off (`pendingExpoRouterNavigation`) that the next React Navigation idle navigation span consumes to set its own `navigation.method` attribute, attributing the transaction back to the originating call.
- A single `__sentryWrapped` flag guards against double-wrapping if `wrapExpoRouter` is called multiple times on the same router.
- PII handling mirrors `reactnavigation.ts`: raw `href` and route `params` (which can encode user identifiers like `/users/123` or `{ id: '123' }`) are only attached to breadcrumbs and span attributes when `sendDefaultPii` is enabled. `method`, `pathname` (the route template), and `route.name` remain unconditional so navigations stay visible and groupable.

Fixes #6158.

Previously only `prefetch` was instrumented, so any programmatic `router.push(...)` / `router.replace(...)` / `router.back()` etc. was invisible in Sentry — no breadcrumb trail, no span, and the resulting React Navigation idle transaction had no way to know which call site initiated it. This PR closes that gap while keeping the existing `prefetch` behavior intact and aligning PII handling with the rest of the navigation tracing.

- Added unit tests in `packages/core/test/tracing/expoRouter.test.ts` covering:
  - prefetch (string href, object href, object without pathname, sync + async success, sync + async errors),
  - `push` / `replace` / `navigate` via `describe.each` — span and breadcrumb shape with string and object hrefs, error reporting, and the pending-navigation hand-off,
  - `back()` and `dismiss(count)` with no/optional args,
  - double-wrap idempotency,
  - PII gating: default-off assertions (no `href`, no `params`) and dedicated `sendDefaultPii: true` cases that verify `href` and `params` reappear on both the breadcrumb and the span.
- Full JS test suite: `yarn jest` — 113 suites / 1481 tests passing.
- Linters: `yarn lint:lerna` clean.
- API report: `yarn api-report:check` up to date.

## :pencil: Checklist

- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPii` is enabled
- [x] All tests passing
- [x] No breaking changes


alwx:
It might look scary but it's mostly tests.

packages/core/src/js/tracing/expoRouter.ts
```diff
@@ -71,21 +125,124 @@ export function wrapExpoRouter<T extends ExpoRouter>(router: T): T {
...
+        ...(safePathname ? { pathname: safePathname } : undefined),
+        // `href` (raw URL form) and `params` may contain user identifiers or
+        // other PII (e.g. `/users/42`, `{ id: '42' }`). Mirror the behavior of
+        // `reactnavigation.ts` and only include them when `sendDefaultPii` is on.
+        ...(sendPii && parsed.href !== undefined ? { href: serializeHref(parsed.href) } : undefined),
+        ...(sendPii && parsed.params ? { params: parsed.params } : undefined),
+      },
+    });
+
+    setPendingExpoRouterNavigation({
```
antonis: nit: The href, pathname, and params seem to be unused

antonis (approved):
LGTM 🚀 
Left a [nit comment](https://github.com/getsentry/sentry-react-native/pull/6221/changes#r3341433719) and added the `ready-to-merge` to run the full test suite.

PR #6288, opened 2026-06-15
Record network request/response bodies in Session Replay

### LINKED ISSUE #4106: [Session Replay] - Record network
**OS:**
- [ ]  Windows
- [x]  MacOS
- [ ]  Linux

**Platform:**
- [x]  iOS
- [x]  Android

**SDK:**

**SDK version:** 0.0.0

**`react-native` version:** 0.0.0

**Are you using Expo?**
- [ ]  Yes
- [x]  No

- [x]  sentry.io (SaaS)
- [ ]  on-premise

[Click here](https://qi-tech-eg.sentry.io/replays/6ebfea0e8ad7402cb8913449c01e5f35/?project=4505018839072768&query=&referrer=%2Freplays%2F%3AreplaySlug%2F&statsPeriod=14d&yAxis=count%28%29&t=0)

**Configuration:**

(`@sentry/react-native`)
```
  const integrations = [
          maskAllImages: false,
          maskAllText: false,
          maskAllVectors: false,
        }),
      ];
  const _experiments = {
        replaysSessionSampleRate: 1.0,
      };
  Sentry.init({
    dsn: SENTRY_DSN,
    _experiments,
    integrations,
    environment: 'production',
    beforeSend: (event) => {
      return event;
    },
  });

```

---
**I have the following issue:**
It doesn't record network payloads, response and others like headers etc...

**Steps to reproduce:**
- use axios for requests
- activate sentry session replay
- make some request and see the log on the website

**Actual result:**
- it shows the screen recording, and the network requests
- but it doesn't show the request and response payloads

**Expected result:**
-  it shows the screen recording, and the network requests
- when i click on one request, see the request, response and headers payload

alwx:
- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Enrich Mobile Session Replay with HTTP request/response **headers** and **bodies**, surfacing them in the Replay network tab (in addition to the URL, method, status code and body sizes already captured).

The feature is **opt-in**: nothing extra is captured until you list URLs in `networkDetailAllowUrls`. Five new options are added on `mobileReplayIntegration`:

| Option | Type | Default | Purpose |
|---|---|---|---|
| `networkDetailAllowUrls` | `(string \| RegExp)[]` | `[]` | URLs to enrich with headers (and bodies, when `networkCaptureBodies` is on). Strings use substring match; regexes use `.test(url)`. |
| `networkDetailDenyUrls` | `(string \| RegExp)[]` | `[]` | URLs to **never** enrich, even if they match the allow list. |
| `networkCaptureBodies` | `boolean` | `false` | Opt in to capturing request/response bodies for allow-listed URLs. |
| `networkRequestHeaders` | `string[]` | `[]` | Extra request headers to capture, in addition to the defaults. |
| `networkResponseHeaders` | `string[]` | `[]` | Extra response headers to capture, in addition to the defaults. |

```ts
Sentry.init({
  dsn: "___PUBLIC_DSN___",
  integrations: [
      networkDetailAllowUrls: ["api.example.com", /^https:\/\/cdn\./],
      networkDetailDenyUrls: [/\/auth\//],
      networkCaptureBodies: true,
      networkRequestHeaders: ["X-My-Header"],
      networkResponseHeaders: ["X-Response-Header"],
    }),
  ],
});
```

### How it works

- **JS layer** (`packages/core/src/js/replay/`):
  - `xhrUtils.ts` registers a `beforeAddBreadcrumb` hook that enriches each `xhr` breadcrumb with `request` / `response` sides containing headers and optionally a serialized body.
  - `networkUtils.ts` handles URL allow/deny matching (regex `lastIndex` is reset on each call so global-flag patterns stay stateless; empty strings in the allow list are ignored), body serialization (strings, `URLSearchParams`, `FormData`, plain JSON, and primitives like `42` / `true`), and a 150 KB truncation cap that surfaces a `MAX_BODY_SIZE_EXCEEDED` warning. Binary payloads (`Blob`, `ArrayBuffer`, typed arrays) are recorded as a placeholder with an `UNPARSEABLE_BODY_TYPE` warning rather than serialized to `"{}"`.
  - Headers are filtered against an allow list (`Content-Type`, `Content-Length`, `Accept` by default, plus user-supplied); authorization-like headers (`Authorization`, `Cookie`, `Set-Cookie`, `X-API-Key`, `X-Auth-Token`, `Proxy-Authorization`) are always stripped.
- **Native layer** (Android Java + iOS ObjC `RNSentryReplayBreadcrumbConverter`):
  - Forward the enriched `request` / `response` dicts to the native rrweb span event for Mobile Replay.
  - Strip the JS-internal `_meta` key before forwarding so the native replay SDKs only see fields they understand.
  - Numeric fields (`start_timestamp`, `end_timestamp`, `status_code`, `request_body_size`, `response_body_size`) accept any `Number` subtype on Android — the RN bridge can surface them as `Long`/`Integer`, which previously crashed (`ClassCastException`) or were silently dropped.

### Scope and limitations

- **XHR only** for now. This covers `axios` and most popular HTTP clients on React Native. Native `fetch` body capture will be added in a follow-up.
- Bodies and headers are PII-scrubbed server-side; users should still be deliberate about which URLs they allow-list.

Closes #4106 — long-standing request to inspect HTTP payloads in Mobile Session Replay alongside what the JS browser Replay SDK already exposes for `fetch`/XHR on the web.

Companion docs PR: getsentry/sentry-docs#18423.

- **JS unit tests** (`packages/core/test/replay/xhrUtils.test.ts`, 20 cases): allow/deny URL matching (string + regex, empty strings, regex `lastIndex`), header filtering (defaults + extras, deny-list always stripped), body serialization (string / `URLSearchParams` / `FormData` / object / number / boolean), binary body placeholders (`Blob`, `ArrayBuffer`, `Uint8Array`) producing `UNPARSEABLE_BODY_TYPE` with a synthetic placeholder body so warnings survive the native `_meta` strip, and the 150 KB truncation cap producing `MAX_BODY_SIZE_EXCEEDED`.
- **Android native tests** (`RNSentryAndroidTester/.../RNSentryReplayBreadcrumbConverterTest.kt`): network breadcrumb forwarding with `_meta` stripped from both sides, empty-after-strip sides dropped, and non-`Double` `Number` subtypes (`Long`/`Integer`) accepted for all five numeric fields.
- **iOS native tests** (`RNSentryCocoaTester/.../RNSentryReplayBreadcrumbConverterTests.swift`): equivalent coverage for the `_meta` strip and empty-after-strip cases.
- Manually verified end-to-end on a sample app: bodies + headers visible in the Replay network tab; binary bodies show the placeholder; sensitive headers (`Authorization`, `Cookie`) never appear regardless of configuration.

## :pencil: Checklist

  - Body / header capture is fully opt-in (empty default allow list + `networkCaptureBodies: false` by default); authorization-like headers are always stripped.
  - See companion PR: getsentry/sentry-docs#18423.
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- Add native `fetch` body capture (currently XHR only — covers `axios` but not raw `fetch`).

### DIFF
--- a/packages/core/android/src/main/java/io/sentry/react/RNSentryReplayBreadcrumbConverter.java
+++ b/packages/core/android/src/main/java/io/sentry/react/RNSentryReplayBreadcrumbConverter.java
@@ -166,17 +166,18 @@ public final class RNSentryReplayBreadcrumbConverter extends DefaultReplayBreadc
     if (breadcrumb.getData("method") instanceof String) {
       data.put("method", breadcrumb.getData("method"));
     }
-    if (breadcrumb.getData("status_code") instanceof Double) {
-      final Double statusCode = (Double) breadcrumb.getData("status_code");
+    // Accept any Number subtype (Double/Long/Integer) — the RN bridge does not guarantee Double.
+    if (breadcrumb.getData("status_code") instanceof Number) {
+      final int statusCode = ((Number) breadcrumb.getData("status_code")).intValue();
       if (statusCode > 0) {
-        data.put("statusCode", statusCode.intValue());
+        data.put("statusCode", statusCode);
       }
     }
-    if (breadcrumb.getData("request_body_size") instanceof Double) {
-      data.put("requestBodySize", breadcrumb.getData("request_body_size"));
+    if (breadcrumb.getData("request_body_size") instanceof Number) {
+      data.put("requestBodySize", ((Number) breadcrumb.getData("request_body_size")).doubleValue());
     }
-    if (breadcrumb.getData("response_body_size") instanceof Double) {
-      data.put("responseBodySize", breadcrumb.getData("response_body_size"));
+    if (breadcrumb.getData("response_body_size") instanceof Number) {
+      data.put("responseBodySize", ((Number) breadcrumb.getData("response_body_size")).doubleValue());
     }
     final Map<Object, Object> requestSide = sanitizeNetworkSide(breadcrumb.getData("request"));
     if (!requestSide.isEmpty()) {
--- a/packages/core/android/src/main/java/io/sentry/react/RNSentryReplayBreadcrumbConverter.java
+++ b/packages/core/android/src/main/java/io/sentry/react/RNSentryReplayBreadcrumbConverter.java
@@ -144,13 +144,16 @@ public final class RNSentryReplayBreadcrumbConverter extends DefaultReplayBreadc
 
   @TestOnly
   public @Nullable RRWebEvent convertNetworkBreadcrumb(final @NotNull Breadcrumb breadcrumb) {
+    // Use Number.doubleValue() rather than a direct (Double) cast: the RN bridge can
+    // surface timestamps as Long/Integer, which pass `instanceof Number` but would
+    // throw `ClassCastException` on a direct cast to Double.
     final Double startTimestamp =
         breadcrumb.getData("start_timestamp") instanceof Number
-            ? (Double) breadcrumb.getData("start_timestamp")
+            ? ((Number) breadcrumb.getData("start_timestamp")).doubleValue()
             : null;
     final Double endTimestamp =
         breadcrumb.getData("end_timestamp") instanceof Number
-            ? (Double) breadcrumb.getData("end_timestamp")
+            ? ((Number) breadcrumb.getData("end_timestamp")).doubleValue()
             : null;
     final String url =
         breadcrumb.getData("url") instanceof String ? (String) breadcrumb.getData("url") : null;
--- a/packages/core/android/src/main/java/io/sentry/react/RNSentryReplayBreadcrumbConverter.java
+++ b/packages/core/android/src/main/java/io/sentry/react/RNSentryReplayBreadcrumbConverter.java
@@ -175,6 +175,12 @@ public final class RNSentryReplayBreadcrumbConverter extends DefaultReplayBreadc
     if (breadcrumb.getData("response_body_size") instanceof Double) {
       data.put("responseBodySize", breadcrumb.getData("response_body_size"));
     }
+    if (breadcrumb.getData("request") instanceof Map) {
+      data.put("request", breadcrumb.getData("request"));
+    }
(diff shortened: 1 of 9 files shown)


alwx:
@cursor review

packages/core/src/js/replay/xhrUtils.ts
```diff
@@ -2,14 +2,55 @@ import type { Breadcrumb, BreadcrumbHint, SentryWrappedXMLHttpRequest, XhrBreadc
...
+interface NetworkBreadcrumbSide {
+  body?: string;
+  headers?: Record<string, string>;
+  _meta?: { warnings: string[] };
+}
+
+const DEFAULT_NETWORK_OPTIONS: ResolvedNetworkOptions = {
+  allowUrls: [],
+  denyUrls: [],
+  captureBodies: true,
```
antonis: q: Should this be false [like above](https://github.com/getsentry/sentry-react-native/pull/6288/changes#diff-15f06061acd36165e8cbd8ed2c25d6f4c3b2880a9d55a5f097ca1d9d2af0b219R193)?

```suggestion
  captureBodies: false,
```

packages/core/android/src/main/java/io/sentry/react/RNSentryReplayBreadcrumbConverter.java
antonis, line 1: nit: It would be nice to add/update the Android native tests if possible

packages/core/ios/RNSentryReplayBreadcrumbConverter.m
antonis, line 1: nit: It would be nice to add/update the iOS native tests if possible

antonis (commented):
Overall LGTM! Added a few comments to consider. Let's also update the PR description for future reference.

antonis (approved):
LGTM 🚀 
Just added the `ready-to-merge` label be be on the safe side since this is a big PR :)

PR #6308, opened 2026-06-17
fix(android): Prevent breadcrumb loss from numeric timestamp ClassCastException

antonis:
## :loudspeaker: Type of change
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
On Android, every breadcrumb synced from JS to the native scope was being discarded since `8.14.0`.

The JS SDK stamps `breadcrumb.timestamp` as a **number** (epoch seconds), which arrives over the bridge as a `java.lang.Double`. Since #6261, `RNSentryBreadcrumb.fromMap` feeds the map to the native `Breadcrumb.Deserializer`, which reads `timestamp` via `MapObjectReader.nextDateOrNull` → `nextStringOrNull` and requires an ISO-8601 **string**. A `Double` throws `ClassCastException`, the breadcrumb is dropped, and an empty fallback (only `origin = "react-native"`) is returned — so native/tombstone crash reports lose breadcrumb context.

```
java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.String
	at io.sentry.util.MapObjectReader.nextStringOrNull(MapObjectReader.java:321)
	at io.sentry.util.MapObjectReader.nextDateOrNull(MapObjectReader.java:146)
	at io.sentry.Breadcrumb$Deserializer.deserialize(Breadcrumb.java:872)
	at io.sentry.react.RNSentryBreadcrumb.fromMap(RNSentryBreadcrumb.java:47)
```

This PR normalizes a numeric `timestamp` to an ISO-8601 string before deserializing, preserving millisecond precision. Existing behavior is unchanged for ISO-string and missing timestamps. iOS is unaffected (Cocoa's `initWithDictionary:` is lenient).

## :bulb: Motivation and Context
Fixes #6306. Regression introduced in `8.14.0` by #6261 ("Use native SDK deserializers for User and Breadcrumb").

## :green_heart: How did you test it?
Added `RNSentryBreadcrumbTest.parsesNumericTimestampWithoutDiscardingBreadcrumb`, which feeds a numeric timestamp and asserts the breadcrumb survives with its fields and correct timestamp. Verified it **fails** (`message` is `null`) without the fix and **passes** with it. Full Android unit suite green.

## :pencil: Checklist
- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] All tests passing
[...]

Issue #6306, opened 2026-06-17
Android: breadcrumb sync throws ClassCastException (Double cannot be cast to String) since 8.14.0


tomekzaw:
### What React Native libraries do you use?

Hermes, RN New Architecture, Expo (mobile only), Expo Router

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### Are you using any other error monitoring solution alongside Sentry?

No

### @sentry/react-native SDK Version

8.13.0 → 8.14.0 (regression introduced in 8.14.0)

### How does your development environment look like?

- macOS, Android (issue is Android-only; iOS is unaffected)
- React Native 0.85.3, Hermes + New Architecture enabled
- Expo SDK 56, Expo Router

### Sentry.init()

```js
Sentry.init({
  dsn: "https://...@o...ingest.de.sentry.io/...",
  enabled: !__DEV__,
  environment: "production",
  enableTombstone: true,
  integrations: [Sentry.reactNavigationIntegration()],
});
```

### Steps to Reproduce

1. Use `@sentry/react-native@8.14.0` on Android.
2. Add any breadcrumb from JS — either manually (`Sentry.addBreadcrumb({ message: "test" })`) or via auto-instrumentation (navigation, console, etc.). The JS scope syncs it to the native scope via `RNSentry.addBreadcrumb`.
3. Watch Logcat.

### Expected Result

The breadcrumb is added to the native Android scope (with its message/category/data/timestamp), so it appears in native/tombstone crash reports. No error logged.

### Actual Result

`RNSentryBreadcrumb.fromMap` logs an error and discards the breadcrumb, returning an empty fallback breadcrumb (only `origin = "react-native"`). The breadcrumb is lost from the native scope, so native crash reports lose breadcrumb context. Logcat (one entry per breadcrumb):

```
Failed to deserialize breadcrumb from map.
java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.String
	at io.sentry.util.MapObjectReader.nextStringOrNull(MapObjectReader.java:321)
	at io.sentry.util.MapObjectReader.nextDateOrNull(MapObjectReader.java:146)
	at io.sentry.Breadcrumb$Deserializer.deserialize(Breadcrumb.java:872)
	at io.sentry.react.RNSentryBreadcrumb.fromMap(RNSentryBreadcrumb.java:47)
[...]

PR #6289, opened 2026-06-15
fix(ios): Remove manual geo handling, use sentry-cocoa native support

antonis:
## :loudspeaker: Type of change
- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

Remove manual `SentryGeo` handling from `RNSentry.mm` now that sentry-cocoa 9.17.0+ natively supports geo deserialization in `SentryUser.initWithDictionary:` (via getsentry/sentry-cocoa#8026).

## :bulb: Motivation and Context

Closes #6266

## :green_heart: How did you test it?

CI, Manual

## :pencil: Checklist
- [ ] I added tests to verify changes - [Verified with the existing tests](https://github.com/getsentry/sentry-react-native/blob/5b7e8a7867a7dea11f0e701eb55d72d676845476/packages/core/RNSentryCocoaTester/RNSentryCocoaTesterTests/RNSentryUserTests.m#L115)
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

PR #6261, opened 2026-06-08
refactor(android,ios): Use native SDK deserializers for User and Breadcrumb

antonis:
## :loudspeaker: Type of change
- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Replace manual field-by-field map-to-object construction with native SDK deserialization APIs

## :bulb: Motivation and Context

Closes https://github.com/getsentry/sentry-react-native/issues/4373

## :green_heart: How did you test it?

CI

## :pencil: Checklist
- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

PR #6309, opened 2026-06-17
fix(core): Forward user geo as an object so the native scope keeps it

antonis:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
User `geo` set from JS never reached the native scope.

`NATIVE.setUser` builds the native `userKeys` by running every required key through `_serializeObject`, which JSON-stringifies any non-string value. `geo` is an object, so it crossed the bridge as a string like `'{"city":"Berlin"}'`. The native SDKs, however, deserialize `geo` from a **nested map**:

- **Android** (`User.Deserializer`, since #6261) calls `Geo.Deserializer`, which throws `IOException: Current token is not an object` on a string → the whole user is discarded.
- **iOS** (`userFrom:`) guards `geo` with `isKindOfClass:NSDictionary.class`, so the string is silently ignored → geo lost (user otherwise kept).

This PR forwards `geo` as a structured object so both platforms deserialize it correctly. The change is JS-only — the native deserializers already handle a nested map.

Pre-existing bug surfaced while investigating #6306 (the breadcrumb-timestamp regression is fixed separately in #6308). The JS double-encoding predates #6261; #6261 only changed the Android failure mode (geo lost → whole user dropped).

Coordinates with #6289 (removes the manual iOS geo handling in favour of `initWithDictionary:`): both expect `geo` as a dictionary, so they are complementary and safe to merge in either order. This PR is the JS-layer half that makes the native geo deserialization actually receive an object.

- Updated `wrapper.test.ts` `setUser` cases to assert `geo` is forwarded as an object (full / partial / empty / undefined). Verified they **fail** against the old stringifying code and **pass** with the fix.
- Manually verified on Android that a nested-map `geo` deserializes via `User.Deserializer` while a JSON-string `geo` drops the user (confirming both the regression and the fix). iOS `userFrom:` already accepts a geo dictionary, which is exactly what this PR now sends.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes


lucas-zimerman (approved):
LGTM!

PR #6314, opened 2026-06-18
feat(core): Use react-native-quick-base64 for envelope encoding when available

### LINKED ISSUE #4884: Use CPP Base64 Encoder for better performance (>10x)
### Description

Hermes offers builtin `btoa`, using it or vendoring `react-native-quick-base64`, will bring 10x+ improvement in the event processing time.

### Things to consider

* Add `carrier.base64Polyfill` to clearly separate the event processing logic and the use encoders (same as for utf8)
* ~~Use ~~`~~String.fromCharCode~~`~~ + ~~`~~btoa~~`~~ both natively supported by Hermes~~
  * \~Hermes Team comment on speed of the builtins [https://github.com/facebook/hermes/discussions/1393\~](https://github.com/facebook/hermes/discussions/1393\~)
  * \~Hermes Base64 source code [https://github.com/facebook/hermes/blob/0760c58acf668f58eb909f7b8efa2b7fc2ff10d1/lib/VM/JSLib/Base64.cpp#L20\~](https://github.com/facebook/hermes/blob/0760c58acf668f58eb909f7b8efa2b7fc2ff10d1/lib/VM/JSLib/Base64.cpp#L20\~)
  * Won't for large arrays (profiling, attachments...)
* (Alternative) vendor JSI lib like [react-native-quick-base64](https://github.com/craftzdog/react-native-quick-base64) to use `base64StringFromByteArray`
* Solution could be also try to avoid the base64 encoding of the envelopes (maybe pass it as an object to the native layers, or as an array (note that we have ran into issue in the past with the array size (link TBA)))

alwx:
## 📢 Type of change

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## 📜 Description

Adds [`react-native-quick-base64`](<https://github.com/craftzdog/react-native-quick-base64>) as an **optional peer dependency** of `@sentry/react-native`. When the package is installed by the consumer, the SDK uses its native JSI base64 encoder (\~10x faster than the bundled JS implementation) for envelope payloads on the `RNSentry.captureEnvelope` hot path. When the package is absent the SDK falls back to the existing `base64-js`-derived encoder in `vendor/base64-js` — no behavior change for users who do not opt in.

Changes:

* New `packages/core/src/js/utils/base64.ts` — `encodeToBase64()` resolves the encoder once and caches it. Tries `require('react-native-quick-base64')` first; falls back to `base64StringFromByteArray` from `vendor/base64-js`.
* `packages/core/src/js/wrapper.ts` — `captureEnvelope` now calls `encodeToBase64(envelopeBytes)` instead of `base64StringFromByteArray(envelopeBytes)` directly.
* `packages/core/package.json` — `react-native-quick-base64 >=3.0.0` added to `peerDependencies` and marked `optional: true` in `peerDependenciesMeta`. No new required deps.
* `CHANGELOG.md` — entry under Unreleased / Features.

## 💡 Motivation and Context

Closes getsentry/sentry-react-native#4884.

## 💚 How did you test it?

Addded tests!

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## 🔮 Next steps

* Real-device benchmark on a large envelope (profile / replay) to confirm the headline \~10x figure in our pipeline.
* Optionally mention `react-native-quick-base64` in the Sentry docs as a recommended performance install for users shipping large payloads.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -70,7 +70,8 @@
   "peerDependencies": {
     "expo": ">=49.0.0",
     "react": ">=17.0.0",
-    "react-native": ">=0.65.0"
+    "react-native": ">=0.65.0",
+    "react-native-quick-base64": ">=3.0.0"
   },
   "dependencies": {
     "@sentry/babel-plugin-component-annotate": "5.3.0",
@@ -129,6 +130,9 @@
   },
   "peerDependenciesMeta": {
     "expo": {
+      "optional": true
+    },
+    "react-native-quick-base64": {
       "optional": true
     }
   }
--- a/packages/core/src/js/utils/base64.ts
+++ b/packages/core/src/js/utils/base64.ts
@@ -1,0 +1,56 @@
+import { debug } from '@sentry/core';
+
+import { base64StringFromByteArray } from '../vendor';
+
+type FromByteArray = (bytes: Uint8Array, urlSafe?: boolean) => string;
+
+let cachedEncoder: FromByteArray | null = null;
+let resolved = false;
+
+/**
+ * Resolves the base64 encoder once. If the optional peer dependency
+ * `react-native-quick-base64` is installed, its native JSI encoder is used
+ * (~10x faster than the pure-JS fallback). Otherwise the bundled JS encoder
+ * from `vendor/base64-js` is used.
+ *
+ * The resolution is cached so the require cost is paid at most once.
+ */
+function resolveEncoder(): FromByteArray {
+  if (resolved) {
+    return cachedEncoder ?? base64StringFromByteArray;
+  }
+  resolved = true;
+
+  try {
+    // Optional peer dependency — only loaded if the consumer installed it.
+    // eslint-disable-next-line @typescript-eslint/no-var-requires, import/no-extraneous-dependencies
+    const quickBase64 = require('react-native-quick-base64') as { fromByteArray?: FromByteArray };
+    if (quickBase64 && typeof quickBase64.fromByteArray === 'function') {
+      cachedEncoder = quickBase64.fromByteArray;
+      debug.log('Using react-native-quick-base64 for envelope encoding.');
+      return cachedEncoder;
+    }
+  } catch (_e) {
+    // Not installed — fall through to JS encoder.
+  }
+
+  cachedEncoder = base64StringFromByteArray;
+  return cachedEncoder;
+}
+
+/**
+ * Encode a byte array to a base64 string. Prefers the native
+ * `react-native-quick-base64` encoder when available, otherwise uses the
+ * bundled JS implementation.
+ */
+export function encodeToBase64(input: Uint8Array): string {
+  return resolveEncoder()(input);
+}
+
+/**
+ * @internal Test helper. Resets the cached encoder so the next call re-resolves.
+ */
+export function _resetBase64EncoderForTesting(): void {
+  cachedEncoder = null;
+  resolved = false;
+}
--- a/packages/core/src/js/utils/base64.ts
+++ b/packages/core/src/js/utils/base64.ts
@@ -9,9 +9,8 @@ let resolved = false;
 
 /**
  * Resolves the base64 encoder once. If the optional peer dependency
- * `react-native-quick-base64` is installed, its native JSI encoder is used
- * (~10x faster than the pure-JS fallback). Otherwise the bundled JS encoder
- * from `vendor/base64-js` is used.
+ * `react-native-quick-base64` is installed, its native JSI encoder is used.
+ * Otherwise the bundled JS encoder from `vendor/base64-js` is used.
  *
  * The resolution is cached so the require cost is paid at most once.
  */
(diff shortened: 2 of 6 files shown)


packages/core/test/utils/base64.test.ts
```diff
@@ -0,0 +1,43 @@
+import { _resetBase64EncoderForTesting, encodeToBase64 } from '../../src/js/utils/base64';
+
+jest.mock(
+  'react-native-quick-base64',
+  () => ({
+    __esModule: true,
+    fromByteArray: jest.fn((bytes: Uint8Array) => `quick:${bytes.length}`),
```
antonis: q: Can we verify with the real output here?

packages/core/src/js/utils/base64.ts
```diff
@@ -0,0 +1,56 @@
...
+
+type FromByteArray = (bytes: Uint8Array, urlSafe?: boolean) => string;
+
+let cachedEncoder: FromByteArray | null = null;
+let resolved = false;
+
+/**
+ * Resolves the base64 encoder once. If the optional peer dependency
+ * `react-native-quick-base64` is installed, its native JSI encoder is used
+ * (~10x faster than the pure-JS fallback). Otherwise the bundled JS encoder
```
antonis: q: Is there a reference for the 10x claim?

packages/core/src/js/utils/base64.ts
```diff
@@ -0,0 +1,56 @@
...
+    return cachedEncoder ?? base64StringFromByteArray;
+  }
+  resolved = true;
+
+  try {
+    // Optional peer dependency — only loaded if the consumer installed it.
+    // eslint-disable-next-line @typescript-eslint/no-var-requires, import/no-extraneous-dependencies
+    const quickBase64 = require('react-native-quick-base64') as { fromByteArray?: FromByteArray };
+    if (quickBase64 && typeof quickBase64.fromByteArray === 'function') {
+      cachedEncoder = quickBase64.fromByteArray;
```
antonis, line 31: Wdyt of probing fromByteArray here to catch a broken native module? 
```suggestion
    if (quickBase64 && typeof quickBase64.fromByteArray === 'function') {
      // Probe the native binding so that a broken native module falls through
      quickBase64.fromByteArray(new Uint8Array([0]));
      cachedEncoder = quickBase64.fromByteArray;
```

packages/core/test/utils/base64.test.ts
alwx: Good call — done in the last commit.

packages/core/src/js/utils/base64.ts
alwx: Applied in 7cd554d5 — probing once at resolution time means a broken native binding (e.g. autolinking failed) fails fast and we permanently fall back to the JS encoder, instead of throwing on every `captureEnvelope`. Added a test for that fallback path too.

packages/core/src/js/utils/base64.ts
alwx: it's true that its not really verifyable, and also depends on a real device that uses it. the claim comes from the original issue but better be removed completely

packages/core/package.json
```diff
@@ -70,7 +70,8 @@
   "peerDependencies": {
     "expo": ">=49.0.0",
     "react": ">=17.0.0",
-    "react-native": ">=0.65.0"
+    "react-native": ">=0.65.0",
+    "react-native-quick-base64": ">=3.0.0"
```
antonis, line 74: Since we do not have any peer dependencies other `react-native` and `expo`, I am still skeptical on [if we should add a dependency](https://develop.sentry.dev/sdk/getting-started/standards/code-quality/#dependencies) for this improvement.
Since `RNSentry` is a TurboModule and could accept the envelope bytes as binary across JSI on the new architecture what do you think of pursuing our own solution without a new dependency?

Looping @lucas-zimerman for more feedback on this 🙇

antonis (changes requested):
Overall the changes on this PR look good. [My main concern is if we should add a dependency](https://github.com/getsentry/sentry-react-native/pull/6314/changes#r3451201831) since this has been avoided by design in the SDK till now.

packages/core/package.json
alwx: The problem here is that it's a bit of reinventing the wheel.
Of course, Claude can do it easily but even the first not really working stage of it is already 1000+ lines of code (https://github.com/getsentry/sentry-react-native/pull/6312) and I'm a bit sceptical about the maintenance load and the fact that a verified, well-tested and well-maintained solution which is react-native-quick-base64 works better than either we or Claude can do. 
In addition, this is an optional peer dependency, people could still use the old approach if they don't want it.

antonis: > The problem here is that it's a bit of reinventing the wheel.

I totally agree with that. My concern is that by design we have avoided dependencies ([ref1](https://develop.sentry.dev/sdk/getting-started/philosophy/), [ref2](https://develop.sentry.dev/sdk/getting-started/standards/code-quality/#dependencies)) even for core functionality.

alwx:
Closing in favor of #6351

PR #6345, opened 2026-06-25
fix(core): Hide dataCollection option until supported in React Native

antonis:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Omits the `dataCollection` option from the public RN options types (`ReactNativeOptions` / `ReactNativeClientOptions`) by adding it to the existing `Omit<…>` on the inherited `@sentry/core` `Options`/`ClientOptions`.

`dataCollection` was introduced upstream in `@sentry/core` v10 as the granular replacement for the (now deprecated) `sendDefaultPii`, and it is inherited into the RN options type surface automatically with the JS SDK bump to 10.60.0. However, it is **not** wired up in React Native:

- The native SDKs (iOS/Android) only read `sendDefaultPii` across the bridge; they do not understand `dataCollection`.
- RN-specific JS gates also read raw `sendDefaultPii` and ignore core's resolved data-collection settings — e.g. `infer_ip` metadata and `addAutoIpAddressToSession` (`client.ts`), and the URL/route-param gating in `reactnavigation.ts`, `expoRouter.ts`, and `deeplink.ts`.

So a user who set `dataCollection` would get a silent partial no-op: native and the RN JS gates would not honor it. Hiding it keeps the option surface honest until full support lands. Users should continue to use `sendDefaultPii`.

Spotted while reviewing the JS SDK bump to 10.60.0. The bump auto-exposes `dataCollection` in the RN types even though no layer honors it, which would be a silent footgun. This is the low-risk interim move; full `dataCollection` support (native + the RN JS gates) is tracked separately and the option will be re-exposed there.

- `yarn build:sdk` passes
- `yarn api-report` regenerated — only change is the `Omit` on `ReactNativeOptions`

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- Re-expose `dataCollection` once the native SDKs and the RN-specific `sendDefaultPii` gates support it.


alwx (approved):
Looks good, let's ship it!

PR #6346, opened 2026-06-25
fix(replay): Guard against non-dictionary touch breadcrumb path elements

### LINKED ISSUE #6342: Session Replay: fatal NSInvalidArgumentException in RNSentryReplayBreadcrumbConverter (objectForKey: on non-dictionary touch-path element)

React Navigation, Hermes, RN New Architecture, Expo Application Services (EAS), Expo (mobile only), Expo Router

sentry.io (SaS)

No

_No response_

8.0.0 (unguarded code also present on main / 8.15.x)

- @sentry/react-native: 8.0.0
- react-native: 0.81.5
- Expo SDK 54 (expo 54.0.33), Hermes enabled
- New Architecture (Fabric + TurboModules): enabled
- Affected platform: iOS (native crash); seen on iOS 17–26.x, physical devices (iPhone15,3)
- Package manager: pnpm (monorepo)

### Sentry.init()

```js
Sentry.init({
  dsn: 'https://<redacted>@<org>.ingest.sentry.io/<project>',
  integrations: [Sentry.reactNavigationIntegration(), Sentry.mobileReplayIntegration()],
  replaysSessionSampleRate: 0.03,
});
```

### Steps to Reproduce

1. Initialize the SDK with Session Replay enabled (mobileReplayIntegration + a non-zero replaysSessionSampleRate) on iOS.
2. Let the app record touch interactions while a replay segment is being captured.
3. When a touch breadcrumb's `path` array contains a non-dictionary element, the native replay breadcrumb converter crashes.

(Intermittent — depends on the touch-breadcrumb payload shape. Observed in production across multiple users; not a controlled repro.)

### Expected Result

The replay breadcrumb converter should skip/ignore a malformed touch-path element instead of crashing the app.

### Actual Result

Fatal `NSInvalidArgumentException` on the replay's background dispatch queue → uncaught → SIGABRT.

```
NSInvalidArgumentException: -[__NSCFString objectForKey:]: unrecognized selector sent to instance
  +[RNSentryReplayBreadcrumbConverter getTouchPathMessageFrom:]
  -[RNSentryReplayBreadcrumbConverter convertTouch:]
  -[RNSentryReplayBreadcrumbConverter convertFrom:]
  SentrySessionReplay.convertBreadcrumbs (SentrySessionReplay.swift)
  SentrySessionReplay.captureSegment / processNewlyAvailableSegment
  SentryOnDemandReplay.createVideoInBackgroundWith (SentryOnDemandReplay.swift)
  _SentryDispatchQueueWrapperInternal dispatchAsyncWithBlock:
```

**Root cause** — in `packages/core/ios/RNSentryReplayBreadcrumbConverter.m`, `getTouchPathMessageFrom:` does:

```objc
NSDictionary *item = [path objectAtIndex:i];
if (item == nil) { return nil; }
id name = [item objectForKey:@"name"];   // crashes if `item` is an NSString
```

There is a `nil` check but no `isKindOfClass:[NSDictionary class]` guard; `convertTouch:` likewise subscripts `breadcrumb.data` guarded only against `nil`. A non-dictionary element throws `unrecognized selector`. Verified present in **8.0.0 and on `main` (latest)**, so an SDK upgrade does not fix it. `convertNavigation:` already does the `isKindOfClass:` check — applying the same guard here resolves it.

Impact: fatal crash, observed 45 times across 3 users in production.

Workaround: disabling Session Replay (removing `mobileReplayIntegration()` and the replay sample rates) avoids it, since the converter only runs while replay is active.

antonis:
## 📜 Description

Fixes a fatal `NSInvalidArgumentException` crash in `RNSentryReplayBreadcrumbConverter` (Session Replay on iOS):

```
-[__NSCFString objectForKey:]: unrecognized selector sent to instance
```

`getTouchPathMessageFrom:` iterated over a touch breadcrumb's `path` array and treated every element as an `NSDictionary`, only guarding against `nil`. When an element was a non-dictionary (e.g. an `NSString`), calling `objectForKey:` on it raised `NSInvalidArgumentException` → `SIGABRT`.

Two gaps are addressed:

1. **`getTouchPathMessageFrom:`** — each element is now validated with `isKindOfClass:[NSDictionary class]` before any dictionary access (the same pattern already used in `convertNavigation:`).
2. **`convertTouch:`** — now guards that `path` is an `NSArray` before use, matching the existing `convertMultiClick:` implementation. Previously a non-array `path` would crash on `[path count]` (`NSString` does not respond to `count`).

## 💡 Motivation and Context

Reported in #6342 — observed as a fatal crash in production on iOS with Session Replay enabled (RN 0.81.5, Expo SDK 54, New Architecture + Hermes). The malformed touch-path element bypassed all existing guards.

Closes #6342

## 💚 How did you test it?

Added unit tests to `RNSentryReplayBreadcrumbConverterTests.swift` that reproduce the crash and assert it no longer occurs:

- `testTouchMessageReturnsNilOnNonDictionaryPathElement` — the exact reported case (a string element).
- `testTouchMessageReturnsNilWhenAnyPathElementIsNotADictionary` — non-dict element mixed with valid dicts.
- `testConvertTouchBreadcrumbWithNonDictionaryPathElementDoesNotCrash` — exercised end-to-end via `convert(from:)`.
- `testConvertTouchBreadcrumbWithNonArrayPathDoesNotCrash` — non-array `path`.

These crash on the previous code and pass with the fix. Ran clang-format on the modified ObjC file.

## 📝 Checklist

- [x] I added tests to verify the changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPii` is enabled.
- [x] All tests passing.
- [x] No breaking changes.

## 🔮 Next steps

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Fixes
 
-- Fix fatal `NSInvalidArgumentException` crash in `RNSentryReplayBreadcrumbConverter` when a touch breadcrumb path contains a non-dictionary element ([#6342](https://github.com/getsentry/sentry-react-native/issues/6342))
+- Fix fatal `NSInvalidArgumentException` crash in `RNSentryReplayBreadcrumbConverter` when a touch breadcrumb path contains a non-dictionary element ([#6346](https://github.com/getsentry/sentry-react-native/pull/6346))
 - Apply `SENTRY_ENVIRONMENT`, `SENTRY_RELEASE` and `SENTRY_DIST` build-time overrides to the JS bundled options to match the native SDKs ([#6330](https://github.com/getsentry/sentry-react-native/pull/6330))
 - Fix user `geo` being dropped from the native scope by forwarding it as a structured object instead of a JSON string ([#6309](https://github.com/getsentry/sentry-react-native/pull/6309))
 - Remove unused `React/RCTTextView.h` import that broke iOS builds on React Native 0.87, where the header was removed as part of the legacy architecture cleanup ([#6322](https://github.com/getsentry/sentry-react-native/pull/6322))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Fixes
 
+- Fix fatal `NSInvalidArgumentException` crash in `RNSentryReplayBreadcrumbConverter` when a touch breadcrumb path contains a non-dictionary element ([#6342](https://github.com/getsentry/sentry-react-native/issues/6342))
 - Apply `SENTRY_ENVIRONMENT`, `SENTRY_RELEASE` and `SENTRY_DIST` build-time overrides to the JS bundled options to match the native SDKs ([#6330](https://github.com/getsentry/sentry-react-native/pull/6330))
 - Fix user `geo` being dropped from the native scope by forwarding it as a structured object instead of a JSON string ([#6309](https://github.com/getsentry/sentry-react-native/pull/6309))
 - Remove unused `React/RCTTextView.h` import that broke iOS builds on React Native 0.87, where the header was removed as part of the legacy architecture cleanup ([#6322](https://github.com/getsentry/sentry-react-native/pull/6322))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### Fixes
 
+- Fix fatal `NSInvalidArgumentException` crash in `RNSentryReplayBreadcrumbConverter` when a touch breadcrumb path contains a non-dictionary element ([#6342](https://github.com/getsentry/sentry-react-native/issues/6342))
```
```suggestion
- Fix fatal `NSInvalidArgumentException` crash in `RNSentryReplayBreadcrumbConverter` when a touch breadcrumb path contains a non-dictionary element ([#6346](https://github.com/getsentry/sentry-react-native/pull/6346))
```

alwx (approved):
Looks good to me

PR #6318, opened 2026-06-18
feat(core): Capture Expo Router ErrorBoundary errors

alwx:
## :loudspeaker: Type of change
- [x] New feature

## :scroll: Description
Adds `Sentry.wrapRouterErrorBoundary` — wraps Expo Router's per-route `ErrorBoundary` so render-phase errors that hit the fallback are captured with route context (`route.name`, `route.path`, `route.params`), the active navigation transaction is marked errored, and a breadcrumb is emitted. Concrete path/params are gated behind `sendDefaultPii`.

## :bulb: Motivation and Context
Closes #6160.

## :green_heart: How did you test it?
New unit tests in `expoRouterErrorBoundary.test.tsx`. Full suite green locally.

## :pencil: Checklist
- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Investigate Metro/Babel auto-instrumentation of `export { ErrorBoundary } from 'expo-router'` re-exports as a follow-up.

PR #6347, opened 2026-06-25
Expo Router ErrorBoundary auto wrapped

alwx:

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Adds a Babel plugin (sentryExpoRouterAutoWrapBabelPlugin) that auto-wraps Expo Router's per-route ErrorBoundary re-exports at build time, removing the manual call-site step introduced in #6318.

 The plugin detects this in user source files:

 ```ts
   export { ErrorBoundary } from 'expo-router';
 ```

and rewrites it to:

 ```ts
   import { ErrorBoundary as __sentryOriginalExpoErrorBoundary } from 'expo-router';
   import { wrapExpoRouterErrorBoundary as __sentryWrapExpoRouterErrorBoundary } from '@sentry/react-native';
   export const ErrorBoundary = __sentryWrapExpoRouterErrorBoundary(__sentryOriginalExpoErrorBoundary);
 ```

 So users get the full ErrorBoundary instrumentation (route context, navigation-span error tagging, breadcrumb) with zero code changes in their route files.

#6318 shipped Sentry.wrapExpoRouterErrorBoundary but required users to edit every route file that re-exports ErrorBoundary from expo-router. That edit is trivial & easy to forget on new routes.

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@
   ```
 
   Render-phase errors that reach the boundary are captured with route context (`route.name`, `route.path`, `route.params`), the in-flight navigation transaction is tagged as errored, and a breadcrumb is emitted. Concrete paths and params are gated behind `sendDefaultPii`.
+- Auto-wrap Expo Router's `ErrorBoundary` re-exports at build time
+
+  `getSentryExpoConfig` now ships a Babel plugin that rewrites `export { ErrorBoundary } from 'expo-router'` into a wrapped re-export, so the boundary is auto-instrumented without changing your route files. Opt out with `autoWrapExpoRouterErrorBoundary: false`. For non-Expo Metro setups, `withSentryConfig` exposes the same option (off by default).
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,20 @@
 ### Features
 
 - Use the runtime's native `btoa` for envelope base64 encoding when available, to improve `captureEnvelope` performance. Falls back to the bundled JS encoder if `btoa` is missing ([#6351](https://github.com/getsentry/sentry-react-native/pull/6351)).
+- Opt-in build-time auto-wrap for Expo Router's per-route `ErrorBoundary` ([#6347](https://github.com/getsentry/sentry-react-native/pull/6347))
+
+  Enable `autoWrapExpoRouterErrorBoundary: true` in `getSentryExpoConfig` (or `withSentryConfig`) and the Sentry Babel plugin rewrites `export { ErrorBoundary } from 'expo-router'` into a `wrapExpoRouterErrorBoundary` call at build time — no route-file edits needed:
+
+  ```js
+  // metro.config.js
+  module.exports = getSentryExpoConfig(__dirname, {
+    autoWrapExpoRouterErrorBoundary: true,
+  });
+  ```
+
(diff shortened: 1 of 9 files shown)


packages/core/src/js/tools/metroconfig.ts
```diff
@@ -154,8 +169,13 @@ export function getSentryExpoConfig(
   });
 
   let newConfig = withSentryFramesCollapsed(config);
-  if (options.annotateReactComponents) {
-    newConfig = withSentryBabelTransformer(newConfig, options.annotateReactComponents);
+  const autoWrapExpoRouterErrorBoundary = options.autoWrapExpoRouterErrorBoundary ?? true;
```
antonis: Should this default to `false`?
```suggestion
  const autoWrapExpoRouterErrorBoundary = options.autoWrapExpoRouterErrorBoundary ?? false;
```

packages/core/src/js/tools/sentryBabelTransformerUtils.ts
```diff
@@ -97,7 +100,12 @@ export function createSentryBabelTransformer(): BabelTransformer {
   const transform: BabelTransformer['transform'] = (...args) => {
     const transformerArgs = args[0];
 
-    addSentryComponentAnnotatePlugin(transformerArgs, options?.annotateReactComponents);
+    if (options?.annotateReactComponents !== undefined) {
```
antonis, line 103: q: Gating on `options?.annotateReactComponents` makes sense. Just wanted to flag the behavior change. Should we communicate this in the changelog or docs?

packages/core/src/js/tools/sentryExpoRouterAutoWrapBabelPlugin.ts
```diff
@@ -0,0 +1,120 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
antonis: nit: I think we can remove the eslint-disable here
```suggestion
```

CHANGELOG.md
```diff
@@ -23,6 +23,9 @@
   ```
 
   Render-phase errors that reach the boundary are captured with route context (`route.name`, `route.path`, `route.params`), the in-flight navigation transaction is tagged as errored, and a breadcrumb is emitted. Concrete paths and params are gated behind `sendDefaultPii`.
+- Auto-wrap Expo Router's `ErrorBoundary` re-exports at build time
```
antonis: We should reformat the changelog and probably shorten it or add an example. Let's also add the PR ref to make the check happy

antonis (commented):
Left a few comments but overall looks good 🚀

packages/core/src/js/tools/metroconfig.ts
alwx: I think the main argument that made me wanna go with `true` is the discoverability — if it's `false` then 99% of users will never know about this feature. But maybe it's the right the thing to do. What do you think?

alwx:
@antonis I've rebased and checked the Sentry Warden / Cursor / whatever-else comments. Please, check again! :)

antonis (approved):
LGTM 🚀 
Let's wait for the CI to become 🟢 to be on the safe side 🙇

PR #6351, opened 2026-06-25
feat(core): Use native btoa for envelope base64 encoding

### LINKED ISSUE #4884: Use CPP Base64 Encoder for better performance (>10x)

alwx:
## 📢 Type of change

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## 📜 Description

Switch envelope base64 encoding on the `RNSentry.captureEnvelope` hot path from the bundled JS encoder to the runtime's native `btoa`. Hermes and modern JSC both expose `btoa`, and the native implementation should be significantly faster than the pure-JS encoder for the envelope payloads going through the bridge (profiles, attachments, replays).

The comes with no native code and no breaking changes for the customers.

## 💡 Motivation and Context

Closes getsentry/sentry-react-native#4884.  <br>Alternative approach to getsentry/sentry-react-native#6314 (which should be closed now)

## 💚 How did you test it?

* Added `test/utils/base64.test.ts`

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/packages/core/src/js/utils/base64.ts
+++ b/packages/core/src/js/utils/base64.ts
@@ -6,10 +6,10 @@ const CHUNK_SIZE = 0x8000;
 
 /**
  * Encodes a byte array to base64. Uses the runtime's native `btoa` when
- * available (Hermes and modern JSC both expose it), which is significantly
- * faster than the pure-JS encoder for the envelope payloads going through
- * `RNSentry.captureEnvelope`. Falls back to the bundled JS encoder when
- * `btoa` is missing (e.g. older JS engines).
+ * available — Hermes exposes it natively, and some JSC builds do too — which
+ * is significantly faster than the pure-JS encoder for the envelope payloads
+ * going through `RNSentry.captureEnvelope`. Falls back to the bundled JS
+ * encoder when `btoa` is missing.
  */
 export function encodeToBase64(bytes: Uint8Array): string {
   const nativeBtoa = (globalThis as { btoa?: (input: string) => string }).btoa;
--- a/packages/core/src/js/utils/base64.ts
+++ b/packages/core/src/js/utils/base64.ts
@@ -1,0 +1,25 @@
+import { base64StringFromByteArray } from '../vendor';
+
+// Chunk size for `String.fromCharCode.apply` — keeps argument count well below
+// engine limits while still amortising the call overhead across many bytes.
+const CHUNK_SIZE = 0x8000;
+
+/**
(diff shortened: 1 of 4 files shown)


alwx:
@cursor review

packages/core/test/utils/base64.test.ts
antonis, line 1: nit: Not sure if this is really testable but if it is it would be nice to test both the btoa available/unavailable cases

antonis (approved):
Added two nits but LGTM 🚀

packages/core/src/js/utils/base64.ts
```diff
@@ -0,0 +1,25 @@
+import { base64StringFromByteArray } from '../vendor';
+
+// Chunk size for `String.fromCharCode.apply` — keeps argument count well below
+// engine limits while still amortising the call overhead across many bytes.
+const CHUNK_SIZE = 0x8000;
+
+/**
+ * Encodes a byte array to base64. Uses the runtime's native `btoa` when
+ * available (Hermes and modern JSC both expose it), which is significantly
```
antonis: nit: I think JSC does not always expose btoa

Issue #6174, opened 2026-05-18
Document the SPM migration playbook for end users


alwx:
Once RNSentry publishes a `Package.swift` and the React Native ecosystem supports SPM autolinking, end users will need clear guidance on:

* Whether to use CocoaPods or SPM for a given app.
* How to migrate an existing CocoaPods-based app to SPM.
* What the support matrix looks like (which RN versions support each path).
* What to do with Expo apps, brownfield iOS apps, and edge cases.

Without explicit guidance, users will hit the transition in an ad hoc way and surface confused support tickets to us.

ismumchow:
## Hitting `Sentry.xcframework-ios.signature couldn't be copied to Signatures`?

  This bites archive builds (CI / TestFlight / Firebase distribution) but **not** `xcodebuild build` or Xcode → Run, because only
  archive packaging consolidates signatures into `<App>.app/Signatures/`. If it works on a device locally but breaks in CI, you're
  hitting this.

  **Why:** With `SENTRY_USE_SPM=1`, RN's `cocoapods/spm.rb` adds Sentry as an SPM product dep on the **RNSentry pod target** in
  addition to your app target's transitive consumption. Xcode collects `Sentry.xcframework-ios.signature` once per target → two files,
  same destination, collision.

  **Workaround** (drop into your Podfile `post_install`):

  ```ruby
  rnsentry = installer.pods_project.targets.find { |t| t.name == 'RNSentry' }
  if rnsentry
    phase_name = 'Strip duplicate Sentry.xcframework signature'
    unless rnsentry.shell_script_build_phases.any? { |p| p.name == phase_name }
      phase = rnsentry.new_shell_script_build_phase(phase_name)
      phase.shell_script = "rm -f \"$BUILT_PRODUCTS_DIR/Sentry.xcframework-ios.signature\" || true\n"
    end
  end

  The app target's canonical signature still lands in the archive. RNSentry compiles normally, Sentry stays statically linked.

  Caveats:
  - Workaround, not a fix — track getsentry/sentry-react-native#6174 for the official RN-SPM playbook and remove this once it ships.
  - Verified on Xcode 26.3 / @sentry/react-native 8.13 / RN 0.80. Worth re-checking on any of those bumps.
  - Fails loud, not silent: if it ever stops applying, you get the same original error back.

PR #6381, opened 2026-06-30
feat(ios): Default to consuming sentry-cocoa as an xcframework

alwx:

- [x] Enhancement

## :scroll: Description

Vendor `sentry-cocoa` as a prebuilt `Sentry-Dynamic.xcframework` (downloaded from the GitHub release and SHA256-verified at `pod install` time), instead of building `Sentry` from source as a CocoaPod. Sidesteps the Xcode 16/26 archive bug that hits when the same xcframework is consumed through Xcode's SPM integration (`"Sentry-Dynamic.xcframework-ios.signature" couldn't be copied to "Signatures"…`).

Override with:
- `SENTRY_USE_XCFRAMEWORK=0` — fall back to the source-built `Sentry` CocoaPod (e.g. offline builds behind a restrictive proxy).

CocoaPods source builds of `sentry-cocoa` are slow and diverge from where the RN community is moving (prebuilt binaries). SPM was the first attempt, but Xcode 16/26 archives fail with a duplicate-signature error on any signed SPM binary xcframework. Consuming the same xcframework through CocoaPods' `vendored_frameworks` avoids the Xcode-SPM archive pipeline and works on the full RN version range.

CI: sample-application (iOS matrix, both `xcframework` default and `cocoapods` fallback), iOS Size Analysis (archive), metrics (fastlane archive).

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,11 +48,11 @@
 ### Changes
 
 - Default `mobileReplayIntegration({ networkCaptureBodies })` to `true`, matching the iOS and Android native SDK defaults ([#6372](https://github.com/getsentry/sentry-react-native/pull/6372))
-- Consume `sentry-cocoa` via Swift Package Manager by default on React Native >= 0.75 ([#6381](https://github.com/getsentry/sentry-react-native/pull/6381))
+- Consume `sentry-cocoa` as a prebuilt xcframework by default on iOS ([#6381](https://github.com/getsentry/sentry-react-native/pull/6381))
 
   **Warning**
 
-  **This may be a breaking change for some setups.** On RN 0.75+, `pod install` now pulls Sentry as a pre-built binary xcframework, instead of building Sentry from source as a CocoaPod. If your iOS build breaks after upgrading (e.g. when another pod also depends on the `Sentry` CocoaPod), set `SENTRY_USE_SPM=0` before `pod install` to restore the previous behavior. On older React Native versions the SPM helper is unavailable and RNSentry transparently falls back to the CocoaPods source build. Use `SENTRY_USE_SPM=1` to require SPM.
+  **This may be a breaking change for some setups.** `pod install` now downloads `Sentry-Dynamic.xcframework` from sentry-cocoa's GitHub release (SHA256-verified) and vendors it, instead of building Sentry from source as a CocoaPod. If your iOS build breaks after upgrading (e.g. when another pod also depends on the `Sentry` CocoaPod), or if your `pod install` environment cannot reach `github.com`, set `SENTRY_USE_XCFRAMEWORK=0` before `pod install` to restore the previous source-build behavior.
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Changes
+
+- Consume `sentry-cocoa` as a prebuilt xcframework by default on iOS ([#6381](https://github.com/getsentry/sentry-react-native/pull/6381))
+
+  **Warning**
+
+  **This may be a breaking change for some setups.** `pod install` now downloads `Sentry.xcframework` from sentry-cocoa's GitHub release (SHA256-verified) and vendors it, instead of building Sentry from source as a CocoaPod. If your iOS build breaks after upgrading (e.g. when another pod also depends on the `Sentry` CocoaPod), or if your `pod install` environment cannot reach `github.com`, set `SENTRY_USE_XCFRAMEWORK=0` before `pod install` to restore the previous source-build behavior.
+
 ## 8.17.2
 
 ### Fixes
@@ -48,11 +58,6 @@
 ### Changes
 
 - Default `mobileReplayIntegration({ networkCaptureBodies })` to `true`, matching the iOS and Android native SDK defaults ([#6372](https://github.com/getsentry/sentry-react-native/pull/6372))
-- Consume `sentry-cocoa` as a prebuilt xcframework by default on iOS ([#6381](https://github.com/getsentry/sentry-react-native/pull/6381))
-
-  **Warning**
-
-  **This may be a breaking change for some setups.** `pod install` now downloads `Sentry.xcframework` from sentry-cocoa's GitHub release (SHA256-verified) and vendors it, instead of building Sentry from source as a CocoaPod. If your iOS build breaks after upgrading (e.g. when another pod also depends on the `Sentry` CocoaPod), or if your `pod install` environment cannot reach `github.com`, set `SENTRY_USE_XCFRAMEWORK=0` before `pod install` to restore the previous source-build behavior.
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,9 +48,11 @@
 ### Changes
 
 - Default `mobileReplayIntegration({ networkCaptureBodies })` to `true`, matching the iOS and Android native SDK defaults ([#6372](https://github.com/getsentry/sentry-react-native/pull/6372))
-- Consume `sentry-cocoa` via Swift Package Manager by default on React Native >= 0.75 ([#5780](https://github.com/getsentry/sentry-react-native/issues/5780))
+- Consume `sentry-cocoa` via Swift Package Manager by default on React Native >= 0.75 ([#6381](https://github.com/getsentry/sentry-react-native/pull/6381))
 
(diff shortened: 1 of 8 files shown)


alwx:
@cursor review

alwx:
It would probably be nice to include it to the next release but no stress

CHANGELOG.md
```diff
@@ -27,6 +27,9 @@
 ### Changes
 
 - Default `mobileReplayIntegration({ networkCaptureBodies })` to `true`, matching the iOS and Android native SDK defaults ([#6372](https://github.com/getsentry/sentry-react-native/pull/6372))
+- Consume `sentry-cocoa` via Swift Package Manager by default on React Native >= 0.75 ([#5780](https://github.com/getsentry/sentry-react-native/issues/5780))
```
antonis: nit: Let's add the PR here

```suggestion
- Consume `sentry-cocoa` via Swift Package Manager by default on React Native >= 0.75 ([#5780](https://github.com/getsentry/sentry-react-native/pull/6381))
```

CHANGELOG.md
```diff
@@ -27,6 +27,9 @@
...
+
+  On RN 0.75+, `pod install` now pulls Sentry as a pre-built binary xcframework through the SPM helper exposed by `react_native_pods.rb`, instead of building Sentry from source as a CocoaPod. On older React Native versions the SPM helper is unavailable and RNSentry transparently falls back to the CocoaPods source build. Override with `SENTRY_USE_SPM=0` to force CocoaPods, or `SENTRY_USE_SPM=1` to require SPM.
```
antonis: I was thinking that we could highlight that this might be a breaking change for some users and suggest the solution. E.g.
```suggestion
  On RN 0.75+, `pod install` now pulls Sentry as a pre-built binary xcframework, instead of building Sentry from source as a CocoaPod. If your iOS build breaks after upgrading (e.g. when another pod also depends on the `Sentry` CocoaPod), set `SENTRY_USE_SPM=0` before `pod install` to restore the previous behavior.
```

antonis (commented):
Overall LGTM other than the failed checks and the changelog suggestions to highlight that this might be a breaking change for some users. Maybe we can even add a warning at the top (e.g. like iOS does [here](https://github.com/getsentry/sentry-cocoa/releases/tag/9.19.0)).

antonis:
> Overall LGTM other than the failed checks and the changelog suggestions to highlight that this might be a breaking change for some users. Maybe we can even add a warning at the top (e.g. like iOS does [here](https://github.com/getsentry/sentry-cocoa/releases/tag/9.19.0)).

We can probably also add a link to the new doc that is part of https://github.com/getsentry/sentry-react-native/issues/6174

CHANGELOG.md
alwx: Changed it a bit.

antonis (commented):
Since the dynamic approach didn't resolve the CI build issue and this partially a breaking change for the users I suggest to try the following:
- Keep SPM opt-in till the React Native framework supports it
- Have RNSentry.podspec consume sentry-cocoa as a vendored_frameworks xcframework downloaded from [sentry-cocoa's GitHub releases](https://github.com/getsentry/sentry-cocoa/releases/tag/9.19.1), instead of either s.dependency 'Sentry' (CocoaPods trunk) or SPM.dependency (spm.rb bridge).
Wdyt?

antonis (commented):
Thank you for iterating on this Alex. Tests look good now 🎉 

Since our plan is to ship the change in a minor it might also make sense to test how this would roll out with an alpha or beta release (even if it uncommon for a minor version).

I've also tried to find scenarios that need to be tested further before shipping:

  1. EAS Build — bare and managed/CNG, on npm and pnpm, including a second build after a version bump (EAS dependency caching), and confirm SENTRY_USE_XCFRAMEWORK=0 via eas.json env reverts. Biggest passive-upgrade population,
  entirely untested.
  2. Symbolication — native crash → dSYM upload → confirm Sentry frames resolve with the vendored xcframework's dSYMs. Silent regression here defeats the SDK's core purpose.
  3. Read-only / pnpm-strict node_modules — the download writes into node_modules/@sentry/react-native/ios/Vendor; verify pnpm strict store / Yarn PnP don't fail on a read-only or symlinked package dir. (Overlaps with #1's pnpm leg —
  can be covered together.)
  4. Signed device archive → TestFlight — confirm the vendored framework is accepted for distribution (beyond CI's CODE_SIGNING_ALLOWED=NO archive).
  5. RN < 0.75 end-to-end — the default is now on for the whole RN range and the Swift stub is gated off there; confirm build + archive on an old RN.

packages/core/scripts/sentry_utils.rb
```diff
@@ -40,3 +40,135 @@ def should_use_folly_flags(rn_version)
...
+# sentry-cocoa version we ship a prebuilt xcframework for.
+SENTRY_COCOA_XCFRAMEWORK_CHECKSUMS = {
+  '9.19.1' => {
+    # `Sentry.xcframework.zip` — the static product. Its enclosing xcframework
+    # name matches the framework name inside (both `Sentry`), which CocoaPods
+    # requires to generate `-framework Sentry` correctly and to resolve the
+    # `Sentry` module. `Sentry-Dynamic.xcframework` would ship the same
+    # `Sentry.framework` inside but under a mismatched enclosing name, so
+    # CocoaPods generates `-framework Sentry-Dynamic` and fails at link.
+    'Sentry' => 'd6d545af17e49851cda2747b0f45cde78ce08ea37709dde5a956c6b4671224e8',
```
itaybre, line 64: `l`: It might be a good idea to have a script to bump this

packages/core/scripts/sentry_utils.rb
```diff
@@ -40,3 +40,135 @@ def should_use_folly_flags(rn_version)
...
+  # Wipe any stale partial extract from a previous interrupted run so we
+  # always start from a clean tree.
+  FileUtils.rm_rf(target_dir)
+  FileUtils.mkdir_p(vendor_dir)
+  zip_path = File.join(vendor_dir, "#{product}.xcframework.zip")
+  url = "https://github.com/getsentry/sentry-cocoa/releases/download/" \
+        "#{version}/#{product}.xcframework.zip"
+
+  Pod::UI.puts "[Sentry] Downloading #{product} #{version} from GitHub Releases…" if defined?(Pod::UI)
+  unless system('curl', '-sSfL', '-o', zip_path, url)
```
itaybre: `m`: Could we use `open-uri` to download the file?
I would prefer to avoid using `system` if possible.

I believe open-uri is installed by default in all compatible ruby version

packages/core/RNSentry.podspec
```diff
@@ -45,41 +45,102 @@ Pod::Spec.new do |s|
...
+    # Point the search paths at the pod-install-time absolute path to the
+    # xcframework. `${PODS_TARGET_SRCROOT}` is only defined in per-pod
+    # xcconfigs, not in aggregate/user-target xcconfigs, and a
+    # `${PODS_ROOT}`-relative fallback works for one Podfile layout but
+    # breaks for another (e.g. the RN sample apps put node_modules at a
+    # different depth from RNSentryCocoaTester). Using the absolute path
+    # avoids the layout-detection dance — the path is regenerated on
+    # every `pod install`, so it's not something anyone commits.
+    xcframework_search_paths = {}
+    sentry_xcframework_slices_by_sdk(sentry_xcframework_dir).each do |sdk, slice_ids|
```
itaybre: `l`: This is fixed, so no need to check for the content of the xcframework.

itaybre (commented):
Looks good, just some small comments.

alwx:
@antonis do you think it makes sense to create a beta with this? The thing is EAS (number 1) and "Signed device archive → TestFlight" (number 4) require the release to be published and I think a beta release could be a way to approach this.

Also: there is a fallback of SENTRY_USE_XCFRAMEWORK=0 in the worst case anyway.

antonis:
> @antonis do you think it makes sense to create a beta with this? The thing is EAS (number 1) and "Signed device archive → TestFlight" (number 4) require the release to be published and I think a beta release could be a way to approach this.

@alwx Yes, I think we should release an alpha or beta to be able to test this change. The EAS scenario is probably the most common one and we should avoid causing any breakage since this is aiming a minor and nor a major release

> Also: there is a fallback of SENTRY_USE_XCFRAMEWORK=0 in the worst case anyway.

Fair, but if users need to change their setup it's a breaking change and cannot be shipped in a minor. Would that option work for the next versions of the Cocoa SDK that do not support cocoapods?

alwx:
in this case let's do an alpha or beta<br>my only concern with those is that nobody seems to be using them but we can try

antonis:
> my only concern with those is that nobody seems to be using them but we can try

I guess we can try it even just for testing the EAS flow internally

PR #6307, opened 2026-06-17
feat(core): Wire TurboModulePerfLogger on iOS and Android

alwx:
## :loudspeaker: Type of change
- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Install a Sentry-owned `facebook::react::NativeModulePerfLogger` on both platforms so the SDK observes every TurboModule lifecycle event:

- `moduleDataCreate{Start,End}`, `moduleCreate{Start,CacheHit,Construct*,SetUp*,End,Fail}`
- `moduleJSRequireBeginning*`, `moduleJSRequireEnding*`
- `syncMethodCall{Start,ArgConversion*,Execution*,ReturnConversion*,End,Fail}`
- `asyncMethodCall{Start,ArgConversion*,Dispatch,End,Fail}`
- `asyncMethodCallBatchPreprocess{Start,End}`
- `asyncMethodCallExecution{Start,ArgConversion*,End,Fail}`

This is the foundation that the next three issues in the [Turbo Modules instrumentation project](https://linear.app/getsentry/project/turbo-modules-instrumentation-improvements-react-native-11aa6f771dd5/overview) build on: JS↔Native crash attribution, per-Turbo-Module spans, and aggregated per-module stats. Each will ship its own `ISentryTurboModulePerfSink` implementation and plug into the hook this PR exposes.

New `enableTurboModuleTracking` option on `Sentry.init`, **default `false`** for this first release so the foundation lands without behavioural change. The native logger is always installed (we never want to miss early lifecycle events); the flag only decides whether forwarded callbacks reach the sink. The option is plumbed through `initNativeSdk` on both platforms.

```ts
Sentry.init({
  dsn: "___PUBLIC_DSN___",
  enableTurboModuleTracking: true, // off by default — will activate the follow-up sinks once they ship
});
```

## :bulb: Motivation and Context

Closes #6162.

## :pencil: Checklist

- [ ] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] I updated the docs if needed.
  - Docs deferred until a sink ships and the option becomes user-visible.
- [ ] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

This is the **foundation** for the Turbo Modules project. Follow-up issues plug in `ISentryTurboModulePerfSink` implementations:

[...]

PR #6396, opened 2026-07-03
fix(android): Force 16 KB ELF alignment for libsentry-tm-perf-logger.so

### LINKED ISSUE #6394: libsentry-tm-perf-logger.so is not 16 KB page aligned after upgrading from 8.16.0 to 8.17.0

Expo (mobile only), Expo Router, Hermes, RN New Architecture, React Navigation

sentry.io (SaS)

No

8.16.0 (old), 8.17.0 (new). The issue started right after the upgrade.

````
System:
  OS: macOS 26.5.2
  CPU: (15) arm64 Apple M5 Pro
Binaries:
  Node: 24.16.0
  npm: 11.16.0
SDKs:
  Android SDK: build-tools 37.0.0
npmPackages:
  expo: 57.0.1
  react: 19.2.3
  react-native: 0.86.0
Android:
  newArchEnabled: true
  hermesEnabled: true
````

Emulator: Pixel 10 Pro AVD with the 16 KB page size system image (`sdk_gphone16k_arm64`, Android 17, API 37, arm64-v8a).

### Sentry.init()

````js
Sentry.init({
  dsn: "https://...@o...ingest.de.sentry.io/...",
  enabled: !__DEV__,
  environment: "production",
  enableTombstone: true,
});
````

### Steps to Reproduce

1. Upgrade `@sentry/react-native` from 8.16.0 to 8.17.0 in a New Architecture app
2. Build and install the app on an Android emulator with a 16 KB page size image
3. Launch the app

### Expected Result

No compatibility warning, all native libraries are 16 KB page aligned (as they were with 8.16.0).

### Actual Result

On every app start, Android shows the "Android App Compatibility" dialog saying the app is not 16 KB compatible and will run in page size compatible mode. It unhelpfully lists every library with "Unknown error", but I pulled the APK and checked the ELF headers of all 32 libs myself. Exactly one is misaligned, and it is the new one introduced in 8.17.0 by the Turbo Module performance tracking feature (#6307):

```
lib/arm64-v8a/libsentry-tm-perf-logger.so   LOAD align: 4096 (all other libs: 16384)
```

Logcat at install time:

```
NativeLibraryHelper: Setting page size compat mode PAGE_SIZE_APP_COMPAT_FLAG_ELF_NOT_ALIGNED
AppWarnings: Showing PageSizeMismatchDialog for package pl.tomekzaw.odjazdowykrk
```

The library is built from source in the consuming app by `android/src/main/jni/CMakeLists.txt`, which does not pass `-Wl,-z,max-page-size=16384`. RN 0.86 / Expo SDK 57 builds with NDK 27, where the linker still defaults to 4 KB alignment (16 KB only became the default in NDK r28), so the resulting `.so` breaks 16 KB compliance for the whole app. This will presumably also fail Google Play's 16 KB requirement, not just show the emulator dialog.

The usual fix would be adding this to the CMakeLists:

```cmake
target_link_options(sentry-tm-perf-logger PRIVATE "-Wl,-z,max-page-size=16384")
```

Screenshot from the emulator:

Android App Compatibility dialog listing libraries that are not 16 KB aligned

antonis:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
`libsentry-tm-perf-logger.so` — the native library introduced in 8.17.0 (#6307) for Turbo Module performance tracking — is compiled from source in the consuming app via `packages/core/android/src/main/jni/CMakeLists.txt`. That CMake target did not pass a `max-page-size` link option, so on NDK r27 and earlier (RN 0.86 / Expo SDK 57 ship with NDK 27) the linker defaulted to 4 KB ELF segment alignment. Every other `.so` bundled by RN/AGP is already 16 KB aligned, so this was the lone misaligned library in the APK.

This PR:

1. Adds `-Wl,-z,max-page-size=16384` to the CMake target (a no-op on NDK r28+, where 16 KB is already the default):

   ```cmake
   target_link_options(
       sentry-tm-perf-logger
       PRIVATE
           "-Wl,-z,max-page-size=16384"
   )
   ```

2. Adds a regression guard — `scripts/check-android-16kb-alignment.sh` unzips a built APK/AAB and asserts every bundled `.so` has ELF LOAD segments aligned to at least 16 KB, failing otherwise. It is wired into the `sample-application` workflow right after the New Architecture Android build (the only place a real APK is produced in CI).

Fixes #6394.

On Android 15+ with a 16 KB page size image, affected apps showed the "Android App Compatibility" dialog on every launch and ran in page-size-compat mode. This also fails Google Play's 16 KB requirement. It affected **all** New Architecture apps on RN ≥ 0.75, regardless of whether `enableTurboModuleTracking` was enabled — the library is packaged whenever the New Architecture is on, and Android's installer scans every bundled `.so` for alignment.

Compiled the actual library sources (`SentryTurboModulePerfLogger.cpp` + `OnLoad.cpp`) against the RN 0.86 source tree with **NDK 27** (the toolchain RN 0.86 / Expo SDK 57 use), both with and without the flag, and inspected the ELF program headers with `llvm-readelf -l`:

| Build | LOAD segment `p_align` | Result |
|-------|------------------------|--------|
| Without the flag (current `main`) | `0x1000` (4 KB) | ❌ reproduces the bug |
| With `-Wl,-z,max-page-size=16384` | `0x4000` (16 KB) | ✅ fixed |

Backward compatibility confirmed from the fixed binary's headers: every LOAD segment satisfies `Offset ≡ VirtAddr (mod 0x4000)`, which implies `mod 0x1000` — so it still loads correctly on 4 KB-page devices (16 KB is a multiple of 4 KB). No file-size change was observed for this library.

The new `check-android-16kb-alignment.sh` guard was also verified locally against two synthetic APKs — one containing the fixed lib (exit 0, pass) and one containing the pre-fix 4 KB lib (exit 1, fail, offending library reported), reproducing #6394 and confirming the check would catch a future regression.

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Fixes
+
+- Force 16 KB ELF alignment for `libsentry-tm-perf-logger.so` so it no longer breaks 16 KB page size compatibility on Android 15+ ([#6394](https://github.com/getsentry/sentry-react-native/issues/6394))
+
+  The native library added in 8.17.0 for Turbo Module performance tracking is built from source with the app's NDK. On NDK r27 and earlier the linker defaults to 4 KB segment alignment, leaving this library misaligned while every other bundled `.so` is already 16 KB aligned. This tripped Android's 16 KB page size compatibility check (and Google Play's 16 KB requirement) for New Architecture apps, even when `enableTurboModuleTracking` was not enabled. The CMake target now passes `-Wl,-z,max-page-size=16384` (a no-op on NDK r28+).
+
 ## 8.17.0
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Fixes
 
-- Force 16 KB ELF alignment for `libsentry-tm-perf-logger.so` so it does not breaks 16 KB page size compatibility on Android 15+ ([#6396](https://github.com/getsentry/sentry-react-native/pull/6396))
+- Force 16 KB ELF alignment for `libsentry-tm-perf-logger.so` so it does not break 16 KB page size compatibility on Android 15+ ([#6396](https://github.com/getsentry/sentry-react-native/pull/6396))
 
 ## 8.17.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,7 @@
 
 ### Fixes
 
-- Force 16 KB ELF alignment for `libsentry-tm-perf-logger.so` so it no longer breaks 16 KB page size compatibility on Android 15+ ([#6394](https://github.com/getsentry/sentry-react-native/issues/6394))
-
-  The native library added in 8.17.0 for Turbo Module performance tracking is built from source with the app's NDK. On NDK r27 and earlier the linker defaults to 4 KB segment alignment, leaving this library misaligned while every other bundled `.so` is already 16 KB aligned. This tripped Android's 16 KB page size compatibility check (and Google Play's 16 KB requirement) for New Architecture apps, even when `enableTurboModuleTracking` was not enabled. The CMake target now passes `-Wl,-z,max-page-size=16384` (a no-op on NDK r28+).
+- Force 16 KB ELF alignment for `libsentry-tm-perf-logger.so` so it does not breaks 16 KB page size compatibility on Android 15+ ([#6396](https://github.com/getsentry/sentry-react-native/pull/6396))
 
 ## 8.17.0
 
--- a/.github/workflows/sample-application.yml
+++ b/.github/workflows/sample-application.yml
@@ -188,7 +188,11 @@ jobs:
           ./scripts/build-android.sh -PreactNativeArchitectures=x86
 
       - name: Check 16 KB native library alignment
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,14 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Fixes
+
+- Force 16 KB ELF alignment for `libsentry-tm-perf-logger.so` so it no longer breaks 16 KB page size compatibility on Android 15+ ([#6394](https://github.com/getsentry/sentry-react-native/issues/6394))
+
+  The native library added in 8.17.0 for Turbo Module performance tracking is built from source with the app's NDK. On NDK r27 and earlier the linker defaults to 4 KB segment alignment, leaving this library misaligned while every other bundled `.so` is already 16 KB aligned. This tripped Android's 16 KB page size compatibility check (and Google Play's 16 KB requirement) for New Architecture apps, even when `enableTurboModuleTracking` was not enabled. The CMake target now passes `-Wl,-z,max-page-size=16384` (a no-op on NDK r28+).
```
```suggestion
- Force 16 KB ELF alignment for `libsentry-tm-perf-logger.so` so it does not breaks 16 KB page size compatibility on Android 15+ ([#6396](https://github.com/getsentry/sentry-react-native/pull/6396))
```

alwx (approved):

antonis:
Converting to draft to check the CI check failures

PR #6405, opened 2026-07-06
fix(ios): Skip source maps upload on Debug builds

### LINKED ISSUE #6399: SDK 8 / sentry-cli 3: local Debug iOS builds hard-fail without SENTRY_ORG ("An organization ID or slug is required") — was warn-and-continue on 7.x
### OS

macOS 15

### Platform

iOS

### SDK

`@sentry/react-native` (`sentry-expo` config plugin path: `@sentry/react-native/expo`)

### SDK version

8.17.1

### React Native version

0.86 (Expo SDK 57)

### Steps to reproduce

1. Expo SDK 57 project with `@sentry/react-native@8.17.1` and the `@sentry/react-native/expo` config plugin.
2. Developer machine with **no** Sentry auth configured — no `SENTRY_ORG` / `SENTRY_AUTH_TOKEN` env, no `sentry.properties` with an org (the common case for a teammate or CI leg that isn't meant to upload).
3. `npx expo prebuild -p ios` + `xcodebuild … -configuration Debug` (local simulator build).

### Expected result

Local Debug builds without upload credentials succeed, skipping source-map/debug-file upload — the behavior we observed on 7.x: the same project, same machine, same missing env built fine before the upgrade, only warning about the skipped upload.

### Actual result

The "Bundle React Native code and images" phase hard-fails the build:

```
error: sentry-cli - error: An organization ID or slug is required (provide with --org, set SENTRY_ORG, or use an org-scoped auth token)
```

So after upgrading, every developer who doesn't export Sentry credentials can no longer produce a local Debug build until they discover `SENTRY_DISABLE_AUTO_UPLOAD=true`.

### Workaround

`export SENTRY_DISABLE_AUTO_UPLOAD=true` before building (or setting it per-scheme). Works, but it's a hidden tribal-knowledge step — nothing in the upgrade guide flags that tokenless local builds now fail instead of warn.

### Ask

Either (a) restore warn-and-continue when no org/token is configured (at least for Debug configuration), or (b) call this out loudly in the v8 migration docs as a breaking change for local builds. The v3 sentry-cli strictness makes sense for release/CI paths where a silent upload skip is a real footgun — but failing the *local Debug* inner loop on a missing org feels like the wrong default.

Yes, if maintainers indicate which direction (a/b) is acceptable.

antonis:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
The iOS source-maps upload build phase (`sentry-xcode.sh`) ran `sentry-cli react-native xcode` for **every** configuration, including `Debug`. This adds a `*Debug*` configuration guard so Debug builds skip the upload and run the React Native bundling step directly.

The guard mirrors the two upload paths that already skip Debug:
- Native debug files upload — `sentry-xcode-debug-files.sh` (`grep -iq "debug"`)
- Android Gradle plugin — `sentry.gradle.kts` (`if (vName.contains("debug", ...)) return`)

so all three upload paths now share one convention: no upload on Debug.

Fixes #6399.

Since SDK 8 bumped the bundled `sentry-cli` from v2 to v3, `sentry-cli react-native xcode` **hard-fails** when no org/auth token is configured:

```
error: An organization ID or slug is required (provide with --org, set SENTRY_ORG, or use an org-scoped auth token)
```

On 7.x (sentry-cli v2) this warned and continued. Because the source-maps script had no Debug guard, every local Debug build without Sentry credentials now fails until the developer discovers `SENTRY_DISABLE_AUTO_UPLOAD=true`. Skipping upload on Debug restores a working local inner loop while keeping v3's strictness on Release/CI paths (where a silent upload skip would be a real footgun).

- Added two unit tests to `sentry-xcode-scripts.test.ts`: Debug skip (asserts a failing `sentry-cli` is **not** invoked and no `error:` is emitted) and case-insensitive `debug`.
- `yarn jest test/scripts/sentry-xcode-scripts.test.ts` — 38/38 pass.
- Manual repro with a v3-style "organization ID required" mock CLI: **Debug** exits `0` (bundles, skips upload); **Release** still exits `1` (hard-fails on missing org).

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps
Minor behavior note: anyone who was relying on iOS source-map upload during *Debug* builds will no longer get it automatically (Release builds and manual `sentry-cli` still work). This matches Android, which already skips Debug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Fixes
+
+- Skip iOS source maps upload on `Debug` builds ([#6399](https://github.com/getsentry/sentry-react-native/issues/6399))
+
 ## 8.17.1
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Fixes
 
-- Skip iOS source maps upload on `Debug` builds ([#6399](https://github.com/getsentry/sentry-react-native/issues/6399))
+- Skip iOS source maps upload on `Debug` builds ([#6405](https://github.com/getsentry/sentry-react-native/pull/6405))
 
 ## 8.17.1
 
--- a/packages/core/scripts/sentry-xcode.sh
+++ b/packages/core/scripts/sentry-xcode.sh
@@ -65,7 +65,18 @@ REACT_NATIVE_XCODE_WITH_SENTRY="\"$SENTRY_CLI_EXECUTABLE\" react-native xcode $A
 
 exitCode=0
 
-if [ "$SENTRY_DISABLE_AUTO_UPLOAD" != true ]; then
+if [ "$SENTRY_DISABLE_AUTO_UPLOAD" == true ]; then
+  echo "SENTRY_DISABLE_AUTO_UPLOAD=true, skipping sourcemaps upload"
+  /bin/sh -c "$REACT_NATIVE_XCODE"
+elif echo "$CONFIGURATION" | grep -iq "debug"; then # case insensitive check for "debug"
+  # Skip the source maps upload for *Debug* configuration, matching the behavior of the
+  # native debug files upload (sentry-xcode-debug-files.sh) and the Android Gradle plugin.
+  # Local Debug builds should not require Sentry credentials. Still run the React Native
+  # bundling step so the build itself succeeds.
+  # See: https://github.com/getsentry/sentry-react-native/issues/6399
+  echo "Skipping source maps upload for *Debug* configuration"
+  /bin/sh -c "$REACT_NATIVE_XCODE"
+else
   # 'warning:' triggers a warning in Xcode, 'error:' triggers an error
   set +x +e # disable printing commands otherwise we might print `error:` by accident and allow continuing on error
   SENTRY_XCODE_COMMAND_OUTPUT=$(/bin/sh -c "\"$LOCAL_NODE_BINARY\" $REACT_NATIVE_XCODE_WITH_SENTRY" 2>&1)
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Fixes
+
+- Skip iOS source maps upload on `Debug` builds ([#6399](https://github.com/getsentry/sentry-react-native/issues/6399))
```
```suggestion
- Skip iOS source maps upload on `Debug` builds ([#6405](https://github.com/getsentry/sentry-react-native/pull/6405))
```

antonis:
@cursor review

antonis:
@sentry review

packages/core/scripts/sentry-xcode.sh
```diff
@@ -65,7 +65,18 @@ REACT_NATIVE_XCODE_WITH_SENTRY="\"$SENTRY_CLI_EXECUTABLE\" react-native xcode $A
 
 exitCode=0
 
-if [ "$SENTRY_DISABLE_AUTO_UPLOAD" != true ]; then
+if [ "$SENTRY_DISABLE_AUTO_UPLOAD" == true ]; then
+  echo "SENTRY_DISABLE_AUTO_UPLOAD=true, skipping sourcemaps upload"
+  /bin/sh -c "$REACT_NATIVE_XCODE"
+elif echo "$CONFIGURATION" | grep -iq "debug"; then # case insensitive check for "debug"
```
lucas-zimerman, line 71: Looking good! But, what about using `ENABLE_TESTABILITY` instead of `CONFIGURATION`?

antonis: Good question @lucas-zimerman 👍 I used configuration mainly [to align with sentry-xcode-debug-files.sh](https://github.com/getsentry/sentry-react-native/blob/e5bb5f68eac91a8e330c1f612d54cbf2397b65dd/packages/core/scripts/sentry-xcode-debug-files.sh#L214). I'll need to check and possibly refactor both places. Wdyt of following up on this separately?

lucas-zimerman: sounds good to me!

lucas-zimerman (approved):
LGTM!

PR #6406, opened 2026-07-06
fix(android): Prevent New Arch build break linking libsentry-tm-perf-logger.so + CI ABI/link coverage

### LINKED ISSUE #6398: Android New Architecture build fails linking libsentry-tm-perf-logger.so with undefined TurboModulePerfLogger::enableLogging on RN 0.86
### Description

After using `@sentry/react-native` 8.17.1 in an Expo / React Native New Architecture app, `npx expo run:android --device` fails during the Android native build for `:sentry_react-native:buildCMakeDebug[armeabi-v7a]`.

The failing target is the new `libsentry-tm-perf-logger.so` library. The linker cannot resolve `facebook::react::TurboModulePerfLogger::enableLogging(...)` from React Native.

This happens even though the app does **not** set `enableTurboModuleTracking` in `Sentry.init`.

### Environment

- `@sentry/react-native`: `8.17.1`
- `expo`: `57.0.2`
- `react-native`: `0.86.0`
- `expo-router`: `~57.0.3`
- New Architecture: `newArchEnabled=true`
- Hermes: `hermesEnabled=true`
- Android compileSdk / targetSdk: `36`
- Android minSdk: `24`
- Gradle: `9.3.1`
- Java: `17.0.19`
- Host OS: macOS `26.5.1`, `aarch64`
- NDK installed: `27.0.12077973`, `27.1.12297006`, `29.0.13599879`, `29.0.14206865`, `30.0.14904198`
- CMake installed: `3.22.1`, `3.30.5`, `4.0.2`, `4.0.3`, `4.1.2`

Android config:

```properties
reactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64
newArchEnabled=true
hermesEnabled=true
```

The Expo CLI command printed at the end includes `-PreactNativeArchitectures=arm64-v8a`, but the failing Sentry task is still `:sentry_react-native:buildCMakeDebug[armeabi-v7a]`.

### Reproduction

1. Use an Expo 57 / React Native 0.86 app with New Architecture enabled.
2. Install/configure `@sentry/react-native@8.17.1` with the Expo plugin.
3. Do not set `enableTurboModuleTracking` in `Sentry.init`.
4. Run:

```sh
npx expo run:android --device
```

### Expected behavior

The Android debug build succeeds. If TurboModule performance tracking is experimental and opt-in, the native perf logger should either link successfully for supported RN versions/ABIs or not break builds when the option is not enabled.

### Actual behavior

The build fails while linking `libsentry-tm-perf-logger.so`:

```text
> Task :sentry_react-native:buildCMakeDebug[armeabi-v7a] FAILED
C/C++: ninja: Entering directory `.../node_modules/@sentry/react-native/android/.cxx/Debug/6a511s2b/armeabi-v7a'
C/C++: ld.lld: error: undefined symbol: facebook::react::TurboModulePerfLogger::enableLogging(std::__ndk1::unique_ptr<facebook::react::NativeModulePerfLogger, std::__ndk1::default_delete<facebook::react::NativeModulePerfLogger>>&&)
C/C++: >>> referenced by SentryTurboModulePerfLogger.cpp:210 (.../node_modules/@sentry/react-native/cpp/SentryTurboModulePerfLogger.cpp:210)
C/C++: >>>               CMakeFiles/sentry-tm-perf-logger.dir/.../node_modules/@sentry/react-native/cpp/SentryTurboModulePerfLogger.cpp.o:(sentry::reactnative::SentryTurboModulePerfController::install())
C/C++: clang++: error: linker command failed with exit code 1 (use -v to see invocation)
C/C++: ninja: build stopped: subcommand failed.

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':sentry_react-native:buildCMakeDebug[armeabi-v7a]'.
```

The full linker invocation includes:

```text
.../Android/sdk/ndk/27.0.12077973/toolchains/llvm/prebuilt/darwin-x86_64/bin/clang++ --target=armv7-none-linux-androideabi24 ... -shared -Wl,-soname,libsentry-tm-perf-logger.so ... -latomic -lm
```

### Additional context

From the installed package, the failing call is in `node_modules/@sentry/react-native/cpp/SentryTurboModulePerfLogger.cpp`:

```cpp
facebook::react::TurboModulePerfLogger::enableLogging(std::make_unique<ForwardingLogger>());
```

I searched existing issues and found the related `libsentry-tm-perf-logger.so` 16 KB alignment issue in 8.17.0, but I did not find an existing report for this undefined symbol link failure.

antonis:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Fixes the Android New Architecture build break in #6398, plus CI coverage for the class of failures around this app-compiled library.

### Root-cause fix — honor `reactNativeArchitectures`
`libsentry-tm-perf-logger.so` (added in 8.17.0, #6307) is compiled from source in the consuming app and links React Native's `reactnative` prefab. Our `build.gradle` **hardcoded** `abiFilters` to all four ABIs, ignoring the app's `reactNativeArchitectures` — unlike `react-native-screens`, `reanimated` and the RN app template, which all honor it.

When the app builds a **subset** (e.g. `-PreactNativeArchitectures=arm64-v8a`, common in Expo/dev/CI builds), React Native only provides its `reactnative` prefab for that subset. Our module still tried to build the other three ABIs, which then can't resolve `TurboModulePerfLogger::enableLogging` at link time → build break. (This matches the reporter's clue: the app requested `arm64-v8a` yet `:sentry_react-native:buildCMakeDebug[armeabi-v7a]` ran and failed.)

```groovy
def reactNativeArchitectures() {
    def value = project.getProperties().get("reactNativeArchitectures")
    return value ? value.split(",") : ["armeabi-v7a", "x86", "x86_64", "arm64-v8a"]
}
// ...
ndk { abiFilters(*reactNativeArchitectures()) }
```

### Defense-in-depth — `-Wl,-z,undefs`
`src/main/jni/CMakeLists.txt` also appends `-Wl,-z,undefs` so that if any built ABI's prefab reference is still unresolved at link time, the build doesn't hard-fail (the NDK/AGP toolchain links with `--no-undefined`). Undefined symbols then resolve at load against `libreactnative.so`; the library is loaded lazily and only when tracking is opted in, and an unresolvable symbol surfaces as an `UnsatisfiedLinkError` the Java side already swallows (`RNSentryTurboModulePerfTracker`) — degrading the experimental feature to a no-op rather than crashing.

### CI coverage
Both #6394 (16 KB alignment) and #6398 escaped because CI only built **x86**, while this is the SDK's only library compiled per-ABI on the user's side:
- **All-ABI sample builds** — the New Architecture sample Android builds (bare React Native and Expo) now build all four ABIs, and the 16 KB alignment check runs across every ABI.
- **`scripts/check-tm-perf-logger-link.sh`** — a lightweight guard (runs on every PR via `buildandtest.yml`) that reproduces the link condition using the real sources and the real `CMakeLists.txt` flags; fails if `-Wl,-z,undefs` is removed.

Fixes #6398. Present since 8.17.0 (still in 8.17.1). Hard build failure with no clean workaround, hitting New Architecture apps that build a subset of ABIs — including Expo — regardless of whether the perf-tracking feature is enabled.

**Root-cause fix — verified locally against a subset build** (`:sentry_react-native:externalNativeBuildDebug -PreactNativeArchitectures=arm64-v8a`):

| build.gradle | ABIs the module builds |
|---|---|
| hardcoded (before) | arm64-v8a + **armeabi-v7a + x86 + x86_64** (builds unrequested ABIs — the bug) |
| `reactNativeArchitectures()` (after) | **arm64-v8a only** |

**Link flag** — built the real sources with NDK 27 against the RN 0.86 prefab: reproduces `ld.lld: undefined symbol: TurboModulePerfLogger::enableLogging` under `--no-undefined` without the flag; links with it; normal builds keep `libreactnative.so` in `DT_NEEDED`. Verified across all four ABIs, all staying 16 KB aligned. `check-tm-perf-logger-link.sh` passes on the fix and fails when `-Wl,-z,undefs` is removed.

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps
- Consider a CI subset-ABI build (e.g. `-PreactNativeArchitectures=arm64-v8a`) to assert the module builds only the requested ABIs.

### DIFF
--- a/.github/workflows/buildandtest.yml
+++ b/.github/workflows/buildandtest.yml
@@ -34,6 +34,28 @@ jobs:
       - name: Test
         run: yarn test
 
+  job_native_link_check:
+    name: Android Native Link Check
+    runs-on: ubuntu-latest
+    needs: [diff_check]
+    if: ${{ needs.diff_check.outputs.skip_ci != 'true' }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - run: corepack enable
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
+        with:
+          package-manager-cache: false
+          node-version: 18
+          cache: 'yarn'
+          cache-dependency-path: yarn.lock
+      - name: Install Dependencies
+        run: yarn install
+      # Guards against regressing the #6398 fix: verifies libsentry-tm-perf-logger.so
+      # still links for armeabi-v7a under --no-undefined when the reactnative prefab
+      # reference is unresolved. Uses the runner's preinstalled NDK (ANDROID_NDK_LATEST_HOME).
+      - name: Check libsentry-tm-perf-logger.so links (armeabi-v7a)
+        run: ./scripts/check-tm-perf-logger-link.sh
+
   job_lint:
     name: Lint
     runs-on: ubuntu-latest
--- a/.github/workflows/sample-application-expo.yml
+++ b/.github/workflows/sample-application-expo.yml
@@ -187,7 +187,11 @@ jobs:
         run: |
           [[ "${{ matrix.build-type }}" == "production" ]] && CONFIG='Release' || CONFIG='Debug'
           echo "Building $CONFIG"
-          ./gradlew ":app:assemble$CONFIG" -PreactNativeArchitectures=x86
+          # Build all ABIs (not just x86): the Sentry perf-logger native library is
+          # compiled from source per ABI, and #6398 (armeabi-v7a link failure) was
+          # reported on an Expo build. Keeps ABI coverage consistent with the bare
+          # RN sample.
+          ./gradlew ":app:assemble$CONFIG" -PreactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64
 
       - name: Export Expo
         working-directory: samples/expo
--- a/.github/workflows/sample-application.yml
+++ b/.github/workflows/sample-application.yml
@@ -185,7 +185,12 @@ jobs:
           [[ "${{ matrix.build-type }}" == "production" ]] && export CONFIG='release' || export CONFIG='debug'
 
           ./scripts/set-dsn-aos.mjs
-          ./scripts/build-android.sh -PreactNativeArchitectures=x86
+          # Build all ABIs (not just x86) so armeabi-v7a and the 64-bit ABIs are
+          # exercised too. The Sentry perf-logger native library is compiled from
+          # source per ABI, and ABI-specific breakage has shipped before (#6394
+          # 16 KB alignment, #6398 armeabi-v7a link) precisely because CI only
+          # built x86. The 16 KB alignment check below then runs across every ABI.
+          ./scripts/build-android.sh -PreactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64
 
       - name: Check 16 KB native library alignment
         # Only Sentry-owned libraries are checked: third-party/React Native
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Fixes
+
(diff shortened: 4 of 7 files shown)


antonis:
@sentry review

alwx (approved):

Issue #816, opened 2020-04-08
How can I disable certain type of errors from receiving. Ex Error: Failed to connect to.cloudfront.net


shubhamdeol:
OS:
- [ ]  Windows
- [X]  MacOS
- [ ]  Linux

Platform:
- [X]  iOS
- [X]  Android

SDK:
- [X]  `@sentry/react-native`
- [ ]  `react-native-sentry`

`react-native` version: 0.61

Are you using Expo?
- [X]  Yes
- [ ]  No

Are you using sentry.io or on-premise?
- [X]  sentry.io (SaaS)
- [ ]  on-premise

Configuration:

```
Sentry.init({
  dsn: 'https://...@sentry.io/...'
});
```

---
I have following issue:

How can I disable receiving errors like these, and what are these errors:

`
Error: Failed to connect to hdashfjkhasfh.cloudfront.net/2600:9000:21c8:e44:1:9ab43:4380:21:443
`

marandaneto:
that comes from your own app, you can filter events like this https://docs.sentry.io/platforms/react-native/configuration/filtering/

PR #6407, opened 2026-07-06
fix(android): Correct New Architecture native-build gate to RN 0.76+

antonis:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
The `libsentry-tm-perf-logger.so` native build (added in 8.17.0, #6307) is gated on the host app's React Native version because it links the `ReactAndroid::reactnative` prefab. That gate was off by one:

```groovy
// before
return major > 0 || minor >= 75
// after
return major > 0 || minor >= 76
```

The **merged `libreactnative.so`** — exposed as the `ReactAndroid::reactnative` prefab selector — was introduced in **React Native 0.76**. On RN **0.75 and earlier**, the native code is split across separate prefab modules (`react_nativemodule_core`, etc.) and `ReactAndroid::reactnative` does not exist. With the old `minor >= 75` gate, an app on **RN 0.75.x with `newArchEnabled=true`** would run our CMake build, which calls `find_package(ReactAndroid REQUIRED CONFIG)` + `target_link_libraries(... ReactAndroid::reactnative)` against a target that isn't there — failing at the **CMake configure step**, before any linking.

Found while investigating the class of build failures around this app-compiled library (#6394, #6398). Unlike the armeabi-v7a link failure (#6406), this one fails at *configure* time, so a linker flag can't rescue it — the gate itself must be correct.

Verified the version boundary directly:
- RN 0.72 AAR: no unified `reactnative` prefab (split modules only)
- RN 0.85.1 / 0.86.0 AARs: `reactnative` prefab present, `TurboModulePerfLogger::enableLogging` exported
- RN docs confirm the merged `libreactnative.so` shipped in 0.76 ([release notes](https://reactnative.dev/blog/2024/10/23/release-0.76-new-architecture), [RFC #816](https://github.com/react-native-community/discussions-and-proposals/discussions/816))

With the fix, RN 0.75.x New Arch builds simply skip the native target (TurboModule perf tracking degrades to a no-op when the library is absent), instead of breaking the build.

Confirmed the prefab-availability boundary against cached RN AARs (0.72 vs 0.85/0.86) and RN's own release documentation. The change is a one-line version-gate correction with updated explanatory comments; RN 0.76+ New Arch behaviour is unchanged (still builds the native target).

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Fixes
+
+- Fix Android New Architecture build failing at CMake configure on React Native 0.75 by gating the `libsentry-tm-perf-logger.so` native build to RN 0.76+ ([#6407](https://github.com/getsentry/sentry-react-native/pull/6407))
+
 ## 8.17.1
 
 ### Fixes
--- a/packages/core/android/build.gradle
+++ b/packages/core/android/build.gradle
@@ -7,11 +7,13 @@ def isNewArchitectureEnabled() {
 }
 
 // `libsentry-tm-perf-logger.so` links against React Native's `reactnative`
-// prefab target, which only ships in the `ReactAndroid` prefab AAR shipped
-// with RN 0.75 and newer. On older RN releases the target does not exist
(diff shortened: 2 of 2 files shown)


antonis:
@sentry review

alwx (approved):
Looks good.
But I think one-line PRs could really be done without Claude and 10 lines of comments that come with Claude :)

PR #6412, opened 2026-07-06
Revert "feat(ios): Default to consuming sentry-cocoa as an xcframework" (#6381)

alwx:
## :loudspeaker: Type of change

- [x] Bugfix

## :scroll: Description

Reverts #6381 (merge commit `fdb06888`), which was accidentally merged. Because the PR was merged as a merge commit that also carried the `8.18.0-alpha.3` release commit on the same feature branch, this revert undoes both the PR content and the `8.18.0-alpha.3` version bump.

## :bulb: Motivation and Context

PR #6381 was merged accidentally. Reverting to restore `main` to the state prior to `fdb06888`.

## :green_heart: How did you test it?

- Verified `git revert -m 1 fdb06888` applied cleanly.
- Diff matches the inverse of the merge (27 files, restores prior `RNSentry.podspec`, drops the vendored xcframework tooling, restores previous version numbers).

## :pencil: Checklist

- [ ] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

- Re-cut `8.18.0-alpha.3` from `main` once the release timing is sorted.
- Re-open #6381 from its feature branch when ready to merge intentionally.

PR #6413, opened 2026-07-06
feat(ios): Consume sentry-cocoa as a prebuilt xcframework by default

alwx:

- [x] Enhancement

## :scroll: Description

Retry of #6381, which was merged into the `release/8.18.0-alpha.3` cut prematurely and then rolled back via #6412. Content re-applied, this time with the pnpm-isolated cache-location fix baked in from the start and validated end-to-end via `8.18.0-alpha.3` on a real bare RN app + EAS Build.

`pod install` now downloads `Sentry.xcframework.zip` from sentry-cocoa's GitHub release (SHA256-verified) and caches it under `~/Library/Caches/sentry-react-native/xcframeworks/<version>/`, instead of building Sentry from source as a CocoaPod. Sentry symbols are then **statically linked** into the app binary (no framework embed, no dylib dep) via `FRAMEWORK_SEARCH_PATHS[sdk=…*]` + clang's `-fmodules-autolink`.

Override with:
- `SENTRY_USE_XCFRAMEWORK=0` — fall back to the source-built `Sentry` CocoaPod (for offline builds, restrictive proxies, or projects with another pod that transitively depends on the `Sentry` CocoaPod).
- `SENTRY_XCFRAMEWORK_CACHE_DIR` — override cache root (defaults to `~/Library/Caches/sentry-react-native/xcframeworks`).

- CocoaPods trunk is winding down; we needed a path off it that also sidesteps the Xcode 16/26 archive bug that hits any signed SPM binary xcframework (`Signatures/*.signature` collision).
- Consuming the same xcframework via CocoaPods' `vendored_frameworks` pipeline goes through a different code path and is unaffected by that bug.
- Cache location is user-writable (not inside `node_modules`) so pnpm's isolated store and Yarn PnP work without changes.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

- Cut a fresh alpha from this branch once CI is green.
- Optionally submit the signed .ipa to TestFlight to close the loop with Apple's server-side validation (not blocking).

### DIFF
--- a/.github/workflows/sample-application.yml
+++ b/.github/workflows/sample-application.yml
@@ -43,7 +43,7 @@ jobs:
       caller_ref: ${{ github.ref }}
 
   build-ios:
-    name: Build ${{ matrix.rn-architecture }} ios ${{ matrix.build-type }} ${{ matrix.ios-use-frameworks }}
+    name: Build ${{ matrix.rn-architecture }} ios ${{ matrix.build-type }} ${{ matrix.ios-use-frameworks }} ${{ matrix.sentry-consumption }}
     runs-on: macos-26-xlarge
     needs: [diff_check, detect-changes]
     if: >-
@@ -60,6 +60,16 @@ jobs:
         rn-architecture: ["new"]
         ios-use-frameworks: ["no-frameworks"]
         build-type: ["dev", "production"]
+        sentry-consumption: ["xcframework"]
+        # `xcframework` is the default: RNSentry vendors a prebuilt
+        # `Sentry-Dynamic.xcframework` downloaded from sentry-cocoa's release.
+        # Keep a single CocoaPods job to catch regressions in the source-build
+        # fallback (`SENTRY_USE_XCFRAMEWORK=0`).
+        include:
+          - rn-architecture: "new"
+            ios-use-frameworks: "no-frameworks"
+            build-type: "production"
+            sentry-consumption: "cocoapods"
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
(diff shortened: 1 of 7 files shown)


alwx:
I've spent quite a lot of time checking different scenarios with embedded xcframework to make sure everything works, and want to share the results here.
One silent regression was caught + fixed before it shipped to real users (that's why I pushed alpha.3 version earlier today which broke things slightly because the PR got merged after that and I needed to revert but at least we discovered a bit more about how the release process works :sweat_smile:)

So here is what I've tried:

1. Native crash → dSYM → symbolicated. Real crash on booted simulator. Sentry frames resolved by name in the dashboard. You can check what got uploaded to Sentry here. The screenshot with the test app (the app itself was quickly drafted by Claude) is below..
2. pnpm strict / read-only node_modules. Reproduced Permission denied locally; moved the xcframework cache to ~/Library/Caches/sentry-react-native/xcframeworks/<version>/ (override with SENTRY_XCFRAMEWORK_CACHE_DIR). This particular test was the reason for alpha.3.
3. RN 0.71 e2e archive. All worked as expected.
4. EAS Build + npm + default xcframework. ~4 min cloud build with no issues.
5. EAS Build + pnpm-isolated + xcframework. Also built with no issues. 
6. EAS Build + SENTRY_USE_XCFRAMEWORK=0. This build used CocoaPods, and it's important to notice that the Source-built Cocoapods build produced functionally identical binary. 
7. EAS second-build after version bump. Identical output as first build (21687 symbols).
8. Last but not least: I've just tried signing the build with EAS — it all worked fine as well :tada:

In addition to that, default vendored static xcframework (number 4 in the list) and fallback source-built CocoaPod (num. 6) produce statically linked binary with the identical shape, and there is no code path in this PR that would've included a dynamic Sentry framework.

antonis (approved):
LGTM 🚀 Awesome work @alwx 🏅
Thank you for getting this across the finish line 🙇

PR #6416, opened 2026-07-07
test(android): Guard reactNativeArchitectures fix with a subset-ABI build

antonis:
- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description
Closes the last CI coverage gap around the app-compiled `libsentry-tm-perf-logger.so` native library. Follow-up to the 8.17.2 fixes (#6406, #6407).

The 8.17.2 root-cause fix for #6398 made `packages/core/android/build.gradle` honour the host app's `reactNativeArchitectures` (`abiFilters(*reactNativeArchitectures())`, matching react-native-screens/reanimated/the RN template). But nothing in CI exercises a **subset-ABI build**, so the existing all-ABI builds pass identically with or without the fix — they cannot catch a regression back to hardcoded `abiFilters`.

This PR splits the bare-RN sample `build-android` matrix so each ABI-coverage concern gets its own entry, and adds a lightweight assertion:

| Matrix entry | ABIs | New guard | Covers |
|---|---|---|---|
| `dev` | `arm64-v8a` only (subset) | `check-tm-perf-logger-abi-subset.sh` asserts the packaged `.so` set matches exactly | Regression of the `reactNativeArchitectures` fix (#6398 root cause) |
| `production` | all four (unchanged) | existing 16 KB alignment check | Cross-ABI alignment (#6394) |

The new script `scripts/check-tm-perf-logger-abi-subset.sh` unzips a built APK/AAB, finds `libsentry-tm-perf-logger.so` per ABI, and asserts the set matches the expected CSV exactly. Same shape as the existing `check-android-16kb-alignment.sh`.

Both #6394 and #6398 escaped because CI only built x86; this module is the SDK's only library compiled per-ABI on the user's side. #6406 + #6407 fixed the code paths; this closes the CI side by ensuring a future regression of the `reactNativeArchitectures` fix is caught pre-merge.

Downstream jobs are unaffected: `Archive Android App` / `Upload Android APK` / `test-android` (Maestro E2E) are all gated on `matrix.build-type == 'production'`, which still builds all four ABIs. Build time for the `dev` entry actually drops slightly (1 ABI vs 4).

Verified locally against a real subset APK from the RN sample (`./scripts/build-android.sh -PreactNativeArchitectures=arm64-v8a`, `SENTRY_DISABLE_AUTO_UPLOAD=true`):

| Case | Expected | Actual | Guard |
|---|---|---|---|
| Subset APK with fix | `arm64-v8a` | `arm64-v8a` | ✅ pass (exit 0) |
| Simulated regression — unrequested `armeabi-v7a` `.so` leaks in | `arm64-v8a` | `arm64-v8a armeabi-v7a` | ❌ fail with root-cause hint (exit 1) |
| `.so` missing entirely | `arm64-v8a` | (none) | ❌ distinct "not found" error (exit 1) |

Workflow YAML validated with `ruby -e 'YAML.load_file(...)'`; shell script passes `bash -n`.

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps

### DIFF
--- a/.github/workflows/sample-application.yml
+++ b/.github/workflows/sample-application.yml
@@ -199,12 +199,24 @@ jobs:
           [[ "${{ matrix.build-type }}" == "production" ]] && export CONFIG='release' || export CONFIG='debug'
 
           ./scripts/set-dsn-aos.mjs
-          # Build all ABIs (not just x86) so armeabi-v7a and the 64-bit ABIs are
-          # exercised too. The Sentry perf-logger native library is compiled from
-          # source per ABI, and ABI-specific breakage has shipped before (#6394
-          # 16 KB alignment, #6398 armeabi-v7a link) precisely because CI only
-          # built x86. The 16 KB alignment check below then runs across every ABI.
-          ./scripts/build-android.sh -PreactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64
+          # Split ABI coverage across the two matrix entries:
+          #   - `dev`        → subset (arm64-v8a only) so we can assert the module
+          #                    builds only the ABIs the app requested — the root
+          #                    cause of #6398 was hardcoded abiFilters ignoring
+          #                    `reactNativeArchitectures`.
+          #   - `production` → all four ABIs so the 16 KB alignment check below
+          #                    runs across every ABI (guards against #6394).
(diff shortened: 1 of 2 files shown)


alwx (approved):
Looks good!

Issue #6415, opened 2026-07-07
Track adoption of opt-in features via `sdk.integrations` markers


antonis:
## Summary

Several opt-in features shipped in the past year leave no distinctive fingerprint on events, so we have no way to measure adoption in our telemetry. Extend the marker-integration pattern already used by [`feedback/lazy.ts`](https://github.com/getsentry/sentry-react-native/blob/main/packages/core/src/js/feedback/lazy.ts) to cover them.

## Background

The RN SDK already tracks feedback adoption by registering empty named integrations (`MobileFeedback`, `AutoInjectMobileFeedback`, etc.) whose names flow through to `event.sdk.integrations` via `@sentry/core`'s `prepareEvent` (Hex table: `sdk_integration_array`).

Confirmed via telemetry:
- `sdk_features_array` is empty for `sentry.javascript.react-native%` — `@sentry/core` has no `setSdkFeature`/`addSdkFeature` API and doesn't populate `event.sdk.features`. Not a viable channel today.
- `sdk_integration_array` is the working channel; e.g., `AppStart` currently surfaces there (~9,983 projects).

## Features to add markers for

| Marker name | Feature | Registration mechanism |
|---|---|---|
| `CaptureAppStartErrors` | Native pre-JS init via `sentry.options.json` (iOS `RNSentrySDK.start`, Android equivalent) — 8.0.0 | Native records the flag on successful bootstrap; JS reads via the bridge at init and calls `client.addIntegration({ name: 'CaptureAppStartErrors' })` |
| `StandaloneAppStart` | `enableStandaloneAppStartTracing` — 8.17.0 | Register at `Sentry.init` when option is set |
| `ExtendedAppStart` | `extendAppStart` / `finishExtendedAppStart` / `getExtendedAppStartSpan` — Unreleased | Register on first call |
| `MobileReplayNetworkDetails` | `mobileReplayIntegration({ networkDetailAllowUrls })` — 8.15.0 | Register at replay integration setup when `networkDetailAllowUrls` is non-empty |
| `MobileReplayNetworkBodies` | `mobileReplayIntegration({ networkCaptureBodies: true })` — 8.15.0 | Register at replay integration setup when option is true |
| `GlobalErrorBoundary` | `Sentry.GlobalErrorBoundary` / `withGlobalErrorBoundary` — 8.9.1 | Lazy — register on component mount |
[...]

PR #6421, opened 2026-07-08
feat(core): Track GlobalErrorBoundary adoption

antonis:
## Type of change

- [ ] Bugfix
- [ ] New feature
- [ ] Refactoring

## Description

Adds a `GlobalErrorBoundary` marker so Sentry can measure adoption of `Sentry.GlobalErrorBoundary` / `withGlobalErrorBoundary` via the `sdk_integration_array` telemetry channel. The marker is registered as a no-op named integration when the component mounts; from there the name is copied onto every outgoing event by `@sentry/core`'s `prepareEvent`.

Also introduces a small shared `registerFeatureMarker` helper. Follow-up PRs for the other opt-in features (NavigationContainer, ExpoRouter error boundary, AppLoaded, ExtendedAppStart, StandaloneAppStart, MobileReplay network detail/bodies) will import the same helper.

## Motivation and Context

Part of #6415. `GlobalErrorBoundary` is opt-in and leaves no distinctive fingerprint on events today, so we cannot measure its adoption in the same way we can for the `AppStart` performance integration or the feedback family. Same pattern as the existing `feedback/lazy.ts` markers.

## How did you test it?

- New `test/utils/featureMarkers.test.ts` (5 tests) — covers no-op, idempotency, no-client, explicit-client, and an end-to-end test that captures an event and asserts the marker actually lands on `event.sdk.integrations` via `prepareEvent`.
- New marker-on-mount test in `test/GlobalErrorBoundary.test.tsx`.
- Full local suite: 1685/1685 pass; lint / circular deps / api-extractor clean.

## Checklist

- [x] I have made corresponding changes to the documentation _(none needed — internal telemetry, no user-facing API)_
- [ ] I have added a CHANGELOG entry _(intentionally omitted per issue #6415: purely internal telemetry, no user-visible change)_

## Next steps

Follow-up PRs (see #6415) will wire the remaining markers using this helper, one PR per file to keep review small:
- `NavigationContainer`
- `ExpoRouterErrorBoundary`
- `AppLoaded` + `ExtendedAppStart` (bundled — same file)
- `StandaloneAppStart`
- `MobileReplayNetworkDetails` + `MobileReplayNetworkBodies` (bundled — same integration setup)

Refs: #6415

### DIFF
--- a/packages/core/src/js/GlobalErrorBoundary.tsx
+++ b/packages/core/src/js/GlobalErrorBoundary.tsx
@@ -7,6 +7,9 @@ import * as React from 'react';
 import type { GlobalErrorEvent } from './integrations/globalErrorBus';
 
 import { subscribeGlobalError } from './integrations/globalErrorBus';
+import { registerFeatureMarker } from './utils/featureMarkers';
+
+export const GLOBAL_ERROR_BOUNDARY_INTEGRATION_NAME = 'GlobalErrorBoundary';
 
 /**
  * Props for {@link GlobalErrorBoundary}. Extends the standard `ErrorBoundary`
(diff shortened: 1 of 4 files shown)


lucas-zimerman (approved):
LGTM!

PR #6425, opened 2026-07-09
feat(tracing): Track wrapExpoRouterErrorBoundary adoption

antonis:
## Type of change

- [ ] Bugfix
- [ ] New feature
- [ ] Refactoring

## Description

Registers a no-op `ExpoRouterErrorBoundary` integration when the user calls `Sentry.wrapExpoRouterErrorBoundary(...)` (i.e., at module-evaluation time of the route file that uses the wrap), so the name flows through to `event.sdk.integrations`. Covers both manual usage and code emitted by the Babel auto-wrap plugin — both paths ultimately call `wrapExpoRouterErrorBoundary` and share this marker.

## Motivation and Context

Part of #6415. The wrap adds route-scoped context to boundary errors, but nothing about the emitted event distinguishes "wrap installed" from "user's global handler caught it" today. Same pattern as `feedback/lazy.ts` and PR #6421.

The registration fires at wrap-call time rather than on the mount of the wrapped boundary, so we count "wrap installed" (not just "wrap engaged with an error"). Expo Router lazy-loads route files during navigation, so `Sentry.init()` has typically completed by the time the wrap runs — if not, `registerFeatureMarker` silently no-ops.

## How did you test it?

- New test in `test/tracing/expoRouterErrorBoundary.test.tsx` asserting the marker is registered synchronously at wrap-call time (before any component mount).
- Existing boundary tests continue to pass.
- Full local suite green.

## Checklist

## Base

Refs: #6415

### DIFF
--- a/packages/core/src/js/GlobalErrorBoundary.tsx
+++ b/packages/core/src/js/GlobalErrorBoundary.tsx
@@ -7,6 +7,9 @@ import * as React from 'react';
 import type { GlobalErrorEvent } from './integrations/globalErrorBus';
 
 import { subscribeGlobalError } from './integrations/globalErrorBus';
+import { registerFeatureMarker } from './utils/featureMarkers';
+
+export const GLOBAL_ERROR_BOUNDARY_INTEGRATION_NAME = 'GlobalErrorBoundary';
 
 /**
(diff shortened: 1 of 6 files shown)


lucas-zimerman (approved):
LGTM!

PR #6426, opened 2026-07-09
feat(tracing): Track appLoaded and extendAppStart adoption

antonis:
## Type of change

- [ ] Bugfix
- [ ] New feature
- [ ] Refactoring

## Description

Registers no-op `AppLoaded` and `ExtendedAppStart` integrations on first call so the names flow through to `event.sdk.integrations`. `ExtendedAppStart` covers all three extend API entry points: `extendAppStart`, `getExtendedAppStartSpan`, and `finishExtendedAppStart`.

Bundled together because both markers live in `packages/core/src/js/tracing/integrations/appStart.ts` — splitting them into separate PRs would produce mechanical merge conflicts on the shared imports.

## Motivation and Context

Part of #6415.
- `Sentry.appLoaded()` explicitly signals app-start end but leaves no distinctive event fingerprint today.
- `Sentry.extendAppStart()` / `getExtendedAppStartSpan()` / `finishExtendedAppStart()` are new experimental APIs that produce standalone app-start transactions indistinguishable from auto-captured ones.

## How did you test it?

- New tests in `test/tracing/integrations/appStart.test.ts`: an `AppLoaded` marker test in the existing `appLoaded() API` describe block and a parameterized `it.each` for the three `ExtendedAppStart` entry points.
- Existing app-start tests continue to pass.
- Full local suite green.

## Note on semantics

`ExtendedAppStart` fires even if the underlying integration no-ops (e.g., user called `extendAppStart` without enabling `_experiments.enableStandaloneAppStartTracing`). This is intentional — measures API adoption, not just successful use.

## Checklist

## Base

Refs: #6415

### DIFF
--- a/packages/core/src/js/GlobalErrorBoundary.tsx
+++ b/packages/core/src/js/GlobalErrorBoundary.tsx
@@ -7,6 +7,9 @@ import * as React from 'react';
 import type { GlobalErrorEvent } from './integrations/globalErrorBus';
 
 import { subscribeGlobalError } from './integrations/globalErrorBus';
+import { registerFeatureMarker } from './utils/featureMarkers';
+
+export const GLOBAL_ERROR_BOUNDARY_INTEGRATION_NAME = 'GlobalErrorBoundary';
 
 /**
  * Props for {@link GlobalErrorBoundary}. Extends the standard `ErrorBoundary`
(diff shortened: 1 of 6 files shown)


lucas-zimerman (approved):
LGTM!

PR #6428, opened 2026-07-09
feat(replay): Track mobile replay network capture adoption

antonis:
## Type of change

- [ ] Bugfix
- [ ] New feature
- [ ] Refactoring

## Description

Registers no-op `MobileReplayNetworkDetails` and `MobileReplayNetworkBodies` integrations at replay setup when the corresponding options are configured, so the names flow through to `event.sdk.integrations`.

Two markers give a clean funnel:
- Replay on → `MobileReplay` (already tracked)
- Network details on (any `networkDetailAllowUrls`) → `MobileReplayNetworkDetails`
- Bodies on (`networkCaptureBodies`, defaults to `true`) → `MobileReplayNetworkBodies`

Bundled together because both markers live in the same `if (networkDetailAllowUrls)` block in `mobilereplay.ts` — splitting them would mean two PRs editing the same lines.

## Motivation and Context

Part of #6415. The parent `MobileReplay` integration name is already reported, but the sub-options (headers vs. bodies) are not distinguishable from `sdk.integrations` alone. This is the more sensitive/opt-in half of replay and worth tracking separately.

## How did you test it?

- New `describe('network detail feature markers')` block in `test/replay/mobilereplay.test.ts` with three cases: no allow URLs → no markers; allow URLs set (bodies default) → both markers; allow URLs + `networkCaptureBodies: false` → only the details marker.
- Existing mobile replay tests continue to pass.
- Full local suite green.

## Checklist

## Base

Refs: #6415

### DIFF
--- a/packages/core/src/js/GlobalErrorBoundary.tsx
+++ b/packages/core/src/js/GlobalErrorBoundary.tsx
@@ -7,6 +7,9 @@ import * as React from 'react';
 import type { GlobalErrorEvent } from './integrations/globalErrorBus';
 
 import { subscribeGlobalError } from './integrations/globalErrorBus';
+import { registerFeatureMarker } from './utils/featureMarkers';
+
+export const GLOBAL_ERROR_BOUNDARY_INTEGRATION_NAME = 'GlobalErrorBoundary';
 
 /**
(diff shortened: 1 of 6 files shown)


lucas-zimerman (approved):
LGTM!

PR #6429, opened 2026-07-09
feat(core): Track CaptureAppStartErrors adoption

antonis:
## Type of change

- [ ] Bugfix
- [ ] New feature
- [ ] Refactoring

## Description

Registers a no-op `CaptureAppStartErrors` integration when the JS runtime detects that `sentry.options.json` is present, so the name flows through to `event.sdk.integrations`.

Detection is entirely on the JS side: the Metro serializer (`sentryOptionsSerializer.ts`) already bundles `sentry.options.json` into the JS bundle as `RN_GLOBAL_OBJ.__SENTRY_OPTIONS__` at build time — shipping that config file is what opts users into the pre-JS crash capture feature. No native or bridge changes are required.

## Motivation and Context

Part of #6415. This was the flagship 8.0 feature and the largest telemetry gap: `sentry.options.json` runs before the JS bundle loads, so today only rare pre-JS crashes signal adoption (and even those look like plain native crashes). The marker lets us see adoption directly in `sdk_integration_array`.

Initial plan called for a native flag + bridge round-trip on iOS and Android. That turned out to be unnecessary: `sdk.tsx` already reads `__SENTRY_OPTIONS__` at init to gate `autoInitializeNativeSdk` (see the existing checks at `sdk.tsx:71,156,190`), so the JS side already has the signal we need.

## Behavior

- Fires in both `__DEV__` and release builds, whenever `__SENTRY_OPTIONS__` is bundled. This measures user *intent* to adopt the feature (matches the choice made for `StandaloneAppStart` in #6427).
- Does not fire when the app has no `sentry.options.json` at all.
- No-ops silently if `Sentry.init()` was somehow called before `initAndBind` — the marker registration runs after the client is bound, so this is essentially guaranteed.

## How did you test it?

Two new tests in `test/sdk.test.ts` under `initialization from sentry.options.json`:
- Asserts the marker is registered when `RN_GLOBAL_OBJ.__SENTRY_OPTIONS__` is set.
- Asserts the marker is NOT registered when `__SENTRY_OPTIONS__` is deleted.

Full local suite green (118/118 in `sdk.test.ts`, 1676/1676 across all suites for this branch), lint clean, `tsc` clean.

## Checklist

- [x] I have made corresponding changes to the documentation _(none needed — internal telemetry, no user-facing API)_

## Base

Refs: #6415

### DIFF
--- a/packages/core/src/js/GlobalErrorBoundary.tsx
+++ b/packages/core/src/js/GlobalErrorBoundary.tsx
@@ -7,6 +7,9 @@ import * as React from 'react';
 import type { GlobalErrorEvent } from './integrations/globalErrorBus';
 
 import { subscribeGlobalError } from './integrations/globalErrorBus';
+import { registerFeatureMarker } from './utils/featureMarkers';
+
+export const GLOBAL_ERROR_BOUNDARY_INTEGRATION_NAME = 'GlobalErrorBoundary';
 
 /**
  * Props for {@link GlobalErrorBoundary}. Extends the standard `ErrorBoundary`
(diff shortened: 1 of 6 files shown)


lucas-zimerman (approved):
LGTM!

PR #6435, opened 2026-07-10
fix(feedback): Guard show/hide against uninitialized _setVisibility

antonis:
## Type of change

- [x] Bugfix

## Description

`FeedbackFormManager._setVisibility` (and the same static on `FeedbackButtonManager` / `ScreenshotButtonManager`) was declared without a default value. It only becomes defined when `FeedbackFormProvider`'s constructor calls `initialize()`. Any call to `showFeedbackForm()` / `hideFeedbackForm()` / `showFeedbackButton()` / `showScreenshotButton()` before that provider has mounted hits an undefined static, and the internal guard `if (this._setVisibility !== NOOP_SET_VISIBILITY)` evaluates truthy, so the code calls `undefined(true)` and throws:

```
TypeError: this._setVisibility is not a function
  at FeedbackManager.show (@sentry/react-native/dist/js/feedback/FeedbackFormManager.js)
```

The `console.warn` fallback that instructs the user to add `Sentry.wrap(RootComponent)` was unreachable in this state.

The fix defaults `_setVisibility` to `NOOP_SET_VISIBILITY` at declaration. The guard now correctly detects the never-initialized state and falls through to the warn branch. Applies to all three subclasses via static-property inheritance.

## Motivation and Context

Reported crashes in the SDK Crashes React Native project — [SDK-CRASHES-REACT-NATIVE-62C](https://sentry.sentry.io/issues/SDK-CRASHES-REACT-NATIVE-62C) (11 users / 42 events across 8.7.0 – 8.16.0, both iOS and Android). All events show the crash originating during React render — matching the case where a user component calls `showFeedbackForm()` in its own render/effect before the provider has finished mounting.

The existing "does not throw" test masked the bug because `beforeEach` calls `resetFeedbackFormManager()`, which sets `_setVisibility` to `NOOP_SET_VISIBILITY` and hides the never-initialized state that production users actually hit.

## How did you test it?

- Added two regression tests under `jest.isolateModules` that load `FeedbackFormManager` fresh (no prior `initialize`/`reset`) and assert `showFeedbackForm()` / `FeedbackFormManager.hide()` do not throw and warn as intended.
- Verified the new tests fail on `main` (produce the exact production `TypeError`) and pass with the fix.
- Full `@sentry/react-native` Jest suite: **1699/1699 pass**.
- `yarn lint:lerna`: clean.
- `yarn circularDepCheck`: clean.

## Checklist

- [x] I reviewed the submitted code
- [x] I added tests to verify the changes
- [x] I updated the docs if needed
- [x] Review from the native team if needed (N/A — TS-only)

## Next steps

_None._


antonis:
@cursor review

antonis:
@sentry review

lucas-zimerman (approved):
LGTM!

PR #6438, opened 2026-07-10
fix(ios): Resolve getNewScreenTimeToDisplay Promise with number not array

### LINKED ISSUE #6433: `getNewScreenTimeToDisplay()` resolves its Promise with an `NSArray` instead of a number, corrupting the fallback TTID timestamp and causing spurious `deadline_exceeded` status

Expo Router

sentry.io (SaS)

No

_No response_

8.16.0

````

System:
  OS: macOS 26.5.2
  CPU: (16) arm64 Apple M3 Max
  Memory: 5.48 GB / 64.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 24.12.0
  Yarn:
    version: 4.12.0
  npm:
    version: 11.6.2
  Watchman:
    version: 2024.10.14.00
Managers:
  CocoaPods:
    version: 1.15.2
    path: /usr/local/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.2
      - iOS 26.2
      - macOS 26.2
      - tvOS 26.2
      - visionOS 26.2
      - watchOS 26.2
  Android SDK: Not Found
IDEs:
  Xcode:
    version: 26.2/17C52
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.11
    path: /usr/bin/javac
  Ruby:
    version: 2.6.10
    path: /usr/bin/ruby
npmPackages:
    installed: 20.2.0
    wanted: latest
  react:
    installed: 19.2.0
    wanted: 19.2.0
  react-native:
    installed: 0.83.6
    wanted: 0.83.6
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: true
iOS:
  hermesEnabled: true
  newArchEnabled: true


````

### Sentry.init()

````js
Sentry.init({
    appHangTimeoutInterval: 4,
    enableAppStartTracking: true,
    enabled: isSentryEnabled,
    enableStallTracking: true,
    ignoreErrors: ignoredErrors,
    integrations: [
      }),
    ],
    maxValueLength: 5000,
    sendDefaultPii: true,
    spotlight:
});
````

### Steps to Reproduce

1. Add a temporary log in `addAutomaticTimeToInitialDisplay` (`dist/js/tracing/integrations/timeToDisplayIntegration.js`), immediately before its `createSpanJSON` call, printing `ttidNativeTimestampSeconds`, `ttidFallbackTimestampSeconds`, `ttidTimestampSeconds`, `transactionStartTimestampSeconds`, `durationMs`, and `isDeadlineExceeded(durationMs)`.
2. Cold launch on iOS Simulator (Expo dev client, not Expo Go) and navigate to any screen with automatic TTID tracking, with no explicit `TimeToInitialDisplay` component so the fallback path is used.
3. Confirm in the log that `ttidFallbackTimestampSeconds` and `ttidTimestampSeconds` print as one element arrays rather than plain numbers, and that the computed `durationMs`/`willBeDeadlineExceeded` look small and correct.
4. Compare against the same span's final state in the Sentry UI.

### Expected Result

The span's final `duration` and `status` in Sentry should match what the JS integration computed locally. If the JS side calculates `durationMs: 318` and `willBeDeadlineExceeded: false`, the span should show a short duration and an ok status.

### Actual Result

The span shows `status: deadline_exceeded` with a duration roughly 8x longer than the JS side computed for the same `rootSpanId` and `start_timestamp`. The root cause is in `ios/RNSentryTimeToDisplay.m`, where `getTimeToDisplay:` resolves using the old-style array-args callback convention (`resolveBlock(@[ @(currentTime) ])`), but it's invoked from `getNewScreenTimeToDisplay`, a real Promise-based bridge method whose `resolve` block expects a raw value, not an array. This means the JS `Promise` resolves with a one-element `NSArray` instead of a `NSNumber` timestamp. The array survives the duration math by accident, since arithmetic coercion via `Array.prototype.toString()` gives the right `durationMs`, but the array gets assigned directly as the span's `timestamp` field in `createSpanJSON`, since object assignment doesn't do that coercion. That malformed span gets handed to the native Cocoa SDK for envelope construction, and ends up with a later finish timestamp and `deadline_exceeded` status, discarding the correct JS-computed values. Also present in the `#else` branch a few lines above, which resolves non-iOS platforms with an empty array instead of `nil`.

### PR DESCRIPTION
**Pull request by antonis**
## 📢 Type of change

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## 📜 Description

* Retype the block (`.h` + `.m` signature + ivar) to `RCTPromiseResolveBlock`.
* Resolve with the raw timestamp `@(currentTime)` instead of `@[ @(currentTime) ]`.
* Resolve with `nil` instead of an empty array `@[]` on the non-iOS (`#else`) branch.
* Add a native regression test asserting the resolve value is a scalar `NSNumber`, never an array.

## 💡 Motivation and Context

Fixes getsentry/sentry-react-native#6433

## 💚 How did you test it?

* Added `testGetTimeToDisplayResolvesWithNumberNotArray` to `RNSentryTimeToDisplayTests.swift`. Verified it **fails** against the previous array-resolving behavior and **passes** with the fix.
* Ran the full `RNSentryTimeToDisplayTests` suite (4/4 pass).
* Ran the affected JS suites (`sentryeventemitterfallback`, `reactnavigation.ttid`, `timeToDisplay` integration) — 32/32 pass. No JS changes were required; consumers already expect a scalar per the declared `Promise<number | undefined | null>` type.

## :pencil: Checklist

- [X] All tests passing.
- [X] No breaking changes.

## 🔮 Next steps

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -12,6 +12,10 @@
 
 - Add `Sentry.reportFullyDisplayed()` imperative API for signaling Time to Full Display ([#6419](https://github.com/getsentry/sentry-react-native/pull/6419))
 
+### Fixes
+
+- Fix iOS time-to-initial-display fallback spans reporting a spurious `deadline_exceeded` status and inflated duration ([#6433](https://github.com/getsentry/sentry-react-native/issues/6433))
```
```suggestion
- Fix iOS time-to-initial-display fallback spans reporting a spurious `deadline_exceeded` status and inflated duration ([#6438](https://github.com/getsentry/sentry-react-native/pull/6438))
```

lucas-zimerman (approved):
LGTM!

PR #6449, opened 2026-07-14
fix(ios): Break RNSentryOnDrawReporterView retain cycle in emit-new-frame block

### LINKED ISSUE #6440: iOS: retain cycle in RNSentryOnDrawReporterView leaks TTID/TTFD views
During memory leak analysis of a normal scrolling session (Instruments ▸ Leaks), the dominant leak cluster was `RNSentryOnDrawReporterView` / `RNSentryFramesTrackerListener` / `CALayer` — roughly half of all leaked allocations in the session (9x each of the first two).

Confirmed by reading `packages/core/ios/RNSentryOnDrawReporter.m`, `RNSentryOnDrawReporter.h`, and `RNSentryFramesTrackerListener.h/.m` at tag `8.7.0` and current `main` (identical, unfixed on both):

- `RNSentryOnDrawReporterView.framesListener` is `(nonatomic, strong)`.
- `RNSentryFramesTrackerListener.emitNewFrameEvent` is `(strong, nonatomic)` and the block passed in is `copy`'d on init.
- `createEmitNewFrameEvent` in the view returns a block referencing `self->isListening`, `self->_parentSpanId`, `self->_fullDisplay`, `self->_initialDisplay` with no `__weak`/`weakSelf` pattern, so ARC captures `self` (the view) strongly.

This forms a closed strong-reference loop: `view → framesListener (strong) → emitNewFrameEvent block (strong copy) → view (implicit strong self capture)`. `RNSentryFramesTrackerListener` does call `[_framesTracker removeListener:self]` after the first frame, which breaks the external tie to `SentryFramesTracker`, but the internal view↔listener↔block loop is self-contained, so ARC can never deallocate any of the three objects regardless of view teardown.

Suggested fix: use a weak self capture inside `createEmitNewFrameEvent`, e.g.:

```objc
__weak typeof(self) weakSelf = self;
return ^(NSNumber *newFrameTimestampInSeconds) {
    RNSentryOnDrawReporterView *strongSelf = weakSelf;
    if (!strongSelf) return;
    ...
};
```

Requested by **Michael Chai** via Junior.

--

[View Junior Session in Sentry](https://sentry.sentry.io/explore/conversations/slack%3AD0ANAGCBY0N%3A1783713170.709959/?project=4510944073809921)

alwx:
## 📢 Type of change

- [X] Bugfix

## 📜 Description

The block returned from `createEmitNewFrameEvent` captures `self` strongly, and the view owns the listener that owns the block — closed loop, nothing gets freed. Switched to weak capture.

## 💡 Motivation and Context

Fixes getsentry/sentry-react-native#6440. Leaks in Instruments showed a bunch of `RNSentryOnDrawReporterView` / `RNSentryFramesTrackerListener` piling up during scrolling.

## 💚 How did you test it?

Added a test that holds only a weak ref to the view and checks it's gone after the autoreleasepool drains. Fails on main, passes here.

## :pencil: Checklist

- [X] All tests passing.
- [X] No breaking changes.

### DIFF
--- a/packages/core/RNSentryCocoaTester/RNSentryCocoaTesterTests/RNSentryOnDrawReporterTests.swift
+++ b/packages/core/RNSentryCocoaTester/RNSentryCocoaTesterTests/RNSentryOnDrawReporterTests.swift
@@ -145,4 +145,22 @@ final class RNSentryOnDrawReporterTests: XCTestCase {
 
         emitNewFrameCallback!(1)
     }
+
+    func testReporterViewIsDeallocatedAfterRelease() {
+        // Guards against the retain cycle described in
+        // https://github.com/getsentry/sentry-react-native/issues/6440 where the
+        // emit-new-frame block captured `self` strongly, keeping the view (and
+        // its frames-tracker listener) alive forever.
+        weak var weakReporter: RNSentryOnDrawReporterView?
+
+        autoreleasepool {
+            let reporter = RNSentryOnDrawReporterView.createWithMockedListener()
+            reporter!.initialDisplay = true
+            reporter!.parentSpanId = spanId
+            reporter!.didSetProps(["initialDisplay", "parentSpanId"])
+            weakReporter = reporter
+        }
+
+        XCTAssertNil(weakReporter, "RNSentryOnDrawReporterView leaked due to a retain cycle")
+    }
 }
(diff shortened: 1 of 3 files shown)


packages/core/RNSentryCocoaTester/RNSentryCocoaTesterTests/RNSentryOnDrawReporterTests.swift
```diff
@@ -145,4 +145,22 @@ final class RNSentryOnDrawReporterTests: XCTestCase {
...
+
+    func testReporterViewIsDeallocatedAfterRelease() {
+        // Guards against the retain cycle described in
+        // https://github.com/getsentry/sentry-react-native/issues/6440 where the
+        // emit-new-frame block captured `self` strongly, keeping the view (and
+        // its frames-tracker listener) alive forever.
+        weak var weakReporter: RNSentryOnDrawReporterView?
+
+        autoreleasepool {
+            let reporter = RNSentryOnDrawReporterView.createWithMockedListener()
```
philprime, line 157: `m`: Force-unwrapping is generally discouraged because it would crash the test runner and not emit a test failure.

Consider using something like this:

```suggestion
            let reporter = try XCTUnwrap(RNSentryOnDrawReporterView.createWithMockedListener())
```

or 


```suggestion
            guard let reporter = RNSentryOnDrawReporterView.createWithMockedListener() else {
                return XCTFail("Failed to create mocked reporter view")
            }
```

philprime (approved):
LGTM when all tests pass. Left one code-style related comment.

lucas-zimerman (approved):
LGTM!

PR #6458, opened 2026-07-16
fix(core): Fix stale route when reactNavigationIntegration uses a route override provider

### LINKED ISSUE #6436: `reactNavigationIntegration`: navigation transaction can be named/attributed for the previous route instead of the destination when using a route override provider (e.g. `expoRouterIntegration`)

Expo Router

sentry.io (SaS)

No

_No response_

8.16.0

````

System:
  OS: macOS 26.5.2
  CPU: (16) arm64 Apple M3 Max
  Memory: 5.48 GB / 64.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 24.12.0
  Yarn:
    version: 4.12.0
  npm:
    version: 11.6.2
  Watchman:
    version: 2024.10.14.00
Managers:
  CocoaPods:
    version: 1.15.2
    path: /usr/local/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.2
      - iOS 26.2
      - macOS 26.2
      - tvOS 26.2
      - visionOS 26.2
      - watchOS 26.2
  Android SDK: Not Found
IDEs:
  Xcode:
    version: 26.2/17C52
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.11
    path: /usr/bin/javac
  Ruby:
    version: 2.6.10
    path: /usr/bin/ruby
npmPackages:
    installed: 20.2.0
    wanted: latest
  react:
    installed: 19.2.0
    wanted: 19.2.0
  react-native:
    installed: 0.83.6
    wanted: 0.83.6
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: true
iOS:
  hermesEnabled: true
  newArchEnabled: true

````

### Sentry.init()

````js
Sentry.init({
    appHangTimeoutInterval: 4,
    enableAppStartTracking: true,
    enabled: isSentryEnabled,
    enableStallTracking: true,
    ignoreErrors: ignoredErrors,
    integrations: [
      }),
    ],
    maxValueLength: 5000,
    sendDefaultPii: true,
    spotlight:
});
````

### Steps to Reproduce

1. Initialize Sentry with `expoRouterIntegration()` (which registers a route override provider that resolves route info via Expo Router's `store.getRouteInfo()`).
2. Navigate to screen A.
3. Before screen A's data/transaction has finished loading, navigate again to a different screen B (e.g. tap a tab bar item or another nav trigger while A is still fetching).
4. Inspect the resulting Sentry transaction for the second navigation (to screen B).

### Expected Result

The transaction's name, `route.name`, and `route.path` reflect screen B — the screen actually navigated to.

### Actual Result

The transaction is named and attributed as screen A (the previous route), even though `route.key` — which is read directly from React Navigation's own `navigationContainer.getCurrentRoute()`, not through the override — correctly shows screen B. So the same span ends up with a `route.key` pointing at the new screen but a `route.name`/`route.path`/transaction name pointing at the old one.

Root cause: `reactNavigationIntegration`'s `updateLatestNavigationSpanWithCurrentRoute` is registered via `navigationContainer.addListener('state', ...)`. Expo Router's own router-store, however, refreshes its cached `routeInfo` (which `store.getRouteInfo()` reads from) through the `NavigationContainer`'s `onStateChange` prop, not through `addListener('state', ...)`. In `@react-navigation/core`'s `BaseNavigationContainer.tsx`, on every state change these fire in this fixed order:

```js
emitter.emit({ type: 'state', data: { state } });   // notifies addListener('state', ...) subscribers — this is what reactNavigationIntegration listens on

if (!isFirstMountRef.current && onStateChangeRef.current) {
  onStateChangeRef.current(hydratedState);          // the onStateChange PROP — this is how Expo Router's own store.onStateChange refreshes its routeInfo cache
}
```

Because `emit('state')` always runs before the `onStateChange` prop callback, `reactNavigationIntegration`'s listener reads Expo Router's `store.getRouteInfo()` one step before Expo Router itself has updated it for that specific transition — so the override returns the templated path for the route being navigated away from. Since `updateLatestNavigationSpanWithCurrentRoute` clears its span reference (`latestNavigationSpan = undefined`) as soon as it runs once, there's no second chance to pick up the corrected value on a later 'state' event for the same navigation. This is deterministic for the affected dispatch, not a flaky timing race — it's most visible when navigating away from a screen quickly, since there's less unrelated state churn to have coincidentally already nudged Expo Router's cache forward beforehand.

alwx:
- [x] Bugfix

## :scroll: Description
`reactNavigationIntegration` subscribes to route changes via `navigationContainer.addListener('state', ...)`. In React Navigation's `BaseNavigationContainer`, on every state change these fire back-to-back in this order:

```js
emitter.emit({ type: 'state', data: { state } });   // → our listener
if (!isFirstMountRef.current && onStateChangeRef.current) {
  onStateChangeRef.current(hydratedState);          // → Expo Router refreshes its store here
}
```

Integrations like `expoRouterIntegration` supply a `RouteOverrideProvider` that reads Expo Router's `store.getRouteInfo()`, and that store is refreshed via the `onStateChange` prop — which runs *after* our listener. So we were reading the override for the previous route on every transition. Since `updateLatestNavigationSpanWithCurrentRoute` clears its span reference after one call, there was no second chance to correct it.

Fix: schedule the update as a microtask so the read happens after the synchronous state-change chain unwinds and downstream caches (Expo Router's store) have been refreshed.

Fixes #6436.

- Tests are added

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 - Fix `TypeError` when `showFeedbackForm`/`showFeedbackButton`/`showScreenshotButton` is called before `FeedbackFormProvider` mounts ([#6435](https://github.com/getsentry/sentry-react-native/pull/6435))
 - Fix orphaned TTID/TTFD spans in the trace view ([#6437](https://github.com/getsentry/sentry-react-native/pull/6437))
 - Fix iOS retain cycle in `RNSentryOnDrawReporterView` leaking TTID/TTFD reporter views and their frame-tracker listeners ([#6449](https://github.com/getsentry/sentry-react-native/pull/6449))
-- Fix `reactNavigationIntegration` reading a stale route from an override provider (e.g. `expoRouterIntegration`), causing navigation transactions to be named/attributed for the previous route ([#6436](https://github.com/getsentry/sentry-react-native/issues/6436))
+- Fix `reactNavigationIntegration` reading a stale route from an override provider (e.g. `expoRouterIntegration`), causing navigation transactions to be named/attributed for the previous route ([#6458](https://github.com/getsentry/sentry-react-native/pull/6458))
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Fix `TypeError` when `showFeedbackForm`/`showFeedbackButton`/`showScreenshotButton` is called before `FeedbackFormProvider` mounts ([#6435](https://github.com/getsentry/sentry-react-native/pull/6435))
 - Fix orphaned TTID/TTFD spans in the trace view ([#6437](https://github.com/getsentry/sentry-react-native/pull/6437))
 - Fix iOS retain cycle in `RNSentryOnDrawReporterView` leaking TTID/TTFD reporter views and their frame-tracker listeners ([#6449](https://github.com/getsentry/sentry-react-native/pull/6449))
+- Fix `reactNavigationIntegration` reading a stale route from an override provider (e.g. `expoRouterIntegration`), causing navigation transactions to be named/attributed for the previous route ([#6436](https://github.com/getsentry/sentry-react-native/issues/6436))
 
 ### Internal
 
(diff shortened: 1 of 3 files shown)


lucas-zimerman (approved):
lGTM!

Issue #4252, opened 2024-11-08
Support for Gradle Configuration Cache


smellouk:
### What React Native libraries do you use?

React Native without Frameworks

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### @sentry/react-native SDK Version

5.25.0

### How does your development environment look like?

OS: mac-os
arch: arm64
react native: 0.74.3-0
hermes enabled: true

### Sentry.init()

```
 sentry.init({
        dsn: sentryConfig.dsn,
        release: sentryConfig.release,
        enabled: sentryConfig.enabled,
        autoSessionTracking: false,
        ignoreErrors: [],
    });
```

### Steps to Reproduce

1. Export  `SENTRY_DISABLE_AUTO_UPLOAD=false`
2. Export  `SENTRY_AUTH_TOKEN=YOUR_TOKEN`
3. enable `org.gradle.configuration-cache=true` in `gradle.properties`
4. set warn level for configuration cache problems `org.gradle.unsafe.configuration-cache-problems=warn` in gradle properties 
5. build release app apk or bundle `./gradlew assembleRelease`

### Expected Result

Building/Assembling APK or bundle tasks should succeed without error. 

### Actual Result

```
See the complete report at .../configuration-cache/arbf12exyse0x2d67zxpdnb38/e07xbrg4hyz0msuzuubqcqzuy/configuration-cache-report.html

FAILURE: Build failed with an exception.

* Where:
Script '.../node_modules/@sentry/react-native/sentry.gradle' line: 104

* What went wrong:
Could not evaluate onlyIf predicate for task ':app:createBundleReleaseJsAndAssets_SentryUpload_PACKAGE_ID@VERSION_NAME+VERSION_CODE'.
> Could not evaluate spec for 'Task satisfies onlyIf closure'.
```

smellouk:
@krystofwoldrich Thanks so much for looking into this matter 🙏!

We’re working on a project that’s growing in complexity, and we’d like to optimize the Gradle configuration to reduce assembly time. Additionally, we have some automation tasks in place, and as a one-time setup, we’d like to enable the automatic upload of sourcemaps and ProGuard mappings. 

It’s not urgent for now, we disabled our gradle configuration cache for now, but we’d appreciate your help when possible.

krystofwoldrich:
[...]

PR #6469, opened 2026-07-17
fix(android): make Sentry Gradle tasks Configuration Cache compatible

### LINKED ISSUE #6466: sentry.gradle.kts still breaks Gradle Configuration Cache on 8.19.0 / Gradle 9 — a small, targeted fix is possible ahead of the SAGP migration (#796)

Hermes, RN New Architecture, React Native without Frameworks

### Are you using [sentry.io](<http://sentry.io>) or on-premise?

[sentry.io](<http://sentry.io>) (SaS)

Yes

Datadog (`@datadog/mobile-react-native`) and Luciq (`@luciq/react-native`). Noted because both ship Gradle scripts with the *same* Configuration Cache incompatibility — this report is scoped to Sentry's.

8.19.0 (latest). Confirmed via a Configuration Cache **build failure on 8.10.0**; the code responsible is present verbatim in 8.19.0's `sentry.gradle.kts` (line references below), so it applies to latest — I have inspected the source but not re-run a build on 8.19.0.

```
Gradle: 9.0.0 (Configuration Cache is the recommended execution mode; Gradle prints this on every build)
AGP: 8.x
React Native: 0.83.9, Hermes enabled, newArchEnabled=true
Build: ./gradlew :app:bundleRelease (release AAB; observed on Linux CI, ubuntu-24.04)
```

### Sentry.init()

```js
Sentry.init({
  dsn: 'https://...@sentry.io/...',
  // standard options; the failure is at build time, independent of runtime init options
});
```

### Steps to Reproduce

1. Set `org.gradle.configuration-cache=true` in `android/gradle.properties`
2. Set `org.gradle.configuration-cache.problems=warn` (to get past warn-level problems and surface the hard failures)
3. `./gradlew :app:bundleRelease` (or any release bundle/assemble)

### Expected Result

The build succeeds with Configuration Cache enabled (stores an entry on the first run, reuses it — and skips the configuration phase — on subsequent runs).

### Actual Result

`problems=warn` does not rescue these — they are hard execution-time exceptions (this is also what [#4252](<https://github.com/getsentry/sentry-react-native/issues/4252>) reported, with warn mode set):

```
* What went wrong:
Could not evaluate onlyIf predicate for task
  ':app:createBundleReleaseJsAndAssets_SentryUpload_<pkg>@<version>+<code>'.
> Could not evaluate spec for 'Task satisfies onlyIf closure'.
```

The same failure also occurs for `:app:copySentryJsonConfiguration`.

---

## Root cause

Line references are to `packages/core/sentry.gradle.kts` on `main` (shipped in 8.19.0); the Groovy `sentry.gradle` in 8.10.0 has the identical constructs.

**1.** `onlyIf` **predicates call closures resolved from** `project`**'s extra properties, and are evaluated at execution time.**

```kotlin
// sentry.gradle.kts
fun shouldCopySentryOptionsFile(): Boolean {
    val closure = extra["shouldCopySentryOptionsFile"] as groovy.lang.Closure<*>   // L87 — reads project state
    return closure.call() as Boolean
}

tasks.register("copySentryJsonConfiguration") {                                     // L103 (unconditional)
    onlyIf { shouldCopySentryOptionsFile() }                                        // L104 — evaluated at execution time → project access
    ...
}
```

Also `onlyIf { shouldCopySentryOptionsFile() }` at L159 and `onlyIf { shouldSentryAutoUploadGeneral() }` at L561 (the `*_SentryUpload` task). `onlyIf` runs at execution time; resolving the extra-property closure there is the CC violation.

**2.** `copySentryJsonConfiguration`**'s** `doLast` **touches** `project` **at execution time.**

```kotlin
doLast {
    val appRoot = project.rootDir.parentFile ?: project.rootDir   // L106 — project at execution time
    ...
    copy { from(sentryOptionsFile); into(androidAssetsDir); ... }  // L113 — project.copy at execution time
}
```

**3.** `*_SentryUpload` **uses** `project.logger` **at execution time** (e.g. L594, L596, and elsewhere in its actions). The CLI invocation itself already uses the CC-safe injected `ExecOperations` (`InjectedExecOps`, defined L76, instantiated L577) — so only these residual `project` accesses remain.

## Why the usual escape hatches don't work

* **Upgrading** to the 8.14.0+ `.kts` rewrite ([#6119](<https://github.com/getsentry/sentry-react-native/issues/6119>)) does not help — verified by reading `sentry.gradle.kts` on `main`: it is a syntax migration, the CC-relevant constructs above are unchanged.
* **Env-var gating** (`SENTRY_DISABLE_AUTO_UPLOAD=true`, `SENTRY_COPY_OPTIONS_FILE=false`) does not help: `copySentryJsonConfiguration` is registered unconditionally (L103) and wired unconditionally via `preBuild.dependsOn("copySentryJsonConfiguration")` in `afterEvaluate` (L781–782), so it is always in the task graph and its `onlyIf` is always *evaluated*. The env var only changes the predicate's *result*, not whether the failing evaluation runs.

## Proposed minimal fix (interim, ahead of [#796](<https://github.com/getsentry/sentry-react-native/issues/796>))

All CC-safe patterns, no behaviour change:

1. `onlyIf` **(L104, L159, L561)** — capture the env read at configuration time and reference the captured boolean:

   ```kotlin
   val copyOptionsFile   = System.getenv("SENTRY_COPY_OPTIONS_FILE") != "false"
   val autoUploadGeneral = System.getenv("SENTRY_DISABLE_AUTO_UPLOAD") != "true"
   // ...
   onlyIf { copyOptionsFile }
   onlyIf { autoUploadGeneral }
   ```

   (`System.getenv` at configuration time is a tracked CC input.)
2. `project.rootDir` **(L106)** — capture at configuration time and use the local in `doLast`.
3. `copy {}` **(L113)** — use an injected `FileSystemOperations` (same injection pattern already used for `ExecOperations`), or `java.nio.file.Files.copy` for the single-file case.
4. `project.logger` **(L594/L596/actions)** — use the task's own `logger` instead of `project.logger` (the task logger is CC-safe).

I'm happy to open a PR with these changes as an interim step before the full SAGP migration in [#796](<https://github.com/getsentry/sentry-react-native/issues/796>) — let me know if you'd prefer to fold it directly into that work instead.

## References

* [#4252](<https://github.com/getsentry/sentry-react-native/issues/4252>) — Support for Gradle Configuration Cache (closed NOT_PLANNED, deferred to the SAGP migration)
* getsentry/sentry-android-gradle-plugin#796 — Migrate `react-native/sentry.gradle` to SAGP (open; last roadmap estimate 2025-12-10 has since slipped)
* [#6119](<https://github.com/getsentry/sentry-react-native/issues/6119>) — Kotlin DSL conversion of `sentry.gradle` (syntax migration; CC behaviour unchanged)

fabiendem:
## 📜 Description

Makes the two Sentry Gradle tasks compatible with the Gradle **Configuration Cache** by removing `project`-state access at execution time.

Fixes the hard failure reported in getsentry/sentry-react-native#6466 (and the earlier [#4252](<https://github.com/getsentry/sentry-react-native/issues/4252>)):

```
Could not evaluate onlyIf predicate for task
  ':app:createBundleReleaseJsAndAssets_SentryUpload_<pkg>@<version>+<code>'.
> Could not evaluate spec for 'Task satisfies onlyIf closure'.
```

...which also occurs for `:app:copySentryJsonConfiguration`. These are hard, non-warnable exceptions — `org.gradle.configuration-cache.problems=warn` does not rescue them.

### What changed in `packages/core/sentry.gradle.kts`

The exec-time `project` accesses were:

1. `onlyIf { shouldCopySentryOptionsFile() }` / `onlyIf { shouldSentryAutoUploadGeneral() }` — these resolve closures stored in `project.extra`, so evaluating the `onlyIf` (at execution time) dereferences `project`.
2. `copySentryJsonConfiguration`'s `doLast`: `project.rootDir` and `copy { … }` (i.e. `project.copy`).
3. The `*_SentryUpload` task actions: `project.logger` and `Project.file(...)`.
4. `resolveSentryCliPackagePath(...)` (called from the upload `doLast`) used `.directory(rootDir)`.

Fixes, all behaviour-preserving and using patterns already present in this file (the CLI itself already runs via an injected `ExecOperations`):

* Env reads captured into config-time `val`s referenced by the `onlyIf` specs.
* `project.rootDir` captured at configuration time.
* `copy { }` → an injected `FileSystemOperations` (`InjectedFsOps`), mirroring `InjectedExecOps`.
* `project.logger` → the task's own `logger`.
* `Project.file("$reactRoot/…")` → `File("$reactRoot/…")` (path is already absolute).
* `resolveSentryCliPackagePath` resolves from `reactRoot` instead of `rootDir`.

This is an interim step ahead of the full SAGP migration tracked in getsentry/sentry-android-gradle-plugin#796.

### Out of scope (called out for transparency)

`resolveSentryReactNativeSDKPath` still runs `node --print require.resolve(...)` as an **external process at configuration time**. That is a separate, *warn-level* Configuration Cache problem shared with several other React Native libraries (netinfo, keychain, keyboard-controller, …) and does **not** fail the build the way the `onlyIf` did. Left for a follow-up so this PR stays focused on the hard failures.

## 💡 Motivation and Context

Gradle 9 recommends the Configuration Cache by default and nags on every build; the configuration phase is a meaningful share of Android build time on large RN apps. Today `sentry.gradle(.kts)` makes the build error out entirely when it's enabled. See getsentry/sentry-react-native#6466 for the full analysis.

## 💚 How did you test it?

Verified end-to-end on a **React Native 0.83.9** app with **Gradle 9.0.0** (`org.gradle.configuration-cache=true`, `problems=warn`).

Because the released `8.x` line still ships the Groovy `sentry.gradle` (the `.kts` is a `8.14.0+` rewrite of the same logic), the fix was proven against the equivalent Groovy script installed via 8.10.0, then ported here 1:1 to Kotlin:

| Build | `sentry.gradle` | Result |
| -- | -- | -- |
| Before | unpatched | ❌ `Could not evaluate onlyIf predicate for task ':app:copySentryJsonConfiguration'` → **BUILD FAILED** |
| After | patched (this change) | ✅ `> Task :app:copySentryJsonConfiguration` runs → **BUILD SUCCESSFUL** |

After the fix, the remaining Configuration Cache problems are only the warn-level config-time external-process ones noted above (no `sentry.gradle` hard failure).

The `*_SentryUpload` task action changes (`logger`, `file()`, `resolveSentryCliPackagePath`) are the same violation classes; they were applied by static analysis (a real upload run needs live credentials), with `copySentryJsonConfiguration` as the empirical anchor for the shared `onlyIf`/`project`-access mechanism.

## 📝 Checklist

- [X] I added a `CHANGELOG.md` entry (Unreleased → Fixes).
- [X] No public API changes; no behaviour change.

Closes getsentry/sentry-react-native#6466

---

*Generated with [Claude Code](<https://claude.com/claude-code>)*


packages/core/sentry.gradle.kts
```diff
@@ -558,7 +570,7 @@ fun processVariant(v: Any) {
 
         val cliTask =
             tasks.register(nameCliTask) {
-                onlyIf { shouldSentryAutoUploadGeneral() }
+                onlyIf { sentryAutoUploadGeneralEnabled }
```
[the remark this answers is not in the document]
fabiendem: Good catch — this was a genuine regression, not a false positive.

`plugin/src/withSentryAndroid.ts` injects `project.ext.shouldSentryAutoUploadGeneral = { -> return false }` immediately after `apply from` when `disableAutoUpload: true`, and the original `onlyIf { shouldSentryAutoUploadGeneral() }` consulted that override at execution time. Capturing `System.getenv("SENTRY_DISABLE_AUTO_UPLOAD")` at apply-time bypassed it, so `disableAutoUpload` would have been silently ignored.

Fixed in 7d293fd:

- The `*_SentryUpload` `onlyIf` now captures `shouldSentryAutoUploadGeneral()` **inside the `onVariants` callback**, which runs after the app `build.gradle` has evaluated — so the post-`apply from` override is in effect.
- `copySentryJsonConfiguration` / `cleanupTemporarySentryJsonConfiguration` likewise capture `shouldCopySentryOptionsFile()` via the closure rather than re-reading the env var, so `project.ext` overrides of it keep working too.

Both `onlyIf` specs still reference a captured boolean, so the tasks stay Configuration Cache compatible and the original override behavior is restored. Thanks for the catch!

packages/core/sentry.gradle.kts
```diff
@@ -100,17 +105,25 @@ val config: Map<String, Any?> =
...
+// Values captured at configuration time so task onlyIf specs and actions do not read
+// `project` state at execution time (required for Gradle Configuration Cache compatibility).
+// The overridable `shouldCopySentryOptionsFile()` closure is still consulted (not the env var
+// directly) so setups that reassign `project.ext.shouldCopySentryOptionsFile` keep working.
+val copyOptionsFileEnabled = shouldCopySentryOptionsFile()
+val rootDirFile = project.rootDir
+
 tasks.register("copySentryJsonConfiguration") {
-    onlyIf { shouldCopySentryOptionsFile() }
+    onlyIf { copyOptionsFileEnabled }
```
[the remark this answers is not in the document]
fabiendem: Fixed in 98dadca. `copyOptionsFileEnabled` is now a `Property` populated in `afterEvaluate` — which runs after the app `build.gradle`, so a `project.ext.shouldCopySentryOptionsFile` override placed after `apply from` is honored — and read via `.get()` in the copy/cleanup `onlyIf`. This mirrors the deferral already applied for `shouldSentryAutoUploadGeneral`, and stays Configuration Cache compatible (the `onlyIf` references a serializable `Property`, not `project`). Thanks!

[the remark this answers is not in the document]
fabiendem: Fixed in 98dadca. `copyOptionsFileEnabled` is now a `Property` populated in `afterEvaluate` — which runs after the app `build.gradle`, so a `project.ext.shouldCopySentryOptionsFile` override placed after `apply from` is honored — and read via `.get()` in the copy/cleanup `onlyIf`. This mirrors the deferral already applied for `shouldSentryAutoUploadGeneral`, and stays Configuration Cache compatible (the `onlyIf` references a serializable `Property`, not `project`). Thanks!

fabiendem:
Update — I ran a full `:app:createBundleDevGooglePlayReleaseJsAndAssets` under `--configuration-cache` (React Native 0.83.9, Gradle 9.0.0), which actually exercises the `*_SentryUpload` task that my earlier revisions had only reasoned about statically. That surfaced a real failure I had missed:

```
Execution failed for task ':app:...JsAndAssets_SentryUpload_...'.
> Could not find method resolveSentryCliPackagePath() for arguments [...] on object of type org.gradle.process.internal.DefaultExecAction_Decorated.
```

The upload task's `doLast` called the top-level script method `resolveSentryCliPackagePath(reactRoot)` at execution time; under the configuration cache a serialized task action can't resolve top-level script methods. Fixed in 98dadca by resolving it at configuration time and capturing the result (mirroring `resolveSentryReactNativeSDKPath`, which was already resolved at config time).

After that fix, a fresh configuration-cache store shows `copySentryJsonConfiguration`, `_SentryUpload`, and `_SentryCollectModules` all executing with no `sentry.gradle` hard failure. The only remaining `sentry.gradle` entries are two **warn-level** config-time `require.resolve` external processes (`resolveSentryReactNativeSDKPath` / `resolveSentryCliPackagePath`) — the same shared React-Native-ecosystem pattern noted as out of scope in the PR description.

So the change now covers both `copySentryJsonConfiguration` and `_SentryUpload` end-to-end (empirically verified, not just by static analysis). Apologies for the earlier over-claim on `_SentryUpload`.

packages/core/sentry.gradle.kts
```diff
@@ -207,7 +221,7 @@ fun resolveSentryCliPackagePath(reactRoot: File): String {
                     "--print",
                     "require.resolve('@sentry/cli/package.json', { paths: [require.resolve('@sentry/react-native/package.json')] })",
                 ),
-            ).directory(rootDir)
+            ).directory(reactRoot)
```
alwx: This has been changed to `.directory(reactRoot)` but several lines (at line 181) `.directory(rootDir)` is still being used. Could you change it for consistency?

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Fixes
+
+- Make `copySentryJsonConfiguration` and the `*_SentryUpload` Gradle tasks compatible with the Gradle Configuration Cache ([#6466](https://github.com/getsentry/sentry-react-native/issues/6466))
```
alwx: This should point to the PR and not to the issue.

packages/core/sentry.gradle.kts
```diff
@@ -100,17 +105,26 @@ val config: Map<String, Any?> =
 val configFile = "sentry.options.json"
 val androidAssetsDir = File("$rootDir/app/src/main/assets")
 
+// Values captured at configuration time so task onlyIf specs and actions do not read
+// `project` state at execution time (required for Gradle Configuration Cache compatibility).
+// `copyOptionsFileEnabled` is a Property populated in `afterEvaluate` (below) rather than at
+// apply-time, so a `project.ext.shouldCopySentryOptionsFile` override placed after `apply from`
+// is still honored. Referencing the Property in `onlyIf` keeps the tasks Config Cache compatible.
+val copyOptionsFileEnabled = objects.property(Boolean::class.java)
```
alwx: This has no default. If `afterEvaluate`at line 117 doesn't run (due to let's say a configuration error), `.get()` inside `onlyIf` will throw an `IllegalStateException`

alwx (changes requested):

packages/core/sentry.gradle.kts
fabiendem: Done in c2fb76f — reverted to `.directory(rootDir)` to match `resolveSentryReactNativeSDKPath`. Both helpers run at configuration time now, so `rootDir` is Configuration Cache safe in both.

CHANGELOG.md
fabiendem: Fixed in c2fb76f — now links #6469.

packages/core/sentry.gradle.kts
fabiendem: Good point — fixed in c2fb76f by adding `.convention(true)`, which preserves the documented default (copy enabled, i.e. `SENTRY_COPY_OPTIONS_FILE != "false"`) so `.get()` returns the default instead of throwing if `afterEvaluate` never runs.

alwx (approved):

PR #6472, opened 2026-07-18
fix(tracing): don't create a duplicate navigation span for Expo Router's `withAnchor` POP_TO bookkeeping dispatch

### LINKED ISSUE #6434: `withAnchor`'s internal `POP_TO` bookkeeping dispatch creates a spurious duplicate navigation span

Expo Router

sentry.io (SaS)

No

_No response_

8.16.0

````

System:
  OS: macOS 26.5.2
  CPU: (16) arm64 Apple M3 Max
  Memory: 5.48 GB / 64.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 24.12.0
  Yarn:
    version: 4.12.0
  npm:
    version: 11.6.2
  Watchman:
    version: 2024.10.14.00
Managers:
  CocoaPods:
    version: 1.15.2
    path: /usr/local/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.2
      - iOS 26.2
      - macOS 26.2
      - tvOS 26.2
      - visionOS 26.2
      - watchOS 26.2
  Android SDK: Not Found
IDEs:
  Xcode:
    version: 26.2/17C52
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.11
    path: /usr/bin/javac
  Ruby:
    version: 2.6.10
    path: /usr/bin/ruby
npmPackages:
    installed: 20.2.0
    wanted: latest
  react:
    installed: 19.2.0
    wanted: 19.2.0
  react-native:
    installed: 0.83.6
    wanted: 0.83.6
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: true
iOS:
  hermesEnabled: true
  newArchEnabled: true


````

### Sentry.init()

````js
Sentry.init({
    appHangTimeoutInterval: 4,
    enableAppStartTracking: true,
    enabled: isSentryEnabled,
    enableStallTracking: true,
    ignoreErrors: ignoredErrors,
    integrations: [
      }),
    ],
    maxValueLength: 5000,
    sendDefaultPii: true,
    spotlight:
});
````

### Steps to Reproduce

1. Initialize Sentry with `expoRouterIntegration({ useDispatchedActionData: true })`.
2. In an Expo Router app, trigger any navigation that passes `withAnchor: true`, e.g.: `router.dismissTo('/screenB', { withAnchor: true })`
3. Enable debug logging (`Sentry.init({ debug: true })`) and watch the console, or inspect the resulting traces in Sentry.
4. Note that `withAnchor: true` causes Expo Router to issue a *second* dispatch after the real navigation, purely to stamp `initial: false` onto the destination route (see expo-router's routing.js: `currentParams.initial = !withAnchor`, only ever set inside `if (withAnchor)`). This second dispatch is a `POP_TO` whose payload looks like: `{ type: 'POP_TO', payload: { name: '<destination route>', params: { initial: false } } }`
5. Observe the transactions/spans produced for this single user-facing navigation.

### Expected Result

A single navigation transaction is created and named for the real destination route (`/screenB`).

### Actual Result

Two navigation transactions are created. The first is the real navigation, correctly named. The second is a spurious duplicate transaction for the `POP_TO` bookkeeping dispatch — `startIdleNavigationSpan` runs unconditionally on every `'__unsafe_action__'` dispatch, and `POP_TO` isn't in the list of action types filtered out when `useDispatchedActionData` is enabled (`PRELOAD`, `SET_PARAMS`, `OPEN_DRAWER`, `CLOSE_DRAWER`, `TOGGLE_DRAWER` — `reactnavigation.ts`, `startIdleNavigationSpan`). This second transaction carries no real work of its own, but if the destination screen is still loading data when this bookkeeping dispatch fires, any in-flight spans (HTTP requests, TTID/TTFD) can get attached to it instead of the first (real) transaction, and it lingers until `childSpanTimeout` (15s default) force-closes it, showing up in Sentry as an extra, confusing transaction — sometimes named generically ("Route Change") if the route name never resolves cleanly, sometimes duplicating the real destination's name, with several of its child spans forced into a `cancelled` status when the idle span is torn down.

### PR DESCRIPTION
**Pull request by Cryptoteep**
## 📢 Type of change

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## 📜 Description

Expo Router issues a second `POP_TO` dispatch after a navigation with `withAnchor: true`, purely to stamp `initial: false` onto the destination it just navigated to (expo-router `getNavigationAction`: `currentParams.initial = !withAnchor`). With `useDispatchedActionData: true` this bookkeeping dispatch started a second idle navigation span for a single user-facing navigation. The spurious span stole in-flight child spans (HTTP, TTID/TTFD) from the real navigation, lingered until `childSpanTimeout` force-closed it, and shipped as a confusing duplicate transaction with several children in `cancelled` status.

Simply adding `POP_TO` to the filtered action list would break tracking of genuine `popTo`/`dismissTo` navigations (on React Navigation 7, `dismissTo` *is* a `POP_TO`). Instead this adds two scoped guards, both active only with `useDispatchedActionData`:

1. **Dispatch time** — skip a `POP_TO` whose target route is already focused somewhere on the active route chain (root → leaf, reusing the same traversal as `getPathFromState`). A genuine `popTo` targets a route *behind* the focused one, and `__unsafe_action__` fires before state changes are applied (documented in React Navigation's types), so its target is never focused at that point. This also protects the real navigation's in-flight span from being discarded by the "noop transaction" branch when the bookkeeping dispatch arrives before the deferred state-change microtask (getsentry/sentry-react-native#6436).
2. **State-change time** — a `POP_TO` span that landed on the *same route key* (the dispatch-time filter can miss when the payload carries a parent navigator name that matches no focused route) is now discarded instead of being abandoned to run out its idle timeout — unless a pending deep link claimed it, preserving the deep-link-to-current-screen attribution added earlier.

No behavior change when `useDispatchedActionData` is disabled (default): the action-type attribute is never set there, so both guards are inert.

## 💡 Motivation and Context

Fixes getsentry/sentry-react-native#6434

## 💚 How did you test it?

Added 7 unit tests covering: a genuine `POP_TO` to a different route still creates a named span; the bookkeeping `POP_TO` (leaf and nested/parent-navigator payload names) creates no span; it does not remove an in-flight navigation span from the scope; a same-route-key `POP_TO` span is discarded unless a deep link claimed it; and `NAVIGATE` landing on the same route keeps its current behavior. Full suite: `yarn jest test/tracing/reactnavigation.test.ts` — 121 passed, plus `pendingDeepLink`/`expoRouterIntegration`/`idleNavigationSpan`/`ttid` suites green. `tsc -p tsconfig.build.json` and `oxlint`/`oxfmt` clean.

## :pencil: Checklist

- [X] All tests passing.
- [X] No breaking changes.

## 🔮 Next steps

If preferred, the same-route-key discard (guard 2) could later be generalized beyond `POP_TO`, but that would change behavior for same-route `NAVIGATE` params updates, so it's deliberately left untouched here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Fixes
+
+- Fix spurious duplicate navigation transaction created by Expo Router's `withAnchor` bookkeeping `POP_TO` dispatch when `useDispatchedActionData` is enabled ([#6472](https://github.com/getsentry/sentry-react-native/pull/6472))
+
 ## 8.19.0
 
 ### Features
--- a/packages/core/src/js/tracing/reactnavigation.ts
+++ b/packages/core/src/js/tracing/reactnavigation.ts
@@ -261,7 +261,8 @@ export const reactNavigationIntegration = ({
   const spanDispatchSeq = new WeakMap<Span, number>();
 
   /**
-   * Attempts to attach the pending deep link to the given span.
+   * Attempts to attach the pending deep link to the given span. Returns `true`
+   * when the link was attached.
    *
    * Warm-open links only attach to spans dispatched *after* the link was
    * received. This prevents an unrelated, already-in-flight navigation from
@@ -274,10 +275,10 @@ export const reactNavigationIntegration = ({
    * Rejected warm-open links are left in the slot to be picked up by the next
    * eligible span.
    */
-  const applyPendingDeepLinkToSpan = (span: Span, maxAgeMs: number): void => {
+  const applyPendingDeepLinkToSpan = (span: Span, maxAgeMs: number): boolean => {
     const pending = peekPendingDeepLink(maxAgeMs);
     if (!pending) {
-      return;
+      return false;
     }
     if (pending.source === 'warm-open') {
       const spanSeq = spanDispatchSeq.get(span);
@@ -285,11 +286,11 @@ export const reactNavigationIntegration = ({
         // Span was dispatched before (or at the same tick as) the link arrived
         // — it cannot be the navigation the link triggered. Leave the link in
         // the slot for the next eligible span.
-        return;
+        return false;
       }
     }
     consumePendingDeepLink(maxAgeMs);
-    tagSpanWithDeepLink(span, pending);
+    return tagSpanWithDeepLink(span, pending);
   };
 
   /**
@@ -443,6 +444,32 @@ export const reactNavigationIntegration = ({
   };
 
   /**
+   * Returns `true` when the given route name is focused at some level of the
+   * current navigation state — i.e. it is on the chain of active routes from
+   * the root navigator down to the leaf. Falls back to comparing against the
+   * leaf route only when the full state is not available.
+   */
+  const isRouteFocused = (routeName: string): boolean => {
+    try {
+      const rootState = navigationContainer?.getState();
+      let currentState: NavigationState | undefined = rootState;
+      while (currentState) {
+        const route: NavigationRoute | undefined = currentState.routes[currentState.index ?? 0];
+        if (route?.name === routeName) {
+          return true;
+        }
+        currentState = route?.state;
+      }
+      if (!rootState) {
+        return navigationContainer?.getCurrentRoute()?.name === routeName;
+      }
+    } catch (e) {
+      debug.warn(`${INTEGRATION_NAME} Failed to read navigation state to check focused route.`, e);
+    }
+    return false;
+  };
+
+  /**
    * To be called on every React-Navigation action dispatch.
    * It does not name the transaction or populate it with route information. Instead, it waits for the state to fully change
    * and gets the route information from there, @see updateLatestNavigationSpanWithCurrentRoute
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by antonis** [comment]
Thanks for taking this on. The state change discard (guard 2) keyed on `route.key` is the right mechanism, and preserving the deep link case is a nice touch.

One issue with guard 1 (the dispatch time `isRouteFocused` skip): it matches by route **name**, so it also suppresses a genuine `popTo` to an earlier same named route in the stack (common in Expo Router with dynamic routes like `[id]`). Example: stack `[Detail(k1), Detail(k2)]` focused on `k2`, then `dismissTo` the earlier `Detail`. The target name equals the focused leaf name, so guard 1 skips a real navigation and its transaction is dropped. I confirmed this against the branch: a `popTo` to an earlier same named route produces 0 transactions instead of 1.

Suggested fix: also require the `withAnchor` marker on guard 1, since `params.initial` is only ever set by `withAnchor` (see expo router `getNavigationAction`), so genuine `popTo`s never carry it:

```ts
const popToParams = (event?.data?.action?.payload as { params?: { initial?: boolean } } | undefined)?.params;
if (
  navigationActionType === 'POP_TO' &&
  targetRouteName &&
  isRouteFocused(targetRouteName) &&
  popToParams?.initial === false
) {
  ...
}
```

I verified this keeps all current tests green (they already pass `initial: false`) and still preserves guard 1's protection of in flight spans, while letting real same named `popTo`s through. Guard 2 still backstops any nested case.

A test that fails on the current branch and passes with the change above:

```ts
test('popTo an earlier same-named route in the stack still creates a span', async () => {
  client.eventQueue = [];

  mockNavigation.emitWithStateChange(
    { data: { action: { type: 'PUSH', payload: { name: 'Detail' } }, noop: false, stack: undefined } },
    { key: 'detail_2', name: 'Detail' },
  );
  await jest.runOnlyPendingTimersAsync();
  client.eventQueue = [];

  mockNavigation.emitWithStateChange(
    { data: { action: { type: 'POP_TO', payload: { name: 'Detail' } }, noop: false, stack: undefined } },
    { key: 'detail_1', name: 'Detail' },
  );
  await jest.runOnlyPendingTimersAsync();
  await client.flush();

  expect(client.eventQueue.filter(e => e.type === 'transaction' && e.transaction === 'Detail').length).toBe(1);
});
```

(Add it inside the `useDispatchedActionData names spans from action payload` describe, which already sets up `client`/`mockNavigation`.)

antonis:
Thank you for you contributions 🙇
Closing in favor of https://github.com/getsentry/sentry-react-native/pull/6439

Issue #6168, opened 2026-05-18
Turbo Modules: Documentation, migration guide, and public API


alwx:
Land the user-facing story once the implementation stabilizes.

**Tasks**

* Public API: confirm the option name (`enableTurboModuleTracking`), the integration name (`turboModulesIntegration`), and the options bag (`slowCallThresholdMs`, `ignoreTurboModules`, top-N cap). Run through `yarn api-report` and lock the surface.
* `CHANGELOG.md` under Unreleased / Features.
* Coordinate a docs PR with the Sentry docs repo covering:
  * How the feature is enabled (new arch vs. old arch).
  * What data lands on spans (attributes, measurements) and on events (tags, contexts).
  * How crash reports attribute the active module + method.
  * Privacy / overhead notes — especially that method names are sent (so users on strict PII setups know what they're shipping).
* Add a section to `packages/core/AGENTS.md` describing the TurboModule perf logger plumbing so future contributors know where to look.
* Add information about new attributes to https://github.com/getsentry/sentry-conventions (https://getsentry.github.io/sentry-conventions/attributes/)
* Decide and document the deprecation story (if any) for the legacy NativeModules path — most users will migrate to the new arch over the next few RN versions, and the legacy path can be marked as best-effort.
*

PR #6478, opened 2026-07-20
feat(core): Attach TurboModule breakdown to active spans on spanEnd

### LINKED ISSUE #6165: JS-side integration: attach TurboModule stats to active spans
Native gives us the data; the JS layer is where it has to land on the right span. Today, idle navigation spans (`reactnavigation.ts`, `expoRouter.ts`) and the user's own `Sentry.startSpan(...)` calls have no visibility into native module work that happened during them.

**Goal**

* New integration: `turboModulesIntegration()` exported from `packages/core/src/js/integrations/exports.ts`.
* On install, the integration:
  * Calls into the native sink (Issue 1) to start a logical "JS observation window" tied to the active span (using `getActiveSpan()`).
  * On span end, pulls a delta of aggregate stats since the window started and attaches them as attributes (`turbo_module.<name>.<method>.call_count`, `..._duration_ms`, `..._error_count`) — capped to top-N modules by duration to keep payload size sane.
  * Adds a breadcrumb (`category: 'native.turbo_module'`, `level: 'info'`) for any async call that exceeded a configurable threshold (`slowCallThresholdMs`, default 500ms) including module + method + duration.
* Surface a tiny JS API for users who want explicit attribution:

  ```ts
  Sentry.startSpan({ name: 'load profile' }, async () => {
    // any TurboModule activity inside this callback is automatically attached
  });
  ```
* Auto-enabled when `enableTurboModuleTracking` (Issue 1) is on; no-op otherwise.

alwx:
## 📢 Type of change

- [X] New feature

## 📜 Description

On root `spanEnd`, attach per-`(module, method)` TurboModule call counts / duration / errors as `turbo_module.<name>.<method>.*` attributes on the span, plus summary keys. Async calls over `slowCallThresholdMs` (default 500ms) emit a `native.turbo_module` breadcrumb.

New options on `turboModuleContextIntegration`: `enableSpanAttribution` (default on), `slowCallThresholdMs`, `maxTopModulesPerSpan`.

Fed by a new per-record observer API on the JS aggregator so open spans can accumulate independently of the transaction / periodic flush drain path.

## 💡 Motivation and Context

Closes getsentry/sentry-react-native#6165

## 💚 How did you test it?

Unit tests for the span-attribution + breadcrumb paths and the observer contract. `yarn test`, `yarn lint`, `yarn circularDepCheck`, `yarn api-report:check` pass locally.

## :pencil: Checklist

- [X] All tests passing.
- [X] No breaking changes.

## 🔮 Next steps

Docs in getsentry/sentry-react-native#6168.


alwx:
@cursor review

packages/core/src/js/integrations/turboModuleContext.ts
```diff
@@ -202,183 +353,117 @@ export const turboModuleContextIntegration = (options: TurboModuleContextOptions
...
+  if (window.writtenPerMethodKeys) {
+    for (const key of window.writtenPerMethodKeys) {
+      if (!nextKeys.has(key)) {
+        attributes[key] = undefined;
       }
     }
   }
-  return out;
-}
+  window.writtenPerMethodKeys = nextKeys;
```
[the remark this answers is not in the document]
alwx: This isn't accurate — the Sentry Span API explicitly documents this pattern:

```ts
// @sentry/core: types/span.d.ts
/**
 * Set multiple attributes on the span.
 * Any attribute set to `undefined` will be removed.
 */
setAttributes(attributes: SpanAttributes): this;
```

And the implementation in `sentrySpan.js`:

```js
setAttribute(key, value) {
  if (value === void 0) {
    delete this._attributes[key];
  } else {
    this._attributes[key] = value;
  }
  // ...
}
setAttributes(attributes) {
  Object.keys(attributes).forEach((key) => this.setAttribute(key, attributes[key]));
}
```

`removeAttribute` isn't part of the public `Span` interface (calling it directly wouldn't type-check), and the test added in this PR (`clears stale per-method keys on re-emit after a late-settling async re-ranks the top-N`) verifies the intended behavior.

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@
 
 ## Unreleased
 
+### Features
+
+- Attach a per-`(module, method)` TurboModule breakdown to active spans on `spanEnd`, plus `native.turbo_module` breadcrumbs for slow async calls ([#6478](https://github.com/getsentry/sentry-react-native/pull/6478))
+
+  When a root span ends (idle nav spans from `reactNavigationIntegration` / `expoRouterIntegration`, or a user's own `Sentry.startSpan(...)`), the integration writes `turbo_module.<name>.<method>.{call_count,duration_ms,error_count}` attributes plus summary keys (`turbo_module.total_call_count`, `turbo_module.total_duration_ms`, `turbo_module.top_module`). Async calls above `slowCallThresholdMs` (default 500ms) additionally record a `native.turbo_module` breadcrumb. Both surfaces enabled by default; new knobs `enableSpanAttribution`, `slowCallThresholdMs`, `maxTopModulesPerSpan` on `turboModuleContextIntegration`.
```
antonis, line 15: Nit: Missing line before the `Fixes` section.

antonis (commented):
Overall the PR LGTM once the CI is 🟢 (apart from the [known issue](https://github.com/getsentry/sentry-react-native/issues/6497) related failures).

I would advocate towards making the changes opt-in / experimental first to gather some feedback from users (track adoption similar to [this](https://github.com/getsentry/sentry-react-native/pull/6428)) and verify stability before releasing for all. Wdyt?

packages/core/src/js/integrations/turboModuleContext.ts
```diff
@@ -180,205 +145,287 @@ export const turboModuleContextIntegration = (options: TurboModuleContextOptions
...
-        out[`${prefix}.${label}`] = count;
+  const spanId = spanToJSON(span).span_id;
+  if (spanId) {
+    if (!pendingSpanAttributes.has(spanId) && pendingSpanAttributes.size >= MAX_PENDING_SPAN_ATTRIBUTES) {
+      const oldest = pendingSpanAttributes.keys().next().value;
+      if (oldest !== undefined) {
+        pendingSpanAttributes.delete(oldest);
       }
     }
+    pendingSpanAttributes.set(spanId, attributes);
```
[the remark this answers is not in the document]
alwx: The suggested fix (merge instead of overwrite in `pendingSpanAttributes`) is functionally a no-op given the current design:

`attachWindowToSpan` iterates every row in `window.counters` and rebuilds the **full cumulative payload** on each call — records don't get removed from `window.counters`, they only accumulate. So the payload written on a late attach is a superset of every earlier payload for the same window. `{ ...existing, ...attributes }` and `set(spanId, attributes)` produce identical results.

The actual race being described — a late-async record settling **after** the transaction event has already flowed through `processEvent` — is a genuine architectural limitation, but not one this fix touches. Once the transaction is sent, no code path here can add attributes to it; `Span#setAttributes` on the frozen span is a no-op, and there's no future `processEvent` for that same event. The best mitigation would be a follow-up event with the late data, which is a substantially bigger change and out of scope here.

In practice, this only bites when an async TurboModule call outlives its owning transaction's send — the transaction still lands with its full sync + in-flight-at-send-time attribution, just missing the residual tail.

packages/core/src/js/integrations/turboModuleContext.ts
```diff
@@ -1,104 +1,54 @@
...
-   * signals with SDK internal noise and (b) allow the periodic flush's own
-   * `captureEvent` to record back into the aggregator and perpetually re-arm
-   * the flush timer in idle sessions. Pass `[]` to opt back in.
-   *
-   * Note: this does NOT disable wrapping — crashes during those calls still
-   * get attributed via `contexts.turbo_module`. It only opts the module out
-   * of the per-(module, method, kind) counters.
-   */
   ignoreTurboModules?: ReadonlyArray<string>;
+  enableSpanAttribution?: boolean;
```
antonis, line 67: q: why we removed the comments? it would be nice to keep the old and new attibutes documented

antonis (commented):
Overall LGTM and the CI is 🟢 
Left [a nit question](https://github.com/getsentry/sentry-react-native/pull/6478#discussion_r3644114086). 
Let's also get a ✅ from the @getsentry/team-mobile since this involves public api changes

packages/core/src/js/integrations/turboModuleContext.ts
alwx: fixed

antonis (approved):
LGTM 🚀 
Worths rechecking the latest cursor comments but nothing blocking on my end. Let's wait for the team review and 🚢 

Also iterating from a previous [previous review](https://github.com/getsentry/sentry-react-native/pull/6478#pullrequestreview-4754744186)

> I would advocate towards making the changes opt-in / experimental first to gather some feedback from users (track adoption similar to https://github.com/getsentry/sentry-react-native/pull/6428) and verify stability before releasing for all. Wdyt?

packages/core/src/js/integrations/turboModuleContext.ts
```diff
@@ -220,183 +299,155 @@ function stripEmptySentinelTags(event: Event): void {
...
+    'turbo_module.unique_methods': rows.length,
+  };
+  const top = rows[0];
   if (top) {
-    event.measurements['turbo_modules.top_module_ms'] = {
-      value: roundMs(top.totalDurationMs),
-      unit: 'millisecond',
-    };
+    attributes['turbo_module.top_module'] = `${safeKeyPart(top.name)}.${safeKeyPart(top.method)}`;
+    attributes['turbo_module.top_module_duration_ms'] = roundMs(top.totalDurationMs);
```
buenaflor, line 376: are these all new attributes? we should also add them here: https://github.com/getsentry/sentry-conventions

buenaflor (commented):
public API LGTM, just one comment about the attributes

packages/core/src/js/integrations/turboModuleContext.ts
alwx: if it's fine for you, I will add them separately as part of this issue: https://github.com/getsentry/sentry-react-native/issues/6168
also because other turbomodules related attributes need to be added there as well

alwx: added the entry about that to the issue above:
"* Add information about new attributes to https://github.com/getsentry/sentry-conventions (https://getsentry.github.io/sentry-conventions/attributes/)"

buenaflor: yeah sounds good

PR #6504, opened 2026-07-23
feat(core): Extend TurboModule instrumentation to legacy NativeModules

### LINKED ISSUE #6166: Legacy `NativeModules` wrapper for the old architecture
`TurboModulePerfLogger` only fires for TurboModules. Apps on the old architecture still talk to native through `NativeModules.*` over the bridge. To keep the feature useful for the long tail of un-migrated apps, mirror the same instrumentation surface on top of the legacy bridge.

getsentry/sentry-react-native#6477

* Two paths, pick the one that gives us cleaner UX:
  1. **Native:** use React Native's `BridgeNativeModulePerfLogger` (analogous to `TurboModulePerfLogger`, available on old arch) and install a Sentry logger from `RNSentryPackage` / `RNSentry.mm` the same way Issue 1 does.
  2. **JS proxy:** wrap `NativeModules` at runtime by replacing the `NativeModules[name]` getter with a `Proxy` that records call start/end. Cheaper to ship but adds a JS-side hot-path cost on every native call.
* Default to path (1); fall back to (2) only on RN versions where the bridge logger isn't accessible.
* Feed the same aggregator + crash-context surface from Issues 2 & 3 — the consumer side should not care whether the data came from a TurboModule or a legacy NativeModule. Add `turbo_module.arch: "new" | "legacy"` so analyses can split them.
* Respect `isTurboModuleEnabled()` / `isFabricEnabled()` detection from `packages/core/src/js/utils/environment.ts` to choose the right code path automatically.

alwx:
## 📢 Type of change

- [X] New feature

## 📜 Description

Extends the TurboModule aggregate + span attribution + slow-call breadcrumb surface (from getsentry/sentry-react-native#6377 and getsentry/sentry-react-native#6478) to Old Architecture apps by wrapping `NativeModules.*` at setup. Consumer side stays identical — only the `arch` field distinguishes new-arch (`'new'`) vs bridge (`'legacy'`) sources.

### Approach

This ships the **JS Proxy** path. The native `BridgeNativeModulePerfLogger` path (option 1) is symmetric but requires new C++/JNI code and a version gate — left for a follow-up. The JS approach works on every RN version we support and has the same shape as the existing TurboModule wrap; hot-path cost is bounded (one property access + function call per bridge invocation).

## 💡 Motivation and Context

`TurboModulePerfLogger` only fires for TurboModules, so apps on the Old Architecture were dark to the entire aggregator / span-attribution / breadcrumb surface. This PR closes that gap so the long tail of un-migrated apps gets the same signals.

Closes getsentry/sentry-react-native#6166.

## 💚 How did you test it?

There are a few tests added!

## 📝 Checklist

- [X] All tests passing.
- [X] No breaking changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 
 - Extend TurboModule instrumentation to the Old Architecture `NativeModules` bridge ([#6504](https://github.com/getsentry/sentry-react-native/pull/6504))
 
-  Apps still on the Old Architecture now flow legacy `NativeModules.*` calls through the same aggregator, span-attribution, and slow-call breadcrumb path as TurboModules on the New Architecture. Records carry `arch: 'new' | 'legacy'` (also surfaced on the flushed aggregate span and each attributed span) so analyses can split the signal by architecture. Auto-wrap is on by default; opt out with `enableLegacyNativeModules: false`, or scope it via `legacyModulesSkip` / `legacyModulesSkipMethods` on `turboModuleContextIntegration`.
+  Apps still on the Old Architecture now flow legacy `NativeModules.*` calls through the same aggregator, span-attribution, and slow-call breadcrumb path as TurboModules on the New Architecture. Records carry `arch: 'new' | 'legacy'` (also surfaced on the flushed aggregate span and each attributed span) so analyses can split the signal by architecture. Lazily-exposed modules stay lazy — they're wrapped on first access instead of being initialised during `Sentry.init`. `RNSentry` and hot RN infrastructure (`Timing`, `UIManager`, animated modules) are skipped by default. Auto-wrap is on by default; opt out with `enableLegacyNativeModules: false`, or scope it via `legacyModulesSkip` / `legacyModulesSkipMethods` on `turboModuleContextIntegration`.
 
 ## 8.21.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 
 - Extend TurboModule instrumentation to the Old Architecture `NativeModules` bridge ([#6504](https://github.com/getsentry/sentry-react-native/pull/6504))
 
-  Apps still on the Old Architecture now flow legacy `NativeModules.*` calls through the same aggregator, span-attribution, and slow-call breadcrumb path as TurboModules on the New Architecture. Records carry `arch: 'new' | 'legacy'` (also surfaced on the flushed aggregate span and each attributed span) so analyses can split the signal by architecture. Lazily-exposed modules stay lazy — they're wrapped on first access instead of being initialised during `Sentry.init`. `RNSentry` and hot RN infrastructure (`Timing`, `UIManager`, animated modules) are skipped by default. Auto-wrap is on by default; opt out with `enableLegacyNativeModules: false`, or scope it via `legacyModulesSkip` / `legacyModulesSkipMethods` on `turboModuleContextIntegration`.
+  Apps on the Old Architecture now get the same aggregate, span attribution and slow call breadcrumbs as TurboModules. An `arch: 'new' | 'legacy'` field distinguishes the two sources. Opt in with `turboModuleContextIntegration({ enableLegacyNativeModules: true })`.
 
 ## 8.21.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@
   });
   ```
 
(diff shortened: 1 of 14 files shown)


alwx:
@cursor review

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
...
 
+- Extend TurboModule instrumentation to the Old Architecture `NativeModules` bridge ([#6504](https://github.com/getsentry/sentry-react-native/pull/6504))
+
+  Apps still on the Old Architecture now flow legacy `NativeModules.*` calls through the same aggregator, span-attribution, and slow-call breadcrumb path as TurboModules on the New Architecture. Records carry `arch: 'new' | 'legacy'` (also surfaced on the flushed aggregate span and each attributed span) so analyses can split the signal by architecture. Lazily-exposed modules stay lazy — they're wrapped on first access instead of being initialised during `Sentry.init`. `RNSentry` and hot RN infrastructure (`Timing`, `UIManager`, animated modules) are skipped by default. Auto-wrap is on by default; opt out with `enableLegacyNativeModules: false`, or scope it via `legacyModulesSkip` / `legacyModulesSkipMethods` on `turboModuleContextIntegration`.
```
antonis: nit: I would advocate toward shortening this and skipping some of the implementation details

packages/core/src/js/integrations/turboModuleContext.ts
```diff
@@ -71,6 +73,24 @@ export interface TurboModuleContextOptions {
 
   /** Cap on per-`(module, method)` rows attributed to a single span. Default: `16`. */
   maxTopModulesPerSpan?: number;
+
+  /**
+   * On Old Architecture, auto-wrap registered `NativeModules.*`. Default: `true`.
```
antonis: q: Is there any risk in defaulting to `true`? Would it make sense to start with false and gather feedback?

antonis (commented):
Left a couple of comments alogn with the agent feedback but overall LGTM. Let's also add the changes of this PR in the sentry-conventions issue

lucas-zimerman:
> Left a couple of comments alogn with the agent feedback but overall LGTM. Let's also add the changes of this PR in the sentry-conventions issue

If so, shouldn't we block this PR until the changes are merged in sentry-conventions?

packages/core/src/js/turbomodule/wrapNativeModules.ts
```diff
@@ -0,0 +1,199 @@
...
+    return false;
+  }
+}
+
+function wrapResolvedModule(name: string, mod: unknown, wrapOptions: WrapOptions): boolean {
+  if (!mod || typeof mod !== 'object') {
+    return false;
+  }
+  wrapTurboModule(name, mod, wrapOptions);
+  return true;
```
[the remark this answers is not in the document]
antonis: This looks valid but is preexisting. Maybe we can handle it in a follow up. Wdyt?

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@
...
+
```
antonis, line 36: Let's also move this to the unreleased section

antonis (commented):
Left a few comments and [a question](https://github.com/getsentry/sentry-react-native/pull/6504/changes#r3665283218) along with the agent feedback but overall looks good 🚀

alwx:
@lucas-zimerman the only new field is arch: 'new' | 'legacy' on records we already emit — no new attribute keys. In addition, we have a conventions issue in this project anyway (https://github.com/getsentry/sentry-react-native/issues/6168)

packages/core/src/js/integrations/turboModuleContext.ts
alwx: Good call. Flipped it to opt-in (enableLegacyNativeModules: false by default)

packages/core/src/js/turbomodule/wrapNativeModules.ts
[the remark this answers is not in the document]
alwx: Created it here: https://github.com/getsentry/sentry-react-native/issues/6542

antonis (approved):
The changes LGTM 🚀 

Approving but let's also bump [the diffMax ](https://github.com/getsentry/sentry-react-native/blob/main/performance-tests/metrics-ios.yml#L14C3-L14C10) since the legacy architecture check [fails on iOS](https://github.com/getsentry/sentry-react-native/actions/runs/30540262442/job/90863271617?pr=6504) on this or a follow up PR.

```
  BinarySizeTest > app size() FAILED
      java.lang.AssertionError: 1782264 should be < 1689600
          at BinarySizeTest.app size(BinarySizeTest.kt:37)
```

PR #6534, opened 2026-07-28
fix(ios): make the RNSentry SPEC CHECKSUM in Podfile.lock machine-independent

### LINKED ISSUE #6467: iOS: RNSentry SPEC CHECKSUM is machine-specific (absolute xcframework path in FRAMEWORK_SEARCH_PATHS) → Podfile.lock churn across devs/CI
### Environment

- `@sentry/react-native`: 8.19.0 (current `latest`)
- `react-native`: 0.85.3
- CocoaPods: 1.17.0
- Platform: macOS, bare React Native app (Podfile.lock committed to git)

### Summary

Since the switch to consuming sentry-cocoa as a prebuilt xcframework by default (#6413), the `RNSentry` entry under `SPEC CHECKSUMS:` in `Podfile.lock` is **machine-specific**. Two developers (or a developer and CI) running `pod install` on the exact same `@sentry/react-native` version get **different** `RNSentry` checksums, so `Podfile.lock` flips back and forth on every install and can never be committed to a stable value.

### Root cause

`RNSentry.podspec` points `FRAMEWORK_SEARCH_PATHS` at the **absolute** pod-install-time path of the cached xcframework:

- `packages/core/scripts/sentry_utils.rb` (`ensure_sentry_xcframework`):
  ```ruby
  cache_root = ENV['SENTRY_XCFRAMEWORK_CACHE_DIR'] ||
               File.expand_path('~/Library/Caches/sentry-react-native/xcframeworks')
  ```
- `packages/core/RNSentry.podspec`:
  ```ruby
  paths = slice_ids.map { |slice| %("#{File.join(sentry_xcframework_dir, slice)}") }
  acc["FRAMEWORK_SEARCH_PATHS[sdk=#{sdk}*]"] = (['$(inherited)'] + paths).join(' ')
  ```

The resulting `pod_target_xcconfig` / `user_target_xcconfig` therefore contains e.g.:

```
FRAMEWORK_SEARCH_PATHS[sdk=iphoneos*] = $(inherited) "/Users/<username>/Library/Caches/sentry-react-native/xcframeworks/9.19.1/Sentry.xcframework/ios-arm64_arm64e"
```

Because CocoaPods computes the pod's `SPEC CHECKSUM` over the *evaluated* podspec (including this xcconfig), the `<username>` component leaks into the checksum. Different home directory → different checksum. The podspec comment notes the path "is regenerated on every `pod install`, so it's not something anyone commits" — but the **checksum derived from it is written into `Podfile.lock`**, which apps do commit.

### Reproduction

1. Install pods on machine A → note `RNSentry: <hashA>` in `Podfile.lock`.
2. Install pods on machine B (different `$HOME`), same `@sentry/react-native` version → `RNSentry: <hashB>`, `hashA != hashB`.
3. Committing either value means the other machine (and CI) re-changes it on the next `pod install`. Endless churn; no committable value.

### Impact

- `Podfile.lock` is non-deterministic across developers and CI, producing spurious diffs on every `pod install`.
- Teams that gate CI on a clean/consistent `Podfile.lock` get repeated failures or auto-commit ping-pong.

This is the same class of problem as [facebook/react-native#31193](https://github.com/facebook/react-native/issues/31193) ("FBReactNativeSpec pod checksum is not stable") — local absolute paths leaking into a committed pod checksum.

### Workaround (works today)

Set `SENTRY_XCFRAMEWORK_CACHE_DIR` to a fixed, username-free absolute path in the `Podfile` so every machine and CI evaluate the same string:

```ruby
ENV['SENTRY_XCFRAMEWORK_CACHE_DIR'] = '/tmp/sentry-react-native-xcframeworks'
```

Verified: with this set, the `RNSentry` checksum is identical and stable across repeated installs; without it, it reverts to the per-`$HOME` value.

### Suggested fix

Keep the fast prebuilt-xcframework path, but stop the machine-specific absolute path from entering the checksummed spec attributes — e.g. reference a build-setting placeholder (like `$(SENTRY_XCFRAMEWORK_DIR)/<slice>`) in `FRAMEWORK_SEARCH_PATHS` and inject the real absolute path via a mechanism that isn't part of the podspec checksum (e.g. a generated `.xcconfig`/build setting written at install time). That yields a deterministic, committable `Podfile.lock` while preserving static linking.

alwx:
## 📢 Type of change

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## 📜 Description

Since getsentry/sentry-react-native#6413, `RNSentry.podspec` points `FRAMEWORK_SEARCH_PATHS` at the absolute pod-install-time path of the cached `Sentry.xcframework` (`~/Library/Caches/sentry-react-native/xcframeworks/…`), which leaks `$HOME` into the `RNSentry` `SPEC CHECKSUM` written to `Podfile.lock`.

This stages the cached bundle behind a machine-independent reference instead:

* `stage_sentry_xcframework_in_pods` (new, `sentry_utils.rb`) symlinks `Pods/sentry-xcframeworks/<sentry-cocoa version>/Sentry.xcframework` → the cached bundle, recreating the link when it's missing or stale (e.g. `SENTRY_XCFRAMEWORK_CACHE_DIR` changed between installs) and pruning links left behind by earlier SDK versions.
* The podspec now writes `"$(PODS_ROOT)/sentry-xcframeworks/<version>/Sentry.xcframework/<slice>"` into `FRAMEWORK_SEARCH_PATHS` — the same string on every machine, so the checksum is deterministic. `$(PODS_ROOT)` is defined in both per-pod and user/aggregate xcconfigs (unlike `$(PODS_TARGET_SRCROOT)`), and because the link lives inside `Pods/` there is no Podfile-layout detection needed.
* Staging only happens inside a real `pod install` (gated on `Pod::Config.instance.podfile_path`). Anywhere else — `pod ipc spec`, `pod lib lint`, a plain `Specification.from_file` — `$(PODS_ROOT)` would point at an unrelated sandbox, so we fall back to the absolute cache path. Same fallback if the symlink can't be created: functional, just with a machine-specific checksum.

The download/cache/SHA256-verification flow is unchanged; the cache stays in `~/Library/Caches` (still shared across projects, still safe under pnpm's read-only store and Yarn PnP).

Note: existing projects will see a **one-time** `RNSentry` checksum change in `Podfile.lock` on their next `pod install`; after that the value is stable across developers and CI. Teams that adopted the `SENTRY_XCFRAMEWORK_CACHE_DIR=/tmp/…` workaround from getsentry/sentry-react-native#6467 can drop it — and mixed setups (some machines with the override, some without) now converge on the same checksum, since the override no longer appears in the spec.

## 💡 Motivation and Context

Fixes getsentry/sentry-react-native#6467.

CocoaPods persists external-source pods as an **evaluated** podspec JSON in `Pods/Local Podspecs/RNSentry.podspec.json` (`ExternalSources::AbstractExternalSource#store_podspec` always calls `sandbox.store_podspec(name, spec, true, true)`), and `Specification#checksum` is the SHA1 of that file. The evaluated `pod_target_xcconfig`/`user_target_xcconfig` therefore carried the `$HOME`-dependent path straight into `Podfile.lock`.

Consequence: two developers (or a developer and CI) installing the exact same `@sentry/react-native` version get different `RNSentry` checksums, so a committed `Podfile.lock` flips on every `pod install` and teams gating CI on a clean lockfile get repeated failures. Same class of problem as [facebook/react-native#31193](<https://github.com/facebook/react-native/issues/31193>).

Supersedes getsentry/sentry-react-native#6474, with two fixes on top of it:

1. That PR fell back to `<cwd>/Pods` when `Pod::Config` was unavailable **and still returned a** `$(PODS_ROOT)` **path**. CocoaPods evaluates podspecs with the CWD set to the podspec's own directory, so that would have staged the link inside `node_modules/@sentry/react-native/` while handing back a search path that dangles at build time — a hard `module 'Sentry' not found` instead of the safe absolute-path fallback.
2. Version-scoped staging directories from previous SDK versions were never cleaned up, accumulating dangling symlinks in `Pods/`.

## 💚 How did you test it?

All on macOS with CocoaPods 1.16.2 / Xcode 26.1.

**Checksum determinism** — reproduced CocoaPods' exact computation (`Digest::SHA1.hexdigest(spec.to_pretty_json)`) while evaluating the podspec under two different `$HOME` values:

* `origin/main`: `842e4309…` vs `8b743fb4…` — bug reproduced.
* This branch: `43ccbce8…` both times, and the two evaluated specs are byte-identical.
* `Pods/Local Podspecs/RNSentry.podspec.json` after a real install contains **zero** `/Users/…` strings (down from the leaking path).
* `SENTRY_XCFRAMEWORK_CACHE_DIR=/tmp/sentry-xcf-alt pod install` yields the **same** checksum as without the override, and re-points the stale symlink.

**Real** `pod install` in `samples/react-native/ios` and `packages/core/RNSentryCocoaTester` (different Podfile depths): `$(PODS_ROOT)` resolves correctly in both the per-pod xcconfig (`PODS_ROOT = ${SRCROOT}`) and the aggregate/user xcconfig (`PODS_ROOT = ${SRCROOT}/Pods`). Repeated installs keep the checksum stable and the symlink survives — `Installer::SandboxDirCleaner` only removes `.xcodeproj` children, target-support and header directories, so the staging dir is untouched.

**Builds**:

* `xcodebuild -scheme RNSentry -sdk iphonesimulator` → **BUILD SUCCEEDED** (`@import Sentry` resolves through the symlink).
* Full sample app `xcodebuild -workspace sentryreactnativesample.xcworkspace` → **BUILD SUCCEEDED**; `nm` on the resulting debug dylib shows 168 Sentry symbols. No `Sentry` CocoaPod exists in `Podfile.lock`, so linking necessarily went through the staged path. `-showBuildSettings` confirms `FRAMEWORK_SEARCH_PATHS` expands to `…/ios/Pods/sentry-xcframeworks/9.19.1/Sentry.xcframework/ios-arm64_x86_64-simulator`.

**Edge cases**: `SENTRY_USE_XCFRAMEWORK=0` stages nothing and keeps the `Sentry` pod dependency; stale/wrong symlinks and leftover version directories are recreated/pruned; `rm_rf` unlinks the symlink without following it (shared cache verified intact afterwards); evaluation without a Podfile falls back to the absolute path and creates no stray directories. `ruby -c` passes on both files.

## :pencil: Checklist

- [X] All tests passing.
- [X] No breaking changes.

Note on tests: the repo has no Ruby test infrastructure, so this is covered by the manual verification above plus the `pod install` jobs in `sample-application.yml` / `native-tests.yml`, which exercise the new staging path end-to-end.

## 🔮 Next steps

Add the `ready-to-merge` label to run the native/E2E/sample-build jobs before merging — those are the ones that exercise `pod install` across RN versions and the `use_frameworks!` variants.


CHANGELOG.md
```diff
@@ -15,10 +15,13 @@
 
   When a root span ends (idle nav spans from `reactNavigationIntegration` / `expoRouterIntegration`, or a user's own `Sentry.startSpan(...)`), the integration writes `turbo_module.<name>.<method>.{call_count,duration_ms,error_count}` attributes plus summary keys (`turbo_module.total_call_count`, `turbo_module.total_duration_ms`, `turbo_module.top_module`). Async calls above `slowCallThresholdMs` (default 500ms) additionally record a `native.turbo_module` breadcrumb. Both surfaces enabled by default; new knobs `enableSpanAttribution`, `slowCallThresholdMs`, `maxTopModulesPerSpan` on `turboModuleContextIntegration`.
 
-- Session Replay network details now capture `fetch` (Blob/ArrayBuffer) response bodies ([#6473](https://github.com/getsentry/sentry-react-native/pull/6473), [#6533](https://github.com/getsentry/sentry-react-native/pull/6533))
-
 ### Fixes
 
+- Make the `RNSentry` SPEC CHECKSUM in `Podfile.lock` machine-independent ([#6474](https://github.com/getsentry/sentry-react-native/pull/6474))
```
antonis: Q: Is this an internal only change or it has user facing area?

If not the later we can remove the changelog or move it in an internal section. Let's also move it to the unreleased section and revert the `- Session Replay network details` item move.

if it is user facing maybe we can use a `Changes` section

antonis (commented):
LGTM once the CI is green and the [changelog fixed](https://github.com/getsentry/sentry-react-native/pull/6534/changes#r3681811305) 🙇

CHANGELOG.md
alwx: Fixed.

User-facing, so I kept it in `Fixes`. It's not just build plumbing: without it a committed `Podfile.lock` flips its `RNSentry` checksum on every `pod install` across developers and CI, which is exactly what was reported in #6467. Users also need the heads-up about the one-time checksum change and that the `SENTRY_XCFRAMEWORK_CACHE_DIR` workaround can be dropped. Happy to move it to `Changes` if you'd rather group it there.

antonis: Thank you for the clarification 🙇

packages/core/scripts/sentry_utils.rb
```diff
@@ -165,3 +165,71 @@ def ensure_sentry_xcframework(version, product = 'Sentry')
...
+  end
+
+  "$(PODS_ROOT)/#{SENTRY_XCFRAMEWORK_STAGING_DIR}/#{version}/#{product}.xcframework"
+rescue StandardError, NotImplementedError => e
+  if defined?(Pod::UI)
+    Pod::UI.warn "[Sentry] Could not link the #{product} xcframework into Pods/ " \
+                 "(#{e.class}: #{e.message}). Falling back to the absolute cache path; " \
+                 "the RNSentry checksum in Podfile.lock will be machine-specific."
+  end
+  nil
```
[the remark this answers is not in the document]
antonis: @alwx The above looks valid 👍 Feel free to handle it in a follow up if the solution is complicated.

antonis (approved):

Issue #6497, opened 2026-07-22
iOS: screenshots no longer captured since 8.19.0 (Feedback Widget, attachScreenshot, captureScreenshot)


antonis:
## Summary

Starting in **8.19.0**, on **iOS** the SDK fails to capture screenshots. `NATIVE.captureScreenshot()` returns empty, which breaks every feature that relies on it:

- **Feedback Widget** with screenshot — tapping the screenshot button and submitting shows an **"Error capturing screenshot. Please try again."** alert and the report can't be sent.
- **`attachScreenshot: true`** — error/crash events are sent **without** the screenshot attachment (fails silently).
- **`Sentry.captureScreenshot()`** — returns empty/null.

## Affected versions

- **8.19.0** and later (regression introduced by #6380).

## Steps to reproduce

1. RN app with `Sentry.init({ attachScreenshot: true, integrations: [Sentry.feedbackIntegration({ enableTakeScreenshot: true })] })` on iOS.
2. Feedback: open the widget → "Take a screenshot" → submit → **"Error capturing screenshot"** alert appears.
3. `attachScreenshot`: capture any exception → the resulting event has **no** `screenshot.png` attachment.

Also see failed Feedback E2E test on main https://github.com/getsentry/sentry-react-native/actions/runs/29898734412/job/88856598124

Confirmed locally by running the RN sample:
- pre 8.19.0 with screenshot: https://sentry-sdks.sentry.io/issues/5005282269/events/eb0d408ff2434e1b83246a6efd90eda8/?project=5428561
- 8.19.0 no screenshot: https://sentry-sdks.sentry.io/issues/5005282269/events/85390da86b1749a1a985645ad89efe63/?project=5428561

PR #6491, opened 2026-07-21
fix(ios): Fix screenshot and view hierarchy capture failing on iOS

antonis:
## :loudspeaker: Type of change
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Since #6380 migrated iOS screenshot/view-hierarchy capture from `PrivateSentrySDKOnly` to the new `SentrySDK.internal` hybrid API, `NATIVE.captureScreenshot()` (and `fetchViewHierarchy()`) started returning empty on iOS.

This affects three features on iOS:
- **Feedback Widget screenshot** — empty result surfaces an "Error capturing screenshot" alert that overlays and blocks the submit button, so the report is never sent (this is what turned the `feedback` E2E flow red).
- **`attachScreenshot: true`** — screenshots silently dropped from error events.
- **`attachViewHierarchy: true`** — view hierarchy silently dropped from error events.

## :bulb: Motivation and Context
The iOS `feedback` E2E flow (`captureFlow-ios.yml`) has been failing since #6380 was merged

## :green_heart: How did you test it?
CI checks, Manual

## :pencil: Checklist
- [ ] I added tests to verify changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] All tests passing.
- [x] Public API changes reviewed by another Mobile SDK team member or implemented according to the [develop docs](https://develop.sentry.dev/) spec.
- [x] No breaking changes.

PR #6380, opened 2026-06-30
refactor(ios): Migrate from PrivateSentrySDKOnly to SentrySDK.internal

alwx:
## 📢 Type of change

- [X] Refactoring

## 📜 Description

Replace deprecated `PrivateSentrySDKOnly` SPI and `SentrySwizzle.h` macro usage in our iOS code with the new `SentrySDK.internal` Swift API exposed by sentry-cocoa 9.19.0, via a thin in-pod ObjC↔Swift bridge (`RNSentryInternal.swift`).

The bridge uses `@_spi(Private) import Sentry` to reach SPI-gated sub-APIs (`performance.currentScreenFrames`, `replay.configure`, `envelope.{store,capture,deserialize}`). Existing `.m`/`.mm` callers route through `[RNSentryInternal …]` via the auto-generated `RNSentry-Swift.h`. The `performSelector:@selector(getReplayIntegration)` workaround in `captureReplayWithReturnValue` is gone — `SentrySDK.internal.replay.capture()` returns `Bool` directly.

Notably, we do *not* use sentry-cocoa's new `SentryObjC` umbrella module (which the issue suggested) — that module isn't shipped via the `Sentry` CocoaPod in 9.19.0, only via SPM as a separate xcframework. The Swift bridge works equally on both consumption paths.

## 💡 Motivation and Context

Closes getsentry/sentry-react-native#6370. `PrivateSentrySDKOnly` is deprecated in sentry-cocoa 9.19.0 and slated for removal in the next major. Migrating now avoids future churn and drops the private header search path / `performSelector:` reliance.

## 💚 How did you test it?

* `yarn build`, `yarn test`, `yarn lint`, `yarn circularDepCheck` — all green
* `xcodebuild` for the `RNSentry` pod target (iphonesimulator) — builds clean
* `xcodebuild test` for `RNSentryCocoaTester` — 158/158 native Cocoa tests pass

## :pencil: Checklist

- [ ] I added tests to verify changes
- [X] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [X] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [X] All tests passing
- [X] No breaking changes

## 🔮 Next steps

PR #6541, opened 2026-07-30
fix(ios): Re-land SentrySDK.internal migration, bump Cocoa SDK to 9.24.0

### LINKED ISSUE #6507: Reapply iOS SentrySDK.internal migration (#6380) after sentry-cocoa fixes screenshotSource lazy-var poisoning
Follow-up to the revert in getsentry/sentry-react-native#6491 (which reverted getsentry/sentry-react-native#6380 to fix the iOS screenshot regression getsentry/sentry-react-native#6497).

## Background

getsentry/sentry-react-native#6380 migrated iOS code from `PrivateSentrySDKOnly` to the new `SentrySDK.internal` Swift API. It shipped in 8.19.0 and broke **all iOS screenshot capture** (Feedback Widget screenshot, `attachScreenshot`, `Sentry.captureScreenshot()`) — see getsentry/sentry-react-native#6497. getsentry/sentry-react-native#6491 reverts getsentry/sentry-react-native#6380 as an interim fix.

## Root cause (sentry-cocoa)

Any access to `SentrySDK.internal` constructs `SentryInternalApi`, whose `init` **eagerly reads** `SentryDependencyContainer.screenshotSource`. That property is a `lazy var` whose builder returns `nil` while `startOptions` is unset, and — being `lazy` — that `nil` is **cached permanently**. Because RN touches `SentrySDK.internal` before `SentrySDK.start` (from `RNSentryStart` and from JS integrations calling native methods like `fetchNativeAppStart`/`fetchNativeSdkInfo` during init), `screenshotSource` gets poisoned to `nil` and stays `nil` for the process lifetime.

A targeted RN-side fix is unreliable (any pre-start `SentrySDK.internal` access poisons it, and JS controls the timing), which is why we reverted the whole migration.

## What needs to happen

1. **Fix in sentry-cocoa** — `screenshotSource` (and the other `startOptions`-gated lazy providers, e.g. `viewHierarchyProvider`) must not permanently cache `nil`. Options:
   * Make `screenshotSource` a computed property instead of a `lazy var` so `getOptionalLazyVar` rebuilds it once `startOptions` is available; and/or
   * Have `SentryInternalApi` read the providers lazily at use-time rather than eagerly in `init`.
2. **Bump the sentry-cocoa dependency** in RN to the fixed version.
3. **Reapply** getsentry/sentry-react-native#6380 (revert the revert), re-verifying the `feedback` E2E flow stays green on iOS.

## Notes

* Reapplying getsentry/sentry-react-native#6380 also restores its two bundled improvements that the revert removed: the `fetchViewHierarchy` nil-guard (returns `null` instead of a zero-byte attachment on capture failure) and the `RNSentryTimeToDisplay` `extern` linkage change.
* A dedicated sentry-cocoa issue should track item (1).

Related: getsentry/sentry-react-native#6380, getsentry/sentry-react-native#6491, getsentry/sentry-react-native#6497

alwx:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Re-lands #6380 (the `PrivateSentrySDKOnly` → `SentrySDK.internal` migration), which shipped in 8.19.0 and was reverted in 8.20.0 by #6491, now that the underlying sentry-cocoa bug is fixed.

Three parts:

1. **Bump sentry-cocoa 9.19.1 → 9.24.0** (podspec + xcframework SHA256, via `scripts/update-cocoa.sh set-version 9.24.0`).
2. **Revert the revert** of #6380 — applied cleanly, no conflicts. The only commits touching `packages/core/ios` since the revert were the two release commits.
3. **New regression guard** — `RNSentryCocoaTester/RNSentryCocoaTesterTests/RNSentryScreenshotSourceTests.swift`.

Re-landing also restores the two improvements #6491 removed along with the migration: the `fetchViewHierarchy` nil-guard (returns `null` instead of a zero-byte attachment when capture fails) and the `RNSentryTimeToDisplay` `extern` linkage change.

> [!IMPORTANT]
> This re-introduces the RN < 0.75 requirement from 8.19.0: on React Native versions where `React-hermes` (or another RN pod) is not modularized by default (e.g. RN 0.71), users must add `use_modular_headers!` to their `ios/Podfile`. CocoaPods refuses to integrate a Swift pod against non-modular ObjC dependencies, and `RNSentryInternal.swift` makes `RNSentry` a Swift pod. Called out in the CHANGELOG.

Fixes #6507.

#6380 broke **all** iOS screenshot capture — Feedback Widget screenshot, `attachScreenshot`, and `Sentry.captureScreenshot()` (#6497).

Root cause was in sentry-cocoa, not in RN: constructing `SentryInternalApi` eagerly reads `SentryDependencyContainer.screenshotSource`, which was a `lazy var` whose builder returns `nil` while `startOptions` is unset. Being `lazy`, that `nil` was cached for the process lifetime. RN reads `SentrySDK.internal` before `SentrySDK.start` — from `RNSentryStart` itself, and from JS integrations calling native methods like `fetchNativeAppStart` / `fetchNativeSdkInfo` during init — so the source was reliably poisoned. A targeted RN-side fix wasn't possible because JS controls the timing of those calls, hence the full revert.

getsentry/sentry-cocoa#8578 makes `screenshotSource` a computed property so `getOptionalLazyVar` re-runs the builder until `startOptions` is available, and only then caches. It shipped in **9.24.0**.

I also confirmed:
- `SentrySDK.internal` is a computed property that builds a fresh `SentryInternalApi` per access, so the eager read in `SentryInternalScreenshotApi.init` is harmless once `screenshotSource` rebuilds.
- `viewHierarchyProvider` is still a `lazy var`, but its builder never touches `startOptions`, so it cannot be poisoned the same way.
- No API drift in `Sources/Swift/HybridSDK/` between 9.19.1 and 9.24.0 — the only diffs are `@_implementationOnly import` → `internal import`. Every symbol `RNSentryInternal.swift` bridges still exists with the same signature.
- Nothing in the 9.20.0–9.24.0 cocoa changelog affects RN. `enableReplayNetworkDetailsCapturing` was removed in 9.22.0 but RN never set it.

**New regression test** (`RNSentryScreenshotSourceTests`): reads `SentrySDK.internal` before `SentrySDK.start`, then starts via `RNSentryStart.start` with `attachScreenshot` and asserts `SentrySDK.internal.screenshot.capture()` is non-nil. Verified it is not vacuous:

| sentry-cocoa | Result |
|---|---|
| 9.19.1 (buggy) | ❌ `XCTAssertNotNil failed` — reproduces #6497 |
| 9.24.0 (fixed) | ✅ passes |

**Full native suite** — `RNSentryCocoaTester`, Release config, iPhone simulator, Xcode 26.1: **163 tests, 0 failures**.

**Linters**: `swiftlint --strict` clean (13 files, 0 violations), `clang-format` clean.

No JS/TS source changed, so no api-report or JS test impact.

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps

Needs the `ready-to-merge` label before merging so the native / E2E / sample-app workflows run — the iOS `feedback` E2E flow is what surfaced #6497 in the first place and should be confirmed green.

Two pre-existing problems found while writing the regression test, both unrelated to this change and left for follow-up issues:

1. **`RNSentryStartTests.swift` and `RNSentryStartFromFileTests.swift` are not in the Xcode test target.** They were dropped from `project.pbxproj` around the v7 merge (`6ab6b160`) and have not compiled or run since. They no longer build (`Options.enableTracing` was removed in cocoa v9), and once that's fixed every test in them crashes — see below. I intentionally did not re-enable them here; that's why the new regression test lives in its own file.
2. **`RNSentrySDK.start` proceeds with `options == nil`** when `sentry.options.json` is absent or unparseable (`packages/core/ios/RNSentrySDK.m` L64-72). The `@{}` fallback also fails validation ("Invalid DSN URL") and returns `nil`, but the code carries on into `updateWithReactDefaults:` / `configureOptions(...)` / `startWithOptions:`. Identical before and after #6380 — both `PrivateSentrySDKOnly.optionsWithDictionary:didFailWithError:` and `SentrySDK.internal.options(fromDictionary:)` funnel into the same `SentryOptionsInternal initWithDict:`, which returns `nil` on failure. From ObjC it silently no-ops; from Swift (`options.dsn = …` on an implicitly-unwrapped `SentryOptions!`) it segfaults.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,19 @@
 
 ## Unreleased
 
+### Changes
+
+- Migrate iOS internals from the deprecated `PrivateSentrySDKOnly` SPI to `SentrySDK.internal` ([#6541](https://github.com/getsentry/sentry-react-native/pull/6541))
+
+  Re-lands [#6380](https://github.com/getsentry/sentry-react-native/pull/6380), reverted in 8.20.0 by [#6491](https://github.com/getsentry/sentry-react-native/pull/6491) because it broke iOS screenshot capture. The underlying sentry-cocoa bug is fixed in 9.24.0.
+
+<!-- prettier-ignore-start -->
+> [!WARNING]
+> **Action required if you target React Native < 0.75.**
+> `RNSentry` now ships Swift code, which makes it a Swift pod under CocoaPods. If your `Podfile` doesn't already modularize React Native's ObjC pods (RN >= 0.75 does this by default), add `use_modular_headers!` to your `ios/Podfile` before running `pod install`. Otherwise `pod install` fails with an error like:
+> `The Swift pod 'RNSentry' depends upon 'React-hermes', which does not define modules.`
+<!-- prettier-ignore-end -->
+
 ### Fixes
 
 - Make the `RNSentry` SPEC CHECKSUM in `Podfile.lock` machine-independent ([#6534](https://github.com/getsentry/sentry-react-native/pull/6534))
@@ -35,14 +48,6 @@
 ### Fixes
 
 - `sentry-expo-upload-sourcemaps` now reads plugin config when the plugin is registered as `@sentry/react-native` ([#6543](https://github.com/getsentry/sentry-react-native/pull/6543))
-
-### Changes
-
-- Migrate iOS internals from the deprecated `PrivateSentrySDKOnly` SPI to `SentrySDK.internal` ([#6541](https://github.com/getsentry/sentry-react-native/pull/6541))
-
-  Re-lands [#6380](https://github.com/getsentry/sentry-react-native/pull/6380), which was reverted in 8.20.0 because it broke iOS screenshot capture ([#6497](https://github.com/getsentry/sentry-react-native/issues/6497)). The underlying sentry-cocoa bug is fixed in 9.24.0.
-
-  The `RNSentry` pod now contains Swift code. On React Native versions where RN pods are not modularized by default (e.g. RN 0.71), add `use_modular_headers!` to your `ios/Podfile`.
 
 ### Dependencies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,11 +36,22 @@
 
 - `sentry-expo-upload-sourcemaps` now reads plugin config when the plugin is registered as `@sentry/react-native` ([#6543](https://github.com/getsentry/sentry-react-native/pull/6543))
 
+### Changes
+
+- Migrate iOS internals from the deprecated `PrivateSentrySDKOnly` SPI to `SentrySDK.internal` ([#6541](https://github.com/getsentry/sentry-react-native/pull/6541))
+
+  Re-lands [#6380](https://github.com/getsentry/sentry-react-native/pull/6380), which was reverted in 8.20.0 because it broke iOS screenshot capture ([#6497](https://github.com/getsentry/sentry-react-native/issues/6497)). The underlying sentry-cocoa bug is fixed in 9.24.0.
+
+  The `RNSentry` pod now contains Swift code. On React Native versions where RN pods are not modularized by default (e.g. RN 0.71), add `use_modular_headers!` to your `ios/Podfile`.
+
 ### Dependencies
 
 - Bump Android SDK from v8.50.1 to v8.51.0 ([#6539](https://github.com/getsentry/sentry-react-native/pull/6539))
   - [changelog](https://github.com/getsentry/sentry-java/blob/main/CHANGELOG.md#8510)
   - [diff](https://github.com/getsentry/sentry-java/compare/8.50.1...8.51.0)
+- Bump Cocoa SDK from v9.19.1 to v9.24.0 ([#6541](https://github.com/getsentry/sentry-react-native/pull/6541))
+  - [changelog](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#9240)
+  - [diff](https://github.com/getsentry/sentry-cocoa/compare/9.19.1...9.24.0)
 
 ## 8.21.0
 
--- a/packages/core/ios/RNSentry+fetchNativeStack.m
+++ b/packages/core/ios/RNSentry+fetchNativeStack.m
@@ -2,7 +2,7 @@
 #import "RNSentryBreadcrumb.h"
 #import "RNSentryHexFormatter.h"
 #import "RNSentryId.h"
-#import <Sentry/PrivateSentrySDKOnly.h>
+#import "SentrySDKWrapper.h"
 @import Sentry;
 
 // This method was moved to a new category so we can use `@import Sentry` to use Sentry's Swift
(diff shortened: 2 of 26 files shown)


packages/core/ios/RNSentry+fetchNativeStack.m
```diff
@@ -2,7 +2,6 @@
 #import "RNSentryBreadcrumb.h"
 #import "RNSentryHexFormatter.h"
 #import "RNSentryId.h"
-#import <Sentry/PrivateSentrySDKOnly.h>
 @import Sentry;
```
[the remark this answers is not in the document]
alwx: Fixed

CHANGELOG.md
```diff
@@ -36,11 +36,22 @@
 
 - `sentry-expo-upload-sourcemaps` now reads plugin config when the plugin is registered as `@sentry/react-native` ([#6543](https://github.com/getsentry/sentry-react-native/pull/6543))
 
+### Changes
+
+- Migrate iOS internals from the deprecated `PrivateSentrySDKOnly` SPI to `SentrySDK.internal` ([#6541](https://github.com/getsentry/sentry-react-native/pull/6541))
+
+  Re-lands [#6380](https://github.com/getsentry/sentry-react-native/pull/6380), which was reverted in 8.20.0 because it broke iOS screenshot capture ([#6497](https://github.com/getsentry/sentry-react-native/issues/6497)). The underlying sentry-cocoa bug is fixed in 9.24.0.
+
+  The `RNSentry` pod now contains Swift code. On React Native versions where RN pods are not modularized by default (e.g. RN 0.71), add `use_modular_headers!` to your `ios/Podfile`.
```
lucas-zimerman: Since this has potential to break users code on old React Native, let's move it to the top of the unreleased header (line 11) and add the following snippet

```
> [!WARNING]
> ⚠️ Action required if you target React Native < 0.75: 
>  RNSentry now ships Swift code, making it a Swift pod under CocoaPods. If your Podfile doesn't already 
> modularize RN's ObjC pods (RN ≥ 0.75 does this by default), add use_modular_headers! 
> to your ios/Podfile before running pod install, otherwise pod install will fail with an error like The Swift > pod 'RNSentry' depends upon 'React-hermes', which does not define modules.

```

This way users using old versions will spot it right away and newer versions will just ignore it

lucas-zimerman: It would be nice to add a follow-up PR on Sentry Docs troubleshooting area, with the error users on old React Native will face with it and how to fix it.

lucas-zimerman (approved):
left some notes on the changelog file, overall the code looks good! lets do the final nits on the changelog.md and ship it!

lucas-zimerman (changes requested):
Lets fix this test before merging https://github.com/getsentry/sentry-react-native/actions/runs/30867551749/job/91862651186?pr=6541

lucas-zimerman (approved):
LGTM!

PR #6545, opened 2026-07-30
fix(core): Attach debug_meta to Hermes JS error events via Debug ID lookup

### LINKED ISSUE #6480: JS exceptions often lack debug_meta on Hermes; profiling path already has a robust Debug ID lookup
## Summary

On React Native + Hermes (including Expo OTA / `expo export` Hermes `.hbc` bundles), **plain JS exception events often ship without `debug_meta.images`**, even when:

1. Metro has injected the Debug ID premodule (`_sentryDebugIds` / `_sentryDebugIdIdentifier`)
2. The matching artifact bundle was uploaded with that Debug ID (`sentry-expo-upload-sourcemaps` / `@sentry/expo-upload-sourcemaps`, `Release: None`)

Profiling already solves this correctly via `getDebugMetadata()` in `dist/js/profiling/debugid.js`, which reads `GLOBAL_OBJ._sentryDebugIds` directly and stamps `code_file` as `app:///index.android.bundle` / `app:///main.jsbundle`. That helper is **not** wired into the exception event pipeline.

## Why it matters

Without `debug_meta`, backends fall back to release + filename matching. Expo/RN rewrites every JS frame to the same canonical filename (`app:///index.android.bundle` / `app:///main.jsbundle`) via `RewriteFrames`. OTA updates keep the **native** `release`, so symbolication can apply the **native build’s** Hermes map to **OTA bytecode** → wrong files/lines.

This is especially painful on Sentry-compatible self-hosts, but the missing `debug_meta` is an SDK-side gap regardless of backend.

Observed with `@sentry/react-native` **8.19.0**, Expo Router, Hermes, OTA via artifact bundles.

## Root cause (SDK)

Exception events rely on `@sentry/core` `prepareEvent` → `applyDebugIds` → `getFilenameToDebugIdMap`:

1. Take the single key in `_sentryDebugIds` (the `Error().stack` string captured when the Debug ID **premodule** ran)
2. Re-parse that stack with the generic stack parser
3. Require an **exact** `filename` string match against exception frames

That path is fragile for Hermes top-level/premodule stacks. If the reparse yields no usable filename (or a different shape than real exception frames), `filenameDebugIdMap` is empty → `applyDebugMeta` no-ops → **no `debug_meta` on the event**.

By contrast, profiling:

```ts
// profiling/debugid.ts (conceptually)
const debugId = Object.values(GLOBAL_OBJ._sentryDebugIds)[0]
return [{ type: 'sourcemap', code_file: DEFAULT_BUNDLE_NAME, debug_id: debugId }]
```

No stack reparse. One bundle, one Debug ID, fixed `code_file`.

## Suggested fix

Reuse the profiling approach for **all** events:

- Extract / share `getDebugMetadata()` outside profiling
- Add a default integration (or extend an existing one) that stamps `event.debug_meta.images` from `_sentryDebugIds` (+ `_sentryDebugIdIdentifier` fallback) using `ANDROID_DEFAULT_BUNDLE_NAME` / `IOS_DEFAULT_BUNDLE_NAME`
- Keep it idempotent if core `applyDebugIds` already succeeded

This matches how RN actually works (single Hermes bundle per platform) and makes Expo OTA Debug ID matching reliable without release/dist workarounds.

## Reproduction sketch

1. Expo app with `getSentryExpoConfig` + `@sentry/react-native/expo` plugin
2. Publish an OTA / `expo export` Hermes bundle; upload with `npx @sentry/expo-upload-sourcemaps dist` (artifact bundle, Debug ID in upload report)
3. On a device running that OTA (`Updates.isEmbeddedLaunch === false`), `Sentry.captureException(new Error('probe'))`
4. Inspect event JSON: `debug_meta` often missing; `release` still native (`*.androidapp@x.y.z+build`); frames use `app:///index.android.bundle`

## Workaround we use today

App-level `Sentry.addEventProcessor` that mirrors profiling’s `getDebugMetadata()` and attaches `debug_meta.images` before send. Happy to upstream that if maintainers agree on the approach.

## Related

- Expo docs assume Debug ID path “just works” after `sentry-expo-upload-sourcemaps` ([Using Sentry](https://docs.expo.dev/guides/using-sentry/))
- Metro Debug ID injection is fine; upload Debug IDs match; the break is event emission

antonis:
## 📢 Type of change

- [X] Bugfix

## 📜 Description

On React Native + Hermes (including Expo OTA / `expo export` Hermes bundles), plain JS exception events often ship **without** `debug_meta.images`, even when Metro injected the Debug ID premodule (`_sentryDebugIds`) and the matching artifact bundle was uploaded with that Debug ID.

The core `prepareEvent` pipeline attempts this via `applyDebugIds` → `applyDebugMeta`, but that path re-parses the premodule `Error().stack` and requires an **exact filename match** against the exception frames. That match is fragile for Hermes premodule stacks, so `filenameDebugIdMap` ends up empty and no `debug_meta` is attached.

Profiling already solves the same problem robustly via `getDebugMetadata()` (`profiling/debugid.ts`): it reads `_sentryDebugIds` directly and stamps a single `sourcemap` image with the canonical bundle `code_file` (`app:///index.android.bundle` / `app:///main.jsbundle`) — no stack reparse. That helper was not wired into the error event pipeline.

This PR adds a default `debugMetaIntegration` that reuses `getDebugMetadata()` to stamp `debug_meta.images` on error events. Key details:

* **Scoped to error events** (`event.type === undefined`) — transactions, profiles and other event types are untouched.
* **Idempotent against core.** Core runs `applyDebugIds` (sets `frame.debug_id`) *before* event processors and `applyDebugMeta` (stamps images) *after* them. The integration detects a successful core match via `frame.debug_id` and backs off, so core's own stamp is not duplicated. When core fails to match (the bug case), the integration fills the gap.
* **Defensive dedupe** by `debug_id`, and preserves unrelated pre-existing images (e.g. native images added by `nativeLinkedErrors`).
* Registered right after `createReactNativeRewriteFrames()` so the canonical bundle filename is already in place.

## 💡 Motivation and Context

Without `debug_meta`, backends fall back to release + filename matching. RN/Expo rewrites every JS frame to the same canonical filename, and OTA updates keep the **native** release, so the native build's Hermes map can be applied to OTA bytecode → wrong files/lines. The missing `debug_meta` is an SDK-side gap regardless of backend.

Fixes getsentry/sentry-react-native#6480

## 💚 How did you test it?

* Added `test/integrations/debugmeta.test.ts` covering: stamping on error events, no-op when no Debug ID, skipping non-error events, backing off when core matched, and preserving pre-existing images.
* Added tests that reproduce the core `prepareEvent` ordering (`applyDebugIds` → event processor → `applyDebugMeta`) and assert exactly one image results whether core succeeds or fails — guarding against a duplicate-image regression.
* Regenerated the API report for the new public `debugMetaIntegration`.

## :pencil: Checklist

- [X] All tests passing.
- [X] No breaking changes.

## 🔮 Next steps

* Confirm naming (`debugMetaIntegration`) — the mechanism is RN/Hermes-specific (single bundle per platform); a more specific name could be used if preferred.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


antonis:
@sentry review

antonis (commented):
Also fixed the duplicate fixes section with https://github.com/getsentry/sentry-react-native/pull/6545/changes/9b3f3a6a1c96562a40a9a2f359389ced4a30e335

packages/core/src/js/integrations/debugmeta.ts
```diff
@@ -0,0 +1,81 @@
...
+  if (hasFrameWithDebugId(event)) {
+    return event;
+  }
+
+  const images = getDebugMetadata();
+  if (!images.length) {
+    return event;
+  }
+
+  event.debug_meta = event.debug_meta || {};
```
lucas-zimerman, line 57:
```suggestion
  event.debug_meta ??=  {};
```

lucas-zimerman (approved):
LGTM!

Issue #4870, opened 2025-05-28
Build fails on RN 0.76.7 + Sentry 6.14.0 (New Architecture) — Unsupported type: StringLiteralUnionTypeAnnotation


mohib-uddin:
### What React Native libraries do you use?

React Native without Frameworks, RN New Architecture, Hermes

### Are you using sentry.io or on-premise?

sentry.io (SaS)

### @sentry/react-native SDK Version

6.14.0

### How does your development environment look like?

I’m encountering a fatal build error when trying to run my React Native app using React Native v0.76.7, Sentry React Native v6.14.0, and New Architecture enabled.

Environment:

* React Native version: 0.76.7

* Sentry React Native version: 6.14.0

* Node.js version: v22.11.0

* Gradle: 8.10.2

* Android Gradle Plugin: 8.6.0

* New Architecture: ✅ Enabled

* OS: Windows

### Sentry.init()

````js
Sentry.init({
  dsn: 'https://...@sentry.io/...'
  // other options
});
````

### Steps to Reproduce

i) Install Sentry React Native v6.14.0

ii) Enable New Architecture in react-native.config.js

iii) Run: npm run android (with react-native run-android)

iv) Observe crash during codegen step

### Expected Result

The project builds and runs without errors using the new architecture.

### Actual Result

C:\Users\nonr\Desktop\Projects\Freelance\consumer\app\node_modules\react-native\node_modules\@react-native\codegen\lib\generators\modules\GenerateModuleH.js:256
      throw new Error(createErrorMessage(realTypeAnnotation.type));
      ^

Error: Unsupported type for param "type". Found: StringLiteralUnionTypeAnnotation
    at translatePrimitiveJSTypeToCpp (C:\Users\nonr\Desktop\Projects\Freelance\consumer\app\node_modules\react-native\node_modules\@react-native\codegen\lib\generators\modules\GenerateModuleH.js:256:13)
    at getCppType (C:\Users\nonr\Desktop\Projects\Freelance\consumer\app\node_modules\react-native\node_modules\@react-native\codegen\lib\generators\modules\GenerateModuleH.js:261:5)
    at C:\Users\nonr\Desktop\Projects\Freelance\consumer\app\node_modules\react-native\node_modules\@react-native\codegen\lib\generators\modules\GenerateModuleH.js:309:28     
    at Array.map (<anonymous>)
[...]

PR #6550, opened 2026-07-31
Convert union type to string to fix codegen error

naftalibeder:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Running `assemble` on Android errors with the message:
> Unsupported type for param "type". Found: StringLiteralUnionTypeAnnotation

Union types are no longer supported in React Native Codegen. I'm not able to find documentation confirming this, but it appears to be an issue others have encountered.

Related:
https://github.com/react/react-native/issues/49566
https://github.com/getsentry/sentry-react-native/issues/4870
https://github.com/react-native-webview/react-native-webview/issues/3767

### DIFF
--- a/packages/core/src/js/NativeRNSentry.ts
+++ b/packages/core/src/js/NativeRNSentry.ts
@@ -108,7 +108,8 @@ export type NativeStackFrames = {
 };
 
 export type NativeAppStartResponse = {
-  type: 'cold' | 'warm' | 'unknown';
+  /** 'cold' | 'warm' | 'unknown' */
+  type: string;
   has_fetched: boolean;
   app_start_timestamp_ms?: number;
   spans: {


alwx:
Hey @naftalibeder and thanks for the PR!

Before we take this, I'd like to pin down the root cause, because I can't reproduce the error from the spec file alone.
I ran `@react-native/codegen`'s TypeScript parser over the unmodified `NativeRNSentry.ts`, and was unable to reproduce the issue.

Could you share the following with me?
1) your `react-native` version
2) the output of yarn why @react-native/codegen (or the npm equivalent),
3) whether the error survives a clean build — `./gradlew clean` plus deleting the generated codegen output and `schema.json` under `android/build` / `node_modules/@sentry/react-native/android/build`

Thanks a lot!

alwx:
Closing due to inactivity.

PR #6561, opened 2026-08-04
fix(core): Measure callback-style native module calls until completion

### LINKED ISSUE #6542: TurboModule instrumentation: callback-style bridge methods are recorded as ~0ms sync calls
`wrapTurboModule` closes a call record as soon as the wrapped method returns, and only defers when the return value is thenable. Bridge methods that complete via success/failure callbacks — the common async shape on the Old Architecture, and still used by some TurboModules — return undefined, so they're recorded as sync with near-zero duration. Consequences: durations in turbo_module.* span attributes and the periodic aggregate are wrong for those methods, and native.turbo_module slow-call breadcrumbs never fire for them.

Possible fix: detect trailing function arguments, wrap them, and close the record when the callback fires (with a cap/timeout so a never-invoked callback can't pin the record). Needs care around methods that take non-completion function args.

Found by Cursor Bugbot on #6504.

alwx:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
`wrapTurboModule` only recognised two completion signals: a plain return and a thenable return. Bridge methods that report completion through success/failure callbacks return `undefined`, so they were closed on return and recorded as sync calls with a near-zero duration — wrong `turbo_module.*` durations, and `native.turbo_module` slow-call breadcrumbs that could never fire.

Trailing completion callbacks are now wrapped, so the record closes when the callback fires. Details:

- **Argument convention.** RN's [`genMethod`](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/BatchedBridge/NativeModules.js) fixes the shape: last argument is the success callback, second-to-last the failure one, and a non-function argument may never follow a function. `'promise'`-typed methods never receive callbacks and are skipped — the thenable path already covers them.
- **Error attribution only where it's guaranteed.** Failure callbacks count as errors on the Old Architecture bridge. New Architecture TurboModules take arbitrary callbacks with no such convention, so guessing there would corrupt `errorCount`; those calls close without an error flag.
- **`kind`** stays `'async'` — no change to `TurboModuleCallKind`, the aggregate keys, or the API report.
- **Crash attribution is unchanged.** The frame is still popped synchronously. Holding it until a callback that may never fire would risk blaming the module for an unrelated later native crash — a regression on the feature the frame exists for. Only the timing record is deferred.
- **Bounded without timers.** A cap on in-flight callback calls plus an amortised age sweep on insert. A call closed out that way is still counted with a zero duration; dropping it would hide the method from the aggregate entirely, which is worse than the ~0ms this fixes. The same clamp covers a callback firing implausibly late — on the New Architecture that is more likely a long-lived subscription handler than a completion callback.
- A method that settles its callback inline stays `kind: 'sync'`, so RN's `'sync'` method type is not mislabelled.

The callback machinery lives in a new `turboModuleCallbacks.ts` because `wrapTurboModule.ts` would otherwise exceed oxlint's `max-lines`. No public API change — `api-extractor` reports the surface unchanged.

Fire-and-forget `type: 'async'` methods that take **no** callback still have no completion signal at all and remain ~0ms. Not addressed here.

Fixes #6542, found by Cursor Bugbot on #6504. Callbacks are the dominant async shape on the Old Architecture bridge and are still used by some TurboModules, so this covered a large share of the calls the instrumentation reported on.

18 new tests — 16 unit (`test/turbomodule/wrapTurboModule.test.ts`) and 2 end-to-end through the integration (`test/integrations/turboModuleContext.test.ts`), covering: duration measured to the callback, inline settle staying `sync`, legacy failure-callback error attribution, no guessing on New Architecture, repeated callback invocations recorded once, receiver/argument/return pass-through, throwing callbacks, cap eviction, age sweep, the late-callback clamp, `'promise'`-typed methods untouched, non-trailing function arguments ignored, slow-call breadcrumbs, and span credit going to the span open at call start.

Full local run: `jest` 1859 passed (+334 tools), oxlint 0/0, oxfmt clean, `tsc -p tsconfig.build.json` clean, `api-extractor` up to date, `madge` no cycles.

Not verified on a real device or simulator — the RN bridge method shape is modelled in tests, not executed.

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps
- `logger` imported from `@sentry/core` in `turbomodule/` is the **Sentry Logs API**, not the debug logger (that's `debug`). The 6 pre-existing `logger.warn` calls in `wrapTurboModule.ts` and 3 in `wrapNativeModules.ts` therefore capture log events from the wrap hot path — with a recursion flavour, since capturing a log goes through the wrapped `RNSentry.captureEnvelope`. New code here uses `debug`; the pre-existing calls are left for a separate fix so this diff stays scoped.
- Never-settling promises leak a tracker frame unbounded today. The registry added here is the natural place to bound that.
- `## Unreleased` in `CHANGELOG.md` has two `### Fixes` headings (pre-existing on `main`).
- Docs, `sentry-conventions` attributes and the public-API lock are tracked in #6168.


lucas-zimerman:
All good! LGTM after dealing with the AI review

packages/core/src/js/turbomodule/wrapTurboModule.ts
```diff
@@ -98,33 +109,60 @@ export function wrapTurboModule<T extends object>(
         logger.warn(`[TurboModuleTracker] notifyStart failed for ${name}.${key}: ${String(e)}`);
```
lucas-zimerman, line 109: Hey, I know this is unrelated to this PR, but these logs on line 104 and 109 are being sent as user logs to `sentry.io` . Shouldn't we send it as debug warn?

lucas-zimerman (approved):
LGTM!

It is unrelated to this PR, we should address https://github.com/getsentry/sentry-react-native/pull/6561/changes#r3753179915 on a follow up PR.

alwx:
@lucas-zimerman here: https://github.com/getsentry/sentry-react-native/pull/6585

Issue #5763, opened 2026-03-05
Expose Android screenshot masking options


antonis:
## Summary

Android SDK 8.34.0 ([changelog](https://github.com/getsentry/sentry-java/blob/main/CHANGELOG.md#8340)) added screenshot masking support using the view hierarchy ([getsentry/sentry-java#5077](https://github.com/getsentry/sentry-java/pull/5077)). This masks sensitive content (text, images) in error screenshots.

The React Native SDK already exposes `attachScreenshot` but doesn't provide masking options, meaning screenshots on Android could contain sensitive data.

## New Android APIs to bridge

```kotlin
options.screenshot.setMaskAllText(true)
options.screenshot.setMaskAllImages(true)
options.screenshot.addMaskViewClass("com.example.MyCustomView")
```

Also configurable via `AndroidManifest.xml`:
```xml
<meta-data android:name="io.sentry.screenshot.mask-all-text" android:value="true" />
<meta-data android:name="io.sentry.screenshot.mask-all-images" android:value="true" />
```

## Proposal

Expose screenshot masking options in the RN SDK options (e.g. `screenshotMaskAllText`, `screenshotMaskAllImages`) and forward them to the native Android SDK during init. Consider parity with iOS if Cocoa SDK adds similar support.

Note: requires `sentry-android-replay` module at runtime for masking to work.

## References

- Android SDK 8.34.0 bump: #5760

PR #6565, opened 2026-08-05
fix(android): apply screenshot masking in captureScreenshot()

OmerToledo:
## 📜 Description

On Android, `RNSentryModuleImpl.captureScreenshot()` (hybrid SDK / `NATIVE.captureScreenshot()`) captured a raw window bitmap via `ScreenshotUtils.takeScreenshot()` with **no masking**, while iOS routes the same API through `PrivateSentrySDKOnly.captureScreenshots()` → `SentryViewPhotographer`, which redacts text and images by default.

Screenshot masking options bridged in [#6007](<https://github.com/getsentry/sentry-react-native/issues/6007>) (`screenshot.maskAllText`, `maskAllImages`, view classes) are applied today only on the `attachScreenshot` **error path** via `ScreenshotEventProcessor`. They are **not** used by the hybrid capture path.

That creates a platform asymmetry for any feature built on `NATIVE.captureScreenshot()`:

* Feedback Widget “Take screenshot”
* Custom feedback / bug-report flows (our production case in Bob Mobile)
* Any direct `Sentry.captureScreenshot()` usage

**iOS:** personal text redacted in the capture primitive  <br>**Android:** cleartext text in the PNG until this fix

This PR mirrors the masking pipeline already used in `ScreenshotEventProcessor`:

1. Capture bitmap (`ScreenshotUtils.captureScreenshot`)
2. Build view hierarchy + `MaskRenderer.renderMasks` when `options.getScreenshot().getMaskViewClasses()` is non-empty
3. Compress to PNG

When masking is configured but fails, the method returns `null` (fail closed) rather than returning an unmasked capture.

Requires `sentry-android-replay` at runtime — already pulled in by `sentry-android`, same as error-screenshot masking.

## 💡 Motivation and Context

We hit this in production while building a custom in-app feedback modal that attaches a screenshot. iOS behaved as documented (“data will be blurred”); Android did not, despite setting `screenshot: { maskAllText: true, maskAllImages: true }` in `Sentry.init()`.

Related:

* getsentry/sentry-react-native#5763 / [#6007](<https://github.com/getsentry/sentry-react-native/issues/6007>) — exposed masking **options** for error screenshots (init bridge only)
* [getsentry/sentry-java#5077](<https://github.com/getsentry/sentry-java/issues/5077>) — Android masking implementation in `ScreenshotEventProcessor`

This closes the gap for the **hybrid capture API**, not just init options.

## 💚 How did you test it?

* Verified the change compiles against current `main` (same imports/pattern as `ScreenshotEventProcessor`)
* Exercised manually in Bob Mobile (custom feedback modal): Android screenshot attachments now redact on-screen text when masking options are enabled

*No new automated test in this PR yet* — happy to add an Android instrumentation/unit test if maintainers point to the preferred harness.

## :pencil: Checklist

- [X] I added a changelog entry
- [ ] I added tests to verify the changes (seeking guidance on preferred Android test location)
- [X] No new PII added / no `sendDefaultPii` changes

Made with [Cursor](<https://cursor.com>)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,9 +27,9 @@
 
 ### Fixes
 
-- Apply screenshot masking in Android `captureScreenshot()` when `screenshot` options are configured
+- Apply screenshot masking in Android `captureScreenshot()` when `screenshot` options are configured ([#6565](https://github.com/getsentry/sentry-react-native/pull/6565))
 
-  Hybrid SDK screenshot capture (`NATIVE.captureScreenshot()`, used by the Feedback Widget and custom integrations) returned unmasked window captures on Android while iOS redacts text and images via `SentryViewPhotographer`. Error-screenshot masking (`attachScreenshot` + `ScreenshotEventProcessor`) was unaffected. Android `captureScreenshot()` now reuses the same view-hierarchy masking pipeline when `options.getScreenshot()` has mask classes configured.
+  Hybrid SDK screenshot capture (`NATIVE.captureScreenshot()`, used by the Feedback Widget and custom integrations) returned unmasked window captures on Android while iOS redacts text and images via `SentryViewPhotographer`. Error-screenshot masking (`attachScreenshot` + `ScreenshotEventProcessor`) was unaffected. Android `captureScreenshot()` now reuses the same view-hierarchy masking pipeline for configured `screenshot` options such as `maskAllText` and `maskAllImages`.
 
 - Attach `debug_meta` to JS error events on Hermes when the Debug ID stack match fails ([#6545](https://github.com/getsentry/sentry-react-native/pull/6545))
 - `sentry-expo-upload-sourcemaps` now reads plugin config when the plugin is registered as `@sentry/react-native` ([#6543](https://github.com/getsentry/sentry-react-native/pull/6543))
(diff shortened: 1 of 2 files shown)


OmerToledo:
Pushed a follow-up commit (`bfb341e4`) addressing review feedback: `screenshotMaskingOptions()` now always returns configured `SentryScreenshotOptions` instead of gating on `getMaskViewClasses().isEmpty()`, so `maskAllText` / `maskAllImages` from `Sentry.init()` reliably enable the masking pipeline.

alwx:
Thanks for looking into this. However, I think the problem is wider than the feedback widget since `src/js/integrations/screenshot.ts` uses `NATIVE.captureScreenshot()` too.

Two things block it for me: 
1) please restore the `getMaskViewClasses().isEmpty()` gate. Without the gate every capture now runs masking and returns null on any failure. 
2) there's no replay-availability check: excluding `sentry-android-replay` is supported (see `packages/core/android/replay-stubs/README.md`), and those apps will hit `NoClassDefFoundError` and end up with no screenshot at all instead of an unmasked one. 

More broadly, I think this fix belongs in `sentry-java` — `ScreenshotUtils.takeScreenshot` is marked "Used by Hybrid SDKs" and other SDKs have the same bug, so a `takeMaskedScreenshot` upstream would fix it everywhere and reduce this diff to one line.

alwx:
Closing due to inactivity. Feel free to reopen if you think it still makes sense.

PR #6567, opened 2026-08-06
ci: Replace curl-pipe-bash with action-setup-cli for Sentry CLI setup

oioki:
## Summary
- Two "Install Sentry CLI" steps in `size-analysis.yml` (Android and iOS jobs) used `curl -sL https://sentry.io/get-cli/ | sh`, piping a remote install script directly into a shell.
- Replaced both with [`getsentry/action-setup-cli`](https://github.com/getsentry/action-setup-cli), which downloads the `sentry-cli` release asset for the runner's OS/arch and verifies its sha256 against the digest GitHub recorded for that asset via the Releases API, before adding it to `PATH`. Uses `gh` only, no `curl`.
- Preserved the existing `if: env.SENTRY_AUTH_TOKEN != ''` condition on both steps.

## Test plan
- [x] Validated `size-analysis.yml` parses correctly
- [ ] Verify both Android and iOS size analysis jobs still install sentry-cli correctly on this PR

Fixes VULN-2348

### DIFF
--- a/.github/workflows/size-analysis.yml
+++ b/.github/workflows/size-analysis.yml
@@ -78,7 +78,7 @@ jobs:
 
       - name: Install Sentry CLI
         if: env.SENTRY_AUTH_TOKEN != ''
-        run: curl -sL https://sentry.io/get-cli/ | sh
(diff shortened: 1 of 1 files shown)


lucas-zimerman (approved):
LGTM! After ci test pass

PR #6577, opened 2026-08-10
chore(deps): bump brace-expansion to ^2.1.4 / ^5.0.9

alwx:
## :loudspeaker: Type of change
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Retargets the existing scoped resolutions so the 2.x and 5.x copies of `brace-expansion` resolve to `2.1.4` and `5.0.9`. The 1.x copy is already on the patched `1.1.18`. Dev-only, transitive via `minimatch`; no direct dependency and no runtime code touched.

## :bulb: Motivation and Context
The `maxLength` cap added in `2.1.2` / `5.0.8` for CVE-2026-14257 only bounded the accumulator in `combine()`. The intermediate `values` array and `expandSequence()` that feed it stayed unbounded, so the mitigation is bypassable — a ~25 KB input still triggers an uncatchable OOM (CVE-2026-69152, GHSA-rgw5-rvv9-x895).

Covers Dependabot alerts 610, 613 and 614:
https://github.com/getsentry/sentry-react-native/security/dependabot/613

## :green_heart: How did you test it?
- Verified against the published tarballs rather than the advisory text: `brace-expansion@2.1.4` and `@5.0.9` both carry the cumulative `valuesLength` bound in `expand_` and pass `maxLength` into `expandSequence`.
- Re-ran `yarn install --mode=update-lockfile` on the committed state; the lockfile is byte-identical, so the `--immutable` check holds.
- Did not run the suite locally — leaving that to CI, since this is a lockfile-only bump of a dev-tooling transitive dep.

## :pencil: Checklist
- [ ] I added tests to verify changes.
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled.
- [ ] I updated the docs if needed.
- [ ] I updated the wizard if needed.
- [ ] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps
Dependabot alerts 620 / 621 (`js-yaml` < 4.3.1 and < 3.15.1) are still open and need the same treatment on the `js-yaml` resolutions.

PR #6578, opened 2026-08-10
chore(deps): bump js-yaml to ^3.15.1 / ^4.3.1

alwx:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Bumps the two `js-yaml` copies in the lockfile to the patched releases: the global resolution moves the 4.x consumers to `4.3.1`, and the three scoped entries keep the 3.x consumers on `3.15.1`. Dev-only transitive dependency; no runtime code touched.

Both lines are bumped in one commit because they are the same advisory (GHSA-5p4m-2wfm-xmqj) in the same lockfile — splitting them would leave one alert open for no benefit.

`resolveYamlOmap()` deduplicated `!!omap` keys with a linear `indexOf` scan inside the per-element loop, making resolution O(n²) in the number of entries. `!!omap` is part of the default schema, so a plain `yaml.load()` is affected with no custom configuration. 5.x fixed this in 5.2.1 (CVE-2026-59870); the backport to the 3.x and 4.x lines only landed now.

Covers Dependabot alerts 620 and 621:
https://github.com/getsentry/sentry-react-native/security/dependabot/621

Verified against the published tarballs rather than the advisory text:

- `3.15.0` / `4.3.0` still contain `objectKeys.indexOf(pairKey)`; `3.15.1` / `4.3.1` replace it with an `Object.defineProperty` hash lookup.
- Ran the advisory PoC against all four versions. Vulnerable versions scale quadratically, patched ones are flat:

  | entries | 3.15.0 | 3.15.1 | 4.3.0 | 4.3.1 |
  |---|---|---|---|---|
  | 10 000 | 50 ms | 31 ms | 52 ms | 32 ms |
  | 20 000 | 128 ms | 30 ms | 134 ms | 32 ms |
  | 40 000 | 463 ms | 41 ms | 465 ms | 47 ms |

- Dependency sets are unchanged across both bumps (`argparse` + `esprima` for 3.x, `argparse` for 4.x), so no new transitive deps enter the tree.
- Re-ran `yarn install --mode=update-lockfile` on the committed state; the lockfile is byte-identical, so the `--immutable` check holds.
- Did not run the suite locally — leaving that to CI, since this is a lockfile-only bump of a dev-tooling transitive dep.

## :pencil: Checklist
- [ ] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps
Companion to #6577 (`brace-expansion`). Together they clear all five open Dependabot alerts on the root lockfile.


lucas-zimerman (approved):
LGTM!

PR #6581, opened 2026-08-10
fix(expo): Resolve config-plugins through the expo package

### LINKED ISSUE #6564: withSentryAndroidGradlePlugin.js imports @expo/config-plugins instead of the expo/config-plugins sub-export

Expo Router

sentry.io (SaS)

No

_No response_

7.11.0 (also confirmed in 8.21.0)

````

expo-env-info 2.1.0 environment info:
    System:
      OS: Windows 11 10.0.26200
    Binaries:
      Node: 20.19.6
      npm: 11.11.0
    IDEs:
      Android Studio: AI-253.29346.138.2531.14876573
    npmPackages:
      expo: ^57.0.10 => 57.0.10
      expo-router: ~57.0.10 => 57.0.10
      expo-updates: ~57.0.12 => 57.0.12
      react: 19.2.3 => 19.2.3
      react-dom: 19.2.3 => 19.2.3
      react-native: 0.86.2 => 0.86.2
      react-native-web: ^0.21.0 => 0.21.2
    Expo Workflow: managed (CNG — no ios/ or android/ committed)

Note: the bug was reproduced on Expo SDK 56 (react-native 0.85.3).
On SDK 57 the incorrect import resolves again by accident, because
@expo/config-plugins is hoisted to the project root.
Builds run on EAS Build (Linux for Android, macOS for iOS).

````

### Sentry.init()

````js
Sentry.init({
  dsn: "https://<redacted>@o<org>.ingest.us.sentry.io/<project>",

  // Keep dev sessions out of the production Sentry project.
  enabled: !__DEV__,

  sendDefaultPii: __DEV__,
  enableLogs: true,
});
````

### Steps to Reproduce

1. Create an Expo SDK 56 project using CNG (no committed ios/ or android/ folders).

2. Install `@sentry/react-native@7.11.0` and add the plugin to app.config.js:

   plugins: [["@sentry/react-native/expo", { organization: "...", project: "..." }]]

3. Confirm `@expo/config-plugins` is NOT hoisted to the project root:

   npm ls @expo/config-plugins

   Under SDK 56 it resolves only to nested copies
   (expo/node_modules/, expo-sharing/node_modules/, expo-splash-screen/node_modules/).

4. Run any command that evaluates config plugins:

   npx expo config --json --full

---

DETERMINISTIC CHECK (independent of hoisting)

The four plugin files disagree about which module to import:

  grep -r "config-plugins" node_modules/@sentry/react-native/plugin/build/*.js

  withSentry.js:                    require("expo/config-plugins")
  withSentryAndroid.js:             require("expo/config-plugins")
  withSentryIOS.js:                 require("expo/config-plugins")
  withSentryAndroidGradlePlugin.js: require("@expo/config-plugins")   <-- inconsistent

`@expo/config-plugins` is not declared in @sentry/react-native's package.json as a
dependency, peer dependency, or optional dependency, so it resolves only when another
package happens to hoist it to the project root.

Confirmed identical in 7.11.0 and 8.21.0.

### Expected Result

`npx expo config --json --full` evaluates the config plugins successfully.

withSentryAndroidGradlePlugin.js should import the `expo/config-plugins` sub-export,
matching its three sibling files, so it resolves through the `expo` package rather than
depending on `@expo/config-plugins` being hoisted to the project root.

### Actual Result

npx expo config --json --full

Error: Cannot find module '@expo/config-plugins'
Require stack:
- node_modules/@sentry/react-native/plugin/build/withSentryAndroidGradlePlugin.js
- node_modules/@sentry/react-native/plugin/build/withSentry.js
- node_modules/@sentry/react-native/plugin/build/index.js
- node_modules/@sentry/react-native/expo.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1421:15)
    at Module._load (node:internal/modules/cjs/loader:1227:37)
    at Module.require (node:internal/modules/cjs/loader:1504:12)
    at Object.<anonymous> (withSentryAndroidGradlePlugin.js:4:26)

This breaks `expo config`, `expo-doctor`, and `expo prebuild`, so EAS builds fail
before any native compilation starts.

Notes:
- `withSentry.js` requires this module eagerly at load time, so it fails even when
  `experimental_android.enableAndroidGradlePlugin` is not enabled.
- This occurs before Sentry.init() runs — it is purely a build-time module resolution
  failure, unrelated to SDK configuration.
- Workaround: declare `@expo/config-plugins` directly in the app's package.json.
  `expo-doctor` then warns that the package "should not be installed directly",
  so it trades one problem for another.
- On SDK 57 the bug is masked again, because `@expo/config-plugins` is hoisted to the
  root by expo, expo-sharing and expo-splash-screen. The incorrect import is unchanged.

alwx:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

`withSentryAndroidGradlePlugin` was the only one of the four Expo config plugin files importing the bare `@expo/config-plugins`; the other three use the `expo/config-plugins` sub-export. It now uses the sub-export too.

`expo/config-plugins.js` is a plain re-export (`module.exports = require('@expo/config-plugins')`), so this is a module-resolution fix only, with no behavioural change.

The test now mocks the same specifier the source imports. Previously it mocked `@expo/config-plugins`, which passes under either import because of the re-export chain, so it could not catch a regression.

Fixes #6564.

`@expo/config-plugins` is not declared in `@sentry/react-native`'s `package.json` as a dependency, peer dependency or optional dependency, so the bare specifier resolved only when some other package happened to hoist it to the project root — true on Expo SDK 57 (via `expo`, `expo-sharing`, `expo-splash-screen`), false on SDK 56. Where it was not hoisted, `expo config`, `expo-doctor` and `expo prebuild` failed with `Cannot find module '@expo/config-plugins'`, so EAS builds died before any native compilation.

`withSentry.ts` requires the module eagerly at load time, so the failure was not gated on `experimental_android.enableAndroidGradlePlugin` being enabled.

- `yarn jest test/expo-plugin` — 67 tests pass.
- `tsc -p plugin/tsconfig.json` clean, and the emitted `plugin/build/*.js` now contains four identical `require("expo/config-plugins")` calls instead of three plus one bare `require("@expo/config-plugins")`.
- Not reproduced end-to-end in an Expo SDK 56 CNG project without hoisting; the fix rests on the specifier plus the contents of `expo/config-plugins.js`.

## :pencil: Checklist

- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps

All four plugin files import `ExpoConfig` from `@expo/config-types`, which is also undeclared. Harmless at runtime (type-only, erased on compile), but the reference survives into the generated `.d.ts`. Not addressed here.

`test/expo-plugin/modifyMainApplication.test.ts` and `modifyAppDelegate.test.ts` still mock `@expo/config-plugins` while their sources import `expo/config-plugins`. Works through the re-export, same inconsistency, left alone.


lucas-zimerman (approved):
LGTM!

lucas-zimerman:
> Warnings
> ⚠️ 	
> ### ⚠️ Android SDK Version Mismatch
> Component 	Version
> `sentry-android` in build.gradle 	**8.52.0**
> `sentry-android` bundled by gradle plugin 6.17.0 	**8.51.0**
> 
> If a user enables AGP `autoInstallation` (default: `true`), this mismatch causes:
> 
> > `IllegalStateException: Sentry SDK has detected a mix of versions`
> 
> Our docs instruct React Native users to set `autoInstallation.enabled = false`, so users following the guide are unaffected. Consider either updating `packages/core/android/build.gradle` to `8.51.0` or waiting for a gradle plugin release that bundles `8.52.0`.
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [5673278](https://github.com/getsentry/sentry-react-native/commit/5673278e631b3b61dfd5c04a37b9d3eeba9d420c)

@alwx might be worth fixing this before shipping a new release, it's unrelated to this PR so it might have slipped into main..

PR #6582, opened 2026-08-10
docs(core): Correct enableTurboModuleTracking docs and document TurboModule plumbing

alwx:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Documentation only. No behaviour change, no public API change.

`enableTurboModuleTracking`'s JSDoc claimed it forwards Turbo Module callbacks "to the higher-level Sentry instrumentation (crash attribution, per-module spans, aggregated stats)". Neither half holds:

- It installs the native `TurboModulePerfLogger`, which forwards to a sink — and `SentryTurboModulePerfController::setSink` has no caller outside `RNSentryCocoaTester`. Enabling it emits nothing.
- Everything the text promised comes from `turboModuleContextIntegration()`, registered whenever `enableNative` is true, which never reads this option.

The option is now `@internal` and documented accurately. Kept rather than removed: the native install path already shipped in 8.17.0 and a sink is the intended consumer.

Adds a `packages/core/AGENTS.md` section covering the two independent stacks and the traps in them — notably that `logger` from `@sentry/core` is the Logs API, not the debug logger, so the existing `logger.warn` calls in `turbomodule/` capture log events from the wrap hot path and recurse through the wrapped `RNSentry.captureEnvelope`. Documented, not fixed, to keep this diff docs-only.

First slice of #6168. Two of that issue's premises were wrong and are corrected here rather than propagated into docs: there is no `turboModulesIntegration` (it is `turboModuleContextIntegration`), and `slowCallThresholdMs` / `ignoreTurboModules` / `maxTopModulesPerSpan` are on `TurboModuleContextOptions`, not on the SDK options.

The attribute rename is **not** here — conventions must merge before an SDK ships them, so it is gated on getsentry/sentry-conventions#564.

No tests added — docs and JSDoc only.

`yarn build:sdk`, `yarn api-report:check` (surface unchanged), `yarn lint` (0 errors), `yarn circularDepCheck`, `yarn test` — 1850 + 334 + 1 passing.

Native and E2E suites not run.

## :pencil: Checklist

- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps

1. getsentry/sentry-conventions#564 — register and unify the attributes. Blocks the rest.
2. SDK rename: apply the agreed keys, emit old + new through the backfill window, then lock the API report.
3. sentry-docs PR.

Separate follow-ups: the `logger` → `debug` fix in `turbomodule/`, and #6542. When #6561 lands, the AGENTS.md section needs the `turboModuleCallbacks.ts` row added back.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,13 @@
 - Make the `RNSentry` SPEC CHECKSUM in `Podfile.lock` machine-independent ([#6534](https://github.com/getsentry/sentry-react-native/pull/6534))
 
   The prebuilt `Sentry.xcframework` is now referenced through a `$(PODS_ROOT)/sentry-xcframeworks/…` symlink instead of the absolute per-user cache path, so `Podfile.lock` no longer churns between developers and CI. Expect a one-time `RNSentry` checksum change on the next `pod install`; the `SENTRY_XCFRAMEWORK_CACHE_DIR=/tmp/…` workaround is no longer needed.
+
+### Internal
+
+- Mark `enableTurboModuleTracking` as internal and correct its documentation ([#6168](https://github.com/getsentry/sentry-react-native/issues/6168))
+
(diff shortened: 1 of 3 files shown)


lucas-zimerman (approved):
LGTM1

PR #6585, opened 2026-08-11
fix(core): Use the debug logger, not the Logs API, for SDK diagnostics

alwx:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
`logger` from `@sentry/core` is the structured **Logs API** — every call captures a log record and ships it to the user's project. `debug` is the internal debug logger, printed to the console only under `debug: true`. Ten internal diagnostics were written against `logger`, so an SDK-internal failure turned into a billable log event the user cannot silence.

Switched to `debug` in multiple Turbo Modules related files.

Follow-up to https://github.com/getsentry/sentry-react-native/pull/6561#discussion_r3753179915, agreed with @lucas-zimerman to do separately. The `logger` calls predate #6561.

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps
- `tools/sentryOptionsSerializer.ts` (7 calls) as described above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@
 
 ### Fixes
 
+- Send SDK-internal diagnostics to the debug logger instead of the Sentry Logs API ([#6585](https://github.com/getsentry/sentry-react-native/pull/6585))
+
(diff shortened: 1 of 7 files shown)


lucas-zimerman (approved):
LGTM!

PR #6612, opened 2026-08-25
fix(replay): Warn when replay sample rates are set but the Replay integration is missing

### LINKED ISSUE #6600: Detect missing mobile replay integration configuration
When `Sentry.mobileReplayIntegration()` is omitted from the React Native SDK integrations array, replay setup fails silently and the release is missing without an error or alert. Consider a way to simplify setup or flag the missing integration to the developer.

[React Native Session Replay setup documentation](<https://docs.sentry.io/platforms/react-native/session-replay/#set-up>)

antonis:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Emit a `debug.warn` at `Sentry.init` when a replay sample rate (`replaysSessionSampleRate` / `replaysOnErrorSampleRate`, or their `_experiments` variants) is set but the matching Replay integration is not present in the resolved integrations list — `MobileReplay` on native, `Replay` (browser) on web.

- Added `warnIfReplayIntegrationMissing(options)` in `sdk.tsx`, called right after `getIntegrationsToSetup` so it inspects the *resolved* integration set.
- Exported `BROWSER_REPLAY_INTEGRATION_NAME` from `browserReplay.ts` so the check matches by name constant instead of a magic string.
- The message is platform-aware and points at the correct factory (`mobileReplayIntegration()` / `browserReplayIntegration()`) plus the setup docs.

Warning-only by design: it **never mutates** the resolved integrations. The wording is "may not work as expected" rather than "will not be recorded" because on native, session recording is actually driven by the sample rates alone — what silently breaks without the integration is error→replay linking, on-error (buffer) flush for JS errors, and masking customization; on web nothing records at all.

Closes #6600.

When `mobileReplayIntegration()` is omitted but sample rates are set, replay setup can fail silently — no error or alert. The SDK already auto-adds the mobile integration via `getDefaultIntegrations`, but that path is skipped when the user supplies their own `defaultIntegrations` (array or `false`) or a function-form `integrations` that filters it out, and the browser integration is never auto-added. In those cases there is currently no signal to the developer.

I considered auto-adding the integration instead of warning, but rejected it: the documented way to disable a Sentry integration is to remove it via `integrations`/`defaultIntegrations`, and re-injecting it would override that explicit intent and could silently turn Session Replay (a PII/cost-sensitive feature) back on — a fail-open regression. `debug.warn` only emits when `debug: true`, so there is no production noise.

Added unit tests in `packages/core/test/sdk.test.ts` (new `replay integration missing warning` describe block):
- warns when session / on-error / `_experiments` sample rates are set but the mobile integration is missing (via `defaultIntegrations: []`)
- does not warn when the mobile integration is auto-added (default path)
- does not warn when no sample rates are set
- warns with the `browserReplayIntegration()` hint on web when the browser integration is missing
- does not warn on web when the browser integration is present

`yarn jest test/sdk.test.ts` → 125 passed. oxlint (type-aware) and oxfmt clean on changed files.

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

### DIFF
--- a/packages/core/src/js/sdk.tsx
+++ b/packages/core/src/js/sdk.tsx
@@ -215,11 +215,12 @@ function warnIfReplayIntegrationMissing(options: ReactNativeClientOptions): void
   // Only a rate > 0 actually enables replay; `0` is a valid way to keep it off, so it must not warn.
   const isEnabledRate = (rate: number | undefined): boolean => typeof rate === 'number' && rate > 0;
   const experiments = options._experiments;
-  const hasReplayOptions =
-    isEnabledRate(options.replaysOnErrorSampleRate) ||
-    isEnabledRate(options.replaysSessionSampleRate) ||
-    isEnabledRate(experiments?.replaysOnErrorSampleRate) ||
-    isEnabledRate(experiments?.replaysSessionSampleRate);
+const hasReplayOptions = [
+    options.replaysOnErrorSampleRate,
+    options.replaysSessionSampleRate,
+    experiments?.replaysOnErrorSampleRate,
+    experiments?.replaysSessionSampleRate,
+].some(isEnabledRate);
 
   if (!hasReplayOptions) {
     return;
--- a/packages/core/src/js/sdk.tsx
+++ b/packages/core/src/js/sdk.tsx
@@ -22,6 +22,8 @@ import { FeedbackFormProvider } from './feedback/FeedbackFormProvider';
 import { getDevServer } from './integrations/debugsymbolicatorutils';
 import { getDefaultIntegrations } from './integrations/default';
 import { shouldEnableNativeNagger } from './options';
+import { BROWSER_REPLAY_INTEGRATION_NAME } from './replay/browserReplay';
+import { MOBILE_REPLAY_INTEGRATION_NAME } from './replay/mobilereplay';
 import { enableSyncToNative } from './scopeSync';
 import { TouchEventBoundary } from './touchevents';
 import { ReactNativeProfiler } from './tracing';
@@ -180,6 +182,7 @@ export function init(passedOptions: ReactNativeOptions): void {
     integrations: safeFactory(userOptions.integrations, { loggerMessage: 'The integrations threw an error' }),
     defaultIntegrations,
   });
+  warnIfReplayIntegrationMissing(options);
   initAndBind(ReactNativeClient, options);
   if (__DEV__) {
     checkSentryJsSdkVersionMismatch();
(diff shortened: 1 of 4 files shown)


antonis:
@cursor review

antonis:
@sentry review

packages/core/src/js/sdk.tsx
```diff
@@ -180,6 +182,7 @@ export function init(passedOptions: ReactNativeOptions): void {
     integrations: safeFactory(userOptions.integrations, { loggerMessage: 'The integrations threw an error' }),
     defaultIntegrations,
   });
+  warnIfReplayIntegrationMissing(options);
```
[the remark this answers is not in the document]
antonis: Good catch — fixed in 1d2cee18. Moved `warnIfReplayIntegrationMissing` to after `initAndBind`, since that's where `@sentry/core` runs `debug.enable()`, so the warning now reaches the console on first init when `debug: true` (matching the nearby Expo Go logs).

packages/core/src/js/sdk.tsx
```diff
@@ -199,6 +204,39 @@ export function init(passedOptions: ReactNativeOptions): void {
...
+ * The mobile integration is auto-added by `getDefaultIntegrations` when sample rates are set, but
+ * that path is skipped when the user supplies their own `defaultIntegrations`. On web the browser
+ * integration is never auto-added. In both cases replay would otherwise fail silently.
+ */
+function warnIfReplayIntegrationMissing(options: ReactNativeClientOptions): void {
+  const hasReplayOptions =
+    typeof options.replaysOnErrorSampleRate === 'number' ||
+    typeof options.replaysSessionSampleRate === 'number' ||
+    typeof options._experiments?.replaysOnErrorSampleRate === 'number' ||
+    typeof options._experiments?.replaysSessionSampleRate === 'number';
```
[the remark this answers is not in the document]
antonis: Good catch — fixed in 9d52b7d. The check now only warns when a sample rate is greater than 0, since `0` is a valid way to keep Session Replay off. It still warns for active configs where just one rate is set (e.g. `session: 0, onError: 1.0`). Added tests covering both cases.

packages/core/src/js/sdk.tsx
```diff
@@ -199,6 +204,42 @@ export function init(passedOptions: ReactNativeOptions): void {
...
+ */
+function warnIfReplayIntegrationMissing(options: ReactNativeClientOptions): void {
+  // Only a rate > 0 actually enables replay; `0` is a valid way to keep it off, so it must not warn.
+  const isEnabledRate = (rate: number | undefined): boolean => typeof rate === 'number' && rate > 0;
+  const experiments = options._experiments;
+  const hasReplayOptions =
+    isEnabledRate(options.replaysOnErrorSampleRate) ||
+    isEnabledRate(options.replaysSessionSampleRate) ||
+    isEnabledRate(experiments?.replaysOnErrorSampleRate) ||
+    isEnabledRate(experiments?.replaysSessionSampleRate);
```
lucas-zimerman: Q: what about 


```suggestion
const hasReplayOptions = [
    options.replaysOnErrorSampleRate,
    options.replaysSessionSampleRate,
    experiments?.replaysOnErrorSampleRate,
    experiments?.replaysSessionSampleRate,
].some(isEnabledRate);
```

lucas-zimerman: Actually a double suggestion, why not in this case
```
... 
].some(Boolean);
```

packages/core/src/js/sdk.tsx
```diff
@@ -199,6 +204,42 @@ export function init(passedOptions: ReactNativeOptions): void {
...
+
+  if (!hasReplayOptions) {
+    return;
+  }
+
```
lucas-zimerman, line 228: lets remember isWeb on a const parameter  and reuse the result on line 228 and 235

lucas-zimerman (approved):
Looking good! 
Left some non-blocking reviews

packages/core/src/js/sdk.tsx
antonis: Thank you Lucas. I've applied the suggestion. I think `some(Boolean)` would drop the > 0

packages/core/src/js/sdk.tsx
antonis: Applied with https://github.com/getsentry/sentry-react-native/pull/6612/commits/776a872bcade7c4148b49a79b01de851293cd302

PR #6615, opened 2026-08-26
fix(ios): Force-load the Sentry static archive to keep ObjC category methods

### LINKED ISSUE #6609: iOS: Session Replay network-detail capture breaks requests on sentry-cocoa 9.25.0
## Summary

After bumping sentry-cocoa `9.24.0 → 9.26.0` (#6409), the iOS E2E flow `captureSpaceflightNewsScreenTransaction.test.ios.auto` began failing **deterministically** on `main` and every PR. The flow fails at `No visible element found: "Load More Articles"` because the article `fetch` never completes, so the list never renders.

Root cause is in **sentry-cocoa 9.26.0**, not in RN JS/bridge code: with 9.26.0, an outbound `URLSession` request to a URL on the Session Replay **network-detail allow-list** (with body capture enabled) does not complete. This is a functional networking regression, not just telemetry loss.

## Why this matters (user impact)

Any app on sentry-cocoa 9.26.0 using Session Replay with `networkDetailAllowUrls` + `networkCaptureBodies` will have its requests to those hosts broken. Bounded to allow-listed network-detail URLs; other traffic is unaffected. In `@sentry/react-native` this surfaces on iOS because the native SDK performs the URLSession instrumentation.

## Evidence — three-way E2E bisect (single variable each)

The sample's `mobileReplayIntegration` in `samples/react-native/src/App.tsx` configures network-detail capture for exactly the failing host:

```ts
  networkDetailAllowUrls: ['https://api.spaceflightnewsapi.net'],
  networkCaptureBodies: true,
})
```

| sentry-cocoa | Replay network-detail capture | Spaceflight iOS flow | CI run |
|---|---|---|---|
| 9.26.0 | on | :x: red (4/4 deterministic) | `32713642451` |
| 9.24.0 | on | :white_check_mark: green | `32722276675` |
| 9.26.0 | **off** | :white_check_mark: green | `32725506585` |

Two independent single-variable flips each turn it green → the trigger is 9.26.0's Session Replay network-detail path. It is the only iOS flow making a real external HTTP request, and the only one that failed; all other iOS flows passed in every run.

## Suspected upstream cause (to pass to sentry-cocoa)

**getsentry/sentry-cocoa#8497** — *"Prevent Session Replay network-detail breadcrumbs from blocking URLSession cancellation on the task monitor"* — the only 9.26.0 change touching both the network-detail path and URLSession lifecycle.

Ruled out: **getsentry/sentry-cocoa#8650** (trace-header copy-and-replace setter) — not gated on replay network-detail, so disabling network-detail would not have fixed it. Last-known-good is **9.24.0** (9.25.0 shipped no networking changes per the cocoa changelog).

**Verification for the cocoa team:** revert #8497 on a 9.26.0 build (or bisect 9.25.0→9.26.0) against the repro below.

## Reproduction

1. `@sentry/react-native` on iOS with sentry-cocoa 9.26.0, Session Replay enabled.
2. Configure `mobileReplayIntegration` with `networkDetailAllowUrls: ['<host>']` and `networkCaptureBodies: true`.
3. From the app, issue a `GET` to a URL on that host (e.g. `https://api.spaceflightnewsapi.net/v4/articles`).
4. Observe: the request does not complete; the response body never reaches the app. Deterministic.

## Current state / workaround

- Interim options: hold the RN cocoa pin at 9.24.0, or (for users) disable `networkCaptureBodies` / drop affected hosts from `networkDetailAllowUrls` on 9.26.0.
- A proof branch demonstrating the third bisect row (removes the two options from the sample; iOS E2E green on 9.26.0) was used to isolate this and has been closed — it was a reproduction, not a merge candidate.

## Next steps

- [ ] File/track the upstream **sentry-cocoa** issue (suspected #8497) — this ticket is the reference.
- [ ] Once cocoa ships the fix, re-apply the 9.26.0 bump and restore the network-detail options in the sample.
- [ ] Watch for cold-start flakiness on re-land and re-add a readiness gate if needed.

antonis:
- [x] Bugfix

## :scroll: Description
RN consumes sentry-cocoa as a prebuilt **static** `Sentry.xcframework`, linked implicitly via `@import Sentry` autolink with no whole-archive linker flag. Objective-C category methods — which is what Swift `@objc extension`s compile to — are dead-stripped from a static library at link time unless the consumer opts in. When cocoa 9.25/9.26 moved `SentryReplayNetworkDetails`' capture methods into such an extension (`SentryReplayNetworkDetails+Capture`), the selector went missing at runtime and crashed with `-[Sentry.SentryReplayNetworkDetails setRequestWithSize:bodyData:contentType:allHeaders:configuredHeaders:]: unrecognized selector`.

This adds a per-SDK `-force_load` of the Sentry binary, placed on whichever target's link actually performs the strip — which depends on linkage:
- **Static** (the RN default): RNSentry and the Sentry archive both land in the app binary, so the strip happens at the app link → flag on `user_target_xcconfig`.
- **Dynamic** (`use_frameworks! :linkage => :dynamic`): `@import Sentry` pulls the static archive into the RNSentry dylib, so the strip happens at the RNSentry link → flag on `pod_target_xcconfig`.

It's placed on exactly one target, never both, which avoids forcing a second copy of Sentry into the app binary (duplicate ObjC classes / two `SentrySDK` singletons). Linkage is detected via `ENV['USE_FRAMEWORKS']`, the convention RN's Podfile setup and our samples already use.

`-force_load` is chosen over `-ObjC`/`-all_load` because it is scoped to the Sentry archive (no whole-app size bloat or duplicate-symbol risk) and, unlike `-ObjC`, also retains Swift-only metadata — so it hardens against future valid-but-strippable upstream changes, not only today's category case.

### Note

Skipping changelog since the bug exposed with the Cocoa bump has not been shipped

Fixes the iOS Session Replay network-capture crash reported in #6609. Apple [QA1490](https://developer.apple.com/library/archive/qa/qa1490/_index.html) documents category stripping from static libraries as the consumer's responsibility to resolve at link time. This is a durable consumer-side fix independent of any producer-side patch in sentry-cocoa (getsentry/sentry-cocoa#8872).

* Multiple CI build configurations
* Linked a minimal ObjC harness the same way RN links Sentry (static, no whole-archive flag) against the unpatched buggy 9.26.0 simulator slice: without `-force_load` the selector is absent (crash reproduced); with the exact flag the podspec now emits, the selector is present. Verified manually via a local `pod install` + sample-app build.
* Ran a real `pod install` on the RN sample in both linkage modes and confirmed `-force_load` lands on exactly the right target: static → app aggregate (present) / RNSentry pod (absent); dynamic → app aggregate (absent) / RNSentry pod (present).

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.


packages/core/RNSentry.podspec
```diff
@@ -150,8 +150,34 @@ Pod::Spec.new do |s|
...
+    force_load_flags = SENTRY_XCFRAMEWORK_SLICES_BY_SDK.each_with_object({}) do |(sdk, slice_ids), acc|
+      loads = slice_ids.map do |slice|
+        %(-force_load "#{File.join(sentry_xcframework_ref, slice, 'Sentry.framework', 'Sentry')}")
+      end
+      acc["OTHER_LDFLAGS[sdk=#{sdk}*]"] = (['$(inherited)'] + loads).join(' ')
+    end
+
     pod_target_xcconfig.merge!(xcframework_search_paths)
-    s.user_target_xcconfig = xcframework_search_paths
+    s.user_target_xcconfig = xcframework_search_paths.merge(force_load_flags)
```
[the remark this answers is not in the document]
antonis: Good catch — fixed in 1ec02c12.

`-force_load` placement is now linkage-aware, keyed off `ENV['USE_FRAMEWORKS']` (the same signal RN's Podfile setup and our samples use):

- **static** (default): RNSentry and the Sentry archive both land in the app binary, so the strip happens at the app link → flag on `user_target_xcconfig`.
- **dynamic** (`use_frameworks! :linkage => :dynamic`): `@import Sentry` pulls the static archive into the RNSentry dylib, so the strip happens at the RNSentry link → flag on `pod_target_xcconfig`.

It's one target or the other, never both, which also avoids the second-copy issue you flagged (duplicate ObjC classes / two `SentrySDK` singletons).

Verified with a real `pod install` on the RN sample in both modes:

| Linkage | App aggregate | RNSentry pod target |
|---|---|---|
| static | `-force_load` present | absent |
| dynamic | absent | `-force_load` present |

packages/core/RNSentry.podspec
```diff
@@ -150,8 +150,39 @@ Pod::Spec.new do |s|
...
+
     pod_target_xcconfig.merge!(xcframework_search_paths)
-    s.user_target_xcconfig = xcframework_search_paths
+    user_target_xcconfig = xcframework_search_paths.dup
+    if ENV['USE_FRAMEWORKS'] == 'dynamic'
+      pod_target_xcconfig.merge!(force_load_flags)
+    else
+      user_target_xcconfig.merge!(force_load_flags)
+    end
+    s.user_target_xcconfig = user_target_xcconfig
```
lucas-zimerman, line 185: @antonis  could you check the two warden warnings?

packages/core/RNSentry.podspec
```diff
@@ -150,8 +150,39 @@ Pod::Spec.new do |s|
...
+    #
+    # The flag must ride whichever target's link does the stripping, which
+    # depends on linkage: static libs (the default) absorb Sentry into the app
+    # binary → app link → `user_target_xcconfig`; `:linkage => :dynamic` absorbs
+    # it into the RNSentry dylib → that link → `pod_target_xcconfig`. Never
+    # both, or a second copy of Sentry lands in the app. Detected via
+    # `ENV['USE_FRAMEWORKS']`.
+    force_load_flags = SENTRY_XCFRAMEWORK_SLICES_BY_SDK.each_with_object({}) do |(sdk, slice_ids), acc|
+      loads = slice_ids.map do |slice|
+        %(-force_load "#{File.join(sentry_xcframework_ref, slice, 'Sentry.framework', 'Sentry')}")
```
NinjaLikesCheez, line 173: `m`: why `force_load` over `ObjC`? I believe force load will disable dead code stripping for all members in the archive, where `ObjC` would only do that for Objective-C members. In reality it probably doesn't matter all that much, but `force_load` could increase binary size more than needed.

antonis: Good point! `force_load` would probably future proof against wider swift refactorings in the cocoa sdk but `ObjC` should be enough for this bug because SentryReplayNetworkDetails+Capture compiles to an ObjC category.
Trying this out with dc780c5a6fb03a99e34d3804ea919c61fb880133

packages/core/RNSentry.podspec
```diff
@@ -150,8 +150,19 @@ Pod::Spec.new do |s|
...
+    # `SentryReplayNetworkDetails+Capture` in sentry-cocoa 9.25/9.26, #6609) is
+    # dropped and crashes at runtime with "unrecognized selector". `-ObjC` is
+    # Apple's documented fix (QA1490). Unlike `-force_load` it names no archive
+    # — it only affects static libs in the current link — so the same flag is
+    # safe on both targets whatever the linkage (app link under static, RNSentry
+    # dylib link under dynamic), pulling no second Sentry copy and needing no
+    # linkage detection.
+    objc_ldflags = { 'OTHER_LDFLAGS' => '$(inherited) -ObjC' }
+    pod_target_xcconfig.merge!(xcframework_search_paths).merge!(objc_ldflags)
+    s.user_target_xcconfig = xcframework_search_paths.merge(objc_ldflags)
```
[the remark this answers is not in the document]
antonis: Verified this empirically — it's a false positive on both points.

I linked the buggy 9.26 slice the same way RN autolinks it (static `Sentry.xcframework`, `@import Sentry`), varying only the linker flag, and probed the exact crashing selector (`class_getInstanceMethod` for `-setRequestWithSize:bodyData:contentType:allHeaders:configuredHeaders:`) at runtime in the simulator:

| Link (buggy 9.26 slice) | Selector present? |
|---|---|
| no flag (RN default) | **NO** — reproduces the crash |
| `-ObjC` | **YES** — rescued |
| `-force_load` / `-all_load` | YES |

**"The default static app link already passes `-ObjC`":** The no-flag link is identical to RN's default, and the selector is *absent* there. If `-ObjC` were already effectively on that link the selector would be present — and crash #6609 could not occur in the default configuration. The RN template doesn't set `-ObjC` globally, and CocoaPods doesn't add it to the aggregate unless a pod requests it, which is what this change does.

**"`-ObjC` does not reliably keep the Swift `@objc extension`":** With `-ObjC` the selector is *present*. The QA1490 "64-bit linker won't load category-only object files" caveat was real in the old `ld` but doesn't apply to the modern `ld64`/ld-prime this builds with — confirmed by the measurement above.

Together with the `pod install` check (the flag lands on the app aggregate under static linkage and on the RNSentry pod target under dynamic), the loop is closed: the flag reaches the strip-point link, and that flag rescues the selector.

packages/core/RNSentry.podspec
NinjaLikesCheez: Thanks! My thinking was that using `force_load` will include all non-ObjC members too (i.e. C/C++) that might otherwise be stripped if not used - and thus inflate the binary unnecessarily. Since we're only calling it on Sentry frameworks it's probably not the end of the world - but when we have KSCrash installed we might see things being pulled in that we don't need.

packages/core/RNSentry.podspec
antonis: I missed those. Thank you @lucas-zimerman 🙇 
Should be fixed with the change in dc780c5a6fb03a99e34d3804ea919c61fb880133 following up to https://github.com/getsentry/sentry-react-native/pull/6615#discussion_r3868939432  🤞

packages/core/RNSentry.podspec
antonis: > Trying this out with dc780c5a6fb03a99e34d3804ea919c61fb880133

[The CI is failing](https://github.com/getsentry/sentry-react-native/actions/runs/33058660585/job/98482259508?pr=6615) with the same [issue](https://github.com/getsentry/sentry-react-native/issues/6609) that started this investigation :(

I'll revert back to the `force_load` solution and iterate on the [warden warning](https://github.com/getsentry/sentry-react-native/pull/6615#discussion_r3867731200)

packages/core/RNSentry.podspec
antonis: I'll tackle the warden comments separately in https://github.com/getsentry/sentry-react-native/pull/6621

packages/core/RNSentry.podspec
NinjaLikesCheez: Huh... weird. Okay thanks for trying it! LGTM

lucas-zimerman (approved):
LGTM!

PR #6617, opened 2026-08-26
chore(sample): Bump React Native sample to 0.87.0

antonis:
> [!NOTE]
> **Stacked on #6625** (`fix(core)`: Resolve Metro from project root). This PR is based on that branch, so its diff is sample-only. Merge #6625 first, then this will retarget to `main`.

## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description
Bumps the React Native sample app (`samples/react-native`) to React Native `0.87.0` (from `0.86.2`), including the Android AGP 9 / Kotlin migration that RN 0.87 requires.

**JS dependencies**
- `react-native` and `@react-native/new-app-screen` → `0.87.0`.
- `@react-native/babel-preset`, `@react-native/jest-preset`, `@react-native/metro-config`, `@react-native/typescript-config` → `0.87.0`.
- `react-native-reanimated` `4.4.1` → `4.6.0` and `react-native-worklets` `0.9.2` → `0.12.1` (required for RN 0.87 compatibility — 4.4.1 fails `pod install` on RN 0.87).
- `yarn.lock` — regenerated.

`react` was already at `19.2.3`, which satisfies RN 0.87's peer requirement.

**Android (AGP 9 / Kotlin, per the [RN 0.87 blog](https://reactnative.dev/blog/2026/08/11/react-native-0.87) and Upgrade Helper)**
- Gradle wrapper `9.3.1` → `9.4.1` (RN 0.87 minimum).
- `compileSdk`/`buildToolsVersion` → `37`, `kotlinVersion` → `2.2.0`.
- `android.builtInKotlin=false` and `android.newDsl=false` in `gradle.properties` (AGP 9 opt-outs).
- `proguard-android-optimize.txt` (AGP 9 no longer supports `proguard-android.txt`).

## :bulb: Motivation and Context
React Native 0.87.0 has been released. This keeps the RN sample app on the latest version and surfaces integration issues early.

Relates to #6527.

## :green_heart: How did you test it?
- iOS: `pod install` succeeds locally with new architecture (reanimated 4.6.0 + worklets 0.12.1).
- Android: Gradle configuration succeeds locally with Gradle 9.4.1 + AGP 9 (all AGP 9 / Kotlin config errors resolved).
- CI: Sample Application workflow (`sample-application.yml`) against RN 0.87.0.
[...]

PR #6625, opened 2026-08-27
fix(core): Resolve Metro from project root for source map generation

antonis:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

The default Metro serializer resolved Metro's internal modules (`baseJSBundle`, `bundleToString`, `sourceMapString`) from the SDK's own location (`__dirname`). In a workspace/monorepo where a different Metro version is nested under `@sentry/react-native` than the one the app uses, that meant source maps were generated with a mismatched Metro version.

The serializer now resolves Metro internals preferring the app's `options.projectRoot`, falling back to the SDK-resolvable Metro. Resolution is deferred to the first serialization (when `projectRoot` is available) and memoized. In a normal single-Metro install both paths resolve to the same Metro, so behavior is unchanged.

Surfaced while bumping the React Native sample to 0.87 (Metro 0.87) while `packages/core` pins Metro `0.84.4` as a devDependency: the nested SDK Metro shadowed the app's Metro 0.87 and broke production source-map generation. The fix also benefits real users who have multiple Metro versions resolvable in a monorepo.

Split out of #6617 so the sample bump carries no core changes.

- Existing Metro serializer unit tests pass.
- Local Android **release** build of the RN 0.87 sample produced a valid production bundle with a Debug ID via this serializer.

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps

- Add a dedicated unit test that asserts Metro internals are resolved from `options.projectRoot` when a different Metro is nested under the SDK.


packages/core/src/js/tools/vendor/metro/utils.ts
```diff
@@ -27,15 +27,97 @@
...
+      }
+    }
+    try {
+      return require(candidate);
+    } catch (e) {
+      lastError = e;
+    }
+  }
+  throw lastError;
+}
```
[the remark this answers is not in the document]
antonis: Good catch — this is a real ordering flaw. The old loop iterated candidates on the outer level and passed `paths: [projectRoot, __dirname]` to a single `require.resolve`, so path-shape (`private` > `src`) took priority over location. The effective order was app/`private` → SDK/`private` → app/`src` → SDK/`src`, which means an app whose Metro only resolves via `metro/src/*` while the SDK's resolves via `metro/private/*` would pick the SDK's Metro — defeating the fix.

Fixed in 06451544 by making resolution **location-first**: each path shape is tried against `projectRoot` before falling back to the SDK's own location (path-shape stays the inner preference). New order: app/`private` → app/`src` → SDK/`private` → SDK/`src`. The bare-`require` last resort now runs only after every located attempt fails, so it can't shadow the app's Metro either.

Added a regression test (`uses the app's Metro even when it only exposes internals via metro/src/* and the SDK exposes metro/private/*`) that reproduces exactly this skew with on-disk fake-Metro fixtures — it fails against the old resolver and passes with the fix.

packages/core/src/js/tools/vendor/metro/utils.ts
```diff
@@ -27,15 +27,107 @@
...
+ * Resolution is location-first: every candidate path shape is tried against the app
+ * (`projectRoot`) before falling back to the SDK's own location. Within a single location the
+ * newer `metro/private/*` path is preferred over the legacy `metro/src/*` path, since Metro moved
+ * its internals behind the `private` export. Ordering locations outside the path shapes is what
+ * guarantees the app's Metro wins even when the two copies expose their internals via different
+ * subpaths (e.g. app on `metro/src/*`, SDK on `metro/private/*`).
+ */
+// oxlint-disable-next-line typescript-eslint(no-explicit-any)
+function requireMetroModule(candidates: string[], projectRoot: string | undefined): any {
+  const roots = projectRoot ? [projectRoot, __dirname] : [__dirname];
```
lucas-zimerman, line 64: could projectRoot be an empty string here? it may silently fail if thats the case

packages/core/src/js/tools/vendor/metro/utils.ts
```diff
@@ -27,15 +27,107 @@
...
+  );
+
+  const sourceMapString: SourceMapStringFunction = toCallable(
+    requireMetroModule(
+      ['metro/private/DeltaBundler/Serializers/sourceMapString', 'metro/src/DeltaBundler/Serializers/sourceMapString'],
+      projectRoot,
+    ),
+    'sourceMapString',
+  );
+  if (typeof sourceMapString !== 'function') {
```
lucas-zimerman: Why have you added this check to `sourceMapString` but not `bundleToString` and `baseJSBundle` ?

packages/core/src/js/tools/vendor/metro/utils.ts
```diff
@@ -27,15 +27,107 @@
...
+ * subpaths (e.g. app on `metro/src/*`, SDK on `metro/private/*`).
+ */
+// oxlint-disable-next-line typescript-eslint(no-explicit-any)
+function requireMetroModule(candidates: string[], projectRoot: string | undefined): any {
+  const roots = projectRoot ? [projectRoot, __dirname] : [__dirname];
+  let lastError: unknown;
+  for (const root of roots) {
+    for (const candidate of candidates) {
+      try {
+        return require(require.resolve(candidate, { paths: [root] }));
```
lucas-zimerman, line 73: Lets add the explanation directly above the return code
```suggestion
        // the line below resolves `candidate` as Node would if
        // required from `root`, without actually requiring it from there. That's what lets us
        // pick up the app's node_modules Metro instead of the SDK's own, even though this code
        // itself lives inside the SDK.
        return require(require.resolve(candidate, { paths: [root] }));
```

lucas-zimerman: If accepted we should also simplify the function description since part of it is already covered here

packages/core/src/js/tools/vendor/metro/utils.ts
```diff
@@ -27,15 +27,107 @@
...
+  const sourceMapString: SourceMapStringFunction = toCallable(
+    requireMetroModule(
+      ['metro/private/DeltaBundler/Serializers/sourceMapString', 'metro/src/DeltaBundler/Serializers/sourceMapString'],
+      projectRoot,
+    ),
+    'sourceMapString',
+  );
+  if (typeof sourceMapString !== 'function') {
+    throw new Error(`
+[@sentry/react-native/metro] Cannot find sourceMapString function in 'metro/src/DeltaBundler/Serializers/sourceMapString'.
```
lucas-zimerman: Isn't it also required on `metro/private/DeltaBundler/Serializers/sourceMapString` ?

packages/core/src/js/tools/vendor/metro/utils.ts
```diff
@@ -69,50 +161,18 @@ export const getSortedModules = (
...
-    sourceMapString = (sourceMapStringModule as { sourceMapString: typeof sourceMapStringType }).sourceMapString;
-  } catch (e) {
-    sourceMapString = require('metro/src/DeltaBundler/Serializers/sourceMapString');
-    if ('sourceMapString' in sourceMapString) {
-      // Changed to named export in https://github.com/facebook/metro/commit/34148e61200a508923315fbe387b26d1da27bf4b
-      // Metro 0.81.0 and 0.80.10 patch
-      sourceMapString = (sourceMapString as { sourceMapString: typeof sourceMapStringType }).sourceMapString;
+  return (entryPoint, preModules, graph, options) => {
+    if (!internals) {
+      internals = resolveMetroInternals(options.projectRoot);
```
lucas-zimerman, line 188: Seems like an edge-case but local AI complained about the following:
```
resolveMetroInternals() (and its sourceMapString shape check/throw) now runs before isHot is 
computed. Previously the throw only fired on non-hot builds; now it can break the dev server too.
 Now it fires unconditionally on the very first bundle serialized, hot or not, so a bad Metro shape that used to only break production/export builds can now break yarn start itself.
```

lucas-zimerman (changes requested):
overall looks good, there are some bits to be polished, and after that we can merge it :D

packages/core/src/js/tools/vendor/metro/utils.ts
antonis: Sounds good 👍 Accepted and simplified the function comment with 5dbc011caa8d4c3d51b85e990ecbfa94cb113d0f

packages/core/src/js/tools/vendor/metro/utils.ts
antonis: That asymmetry was inherited from the original code. I've folded the guard into toCallable, which every internal goes through with 79ea6094b3862c62d96f5af5e4214de662847315

packages/core/src/js/tools/vendor/metro/utils.ts
antonis: Good point 👍  Update with 79ea6094b3862c62d96f5af5e4214de662847315 to be path agnostic

packages/core/src/js/tools/vendor/metro/utils.ts
antonis: I think this should work since an empty string should fallback to `[__dirname]`

packages/core/src/js/tools/vendor/metro/utils.ts
antonis: Good catch 👍 Updated with b585061a2e0f85ff5428775017c7014a780cebcf

lucas-zimerman (approved):
LGTM!

PR #6210, opened 2026-05-22
fix(tracing): Bound TTID/TTFD to prevent inflated transactions

antonis:
## :loudspeaker: Type of change
- [x] Bugfix

## :scroll: Description

Apply the existing 30s `TIME_TO_DISPLAY_TIMEOUT_MS` deadline to TTID spans (both automatic and manual) and to the transaction timestamp extension logic. Previously only TTFD had a deadline guard, and even TTFD's guard didn't prevent the transaction `event.timestamp` from being extended.

## :bulb: Motivation and Context

Fixes #6208

On Android with `enableTimeToInitialDisplay: true`, when a `ScreenStackFragment` view is destroyed (e.g., navigating away in a `Stack > Tab > Stack` hierarchy) and later recreated (e.g., POP back), `RNSentryReactFragmentLifecycleTracer.onFragmentViewCreated` re-registers a `FirstDrawDoneListener`. The captured draw timestamp can be minutes or days after the navigation span started. `processEvent` then unconditionally extended `event.timestamp` with this stale TTID, inflating navigation transactions to hours or days.

TTFD already had a `deadline_exceeded` status guard, but it wasn't applied to the transaction timestamp extension. TTID had no deadline guard at all. This change applies the same 30s deadline consistently to both TTID and TTFD across all code paths.

## :green_heart: How did you test it?

Manual, CI Checks

## :pencil: Checklist
- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPII` is enabled
- [x] I updated the docs if needed.
- [x] I updated the wizard if needed.
- [x] All tests passing
- [x] No breaking changes

PR #6626, opened 2026-08-27
fix(tracing): Keep launch-screen TTID/TTFD anchored to navigation on delayed first navigation

### LINKED ISSUE #6618: Launch-screen TTID/TTFD inflated by pre-navigation time
### SDK versions
Regression can appear from **~8.3.0–8.5.0 onward**, as app-start attachment to the first navigation transaction became reliable (#5756 in 8.3.0, #5833 in 8.5.0); symptom shaped/censored by the 30s cap added in **8.13.0**. Baseline behavior last correct in **8.1.0**.

### Platform
iOS and Android (JS-layer mechanism, platform-agnostic; reproductions captured on Android).

### Summary
When an app's **first React-Navigation transaction fires well after process start** (a delayed first navigation — splash / auth / loading / deferred nav), the app-start integration still attaches app-start data to it and rewrites the transaction's `start_timestamp` to process-init time. `timeToDisplayIntegration` then computes `time_to_initial_display = ttid_end − event.start_timestamp`, so TTID becomes *process_init → first_frame* and absorbs the entire pre-navigation gap. The launch screen's actual render is a few hundred ms, but the reported TTID is 20–30s.

This is **not** a change to the TTID formula — the rewrite code is byte-identical to 8.1.0. What changed is that app-start data now reliably *reaches* the first navigation transaction, which it did not in 8.1.0.

### Evidence (launch screen, Android)
Three slow events, identical shape:

| Root `ui.load` duration | `app_start_cold` | `navigation.processing` (real render) | `ui.load` self_time (dead air) |
|---|---|---|---|
| 20.04s | 1.57s | 186ms | **18.15s** |
| 25.86s | 2.37s | 308ms | **21.00s** |
| 29.75s | 1.03s | 30ms | (same shape, 38 spans) |

In every trace: cold start completes ≤2.4s → **~18–21s uninstrumented gap** → navigation + real work in a thin sliver at the end. The `ui.load` span's own `self_time` (18–21s) *is* that gap.

Aggregate P95 by `sdk.version`, same screen/traffic:
- `time_to_initial_display`: **906ms (8.1.0) → 23.40s (8.13.0)**
- `app_start_cold`: *(not present on 8.1.0)* → **2.13s (8.13.0)** — app-start itself is healthy; the inflation is the pre-nav gap, not a large app start
- `app_start_warm`: *(not present on 8.1.0)* → **1.07s (8.13.0)**

The absence of app-start measurements on 8.1.0 is the direct confirmation that app-start attachment to this transaction is new.

### Root cause
`packages/core/src/js/tracing/integrations/appStart.ts`
- `:727` `event.start_timestamp = appStartTimestampSeconds;` (process init)
- `:736-738` re-anchors the `ui.load.initial_display` span to app-start and recomputes `time_to_initial_display`
- `:684-685` age guard `MAX_APP_START_AGE_MS = 60_000` — permits attachment when the first navigation is up to 60s after process-init

`packages/core/src/js/tracing/integrations/timeToDisplayIntegration.ts`
- `:51` `transactionStartTimestampSeconds = event.start_timestamp`
- `addAutomaticTimeToInitialDisplay`: `durationMs = (ttidTimestampSeconds − transactionStartTimestampSeconds) * 1000`
- `appStartIntegration` is registered before `timeToDisplayIntegration` (`integrations/default.ts`), so the start is already rewritten before TTID is computed.

### Why it correlates with SDK version (enabler chain)
The rewrite existed in 8.1.0 but never fired on the launch screen (a premature route-less idle span absorbed the app-start lock, then was discarded). These fixes progressively routed app-start onto the first sampled, non-discarded navigation transaction:

| PR | Version | Change |
|----|---------|--------|
| #5756 | 8.3.0 | App start data not attached to sampled transactions when preceded by unsampled transactions |
| #5789 | 8.4.0 | Defer initial navigation span creation until navigation container is registered |
| #5833 | 8.5.0 | Fix app start data lost when first navigation transaction is discarded |
| #6051 | 8.10.0 | Discard invalid navigation/interaction transactions via an event processor |

### The ~23s plateau
#6210 (8.13.0) added a `>30s` gate in `timeToDisplayIntegration.ts` (`:18-19`, `:80-87`) that drops inflated samples entirely. A distribution right-censored at 30s pushes high percentiles against the ceiling → observed ~23s P95. Before 8.13 the same inflation was reported uncapped.

### Impact
Silently changes the meaning of a core mobile vital (launch-screen TTID/TTFD) for any app whose first navigation is delayed after process start. Corrupts P95 dashboards/alerting and makes auto-TTID incomparable across the 8.1→8.13 range, with no user-facing signal that the baseline moved.

### Proposed fix direction
- Compute/attach auto-TTID relative to the **navigation transaction's own (dispatch) start**, not the app-start-rewritten `event.start_timestamp` — e.g. snapshot the nav start before app-start attachment, or compute TTID before `appStart`'s event processor runs.
- Tighten `MAX_APP_START_AGE_MS` (and/or only attach app-start when the first navigation occurs within a small window of process-init) so a long-delayed first navigation doesn't inherit app start.
- Consider not conflating "app cold start" with "first screen initial display" when they're separated by a large uninstrumented gap.

### Reproduction
App with a delayed first React-Navigation navigation (splash / auth / loading taking more than ~5s before the first `navigation` transaction). Observe launch-screen `time_to_initial_display` ≈ process-init → first frame, with `ui.load` self_time ≈ the pre-navigation gap and `navigation.processing` ≈ the true render.

**Amplifier:** a high `finalTimeoutMs` on `reactNativeTracingIntegration` (e.g. `finalTimeoutMs: 45000`, `idleTimeoutMs: 5000`) lets the idle navigation transaction stay open long enough to encompass a large pre-navigation gap, making the inflation more pronounced.

antonis:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
On non-standalone app start tracking, attaching app start to the first navigation transaction rewrites its `start_timestamp` to process init and re-anchors the TTID/TTFD spans to it; `timeToDisplay` (which runs after) then derives `time_to_initial_display`/`time_to_full_display` from that rewritten start.

On a **delayed** first navigation — a splash / auth / loading screen held for several seconds — this makes TTID/TTFD span the uninstrumented app-ready → navigation gap instead of the screen render, inflating them by an order of magnitude (e.g. ~0.9s → ~23s in #6618; the 30s cap from #6210 flattens the overflow into a plateau).

This adds a gap check: when `first navigation start − app start end > 5s`, the SDK still reports the `app_start_cold`/`app_start_warm` measurement but leaves `start_timestamp` (and therefore TTID/TTFD) at the navigation start, so they measure the real render. Prompt cold-start displays (gap ≤ 5s) are unchanged. Scoped to non-standalone, non-`__DEV__`, and runs after the existing age/duration guards.

The `5s` cutoff is a heuristic; a root-cause fix would use a native initial-display signal instead of elapsed time (see Next steps).

React Native decouples JS navigation from native app start, so the first `navigation` transaction can fire arbitrarily late. The [TTID/TTFD spec](https://develop.sentry.dev/sdk/telemetry/traces/time-to-initial-full-display/) anchors first-screen TTID to app start with no gap guard, because on native the first Activity *is* the app start — they can't be seconds apart. On RN they can, producing a gap the spec never contemplated.

Fixes #6618

- New unit tests in `appStart.test.ts`: delayed nav keeps `start_timestamp` at the navigation start and adds no `Cold Start`/`Warm Start` span while still reporting `app_start_cold`/`warm`; prompt nav (< 5s) keeps the existing app-start-anchored behavior; dev build skips the branch.
- `appStart.test.ts` (94 passing, incl. snapshots) and `timeToDisplayIntegration.test.ts` pass; `yarn build`, TS/JS lint, circular-dep check, and `api-report:check` all clean (no public API change).
- Manual testing

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.


antonis:
@cursor review

antonis:
@sentry review

packages/core/src/js/tracing/integrations/appStart.ts
```diff
@@ -722,6 +732,31 @@ export const appStartIntegration = ({
...
+    // adding the process-init-anchored breakdown spans — would make TTID/TTFD absorb the whole gap.
+    // Skipped for standalone (the transaction *is* the app start) and in dev builds (long dev app starts).
+    const appReadyToFirstDisplayGapMs = event.start_timestamp
+      ? event.start_timestamp * 1000 - originalAppStartEndTimestampMs
+      : 0;
+    if (!standalone && !__DEV__ && appReadyToFirstDisplayGapMs > MAX_APP_START_TO_FIRST_DISPLAY_GAP_MS) {
+      debug.log(
+        '[AppStart] First navigation is delayed past the app start end. Reporting the app start measurement ' +
+          'only and leaving the screen TTID/TTFD anchored to the navigation start.',
+      );
```
lucas-zimerman: If the message is too big for who is reading the SDK code it may also make sense for the layout to also reflect to the end users.

```suggestion
      debug.log(`[AppStart] First navigation is delayed past the app start end.
Reporting the app start measurement only and leaving the screen TTID/TTFD anchored to the navigation start.`);
```

packages/core/src/js/tracing/integrations/appStart.ts
```diff
@@ -722,6 +732,31 @@ export const appStartIntegration = ({
...
+      const measurementKey = appStart.type === 'cold' ? APP_START_COLD_MEASUREMENT : APP_START_WARM_MEASUREMENT;
+      event.measurements = event.measurements || {};
```
lucas-zimerman: the amount of measurement initialization that uses this pattern is increasing.

What do you think if we create a small helper like 
```
function addMeasurement(
  event: Event,
  key: string,
  measurement: Measurement,
): void {
  event.measurements ??= {};
  event.measurements[key] = measurement;
}
```
And replace any part of the code that we are initializing the measurements like that.

packages/core/src/js/tracing/integrations/appStart.ts
antonis: Sounds good 👍

packages/core/src/js/tracing/integrations/appStart.ts
antonis: Good idea 👍 Added with 7735008585684ce7792c5b9246a7fef36a13e3b5

lucas-zimerman (approved):
LGTM!

PR #6683, opened 2026-09-07
test(e2e): Restore iOS replay assertion in captureReplay test

### LINKED ISSUE #6082: Add back replay assertion from captureReplay E2E test
### Description

Evaluate and add back the validation when we resolve the current CI stability issues

* Add back iOS replay assertion from captureReplay test removed with https://github.com/getsentry/sentry-react-native/pull/6072
* Evaluate adding an equivalent Android check

antonis:
- [x] Bugfix

## :scroll: Description

Re-adds the `assertReplay.yml` iOS check to the `captureReplay` e2e test, removed in #6072, and makes it deterministic instead of flaky.

The assertion was flaky for a structural reason, not CI noise: buffer-mode replay (`replaysOnErrorSampleRate`) only attaches a `replay_id` to the error event if the native replay buffer captured at least one frame **before** the error fired. On slow CI the exception was captured while the buffer was still empty, so the event was sent **permanently** without a `replay_id` — and no server-side retry could recover it (`sentryApi.js` already retries non-200s for ~10 min, so replay processing lag was never the real problem).

Changes:
- **`src/EndToEndTests.tsx`** — add a side-effect-free `Replay Ping` counter (`testID='replayPing'`) that mutates the view hierarchy without sending any events.
- **`maestro/utils/primeReplayBuffer.yml`** (new) — tap `replayPing` 8× to churn the view (and spend a few seconds while Maestro re-reads the UI between taps) so the native capture records frames before the exception.
- **`maestro/utils/assertReplay.yml`** — restored verbatim from before #6072 (asserts `replay_id`, duration, segments, `ftypmp42` codec via the Sentry API).
- **`maestro/captureReplay.yml`** — run the prime step, then re-wire `assertReplay.yml`, both gated `platform: iOS`.

`dist/` is gitignored and rebuilt by `cli.mjs` on create, so the harness change flows into CI automatically.

Closes #6082.

[CI](https://github.com/getsentry/sentry-react-native/actions/runs/34131281750/job/101776247314?pr=6683)

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps

https://github.com/getsentry/sentry-react-native/pull/6684

### DIFF
--- a/dev-packages/e2e-tests/maestro/captureReplay.yml
+++ b/dev-packages/e2e-tests/maestro/captureReplay.yml
@@ -5,5 +5,16 @@ jsEngine: graaljs
     file: utils/launchTestAppClear.yml
     env:
       replaysOnErrorSampleRate: 1.0
+# Prime the native replay buffer before capturing the exception so a replay_id
+# is attached to the event (iOS only; Android does not reliably capture replays
+# in CI — see https://github.com/getsentry/sentry-react-native/pull/4277).
+- runFlow:
+    file: utils/primeReplayBuffer.yml
+    when:
+      platform: iOS
 - tapOn: "Capture Exception"
 - runFlow: utils/assertEventIdVisible.yml
+- runFlow:
+    file: utils/assertReplay.yml
+    when:
+      platform: iOS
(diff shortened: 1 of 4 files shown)


lucas-zimerman (approved):
LGTM!

PR #6685, opened 2026-09-08
fix(replay): Defer buffered replay upload until after error sampling

### LINKED ISSUE #6598: On-error mobile replay orphaned when its error is dropped by `sampleRate`
## Root cause

Since `@sentry/core` 10.70.0 ([getsentry/sentry-javascript#22819](https://github.com/getsentry/sentry-javascript/pull/22819)), the error `sampleRate` roll runs **after** `beforeSend`. The mobile-replay on-error flush lives in a `beforeSend` wrapper ([`mobilereplay.ts`](https://github.com/getsentry/sentry-react-native/blob/main/packages/core/src/js/replay/mobilereplay.ts)), so it now uploads the buffered replay **before** the sampling decision. When the error is then dropped by `sampleRate`, the replay is uploaded with no corresponding error event.

## Who is affected

Only the intersection of:

- **Buffer / on-error replay** (`replaysOnErrorSampleRate`)
- **`sampleRate < 1.0`** (error events are being sub-sampled)
- A specific error that flushes the buffer and then loses the `sampleRate` roll

Full-session replays and default-`sampleRate` apps are **not** affected.

## Proposed fix

- **JS:** in `beforeSend`, only *link* the event to the buffered replay id (no upload); defer the native flush to the **`afterSendEvent`** client hook, which fires only for events that survive sampling and are actually sent. `afterSendEvent` already exists in core — no `@sentry/core` change needed.
- **Android:** ✅ RN-repo-only. Read the buffered id from `ReplayController.getReplayId()` (public sentry-java API).
- **iOS:** requires an upstream **sentry-cocoa** change. `PrivateSentrySDKOnly.getReplayId` reads the scope (empty while buffering) in all released versions incl. 9.27.0 / `main`; the buffered `sessionReplayId` is only reachable via internal APIs (`getReplayIntegration`, not in the public header). Fix: have `getReplayId` fall back to the integration's buffered `sessionReplayId`. iOS then follows with a cocoa bump.

antonis:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

In buffer / on-error Session Replay mode (`replaysOnErrorSampleRate`), the buffered replay was flushed inside the mobile-replay integration's `beforeSend` wrapper. Since `@sentry/core` 10.70.0 the error `sampleRate` roll runs **after** `beforeSend`, so a replay was uploaded even for errors that were then dropped by `sampleRate` — orphaning the replay and burning replay quota.

The work is split in two:

1. **`tagEventWithReplayId`** (in the `beforeSend` wrapper) only links the event to the current buffered replay id — a read via `NATIVE.getCurrentReplayId()`, no flush.
2. **`flushReplayForSentEvent`** (in a new `afterSendEvent` client hook) performs the native `captureReplay()` flush, and only fires for events that survive sampling and are actually sent.

Reading the id *before* the flush requires the current replay id to be available while a buffer replay is still recording:

- **Android:** `getCurrentReplayId()` now prefers `ReplayController.getReplayId()` (assigned at record start) and falls back to the scope. Self-contained — sentry-java already exposes it.
- **iOS:** relies on the matching sentry-cocoa change to `SentrySDK.internal.replay.replayId` (getsentry/sentry-cocoa#8976). **Until `sentry_cocoa_version` in `RNSentry.podspec` is bumped to a release containing that fix, on-error replays are not linked or uploaded on iOS.** This is the blocker for taking this PR out of draft.

Fixes getsentry/sentry-react-native#6598.

- Manual testing
- Rewrote `mobilereplay.test.ts` to assert the new split (no flush in `beforeSend`; flush in `afterSendEvent`).
- Added `mobilereplay.sampling.test.ts`, which drives a real `@sentry/core` client and proves `sampleRate: 0` → `captureReplay` is never called.
- Added `RNSentryReplayIdTest.java` covering the controller-first `getCurrentReplayId()` lookup (4 cases).
- Local: JS replay suites pass (48), `yarn build:sdk`, `yarn circularDepCheck`, oxlint, and Java format all clean.

## :pencil: Checklist

- [x] All tests passing.
- [x] Public API changes reviewed by another Mobile SDK team member or implemented according to the develop docs spec.
- [x] No breaking changes.


antonis:
@cursor review

antonis:
@sentry review

antonis:
Blocking till https://github.com/getsentry/sentry-cocoa/pull/8976 is released

packages/core/src/js/replay/mobilereplay.ts
```diff
@@ -370,39 +405,54 @@ export const mobileReplayIntegration = (initOptions: MobileReplayOptions = defau
...
+        debug.log(
+          `[Sentry] ${MOBILE_REPLAY_INTEGRATION_NAME} not sampled for event ${eventId} (replaysOnErrorSampleRate).`,
+        );
       }
+    } catch (error) {
+      debug.error(`[Sentry] ${MOBILE_REPLAY_INTEGRATION_NAME} Failed to flush replay for sent event ${eventId}`, error);
     }
-
-    return event;
   }
```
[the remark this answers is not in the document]
antonis: Good catch — fixed in a676da9a. On the null path, `flushReplayForSentEvent` now re-reads `NATIVE.getCurrentReplayId()` and updates the cache, so it resolves to the still-active buffer id (or null) instead of leaving the non-uploaded id exposed to DSC/metrics/`getReplayId()`. Covered by the new test "re-reads the current recording id when the flush uploads nothing (on-error sampling miss)".

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -485,6 +485,17 @@ public void captureReplay(boolean isHardCrash, Promise promise) {
...
+    // Prefer the replay controller's id: it is assigned when recording starts
+    // (buffer or session) and is therefore available BEFORE a replay is
+    // flushed. The scope's replayId is only populated once a replay is sent, so
+    // it stays empty while a buffer (on-error) replay is recording (issue
+    // #6598).
+    final @NotNull SentryId controllerId =
+        Sentry.getCurrentScopes().getOptions().getReplayController().getReplayId();
+    if (controllerId != SentryId.EMPTY_ID) {
+      return controllerId.toString();
+    }
```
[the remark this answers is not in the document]
antonis: Fixed in a676da9a. `captureReplay` now resolves via a dedicated `getReplayIdFromScope()` — the scope's replayId is populated only when a replay is actually sent, so an on-error sample miss resolves null, matching the iOS contract (`captureReplayWithReturnValue`). `getCurrentReplayId()` still prefers the controller id so a buffered replay can be linked in `beforeSend`. New tests: `captureReplayResolvesNullOnSamplingMissEvenWhileBuffering` and `captureReplayResolvesScopeIdWhenReplayWasSent`.

packages/core/test/replay/mobilereplay.test.ts
```diff
@@ -480,14 +529,18 @@ describe('Mobile Replay Integration', () => {
...
 
+      // No buffered recording => nothing to link...
       expect(result).toBeDefined();
-      expect(mockCaptureReplay).toHaveBeenCalledWith(true); // isHardCrash
       expect(result?.contexts?.replay?.replay_id).toBeUndefined();
+
+      // ...and nothing to flush, even after the event is sent.
+      await fireAfterSendEvent(result as Event);
+      expect(mockCaptureReplay).not.toHaveBeenCalled();
     });
```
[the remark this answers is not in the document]
antonis: Fixed in a676da9a. Removed the bounded `eventsPendingReplayFlush` set (and its cap/marker) entirely; the flush decision now lives on the event's own linked replay context (`event.contexts.replay.replay_id`), so no shared bookkeeping can evict a pending event. New test links 200+ events in `beforeSend` before sending the first and asserts it still flushes exactly once.

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
```diff
@@ -481,10 +481,29 @@ public void fetchNativeFramesDelay(
...
 
   public @Nullable String getCurrentReplayId() {
+    // Prefer the replay controller's id: it is assigned when recording starts
+    // (buffer or session) and is therefore available BEFORE a replay is
+    // flushed, so a buffered (on-error) replay can be linked to the event in
+    // `beforeSend`. The scope's replayId is only populated once a replay is
+    // sent, so it stays empty while a buffer replay is recording (issue #6598).
+    final @NotNull SentryId controllerId =
+        Sentry.getCurrentScopes().getOptions().getReplayController().getReplayId();
+    if (controllerId != SentryId.EMPTY_ID) {
```
alwx: Shouldn't it be something like that to avoid checking for object equality and check for value instead? 
```
if (!SentryId.EMPTY_ID.equals(controllerId)) {
```

packages/core/src/js/replay/mobilereplay.ts
```diff
@@ -345,7 +345,25 @@ export const mobileReplayIntegration = (initOptions: MobileReplayOptions = defau
...
+  //      the event to the buffered replay id (no flush).
+  //   2. `flushReplayForSentEvent` runs in `afterSendEvent`, which fires only
+  //      for events that survive sampling and are actually sent, and performs
+  //      the native flush there.
+  //
+  // Trade-off: the link is tagged optimistically from the buffered id before the
+  // native `replaysOnErrorSampleRate` roll (which still happens at flush time in
+  // `captureReplay`). If that roll misses, the event carries a `replay_id` for a
+  // replay that is never uploaded. A fully-correct fix requires the native SDKs
+  // to decouple the on-error sampling decision from the buffer upload.
```
alwx, line 364: > If that roll misses, the event carries a `replay_id` for a replay that is never uploaded.

Maybe I'm wrong but seems like the fix introduced in this PR still produces orphaned `replay_id` links when `replaysOnErrorSampleRate` misses which sounds a bit like a bug, not really a trade off. 
IMO it's better to fix it, or at least create an issue instead of writing it down as a block comment in production code.

alwx (changes requested):
One question here (except for review comments): what happens if If two errors survive sampling simultaneously (or near-simultaneously, with a super minor difference in time)? 

I think it would mean both will reach `afterSendEvent` before either flush completes and then both will call `captureReplay`.

Maybe some sort of guard needs to be added here.

packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java
antonis: Good catch. fixed with 649f560894dbc10a931a578e104578fafada0502

packages/core/src/js/replay/mobilereplay.ts
antonis: Good point 👍 Created an issue to follow up on this https://github.com/getsentry/sentry-react-native/issues/6696

antonis:
> One question here (except for review comments): what happens if If two errors survive sampling simultaneously (or near-simultaneously, with a super minor difference in time)?
> 
> I think it would mean both will reach `afterSendEvent` before either flush completes and then both will call `captureReplay`.
> 
> Maybe some sort of guard needs to be added here.

Good question 👍 I traced both native paths and should be safe. Both errors share the same buffered replayId, and the captureReplay bridge calls run on a single serial thread on each platform (iOS methodQueue = main queue; Android's Promise @ReactMethod = native-modules thread), so they execute in sequence.

- iOS: captureReplay() starts with guard !isFullSession else { return true }; the first call flips isFullSession before returning, so the second no-ops.
- Android: the first call converts BufferCaptureStrategy → SessionCaptureStrategy, whose captureReplay is a no-op ("already running in 'session' mode").

Ready for another pass 🙇

alwx (approved):

PR #6692, opened 2026-09-09
fix(android): identify the screen appear event by name instead of class

### LINKED ISSUE #6691: Android: time to initial display is understated in R8-optimized builds because the screen appear event is matched by class name

React Navigation, Hermes, RN New Architecture

sentry.io

Yes

FullStory

8.25.0

```
React Native 0.86.3, Hermes, New Architecture enabled
@sentry/react-native 8.25.0
react-native-screens 4.27.0
@react-navigation/native-stack 7.18.10
Android Gradle Plugin with its bundled R8 8.12.14, no pinned compiler
JDK 17
Release build with minifyEnabled true and proguard-android-optimize.txt, which is the React Native template default
```

### Sentry.init()

```js
const navigationIntegration = Sentry.reactNavigationIntegration({
});

Sentry.init({
  dsn: 'https://examplePublicKey@o0.ingest.sentry.io/0',
  tracesSampleRate: 1.0,
  enableAutoPerformanceTracing: true,
});
```

### Steps to Reproduce

Minimal reproduction: https://github.com/TaduJR/sentry-ttid-r8-repro. Four dependencies, no patched libraries, default React Native release configuration, the R8 that AGP ships, no pinned compiler.

1. `npm install`
2. `cd android && ./gradlew assembleRelease`
3. `grep ScreenAppearEvent android/app/build/outputs/mapping/release/mapping.txt`
4. `apkanalyzer dex code --class G2.a android/app/build/outputs/apk/release/app-release.apk`

Step 3 returns a single method line that sits inside the entry for `com.facebook.react.views.drawer.events.DrawerClosedEvent`, because a class that is merged away gets no header of its own. Step 4 shows that class carrying a synthetic discriminator field and a switch over 22 event names.

Equivalent by hand: use a React Navigation native stack so react-native-screens creates `ScreenStackFragment`, enable `reactNavigationIntegration({enableTimeToInitialDisplay: true})`, build a release with `proguard-android-optimize.txt`, and disassemble the listener that `RNSentryReactFragmentLifecycleTracer` installs.

### Expected Result

The listener waits for the screen appear event, so `ui.load.initial_display` closes when the screen is actually drawn.

### Actual Result

`RNSentryReactFragmentLifecycleTracer` identifies the event by canonical class name:

```java
if ("com.swmansion.rnscreens.events.ScreenAppearEvent"
    .equals(event.getClass().getCanonicalName())) {
```

Class names are not stable under R8. Horizontal class merging collapses identically shaped classes into one, and React Native `Event` subclasses share a superclass and a constructor signature, which makes them candidates for merging.

In the linked reproduction, `ScreenAppearEvent` is merged into React Native's own `DrawerClosedEvent`, obfuscated to `G2.a`. The merged class carries a synthetic discriminator field and a switch over 22 event names:

```
.field public final synthetic l:I

iget v0, p0, LG2/a;->l:I
packed-switch v0, :pswitch_data_42
    "topAppear"  "topWillAppear"  "topDisappear"  "topWillDisappear"  "topDismissed"
    "topFocus"   "topAttached"    "topDetached"   "topOpen"           "topClose"
    "topSearchFocus"  "topSearchBlur"  "topFinishTransitioning"
    "topHeaderBackButtonClicked"  "topClick"  "topRefresh"  "topShow"
    "topRequestClose"  "topDrawerOpen"  "topDrawerClose"  "dismissCount"  "target"
```

R8 rewrote the string literal to follow the merge, so the shipped comparison is against `G2.a`.

All 22 are the same class at runtime, so the check matches on whichever dispatches first after the fragment view is created. `topWillAppear` always precedes `topAppear`, and `topFocus` and `topAttached` can precede both. The first draw listener is registered too early and the initial display span closes early, so time to initial display is reported shorter than it was.

This is silent. Every diagnostic on this path logs at `SentryLevel.DEBUG`, and nothing throws.

Two notes for anyone verifying this. The mapping file shows nothing at the class level, so `grep ScreenAppearEvent mapping.txt` looks clean unless you read the method lines inside other classes, and only the dex is reliable. react-native-screens ships no consumer ProGuard rules, so nothing protects the event class in a consuming app.

The sibling check against `com.swmansion.rnscreens.ScreenStackFragment` is not affected. androidx.fragment ships a consumer rule keeping public `Fragment` subclasses that have a public no-arg constructor, so that class is never merged.

TaduJR:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
`RNSentryReactFragmentLifecycleTracer` waited for the react-native-screens appear event by comparing its canonical class name:

```java
if ("com.swmansion.rnscreens.events.ScreenAppearEvent"
    .equals(event.getClass().getCanonicalName())) {
```

Class names are not stable once R8 optimization is on. React Native `Event` subclasses share a superclass and a constructor signature, which makes them candidates for merging, and R8 collapses a group of them into one class that answers to a single name. It also rewrites the compared string literal to that name, so the comparison then matches every event in the group rather than one.

The event is now identified by `Event.getEventName()`, which React Native documents as "the name of this event as registered in JS". react-native-screens registers `topAppear` as the `onAppear` prop, so the name is part of its public contract and cannot change without breaking it. Unoptimized builds behave exactly as before.

The sibling check against `com.swmansion.rnscreens.ScreenStackFragment` is left as is. androidx.fragment ships a consumer rule that keeps public `Fragment` subclasses with a public no-arg constructor, and Android requires that constructor for fragment recreation, so that class is never merged.

Fixes https://github.com/getsentry/sentry-react-native/issues/6691
With R8 optimization enabled, `ui.load.initial_display` closes on whichever merged event dispatches first rather than on the screen appear event. `topWillAppear` always precedes `topAppear`, and events such as `topFocus` and `topAttached` can precede both, so time to initial display is reported shorter than it was. Nothing throws, and every diagnostic on this path logs at `SentryLevel.DEBUG`, so the data is wrong silently.

This affects any app that enables optimization, which AGP 9 makes the default by dropping support for the ProGuard file that carried `-dontoptimize`.

Reproduction: [REPRO_URL](https://github.com/TaduJR/sentry-ttid-r8-repro). Four dependencies, no patched libraries, the R8 that AGP ships. There `ScreenAppearEvent` is merged into React Native's own `DrawerClosedEvent`, and the merged class holds 22 event names behind one class name.

Added unit tests to `RNSentryReactFragmentLifecycleTracerTest`, which previously covered listener registration but not which event triggers the first draw. One asserts that `topAppear` registers the first draw listener and removes the event listener, one asserts that `topWillAppear` does neither.

The reported behaviour was confirmed by disassembling the release dex of the reproduction, where the compared literal had been rewritten to the merged class name.

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps
An alternative that needs no code change is to ship `-keep class com.swmansion.rnscreens.events.ScreenAppearEvent` as a consumer ProGuard rule in the SDK. That keeps the class name intact but leaves the identity check dependent on the optimizer, which is why matching by event name is proposed instead.


packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
```diff
@@ -22,6 +22,7 @@
 import org.jetbrains.annotations.Nullable;
 
 public class RNSentryReactFragmentLifecycleTracer extends FragmentLifecycleCallbacks {
+  private static final String SCREEN_APPEAR_EVENT_NAME = "topAppear";
```
alwx, line 29: Could you please add a code comment explaining what this constant is? You did that in the PR itself but nobody will remember about this PR once it's merged :)

alwx (changes requested):

alwx:
Hey @TaduJR,

Thanks for submitting this PR! I've added a small comment to the code, and I think that except this really small thing, everything looks good for me.

packages/core/android/src/main/java/io/sentry/react/RNSentryReactFragmentLifecycleTracer.java
TaduJR: Thanks @alwx 

Addressed

TaduJR:
Hi @alwx 

Thanks so much for reviewing.

Addressed your concern.

TIA.

alwx (approved):

PR #6707, opened 2026-09-11
fix(core): Fix process.exit typo in debugid script

friederbluemle:
has-sourcemap-debugid.js called process.exist(1), which is undefined. Both failure branches (unreadable sourcemap, missing debugId) threw a TypeError with a stack trace instead of exiting cleanly with status 1.


antonis (approved):

antonis:
Thank you for catching this and opening a PR 🙇 LGTM 🚀

PR #6719, opened 2026-09-14
ci(e2e): Fix broken idb-companion install on iOS e2e

antonis:
- [x] Bugfix

## :scroll: Description
The `Install iDB Companion` step in the iOS e2e job started hard-failing on **both `main` and PRs**:

```
Invalid formula (...): .../facebook/homebrew-fb/buck.rb
Invalid formula (...): .../facebook/homebrew-fb/xar.rb
Cannot tap facebook/fb: invalid syntax in tap!
Process completed with exit code 1.
```

The step ran `brew tap facebook/fb && brew install facebook/fb/idb-companion`. Homebrew **auto-updates on the runner** (`==> Auto-updated Homebrew!` in the logs), and the current Homebrew now rejects the unmaintained `facebook/fb` tap: `buck.rb`/`xar.rb` use disabled `depends_on :java` / `depends_on :osxfuse` syntax. Because `brew tap` validates the *entire* tap, it aborts before `idb-companion` is even reached — so an install we do want dies on two formulae we don't use.

This is upstream/infra rot, not a code change here — the step was unchanged since Oct 2024 (#4210). It only surfaces on `ready-to-merge` runs, which is why it went unnoticed.

**This PR removes the step.** Recent Maestro ships its own iOS driver and no longer needs `idb_companion` — so the install was dead weight even before it broke.

Unblocks all iOS e2e runs, which currently fail deterministically at the idb install step.

Added the `ready-to-merge` label to run the full e2e suite. **Both iOS e2e Test jobs pass with the step removed:**
- ✅ `Test RN 0.87.1 new hermes ios production no`
- ✅ `Test RN 0.87.1 legacy hermes ios production no`

confirming Maestro drives the iOS simulator without `idb_companion`.

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.


lucas-zimerman (approved):
LGTM!

PR #6720, opened 2026-09-14
fix(tracing): background root spans overwrite native propagation context

### LINKED ISSUE #6697: fix(tracing): background root spans overwrite native propagation context
## Problem

`syncPropagationContextToNative` (added in #6686) fires on every root span via the `spanStart` hook. Background/inactive roots created with `startInactiveSpan({ forceTransaction: true })` — app-start in `integrations/appStart.ts` — also trigger a sync, overwriting the native scope with their `traceId`/`spanId` while a navigation root may already be active. When these background roots end, there is no restore, so in-flight native HTTP requests (OkHttp / URLSession) get attributed to the wrong trace.

A secondary issue: `sampleRand` is read from `getCurrentScope().getPropagationContext()` at sync time, which can be stale for these background roots since they don't update the propagation context before starting (unlike idle navigation spans).

## Steps to reproduce

1. App-start span fires mid-navigation
2. Native HTTP request starts during the navigation transaction
3. The request is linked to the app-start trace instead of the navigation trace

## Expected

Background/inactive root spans should not overwrite the native propagation context when an active navigation trace is already set.

## Possible fix

Move the sync out of the generic `spanStart` hook and into `startIdleNavigationSpan` directly, after the span is made active. This naturally excludes background roots without needing to filter by op or attributes.

alwx:
- [x] Bugfix

## :scroll: Description
`syncPropagationContextToNative` used to run on every root span's `spanStart`. Background roots (app-start, expo-updates), which never become the active span, would still fire it and overwrite the native scope's trace, clobbering an already-active navigation trace with no way to restore it.

Moved the sync out of the generic hook into `startIdleSpan`, called right after the idle span (navigation or interaction) becomes active. Background roots never go through `startIdleSpan`, so they're excluded by construction rather than by filtering on op/attributes.

Fixes #6697

- Rewrote `nativeScopeSync.test.ts` for the new call-site-based behavior, including a regression test that starts a navigation span then an app-start-like background span and asserts the native context isn't overwritten.
- Added coverage in `idleNavigationSpan.test.ts` and `appStart.test.ts` for the same scenario.
- Full test suite, build, lint, circular dep check, and API report all pass.

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.

## :crystal_ball: Next steps


alwx:
@cursor review

antonis (approved):
LGTM 🚀 
The failed iOS e2e should be fixed when we merge https://github.com/getsentry/sentry-react-native/pull/6719

PR #6722, opened 2026-09-14
fix(android): Prevent fatal 'JavaCallback was already settled' crash from getNewScreenTimeToDisplay

### LINKED ISSUE #6709: Fatal 'JavaCallback was already settled' from getNewScreenTimeToDisplay under rapid navigation (Android, Expo, new arch)
### OS

Android

### Platform

Hermes, New Architecture

### SDK Version

8.19.0

### React Native Version

0.85.3 (Expo SDK 56, expo-modules-core 56.0.19)

### Steps to Reproduce

Not deterministic — it is a race, but the shape that reproduces it in our production fleet:

1. `Sentry.init` with `reactNavigationIntegration({enableTimeToInitialDisplay: true})`, container registered via `registerNavigationContainer` (expo-router).
2. App with 4 native bottom tabs (`react-native-bottom-tabs`) sharing a header.
3. Switch tabs rapidly and/or push a modal right after a tab switch (our events consistently show a paywall modal being pushed immediately after navigation — breadcrumbs end with the new screen's mount logs).

### Expected Result

TTID measurement completes or is dropped silently when navigations race.

### Actual Result

Fatal crash reported through `onerror`:

```
Error: Non-js exception: JavaCallback was already settled. Cannot invoke it again
```

Event tags point at the SDK's own TurboModule call:

- `turbo_module.name: RNSentry`
- `turbo_module.method: getNewScreenTimeToDisplay`
- `mechanism: onerror`, `handled: no`, `level: fatal`, Hermes, Android 12–16

Volume: 23 users in 4 days in production (it likely existed longer — our DSN was broken before that, so the SDK could not report its own crash).

### Analysis

`reactnavigation.ts` (`updateLatestNavigationSpanWithCurrentRoute`) calls `NATIVE.getNewScreenTimeToDisplay()` on **every** navigation state change while `enableTimeToInitialDisplay` is enabled. The Android implementation appears to keep a single stored callback to be settled on the next frame; when two navigation state changes race, the same callback gets settled twice. On Expo (new architecture) the promise guard in `expo-modules-core` (`PromiseImpl` — "was already settled") turns the second settle into the fatal above.

No matching fix found in the changelog up to 8.26.0, and no existing issue mentions `getNewScreenTimeToDisplay`, so filing this one.

### Workaround

`enableTimeToInitialDisplay: false` removes the crashing path entirely (confirmed: the call is gated only by that option).

antonis:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
On Android, `RNSentryTimeToDisplay.getTimeToDisplay` (reached from the `getNewScreenTimeToDisplay` TurboModule method) settles the JS promise from a **deferred `Choreographer` frame callback**. Under rapid navigation the bridge callback backing that promise can already have been settled/invalidated by the runtime by the time the frame callback runs — on Expo bridgeless (new architecture) this surfaces as `JavaCallback was already settled. Cannot invoke it again`. Because the settle happens *inside* the frame callback, the throw was uncaught and reached the host app as a **fatal crash**.

This routes every settle through guarded helpers (`resolveSafely` / `rejectSafely`) that:
- settle the promise **at most once** (`AtomicBoolean` guard), and
- **swallow any settle failure**, logging at `WARNING` instead of propagating.

A late/duplicate settle now drops the time-to-display measurement (the expected degradation) instead of crashing the app. The implementation is shared between the old and new architecture, so the fix covers both.

Fixes #6709

New Android unit tests in `RNSentryTimeToDisplayTest`:
- Manula testing
- Added tests

## :pencil: Checklist
- [x] All tests passing.
- [x] No breaking changes.


antonis:
@cursor review

antonis:
@sentry review

packages/core/android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
```diff
@@ -69,11 +85,49 @@ public static void getTimeToDisplay(Promise promise, SentryDateProvider dateProv
...
+      @Nullable Throwable cause,
+      ILogger logger) {
+    if (!settled.compareAndSet(false, true)) {
+      return;
+    }
+    try {
+      if (cause != null) {
+        promise.reject(message, cause);
+      } else {
+        promise.reject(message);
```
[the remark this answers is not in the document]
antonis: The `reject(String)` overload does exist on the React Native `Promise` interface — it's declared at `Promise.java:123` (`@Deprecated`, maps to the generic `EUNSPECIFIED` code) and `PromiseImpl.reject(String)` (`PromiseImpl.java:253`) delegates to `reject(null, message, null, null)` without throwing. So this compiled and rejected normally — no silent failure. The usage also predates this PR: the `mainLooper == null` branch has called `reject(String)` since 6.2.0.

The deprecation is a fair point, though — switched the null-cause branch to the coded `reject("SentryTimeToDisplay", message)` overload in 8ee98f8a so JS gets a specific error code instead of `EUNSPECIFIED`.

packages/core/android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
```diff
@@ -69,11 +85,51 @@ public static void getTimeToDisplay(Promise promise, SentryDateProvider dateProv
...
+    if (!settled.compareAndSet(false, true)) {
+      return;
+    }
+    try {
+      if (cause != null) {
+        promise.reject(message, cause);
+      } else {
+        // Use the coded overload rather than the deprecated reject(String) so JS receives a
+        // specific error code instead of the generic "EUNSPECIFIED".
+        promise.reject("SentryTimeToDisplay", message);
```
lucas-zimerman: Q: why not `promise.reject("SentryReactNative",...  );` ?  all promise reject that are in the code uses the same string

packages/core/android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
```diff
@@ -69,11 +85,51 @@ public static void getTimeToDisplay(Promise promise, SentryDateProvider dateProv
...
   }
+
+  @VisibleForTesting
+  static void resolveSafely(Promise promise, AtomicBoolean settled, double value, ILogger logger) {
+    if (!settled.compareAndSet(false, true)) {
+      return;
+    }
+    try {
+      promise.resolve(value);
+    } catch (Throwable t) { // NOPMD - We don't want to crash the host app
```
lucas-zimerman, line 108: Q: `resolveSafely/rejectSafely` catch `Throwable` and treat any failure as "already settled", could that also swallow something unrelated (e.g. a torn-down `CatalystInstance`)? In that case the promise was never actually settled and nothing settles it after, so whatever's awaiting `getNewScreenTimeToDisplay` just hangs instead of degrading. Not blocking, but maybe log unrecognized causes at a level above WARNING so it's distinguishable from the expected "already settled" case, the tests currently only cover that literal message.

packages/core/android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
antonis: Good point @lucas-zimerman! Updated with fd551773e840fe3b622eb82df19da76a57f76831

packages/core/android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
antonis: Updated with fd551773e840fe3b622eb82df19da76a57f76831

packages/core/android/src/main/java/io/sentry/react/RNSentryTimeToDisplay.java
```diff
@@ -69,11 +85,51 @@ public static void getTimeToDisplay(Promise promise, SentryDateProvider dateProv
...
+      AtomicBoolean settled,
+      String message,
+      @Nullable Throwable cause,
+      ILogger logger) {
+    if (!settled.compareAndSet(false, true)) {
+      return;
+    }
+    try {
+      if (cause != null) {
+        promise.reject(message, cause);
```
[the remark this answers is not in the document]
antonis: Updated with fd551773e840fe3b622eb82df19da76a57f76831

lucas-zimerman (approved):
LGTM!

PR #6749, opened 2026-09-17
fix(ios): Honor shutdownTimeout on iOS

### LINKED ISSUE #6748: shutdownTimeout is silently ignored on iOS (key/unit mismatch with cocoa shutdownTimeInterval)
## Summary

The documented `shutdownTimeout` option is honored on Android but has **no effect on iOS**. The native cocoa SDK always uses its default 2s shutdown flush interval regardless of what the user sets.

## Root cause

The JS layer forwards options to native by key name (`wrapper.ts` → `RNSentry.initNativeSdk(filteredOptions)`), and `shutdownTimeout` passes through unchanged. On native:

- **Android** — `RNSentryStart.java` maps `shutdownTimeout` → `setShutdownTimeoutMillis(...)`. ✅ Works (both are milliseconds).
- **iOS** — cocoa's dictionary parser (`SentryOptionsInternal`) only reads the key **`shutdownTimeInterval`** (an `NSTimeInterval`, i.e. **seconds**). It has no `shutdownTimeout` key, and `RNSentryStart.m` does no explicit handling. So the value is dropped and the native default (2s) is always used.

Two mismatches, not one:

1. **Key name:** RN sends `shutdownTimeout`; cocoa reads `shutdownTimeInterval`.
2. **Unit:** RN's `shutdownTimeout` is **milliseconds** (per its JSDoc and Android's `...Millis`); cocoa's `shutdownTimeInterval` is **seconds**. A rename alone would be a 1000× error — the value must be divided by 1000.

## Evidence

- `packages/core/src/js/options.ts` — `shutdownTimeout?: number` documented as "milliseconds … default 2 seconds".
- `packages/core/src/js/wrapper.ts` — `shutdownTimeout` is not filtered out, so it's forwarded verbatim.
- `packages/core/ios/RNSentryStart.m` — no `shutdown*` handling.
- cocoa `SentryOptionsInternal` — parses only `shutdownTimeInterval` (seconds).

## Expected

Setting `shutdownTimeout` affects the native iOS shutdown flush interval, matching Android behavior.

## Proposed fix

In `RNSentryStart.m` (`createOptionsWithDictionary`), read `shutdownTimeout` (ms) and set `sentryOptions.shutdownTimeInterval = value / 1000.0`. Add test coverage on both platforms.

## Environment

Affects all RN SDK versions on the current native binding; verified against bundled cocoa 9.29.0.

antonis:
#### 📢 Type of change

- [x] Bugfix

#### 📜 Description

The documented `shutdownTimeout` option was silently ignored on iOS, so the native SDK always used its default 2s flush interval on app close.

Two mismatches with Cocoa caused it:
- **Key:** RN forwards `shutdownTimeout`, but Cocoa's dictionary parser only reads `shutdownTimeInterval`.
- **Unit:** RN's `shutdownTimeout` is milliseconds (matching Android's `setShutdownTimeoutMillis`), while Cocoa's `shutdownTimeInterval` is an `NSTimeInterval` in seconds.

`RNSentryStart.m` now reads `shutdownTimeout` (ms) and sets `shutdownTimeInterval = value / 1000.0`. Android was already correct.

The option is already documented for React Native ([Transport Options → `shutdownTimeout`](https://docs.sentry.io/platforms/react-native/configuration/options/#shutdownTimeout)); this fix simply makes iOS honor it, so no docs change is needed.

#### 💡 Motivation and Context

`shutdownTimeout` controls how long the SDK waits to flush events before the app terminates — directly affecting whether events survive termination. It was a documented, promised option that was broken on iOS. Fixes #6748.

#### 💚 How did you test it?

Added `RNSentryCocoaTester` unit tests asserting the ms→s mapping (`shutdownTimeout: 5000` → `shutdownTimeInterval == 5.0`) and the default (`shutdownTimeInterval == 2.0`).

#### 📝 Checklist

- [x] I added tests to verify the changes.
- [x] No breaking changes.
- [x] I updated the docs if needed. (already documented — no change needed)

#### 🔮 Next steps

None.


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
 > make sure you follow our [migration guide](https://docs.sentry.io/platforms/react-native/migration/) first.
 <!-- prettier-ignore-end -->
 
+## Unreleased
+
+### Fixes
+
+- Honor `shutdownTimeout` on iOS ([#6748](https://github.com/getsentry/sentry-react-native/issues/6748))
```
```suggestion
- Honor `shutdownTimeout` on iOS ([#6749](https://github.com/getsentry/sentry-react-native/pull/6749))
```

antonis:
@cursor review

antonis:
@sentry review

alwx (approved):

